PR #258, opened 2012-03-19
Add support for concatenated depfiles

PetrWolf:
This adds a new feature in ninja - the ability to read dependency information from "depfile groups", which are text files created by simple concatenation of several depfiles.

This is to speed up ninja start-up time, especially on Windows, where reading large amount of "little" depfiles is inefficient.

Note that this also includes a refactoring of Edge::StatIfNecessary, because I needed to separate the notion of a stat-ed and visited node.

For more info, see the commit in asciidoc for high-level description (478d765f5f) and the following e-mail threads on previous discussions:
* .d files on Windows again
* Building ninja with MSVC

Please, raise questions and comments!


martine:
We think this is no longer necessary due to the depslog.  Please reopen if we're wrong!

PR #274, opened 2012-04-19
Max load average

nicolasdespres:
Add -l N option to limit the load average.

This is similar to GNU make -l/--load-average option.  It limits the number of job started if the load average exceed the given value.  It can be very useful when running ninja on a continuous integration server where we want to use parallelism as much as possible without overloading the server.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -452,7 +452,9 @@ void RealCommandRunner::Abort() {
 }
 
 bool RealCommandRunner::CanRunMore() {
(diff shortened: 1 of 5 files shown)


lilydjwg:
Is there any way to configure this globally (like `MAKEFLAGS` for make)? Or I have to tell everyone to not forget it? (At the moment two ninjas from different users are pushing the load very high....)

ocroquette:
@lilydjwg See https://github.com/ninja-build/ninja/pull/1399

PR #303, opened 2012-05-08
don't delete response file when verbose so printed cmdline is runnable

sgraham:


sam-github:
What happened to never delete?

I'm trying to figure out wha compiled options are being used to build a .c file, and since they are all hidden in a response file I've no idea. The compilation is succeeding, but I need to know, because they are in some way conflicting with the compilation options of a different .c file, and its only at the link stage errors occur.

I'm going to deliberately introduce a syntax error to cause the compile to fail and the response file to be left, but that's pretty crude, a "leave the response files around" option would be greatly appreciated.

evmar:
Try `ninja -d list`

sam-github:
@evmar thanks

PR #440, opened 2012-10-05
Add -d depcheck to check for unsafe dependencies

wolfp:
The (optional) -d depcheck activates a mode, in which ninja would
perform additional safety checks while reading the depfiles. If a file
mentioned in a depfile is at the same time a target of some rule, ninja
will check, whether this dependency is mentioned in the .ninja files
(i.e. whether ninja would know about it in a clean run, without
depfiles). If not, the dependency is not safe and ninja will raise an
error.

Becasue this option has a runtime cost and makes only sense after a
complete build (we must have depfiles to inspect), the typical usage
would be to run this as a sanity chec of the integrity of the build
after the build itself, e.g. on a central CI server.


jhasse:
In order to reduce the number of open PRs I'm closing this one. If you're interested in bringing it up-to-date, feel free to reopen :)

PR #452, opened 2012-10-29
 Adding a deps tool that returns recursive inputs/outputs

qsr:
The deps tool will return a list of inputs/outputs for the given target
in the format of a python dictionary.

 Contrary to query, the inputs are computed recursively.


martine:
It sounds like this is no longer necessary.  Please correct me if I'm wrong.

PR #456, opened 2012-11-09
Add "workingdir" as special global build variable, equivalent to -C

rgeary1:
This is used by the Premake ninja generator in https://github.com/rgeary1/premake-dev-rgeary

Similar to the ninjadir variable, the workingdir variable sets the current working directory equivalent to specifying -C on the command line. 

This is very useful in the following case. Suppose you have a master build script with all the edges (buildedges.ninja), and many build.ninja scripts in the code subdirectories. The build.ninja script sets default build edges & subninja's to the buildedges.ninja script. For the build.ninja script to run, it needs to change the working dir to the buildedges.ninja directory. So now the build script generator can insert a "workingdir" variable in the build script to embed this information.

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -608,6 +608,19 @@ bool DebugEnable(const string& name, Globals* globals) {
   }
 }
 
+bool SetWorkingDirFromLog(Globals* globals) {
+  const string working_dir =
+      globals->state->bindings_.LookupVariable("workingdir");
+
+  if (!working_dir.empty()) {
+    if (chdir(working_dir.c_str()) < 0) {
+      Fatal("chdir to '%s' - %s", working_dir.c_str(), strerror(errno));
+    }
+    return true;
+  }
+  return false;
+}
+
 bool OpenLog(BuildLog* build_log, Globals* globals,
              DiskInterface* disk_interface) {
   const string build_dir =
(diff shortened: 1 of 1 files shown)


nico:
We're cleaning out all issues. This looks like it's not fully thought through (what if a subninja has a workingdir? build.ninja is also usually relative to the workingdir) and hasn't seen many requests in 3 years, so we're going to close this for now. Shout if your really really need something like.

chisophugis:
I'd just like to say that something like this would be nice. I don't think this is the right design though. I would phrase the problem in a simpler way as that it is useful to be able to compose ninja builds without recursively running `ninja -C subdir`.

Concretely, consider the following directory tree:
```
out/
    Release/
        build.ninja
        ...
    ReleaseLTO/
        build.ninja
        ...
```

I would like a way of adding `out/both.ninja` with, say, the following contents (strawman syntax):
```
scoped subninja Release/build.ninja
scoped subninja ReleaseLTO/build.ninja
```
Depending on the nature of the build, this can be a huge speedup because ReleaseLTO has a very long link job which leaves a bunch of cores free. If ninja can see the entire build graph then it can get other work done while the long LTO job is running.

The reason that this doesn't currently work with `subninja` (or `include` for that matter) is that the original cwd is preserved and in general programs do rely on the cwd.

PR #477, opened 2012-12-30
Include subninja/included files in dependency graph

ogait87:
This commits should add initial support for the feature https://github.com/martine/ninja/issues/370


orenbenkiki:
An alternative approach, which is aligned with the way `ninja` works wrt. the main input file, is https://github.com/ninja-build/ninja/pull/2232

PR #520, opened 2013-03-17
Fix error in link time if link command line too long

gendalph:
Cool build project for construct huge projects from source, but...

Not true solution, but work fine for my project.

Use execv instead of execl (and 100% error)
in real huge project linking command more than MAX_ARG_STRLEN.
We can split in in this case and try link.


nico:
I think you can get this working more nicely by using `rspfile` and `rspfile_contents` for your link lines.

PR #561, opened 2013-04-29
Add "environment" as a special attribute

sgraham:
Oops, forgot to click pull.

This optionally sets the environment block for Windows subprocesses. Do you think this is a reasonable approach to getting rid of the wrapper on win?

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -777,7 +777,8 @@ keys.
   is a filename. When `command` is started via
   http://msdn.microsoft.com/en-us/library/windows/desktop/ms682425.aspx[
   CreateProcess], the contents of the named file will be used as the
-  `lpEnvironment` argument. This key is only supported on Windows.
+  `lpEnvironment` argument, and so will completely replace the environment
+  inherited by `command`. This key is only supported on Windows.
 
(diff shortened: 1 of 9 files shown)


nico:
I always said that I thought just setting the env via e.g. `ninja -t msvc -e` would be fine, but iirc @ngladitz collected data for me that shows that the subprocess overhead from that is actually measurable. (I think that data is still in a tab on another computer, but I can't find it now.) So this might be something we might want to do in the future, but it's a bit messy and the win wasn't _that_ huge iirc. But let's keep this open for now.

nico:
Here's the link ngladitz sent me: https://github.com/ngladitz/cmake/commits/ninja-cl-deps-performance

martine:
I suspect if we went down the route of the inline environment block we'd run into requests for:
- a way to express just additions/removals to the outer environment
- a more convenient way to express them than as a long block of `foo$0bar$0baz$0$0`
Perhaps we can punt both of those to just say it's the generator's problem, though.

This is a longer way of saying: it'd be nice to collect all the requirements of users before we commit to any new feature.

jhasse:
Closing due to merge conflicts. If you still think this is relevant, feel free to reopen :)

PR #567, opened 2013-05-03
Fix query tool with implicit deps

riannucci:
Two basic changes:

 * Move the RUN_AFTER_LOAD to after the Builder is constructed but before the build is run. Also pass Builder to tools.
 * When loading implicit dependencies, assign the dependent node as an output of its dependency.

I've got a generic ninja https://github.com/Valloric/YouCompleteMe `.ycm_extra_conf.py` file  in the works which relies on query actually producing useful information for all files (the file returns the flags necessary to build a given file path, which I would like to have ninja accurately provide for me).


martine:
Is this still needed?

riannucci:
No idea! Probably not though.

PR #581, opened 2013-05-22
Add NINJA_SMART_TERMINAL environment variable, with on, off, and hide_newline options

rgeary1:
This patch allows you to manually enable or disable the smart terminal behaviour, and adds a hide_newline smart-terminal mode to allow sed processing.

Like NINJA_STATUS, I have added this optional feature as an environment variable rather than a typical command line option. Since it only affects how the progress is displayed, this would appear to be a good place for it, and minimises command line option clutter.

Use cases : Currently ninja will turn off smart terminal mode if you are piping to a file. Occasionally you want to force this behaviour on, to replicate the smart line display through a pipe, or you may want to force it off if your terminal emulator doesn't support special escape characters (eg. emacs)

Additionally, you can export NINJA_SMART_TERMINAL="hide_newline" :
This feature prints a \n newline and move-up-1-line escape-char \e[1A at the end of each line to allow sed processing. The stream editor "sed" usually processes the stream one line at a time, so it waits for a newline \n character before printing the piped output. The "hide_newline" feature allows the smart terminal behaviour whilst using sed.

 This allows the following script, which highlights directories in blue & files in red :

~/inet/ninja-master/ninja | sed  -e 's,^, ,' -e 's,$, ,' -e "s,([^+ ][A-Za-z/][-A-Za-z0-9/._]_/)([^ :]_.[ :]),\x1b[0;34m\1\x1b[m\x1b[1;31m\21b[m,g" -e 's,^ ,,' -e 's, $,,'


jhasse:
Thanks for the PR! Sorry for the long silence.

> This allows the following script, which highlights directories in blue & files in red

Wouldn't this work anyway since you're piping to sed which would disable the smart terminal features?

jonesmz:
@rgeary1 Can you rebase this pull request?

jhasse:
Closing due to inactivity. Also there's now `CLICOLOR_FORCE`, so I'm not sure if this is still needed.

PR #634, opened 2013-08-11
Fix for issue #478. Mtime of output of phony edges should be the latest mtime of its inputs

rgeary1:
Fix for issue #478 

Observed : Real edges depending on a phony edge will not be marked as dirty or rebuilt if the phony's (real) inputs are updated
Expected : An edge should always be rebuilt if its inputs or transitive inputs are newer than the output's mtime

Change : Node::mtime_ was overloaded, 0 represented "does not exist". This change disambiguates it by adding Node::exists_. Then to fix the observed behaviour, Node::UpdatePhonyMtime was added to update the mtime if the node does not exist.

I've added tests BuildTest.PhonyUseCase# GraphTest.PhonyDepsMtimes
Unit tests will also test for always-dirty behaviour if a phony rule has no inputs

### DIFF
--- a/src/build_test.cc
+++ b/src/build_test.cc
@@ -942,6 +942,151 @@ TEST_F(BuildTest, PhonyNoWork) {
   EXPECT_TRUE(builder_.AlreadyUpToDate());
 }
 
+// There are 6 different cases for phony rules:
+//
+// 1. output edge does not exist, inputs are not real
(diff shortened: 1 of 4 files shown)


martine:
Gonna close this in favor of #1021, the rebase.  (The comment history here will be readable via a link from there.)

PR #656, opened 2013-09-10
Proposed/fix redundant restat with deps ticket 652

maximuska:
Fixing https://github.com/martine/ninja/issues/652


jhasse:
Hi! I didn't take a close look yet, sorry. Is this something that still make sense? If so, please reopen and rebase on master.

PR #659, opened 2013-09-20
Use ANSI escapes to output to Cygwin-based terminals

rnk:
Cygwin detects ttys by looking at the name of the pipe the parent
process created for it.  It also uses the NtQueryObject API to get the
pipe name.

This enables terse output with mintty avoids stripping ANSI escape codes
from the output of subprocesses like clang.

This does _not_ enable the ellipsis insertion code because that code
doesn't compile on Windows.

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -190,10 +190,11 @@ you don't need to pass `-j`.)
 Environment variables
 ~~~~~~~~~~~~~~~~~~~~~
 
-Ninja supports one environment variable to control its behavior:
-`NINJA_STATUS`, the progress status printed before the rule being run.
+The output of ninja can be controlled with two environment variables.
+`NINJA_STATUS` controls the progress status printed before the rule being run,
+and `NINJA_TERM` can be used to override ninja's tty detection logic.
 
-Several placeholders are available:
+Several `NINJA_STATUS` placeholders are available:
 
 `%s`:: The number of started edges.
 `%t`:: The total number of edges that must be run to complete the build.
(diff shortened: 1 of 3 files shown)


TheOneRing:
As Windows 10 has native support for ANSI escape sequences nowadays this would also be interesting for non Cygwin users.
The use case for ANSI over GetConsoleScreenBufferInfo is that it also works when piped through python ;) .

nico:
Is there a way to detect if ninja's running in a Windows 10 environment with ANSI support? Or is just checking for Windows 10 at runtime enough since Windows 10 shells always support ANSI?

TheOneRing:
I couldn't  find anything in https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences regarding the supported Windows versions, but I guess a call to ```SetConsoleMode``` with 
https://docs.microsoft.com/en-us/windows/console/setconsolemode#ENABLE_VIRTUAL_TERMINAL_PROCESSING would fail on a platform not supporting it?

On the other hand Ninja should probably not set that flag itself so it could just check ```GetConsoleMode``` for ```ENABLE_VIRTUAL_TERMINAL_PROCESSING``` ?

nico:
GetConsoleMode() requires a HANDLE as returned by e.g. GetStdHandle(), meaning you run in cmd. In that case, ninja already doesn't strip ansi codes. So I think if you're in Win10 cmd, ansi codes will work if you tell the commands called by ninja to produce ansi output (by e.g. passing `-fcolor-diagnostics` to clang). So all the build files that do `if (!win) cflags = fcolor-diagnostics` need to be changed to say `if (!win || win10) cflags = fcolor-diagnostics` and things will be fine in Win10 cmd. I don't have a win10 box at hand to test, but it's possible this Just Works with current ninja.

(Reid's patch is for having the same behavior in non-cmd shells.)

TheOneRing:
Ok then I guess no change to ninja is needed besides one of the force color patches, as GetStdHandle will fail similar to isatty is ninja's output is piped.
thx

jhasse:
You can now disable escape code stripping using `CLICOLOR_FORCE=1`. For cygwin, it makes sense to implement option 1 suggested here: https://github.com/ninja-build/ninja/pull/1319#issuecomment-327451922 (msys2 also uses mintty and is very similar to cygwin IIRC)

PR #660, opened 2013-09-24
add option to limit job execution dependant on memory consumption (-m)

metti:
Setting a value in range [0,100] for -m limits starting of new jobs. once
the limit has been exceeded at most one single job is run at a time.

This aims at C++ make projects that run into memory limitations when
building expensive (in regards to memory) compilation units, e.g. caused
by massive template instantiations.

The implementation currently covers support for Apple, Linux and Windows.

Signed-off-by: Matthias Maennich matthias@maennich.net

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -186,6 +186,14 @@ match those of Make; e.g `ninja -C build -j 20` changes into the
 Ninja defaults to running commands in parallel anyway, so typically
 you don't need to pass `-j`.)
 
+In certain environments it might be helpful to dynamically limit the
+amount of running jobs depending on the currently available resources.
+Thus `ninja` can be started with `-l` and/or `-m` options to prevent
+from starting new jobs in case load average or memory usage exceed
+the defined limit. `-l` takes one single numeric argument defining
+the limit for the typical unix load average. `-m` takes one single
+numeric argument within [0,100] as a percentage of the memory usage
+(reduced by caches) on the host.
 
 Environment variables
 ~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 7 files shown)


maxim-kuvyrkov:
@metti , I don't quite understand why you closed this submission.  Most of the feedback was among the lines of "this is useful, it solves part of the problem, and $myproject may need another feature to solve memory problem completely".

Let me chime in with my use-case.  I'm looking after LLVM buildbots for AArch64 and AArch32 architectures.  I'm considering migration of buildbots from several 8-core 32GB machines to 96-core 128GB machine.  Note that RAM per core ratio goes down from 4GB to ~1GB.  During peak memory usage the machine runs out of RAM.  However, most of the time RAM utilization is ~20%.  Peak memory usage is caused by ~50 compilation jobs each consuming 1-2GB of RAM.

The approach I'm working on is to not start new jobs when RAM utilization is beyond 75%.  Peak memory usage occurs rather gradually, so 25% available RAM (32GB) is plenty for all existing jobs to finish.

HFTrader:
@metti I have to second @maxim-kuvyrkov here. I landed here by looking at ways to dynamically limit job creation by memory usage. My case is as simple as compiling clang on a 6-core virtual machine and 8GB total memory. If I could just say 'avoid creating new jobs if the memory usage at that point is over 60%' or 'only create new jobs if we have at least X Gb of memory available' it would be a wonderful flag and perhaps not so hard to implement.

maxim-kuvyrkov:
@HFTrader , I've taken @metti 's patch, reworked and improved it for my needs (which is 10x clang builds with -j64).  Might be useful to you: https://github.com/maxim-kuvyrkov/ninja , https://git-us.linaro.org/ci/dockerfiles.git/tree/tcwg-base/tcwg-llvmbot/run.sh#n152 .

Issue #413, opened 2012-09-10
Automatically use pager for failed output


mgaunard:
When compilation fails, the error message may be very long. I frequently pipe it to 'less' manually.
It would be nice if ninja, like git, could use the less pager automatically.

PR #672, opened 2013-10-18
Address FIXME: Add --no-strip-ansi-escapes option

chisophugis:
This option allows overriding Ninja's (usually useful) behavior of
stripping ANSI escapes when its output is not a smart terminal (e.g., a
dumb terminal, or redirected to a file).

I ran into a situation where I needed this, and lo and behold there was a FIXME in the source, so I thought I'd give it a shot.

Some things I would specifically like a second opinion on:
- The way I added the configuration option (is this the "right" way?)
- The option name
- The help text

This is a follow-up to #198, and now that I see it replicates the functionality in #435, but in a far simpler way (just enough to address the FIXME). In the future, it would be possible to add a corresponding `--strip-ansi-escapes` (following the `--foo`/`--no-foo` pattern) if it is desirable to be able to force this option on (but in the grand Ninja tradition, I've implemented the simplest thing that addresses the use case).

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -134,9 +134,10 @@ void BuildStatus::BuildEdgeFinished(Edge* edge,
     // (Launching subprocesses in pseudo ttys doesn't work because there are
     // only a few hundred available on some systems, and ninja can launch
     // thousands of parallel compile commands.)
-    // TODO: There should be a flag to disable escape code stripping.
     string final_output;
-    if (!printer_.is_smart_terminal())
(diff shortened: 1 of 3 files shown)


mgaunard:
Currently, colors are stripped whenever piping to less, so this also affects issue #413

jhasse:
I think the solution, which would accommodate all needs, would be something like #1210.

As for the disabling of output stripping, please see https://github.com/ninja-build/ninja/pull/1268#issuecomment-438289348.

PR #676, opened 2013-10-21
configure.py/bootstrap.py — account for CXXFLAGS.

geoff-codes:
Take into account CXXFLAGS in the environment in the bootstrap and configure scripts. I've tried to ensure this doesn't interfere with existing workflows which depend on using CFLAGS (only) by appending to, rather than replacing, CFLAGS; this may lead to some extraneous duplicates, but it seems better than the alternative.

Also, CXXFLAGS (and CFLAGS) are added to LDFLAGS with this change; this is to account for some special flags (like clang's `-stdlib`) which modify the behavior of both the compiler and the linker.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -53,6 +53,10 @@ def run(*args, **kwargs):
 # g++ call as well as in the later configure.py.
 cflags = os.environ.get('CFLAGS', '').split()
 ldflags = os.environ.get('LDFLAGS', '').split()
+cxxflags = os.environ.get('CXXFLAGS', '').split()
+ldflags.extend(cflags)
+cflags.extend(cxxflags)
+ldflags.extend(cxxflags)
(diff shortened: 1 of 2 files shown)


ryandesign:
Can this, or any other fix for the issue, be merged, please?

ryandesign:
Here is an updated/simplified version of this change which works for me with ninja 1.8.2 and which I have added to MacPorts:

https://github.com/macports/macports-ports/blob/53f57e242a7dc2bc4692db47aad18d4e35bd82a1/devel/ninja/files/patch-configure.py-CXXFLAGS.diff

nico:
@ryandesign If you send me a pull request for the diff you linked to, I'll merge it.

Closing this one out in the meantime since it doesn't apply anymore (and looked a bit more complex).

PR #685, opened 2013-10-31
Update search to work with latest python

Ation:
There is a problem. On windows 7 station, with Python 3.3.2 and russian VisualStudio installed, if you are trying to build ninja, you have an error  message
"File "platform_helper.py", line 65, in msvc_needs_fs return '/FS ' in out TypeError: Type str doesn't support the buffer API".
So there is attempt to fix this by decoding byte array and ignoring symbols, that could not be decoded. Apparently only non english letters will  be decoded and function will return required result.

### DIFF
--- a/platform_helper.py
+++ b/platform_helper.py
@@ -62,7 +62,7 @@ class Platform(object):
                                  stdout=subprocess.PIPE,
                                  stderr=subprocess.PIPE)
         out, err = popen.communicate()
-        return '/FS ' in out
+        return '/FS ' in out.decode("utf-8", "ignore")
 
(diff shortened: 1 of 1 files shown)


martine:
This was modified for Python 3 here:
https://github.com/ninja-build/ninja/pull/691

nico:
Thanks, I tried to rebase this and changed the way how exactly the fix works (use `b'/FS'` on the lhs instead). Please let me know if it's better now – it's possible that other things are broken with a Russian cmd.exe too. If there's more, merging will be faster next time, sorry about the long delay :-(

PR #686, opened 2013-11-13
Order-only dependencies trigger unneeded actions

usovalx:
Here is very artificial example:

```
rule CC
    command = gcc -c -o $out $in
build b.o: CC b.c
build c.o: CC c.c || b.o
```

Example of what's going wrong:

``` shell
touch b.c c.c
ninja
touch b.c
ninja -v c.o
```

Second ninja invocation triggers completely unnecessary rebuild of `c.o`


jhasse:
If I understand it correctly this is indeed intentional to match Make's behavior.

PR #693, opened 2013-12-13
remove abandoned outputs (mostly done)

fifoforlifo:
This implements "remove abandoned outputs" (RAO) mostly as I'd intend.
Missing are commandline argument to control it, and a merge squash for less messy commit history.
I was thinking RAO could be default enabled, thus a command switch would disable it.

Please also see discussion in https://groups.google.com/forum/#!topic/ninja-build/zQa5-K8XJfE

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,5 @@ TAGS
 # Ninja output
 .ninja_deps
 .ninja_log
+.ninja_regen
+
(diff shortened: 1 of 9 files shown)


Qix-:
+1 on this. This is like the last missing link from coming to Ninja from Tup.

jhasse:
Thanks for the PR and sorry for the long silence!

I think this is a very nice feature (one I miss coming from Tup, too), but I would prefer if this was a separate tool which would be called by the generator instead of a new command line flag. (I also wouldn't mind if it wasn't Ctrl+C safe)

I'm closing this due to the merge conflict and in favor of #1432.

PR #704, opened 2014-01-15
Support dependencies defined with one rule per dependency

dpwright:
ARM's RVCT compiler outputs dependency files with one rule for each dependency, like so:

``` make
foo.o: foo.cpp
foo.o: foo.h
foo.o: bar.h
# ...etc
```

The existing dependency file parser would treat these repeated instances of foo.o as dependencies, and identify this as a circular dependency.

This patch resets the "parsing_targets" flag when we encounter a non-indented line after a non-escaped newline in a rule, and if the later targets match the first one, drops them and continues reading dependencies.  If they don't match (i.e. the dependency file defines dependencies for more than one target), the parser fails as before.

### DIFF
--- a/src/depfile_parser.cc
+++ b/src/depfile_parser.cc
@@ -33,9 +33,11 @@ bool DepfileParser::Parse(string* content, string* err) {
   // in: current parser input point.
   // end: end of input.
   // parsing_targets: whether we are parsing targets or dependencies.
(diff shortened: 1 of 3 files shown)


nico:
Once #1050 is done, depfile parsing speed might be less of an issue.

jhasse:
Fixed by #1039.

PR #723, opened 2014-03-26
Add a new ninja tool: finddeps.

ar1nd4m:
This will find and output the dependency path from one target to
another.

It uses DFS to search for nodes starting from the [from] node, and then
going up the dependency tree until it finds the first path to the [to]
node.

Due to memoization, it's not trivially possible to find all paths, and
also by definition this returns the smallest path.

A sample output when run on the ninja project is:

```
$ ninja -t finddeps ninja src/browse.cc
Path: ninja
    > build/libninja.a
    > build/browse.o
    > src/browse.cc

$
```

I've found this very useful navigating the dependency tree in order to break dependency of one target from another.

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -27,6 +27,11 @@
 #include <unistd.h>
 #endif
 
+#include <algorithm>
+#include <map>
+#include <stack>
+#include <vector>
+
 #include "browse.h"
 #include "build.h"
 #include "build_log.h"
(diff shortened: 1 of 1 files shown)


martine:
By the way, you can also use `-d explain` if you're trying to diagnose why something is being built.

ar1nd4m:
I haven't used ninja for a while now. If similar features have gone in, this is probably outdated. I will let you deal with the patch as you feel best, but do let me know if you need any help from my side.

However, last I checked, `-d explain` option was present, but it was not sufficient for my task. I was trying to see *all* the modules that depend on a particular library, in my attempt to eradicate that library.

martine:
Another idea is that we already have this support for the "query" tool and the graph so it's sorta something that belongs in there.

I have experience with this kind of dependency-breaking ("why does module X pull in module Y?") with Google's build system (now open source as "bazel") and I know it has a pretty sophisticated querying language.  I worry if we provide some functionality now we'll slowly slide into their complex requirements.

Which makes me think: perhaps this kind of logic belongs in the generator program, since it ought to have more knowledge about relationships between parts of the project.

jhasse:
In order to reduce the number of open PRs I'm closing this one, feel free to reopen if you still think this should be merged.

PR #731, opened 2014-04-14
Support building Ninja on Cygwin

rossy:
On Cygwin I get this error, and I have to run configure.py manually:

```
$ ./bootstrap.py
Building ninja manually...
Building ninja using itself...
Usage: configure.py [options]

configure.py: error: option --platform: invalid choice: 'cygwin' (choose from 'linux', 'darwin', 'freebsd', 'openbsd', 'solaris', 'sunos5', 'mingw', 'msvc', 'gnukfreebsd8', 'bitrig')
```

This adds 'cygwin' to the list in `platform_helper` so bootstrapping works for native Cygwin builds.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -99,7 +99,7 @@ def cc(name, **kwargs):
 def cxx(name, **kwargs):
     return n.build(built(name + objext), 'cxx', src(name + '.cc'), **kwargs)
 def binary(name):
-    if platform.is_windows():
+    if platform.is_windows() or platform.is_cygwin():
         exe = name + '.exe'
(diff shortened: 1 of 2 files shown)


martine:
Nico says: the main worry about this is that you end up with a less-awesome Ninja binary.  If we used `posix_spawn` to spawn processes that would help.

I wonder if we made configure require a `--slower-ninja=True` flag to cover this that would be enough.

jhasse:
If I understand it correctly `--platform=linux` should work on Cygwin? If not, please reopen.

PR #738, opened 2014-04-15
Solve the line-wrapping performance problem by shortening lines.

bratell-at-opera:
Avoid joining a list just to split it again.

The main performance problem in the line wrapping came from lines that were
thousands of characters long, and those came from a single location, build.

This patch jumps past _line and the slow code, by emitting the file
list with one file per line. It even becomes easier to read that way.
Everyone is a winner.

This saves about 1 CPU second when running gyp for chromium.

(Not wrapping at all is slightly faster but only by 0.1 second or so).

### DIFF
--- a/misc/ninja_syntax.py
+++ b/misc/ninja_syntax.py
@@ -72,8 +72,10 @@ class Writer(object):
             all_inputs.append('||')
             all_inputs.extend(order_only)
 
-        self._line('build %s: %s' % (' '.join(out_outputs),
-                                     ' '.join([rule] + all_inputs)))
+        self.output.write('build ' +
+                          ' $\n    '.join([""] + out_outputs) +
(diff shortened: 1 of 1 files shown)


martine:
I think we don't want to always line-wrap.

I understand the problem you're having with massive Ninja files, but maybe in that case you could fork `ninja_syntax.py` in your generator to meet your specific needs.  The module mostly is intended as some example code.

bratell-at-opera:
I believe gn are not wrapping lines, and it's certainly not using the ninja_syntax.py code and since gn will eventually replace gyp in the Chromium project (where this mattered) I believe this pull request is not important anymore. 

So I won't object if you close this.

nicolasdespres:
Indeed, I am fine with forking `ninja_syntax.py` in my own project.

PR #743, opened 2014-04-17
Handle SIGTERM properly.

nicolasdespres:
SIGTERM is the default signal sent by most standard program: kill(1), task managers, etc...

It is nice if ninja can handle it properly.

### DIFF
--- a/src/subprocess-posix.cc
+++ b/src/subprocess-posix.cc
@@ -25,20 +25,6 @@
 
 #include "util.h"
 
-namespace {
-
-bool HasPendingInterruption() {
-  sigset_t pending;
-  sigemptyset(&pending);
-  if (sigpending(&pending) == -1) {
-    perror("ninja: sigpending");
-    return false;
-  }
-  return sigismember(&pending, SIGINT);
-}
-
-}  // anonymous namespace
-
 Subprocess::Subprocess(bool use_console) : fd_(-1), pid_(-1),
                                            use_console_(use_console) {
 }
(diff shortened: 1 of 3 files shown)


nico:
What's wrong with how ninja currently handles sigterm?

nicolasdespres:
It currently has no handler for SIGTERM so the process is terminated, the cleanup is not done and sub-processes are not stopped either. The PR handles SIGTERM exactly like SIGINT.

nico:
Ok, so the effect is that if I `killall ninja`, ninja will then forward this to its children and it might quit faster and in a more orderly way, right?

That seems fine to me.

@pcc @tzik , any concerns about this?

pcc:
Seems reasonable.

nico:
Cool, I'll merge this then.

@nicolasdespres Out of interest, in which cases do you run into this in practice?

nicolasdespres:
In practice, I run into this when a the CI server running ninja shutdown or when I stop ninja from OSX Activity Monitor or htop (default is SIGTERM). I admit that if you always run ninja as a foreground process SIGINT is enough.

Thanks for the merge.

martine:
When Emacs kills subprocesses it uses SIGHUP for some reason, which in my experience breaks all sorts of apps.  Thanks for caring about details like this, Nicolas!

nicolasdespres:
Emacs' compile mode emulates a "dumb" terminal. When you close the
compile-mode buffer, it closes the terminal which sends SIGHUP. When you
close an ssh connection or a terminal window SIGHUP is also sent.

I added a SIGHUP handler: https://github.com/martine/ninja/pull/964

On Tue, Apr 28, 2015 at 7:16 PM, Evan Martin <notifications@github.com>
wrote:

> When Emacs kills subprocesses it uses SIGHUP for some reason, which in my
> experience breaks all sorts of apps. Thanks for caring about details like
> this, Nicolas!
>
> —
> <https://github.com/martine/ninja/pull/743#issuecomment-97141620>.
>



-- 
Nicolas Desprès

PR #744, opened 2014-04-17
Suspend sub-processes when control-z is hit.

nicolasdespres:
When control-z is hit the shell sends SIGTSTP to the front process group.
    Generally it suspends all the process tree started by the command. But ninja
    moves its sub-process to their own group so that it can stop them all on
    interruption. This is a great improvement compared to make(1). However in such
    a case sub-processes are not automatically suspended by the shell when you
    hit control-z. Only ninja is and the command it has started keep running.
    This patch addresses this issue by catching SIGTSTP and forwarding it to its
    sub-processes groups. Thus, the entire ninja's process tree is suspended as
    expected.

This pull-request is based on the #743.

### DIFF
--- a/src/subprocess-posix.cc
+++ b/src/subprocess-posix.cc
@@ -87,7 +87,7 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
 
   if (!use_console_) {
     // Put the child in its own process group, so ctrl-c won't reach it.
-    // This may be useful if we want to kill subsets or running subcommands.
+    // This may be useful if we want to kill subsets of running subcommands.
     flags |= POSIX_SPAWN_SETPGROUP;
     // No need to posix_spawnattr_setpgroup(&attr, 0), it's the default.
 


nico:
Can you rebase this on top of trunk now that your SIGTERM change is merged?

nicolasdespres:
Done

nicolasdespres:
@nico I have tested this PR on OSX 10.10.5, Linux 2.6.32-35 and Linux 3.16.0-4-amd64 and it passes. I think the reason why the travis job fails is because it uses Linux 2.6.32-042stab090.5. Unfortunately, I did not succeed to reproduce Travis environment locally to debug it because all the VM images I found on Travis' web site have a 3.12 kernel version. Apparently, our travis jobs uses the legacy infrastructure so maybe that's why I cannot find older image.

nicolasdespres:
If anyone have tried this PR on its setup, I will be glade to have some feedback whether 1) it works 2) the test suite passes. TIA.

nicolasdespres:
@nico I have improved the test robustness and now it passes travis' check.

mgorny:
Bump. @nicolasdespres, could you rebase it, please?

nicolasdespres:
I will try to do it shortly.

nicolasdespres:
@mgorny rebase done. Have fun.

src/subprocess-posix.cc
```diff
@@ -330,9 +351,81 @@ void SubprocessSet::Clear() {
...
+  // Suspend children.
+  for (vector<Subprocess*>::iterator i = running_.begin();
+       i != running_.end(); ++i)
+    if (!(*i)->use_console_)
+      Xkill(-(*i)->pid_, SIGTSTP);
+
+  // Tell users what we have done.
+  {
+    // We use write to bypass the buffer that may not be flush before
+    // we re-raise the SIGTSTP for us.
```
mgorny: Strictly speaking, this does not bypass the buffer for me, and the output happens only after resuming. If you really want it to appear instantly, I suppose you need to explicitly flush it. Note that I haven't tested it though.

src/subprocess-posix.cc
```diff
@@ -330,9 +351,81 @@ void SubprocessSet::Clear() {
...
+    const char* msg = "\nninja: entire build suspended.\n";
+    if (write(STDOUT_FILENO, msg, strlen(msg)) < 0) {
+      // If the write fails, there's nothing we can do.
+      // But this block seems necessary to silence the warning.
```
mgorny: If I recall correctly, the canonical way of silencing the warning is to use:

    (void) write(...);

src/subprocess-posix.cc
nicolasdespres: I have tested on OSX Sierra 10.12.6 and Ubuntu 16.04.3 and the message appears after suspending and before resuming as expected. 

The buffer I am speaking about is the internal buffer that is used by `printf` or `fprintf`. That's why I use `write(1)` directly. I can update the comment if it is not clear.

Here what I have when testing:
```
$ ./ninja -f misc/long-slow-build.ninja
[  0%] SLEEP 5^Z
ninja: entire build suspended.

zsh: suspended  ./ninja -f misc/long-slow-build.ninja
$ fg
[1]  + continued  ./ninja -f misc/long-slow-build.ninja
[ 60%] SLEEP 23^Z
ninja: entire build suspended.

zsh: suspended  ./ninja -f misc/long-slow-build.ninja
$ fg
1]  + continued  ./ninja -f misc/long-slow-build.ninja
[100%] SLEEP 29
```

src/subprocess-posix.cc
nicolasdespres: I think the `if` was more to encapsulate the comment... Removing the `if` entirely does not generate any warnings.

src/subprocess-posix.cc
mgorny: Indeed you are correct. I was testing through Portage, so I suppose Portage's VT handling causes the additional buffering.

src/subprocess_test.cc
```diff
@@ -159,7 +163,81 @@ TEST_F(SubprocessTest, Console) {
...
+  } else { // parent process
+    int status;
+    if (waitpid(pid, &status, 0) < 0)
+      Fatal("waitpid(%d): %s", pid, strerror(errno));
+    if (WIFEXITED(status)) {
+      int exit = WEXITSTATUS(status);
+      // ExitSuccess must be returned by Subprocess::Finish()
+      ASSERT_EQ(0, exit);
+      return; // end of the test
+    } else if (WIFSIGNALED(status)) {
```
jhasse: No need for else after return.

nicolasdespres:
TravisBuild fails on Linux (4.4.0-101-generic). I cannot reproduce it on my Linux box (kernel 4.9.0-8-amd64). Any idea how to have access to the Travis environment without the [Debug Mode](https://docs.travis-ci.com/user/running-build-in-debug-mode/) ?

jhasse:
Have you tried Debug Mode already?

nicolasdespres:
It is not enabled on this repository. I think one of the maintainer should do it. See the link in my previous post.

jhasse:
You can enable Travis CI for your fork on GitHub :)

nicolasdespres:
I do not now how to this and Travis only run on the master repository not mine.

nicolasdespres:
Debug mode enabled. Sorry for the noise.

nicolasdespres:
The build passes in debug mode. If have pushed multiple debug patches to investigate. There is something wrong with SIGTSTP which did not get caught when sent from Ninja... That is very unfortunate. I have double checked the timestamp and run the test a hundred times to make sure there is not a race condition going one. VFORK does not interfere either. Honestly I do not know what to try now. I think I would merge the PR and disable the test on Travis and see if there is a people complaining about this test. If there is we can still revert the commit during the candidate release process. WDYT?

jhasse:
I would prefer if the test passes on Travis CI.

Have you tried reproducing on Ubuntu 14.04?

nicolasdespres:
No. I tried on Debian 9.8 and it succeeds. But I just found a Ubuntu 16.04.6 host where I can reproduce it. I'll try to find some time to investigate.

src/subprocess_test.cc
jonesmz: The return only triggers if `WIFEXITED(status)` evaluates as `true`

The `else if(WIFSIGNALED(status)` clause and the `else` class would evaluate otherwise.

I think though that `return` is unnecessary here, since the `else if` and `else` clause won't run if `WIFEXITED(status)` evaluates to true, and `ASSERT_FALSE("We should have returned before this point");` should be removed.

jonesmz:
Something to consider for a future PR (not this one) is that on newer Linux kernels, the cgroups functionality has support for assigning child processes to a cgroup, which also ensures that all child processes of the child process will be in the same cgroup.

The main benefit for underlying concept of this PR is that cgroups allow all processes inside of a cgroup to be "frozen" simultaneously.

https://www.kernel.org/doc/Documentation/cgroup-v1/freezer-subsystem.txt

I can't speak to non-Linux operating systems having support for something like this, but for Linux, using cgroups (when detected at runtime as available) would likely be a superior way to implement this functionality.

src/subprocess-posix.cc
```diff
@@ -343,9 +364,79 @@ void SubprocessSet::Clear() {
...
+  // the default handler of SIGTSTP we just restored. Thus, we get stopped
+  // right after sigprocmask call.
+  sigset_t tstp_mask;
+  sigset_t old_mask;
+  sigemptyset(&tstp_mask);
+  sigaddset(&tstp_mask, SIGTSTP);
+  if (sigprocmask(SIG_UNBLOCK, &tstp_mask, &old_mask) < 0)
+    Fatal("cannot unblock SIGTSTP: %s", strerror(errno));
+
+  // We are stopped here until we receive SIGCONT.
```
jonesmz: Could you document how you know the program stops at exactly this line?

The documentation for sigprocmask http://man7.org/linux/man-pages/man2/sigprocmask.2.html doesn't give any indication that the program will process signals immediately upon unblocking them.

nicolasdespres: From [signal(7)](http://man7.org/linux/man-pages/man7/signal.7.html)

```A signal may be blocked, which means that it will not be delivered until it is later unblocked.```

Apart from that and/or reading the source code I have no clue about when it is exactly delivered.

jonesmz: Thanks for researching that.

Mind adding this information straight into the code?

lersek: > The documentation for sigprocmask http://man7.org/linux/man-pages/man2/sigprocmask.2.html doesn't give any indication that the program will process signals immediately upon unblocking them.

That's an omission in the manual page then; see [POSIX](https://pubs.opengroup.org/onlinepubs/9799919799/functions/sigprocmask.html):
> If the argument _set_ is not a null pointer, after _pthread_sigmask()_ changes the currently blocked set of signals it shall determine whether there are any pending unblocked signals; if there are any, then at least one of those signals shall be delivered before the call to _pthread_sigmask()_ returns. [...] The _sigprocmask()_ function shall be equivalent to _pthread_sigmask()_, except that its behavior is unspecified if called from a multi-threaded process, and on error it returns `-1` and sets _errno_ to the error number instead of returning the error number directly.

PR #749, opened 2014-04-27
Show build time

AoD314:
Show build time:

``` bash
$ ./ninja --buildtime
ninja: no work to do.
build time: 00:00:00.000 (0 ms)

$ ./ninja --buildtime
[23/23] LINK ninja
build time: 00:00:00.113 (113 ms)
```


Trass3r:
Why was this dropped?

PR #751, opened 2014-04-29
escape_path performance.

bratell-at-opera:
escape_path is fast, about 1 microsecond per call on my computer,
but some have systems where it's called hundreds of thousands of times
(chromium) or even more (whatever nicolasdespres is doing).

For input which do not need escaping this is three times as fast in my
profiling. For chromium 48 out of 371000 paths needed escaping so
it was a 0.25 second improvement in profiling.

Note: The any() function made it slower or I would have used that.

### DIFF
--- a/misc/ninja_syntax.py
+++ b/misc/ninja_syntax.py
@@ -11,7 +11,9 @@ import textwrap
 import re
 
 def escape_path(word):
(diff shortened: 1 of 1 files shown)


martine:
@nico this seems fine to me, what do you think?

nico:
I think it's a relatively minor win and it duplicates the strings. The 0.25s win are on top of a > 60s base. So I think we don't want this at this point.

PR #763, opened 2014-05-08
Allow paths with international characters in depfiles

drbo:
While C does not allow non-ASCII characters in #include filenames, other languages may have dependencies that work in a similar way and that do allow such characters. I believe that D and Python are two examples.

This patch treats characters greater than 127 as text, rather than as whitespace, in depfiles. A test is also added to verify that these characters are handled correctly.


martine:
@nico this looks good to me, what do you think?

nico:
Yup, thanks! Sorry for the delay.

PR #771, opened 2014-05-31
Replacing '$ ' with '$$ ' breaks the case where the original '$' was

drbo:
Replacing '$ ' with '$$ ' breaks in the case where the original '$' was escaped, by adding an additional '$'.


martine:
I think without a specific use case (like a failing test) I'd prefer to just leave this alone.

PR #816, opened 2014-09-18
add an environment variable to set the default number of jobs used in parallel

glehmann:


martine:
A general way to handle this and some other similar requests might be to add a NINJA_DEFAULT_FLAGS env var that is interpreted as command line flags.  But we'd then probably need some way to *undo* those values (e.g. a way to opt back in to the default `-j` behavior when there was a `-j` set in the environment).

martine:
I think even if we want this feature (some way to control -j, see the bug), we won't want this implementation of it.  So I'm closing this and leaving the bug open.

PR #827, opened 2014-10-03
Allow the user to specify what shell to use for commands

sorbits:
This is done by assigning to the ‘shell’ variable, e.g.:

```
shell = /bin/bash
```

The commit uses the GetBinding function to obtain the value of the shell variable, but ideally it should obtain the value without going through shell escaping.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -495,7 +495,7 @@ bool RealCommandRunner::CanRunMore() {
 
 bool RealCommandRunner::StartCommand(Edge* edge) {
   string command = edge->EvaluateCommand();
-  Subprocess* subproc = subprocs_.Add(command, edge->use_console());
+  Subprocess* subproc = subprocs_.Add(command, edge->use_console(), edge->GetBinding("shell"));
   if (!subproc)
     return false;
   subproc_to_edge_.insert(make_pair(subproc, edge));
--- a/src/subprocess-posix.cc
+++ b/src/subprocess-posix.cc
@@ -36,7 +36,7 @@ Subprocess::~Subprocess() {
     Finish();
 }
 
-bool Subprocess::Start(SubprocessSet* set, const string& command) {
+bool Subprocess::Start(SubprocessSet* set, const string& command, const string& shell) {
   int output_pipe[2];
   if (pipe(output_pipe) < 0)
     Fatal("pipe: %s", strerror(errno));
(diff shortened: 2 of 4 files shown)


itensionanders:
Funny, didn't find this thread/PR until I had submitted by own (#1058)

itensionanders:
I find this feature useful to use a _less_ featureful shell (`/bin/dash`).
But maybe that is just me... It's quicker than `/bin/bash`, though ?

itensionanders:
@sorbits: what do you think about my variant, which also adds "shellflags" (89f4da8225afbd5415318dc422425be7f9048ef8) ?
It seems a common use case for setting Make's .SHELLFLAGS was to use `-ec` (errexit)

sorbits:
> @sorbits: what do you think about my variant, which also adds "shellflags" (89f4da8) ?
> It seems a common use case for setting Make's .SHELLFLAGS was to use -ec (errexit)

Adding `shellflags` does sound like a useful thing to include. Would be great if @nico will review his opinion on this feature, as I am still running my custom fork, and very happy not having to use a “lowest common denominator” shell when I am writing build files.

nicolasdespres:
Like @nico mentioned, it probably adds to much flexibility to allow to set the `shell` (or `shellflags`) variable at the level of a rule / build statement. On the other hand, generators would be a nightmare to maintain if they have to deal with quote escaping because they invoke bash in all their rules. Such generated ninja files would be hard to understand too. The autotools are a good example of such a quote escaping nightmare.

I think a middle ground would be to accept `shell` (and `shellflags`) as top-level variables like `build_dir` and `ninja_required_version`.

sorbits:
I didn’t realize @nico was specifically talking about the ability to 
set the variable in the edge nodes.

Only being able to set the shell variable at the root level would 
certainly be enough to satisfy my needs.

@nico: Would a PR for such feature be acceptable?

That said, I could probably imagine some use-cases for changing shell at 
the edge-level, but it would mainly be for debug purposes or playing 
around. Seems to me that given the code complexity isn’t lower by only 
allowing the variable at the top-level, limiting it in this way makes 
little sense.

jhasse:
> instead there should probably be a way to have raw, unshelled commands (like on Windows, due to command-line length concerns I believe).

I think this would be the way to go. You could call `$shell` in your rule to mimic the behavior of this PR.

Feel free to reopen if I missed something.

PR #883, opened 2014-12-12
configure.py: enable separate build directories

sbc100:
With this change it is now possible to run configure.py
from a separate directory and have build.ninja and all
build projects written there rather than within the source
checkout. This works in a similar way to autoconf-generated
configure scripts.

This allow multiple different build trees to exists
based on a single ninja checkout.

This change was useful when building ninja for NaCl where
we have multiple toolchains that we use to build multiple variants
of the ninja binary.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -28,7 +28,9 @@ import string
 import subprocess
 import sys
 
-sys.path.insert(0, 'misc')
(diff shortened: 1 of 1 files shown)


nico:
Someone else sent the same patch at #971. Since several people want this (you and #971), I merged this.

PR #887, opened 2014-12-31
Fix compilation errors on Visual Studio 2015 (_MSC_VER 1900)

berenm:
Latest version of ``cl.exe`` complains about ignored typedef, hidden "path" local variable and hash_map being deprecated.

With C++11-compliant compilers (and MSC >= 1900), it is possible to use the standard unordered_map container instead of vendor-specific structures.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -284,6 +284,8 @@ if platform.is_msvc():
               '/wd4512', '/wd4800', '/wd4702', '/wd4819',
               # Disable warnings about passing "this" during initialization.
               '/wd4355',
+              # Disable warnings about ignored typedef in DbgHelp.h
+              '/wd4091',
               '/GR-',  # Disable RTTI.
               # Disable size_t -> int truncation warning.
               # We never have strings or arrays larger than 2**31.
--- a/src/hash_map.h
+++ b/src/hash_map.h
@@ -50,7 +50,22 @@ unsigned int MurmurHash2(const void* key, size_t len) {
   return h;
 }
 
-#ifdef _MSC_VER
+#if (__cplusplus >= 201103L) || (_MSC_VER >= 1900)
+#include <unordered_map>
+
+namespace std {
+template<>
+struct hash<StringPiece> {
+  typedef StringPiece argument_type;
+  typedef std::size_t result_type;
+
+  result_type operator()(argument_type const& s) const {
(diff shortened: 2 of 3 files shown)


configure.py
```diff
@@ -284,6 +284,8 @@ def binary(name):
               '/wd4512', '/wd4800', '/wd4702', '/wd4819',
               # Disable warnings about passing "this" during initialization.
               '/wd4355',
+              # Disable warnings about ignored typedef in DbgHelp.h
+              '/wd4091',
```
nico: (Mostly note to self: This was reported upstream at https://connect.microsoft.com/VisualStudio/feedbackdetail/view/888527/warnings-on-dbghelp-h )

nico:
Thanks!

src/hash_map.h
```diff
@@ -50,7 +50,22 @@ unsigned int MurmurHash2(const void* key, size_t len) {
...
 
-#ifdef _MSC_VER
+#if (__cplusplus >= 201103L) || (_MSC_VER >= 1900)
+#include <unordered_map>
+
+namespace std {
+template<>
+struct hash<StringPiece> {
+  typedef StringPiece argument_type;
+  typedef std::size_t result_type;
```
nico: size_t isn't in `std`, is it? I'll change this.

berenm: It should be, but it's also in the global namespace.

nico: Ah, it's in std from cstddef but it's only available globally in stddef.h (like with the other cfoo headers). Ninja uses size_t in an unqualified way everywhere else. Thanks :-)

nico: Another question: I was looking at the C++ standard, hoping that result_type and argument_type aren't necessary (libc++ doesn't need them, but it looks like the standard asks for them :-/), and it says "the instantiation hash<Key> shall … satisfy the Hash requirements (17.6.3.4), with Key as the function call argument type". Key is StringPiece here, so should the parameter be per value instead of per const reference? StringPiece is small and can be passed by value efficiently, so that's not a concern.

berenm: I think it's OK for the ``operator()`` to take the parameter by ``const&``. I basically copy-pasted the custom ``std::hash`` from http://en.cppreference.com/w/cpp/utility/hash.

PR #894, opened 2015-01-10
Cleanup: Fix 'hasIdent' variable name/style.

tfarina:
Seems more correct to name it has_indent_token and to use the
unix_hacker style.

Signed-off-by: Thiago Farina <tfarina@chromium.org>


nico:
Thanks! Optional, tiny nits: remove "_token" suffix (it's a local var), remove "Signed-off-by" line. I'll merge this tonight.

PR #908, opened 2015-01-27
Allow manifest rebuild to loop up to 100 times

colincross:
Ninja generators that bootstrap themselves with Ninja may need
to rebuild build.ninja multiple times.  Replace the 2 cycle loop
with a 100 cycle loop, and print the pass number each time it
restarts.

As discussed in https://groups.google.com/forum/#!searchin/ninja-build/jamie/ninja-build/YHOpOIS3FSo/ProjUh-DUxkJ

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -1061,9 +1061,9 @@ int real_main(int argc, char** argv) {
     return (ninja.*options.tool->func)(argc, argv);
   }
 
-  // The build can take up to 2 passes: one to rebuild the manifest, then
-  // another to build the desired target.
-  for (int cycle = 0; cycle < 2; ++cycle) {
+  // Limit number of rebuilds, to prevent infinite loops.
+  const int kCycleLimit = 100;
+  for (int cycle = 1; cycle <= kCycleLimit; ++cycle) {
     NinjaMain ninja(ninja_command, config);
 
     RealFileReader file_reader;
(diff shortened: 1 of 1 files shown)


src/ninja.cc
```diff
@@ -1061,9 +1061,11 @@ int real_main(int argc, char** argv) {
     return (ninja.*options.tool->func)(argc, argv);
   }
 
-  // The build can take up to 2 passes: one to rebuild the manifest, then
-  // another to build the desired target.
-  for (int cycle = 0; cycle < 2; ++cycle) {
+  // The build can take up to 100 passes to rebuild the manifest and then to
```
nico: `// Limit number of rebuilds, to prevent infinite loops.`

src/ninja.cc
```diff
@@ -1061,9 +1061,11 @@ int real_main(int argc, char** argv) {
...
+  // build the desired target, far more times than expected in any normal
+  // use case, but will prevent infinite loops.
+  const int numPasses = 100;
```
nico: `kCycleLimit`

src/ninja.cc
```diff
@@ -1086,16 +1088,17 @@ int real_main(int argc, char** argv) {
...
-        // Start the build over with the new manifest.
-        continue;
-      } else if (!err.empty()) {
-        Error("rebuilding '%s': %s", options.input_file, err.c_str());
-        return 1;
+    // Attempt to rebuild the manifest before building anything else
+    if (ninja.RebuildManifest(options.input_file, &err)) {
+      // Start the build over with the new manifest.
+      if (cycle < numPasses) {
+        printf("%s updated, restarting ninja for pass %d\n",
```
nico: Remove ("Developers should design programs so that they do not print unnecessary output" etc)

src/ninja.cc
```diff
@@ -1086,16 +1088,17 @@ int real_main(int argc, char** argv) {
...
+               options.input_file, cycle + 1);
       }
+      continue;
+    } else if (!err.empty()) {
+      Error("rebuilding '%s': %s", options.input_file, err.c_str());
```
nico: if it's important in your experience, log the cycle here (`"rebuildling '%s' (cycle %d): …`)

nico:
Thanks! Couple nits, else looks good.

colincross:
Done

PR #910, opened 2015-01-29
POSIX: detach background subprocesses from terminal.

jlnt:
Put background subprocesses (i.e. subprocesses with no access
to the console) in their own session and detach them from the
terminal.

This fixes martine/ninja#909.

### DIFF
--- a/src/subprocess-posix.cc
+++ b/src/subprocess-posix.cc
@@ -80,8 +80,11 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
         break;
 
       if (!use_console_) {
-        // Put the child in its own process group, so ctrl-c won't reach it.
-        if (setpgid(0, 0) < 0)
+        // Put the child in its own session and process group. It will be
+        // detached from the current terminal and ctrl-c won't reach it.
+        // Since this process was just forked, it is not a process group leader
+        // and setsid() will succeed.
+        if (setsid() < 0)
           break;
 
         // Open /dev/null over stdin.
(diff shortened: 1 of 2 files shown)


nico:
Thanks for the patch. I tested that this doesn't have a measurable perf impact on either Linux or OS X (maybe allocating sessions are expensive somehow, but apparently not).

nico:
Oh, and sorry about the delay!

jlnt:
Thanks for the merge and the thorough testing!

nico:
This makes it impossible to move ninja from fork/exec to posix_spawn(), which is faster on some platforms (notably, OS X – but probably also the BSDs, even though I haven't measured it, and in posix-on-win layers). Using posix_spawn() for subprocesses seems like a very nice change, but posix_spawn() doesn't support setsid().

Is there any way to do this explicitly in the few cases that need it? Are there any programs that run an arbitrary command in a new session?

nicolasdespres:
I think you might be interested by rereading this [thread](https://groups.google.com/forum/#!topic/ninja-build/Uq6pTV3FC2Y).

I think we could consider the benefit of not to putting child processes in their own session:
* I have never felt the need for Ninja to stop a set of sub-commands and keep running afterward as @pcc mentioned in the thread.
* The last time I checked, make(1) does not move its child processes in their own session.
* Users who really need their own session can write a wrapper.
* Signal handling would be much simpler (removing the need for PR #744).
* We can still remove partially created output files on interruption.

If you want to improve the sub-process spawning task you might also be interested by vfork(2).

PR #911, opened 2015-02-01
Use a monotonic clock for measuring time.

anders:
gettimeofday() isn't accurate for measuring time because it can jump backwards/forwards.

I link to librt on Linux, I think it's only necessary on Linux. The FreeBSD man page says it's available in libc.

Tested on Linux and OS X.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -481,6 +481,9 @@ if platform.is_msvc():
 else:
     libs.append('-lninja')
 
+if platform.is_linux():
+    libs.append('-lrt')
+
 all_targets = []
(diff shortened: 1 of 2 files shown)


martine:
Out of curiosity, did you encounter this because you saw the problem or were you just looking at the code's use of time functions?

anders:
I spotted the problem while randomly reading the sources.

martine:
This is only used in the metrics feature (which is for debugging purposes) and it's more code, so I think we don't want this.

Issue #486, opened 2013-01-15
Use SetConsoleTextAttribute on Windows to simulate color


evmar:
Rather than stripping ANSI codes, we could implement them.  Not sure if that serves a user need but it would at least be a cool hack.

Suggested by TheOneRing on IRC.

nico:
Clang can output ansi on windows by passing `-fcolor-diagnostics -fansi-escape-codes`.

Currently, ninja strips them in cygwin (where they would work) because that terminal isn't detected as smart, and ninja prints them in cmd.exe (where they don't work – it'd be cool if ninja could interpret them and make them do the right thing).

sgraham:
I have ansicon installed in cmd, and I noticed after I did that some tools started magically outputting color. Maybe there's some easy way to probe for that.

sgraham:
Hmm, maybe not, https://github.com/ggreer/the_silver_searcher/blob/master/src/options.c just does getenv("ANSICON"). The other one was 'less', didn't look at its code.

nico:
`-fcolor-diagnostics -fansi-escape-codes` + ansicon already works today. However, I'd guess that most people don't use ansicon, and I feel it'd be nice if things just worked without requiring everyone to get ansicon (?)

sgraham:
Yes that would be nice, I don't think ansicon is very popular (and I've found it crashes cmd occasionally too).

Interpreting ANSI and outputting using console functions sgtm if you think that belongs in ninja.

jhasse:
Now that Windows supports ANSI codes natively, I'm closing this in favor of #1462.

PR #912, opened 2015-02-02
Zero width format

### LINKED ISSUE #713: Support color in NINJA_STATUS

mathstuf:
I decided to not do `%0{` and `%}` because the logic for control sequences more than just a single character is a messier. I can also change it over to `%{` and `%}` if separate open/close sequences are preferable. It is not (currently) an error to have mismatched open/close sequences, but it is also not a zero-width section and `%0` will be literal in that case.

---

Fixes #713.

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -207,6 +207,9 @@ Several placeholders are available:
 specified by `-j` or its default)
 `%e`:: Elapsed time in seconds.  _(Available since Ninja 1.2.)_
 `%%`:: A plain `%` character.
+`%0`:: Marks the start and end of a zero-width region (useful to exclude the
+width of color control codes from the elision width computation). _(Available
+since Ninja 1.6.)_
 
 The default progress status is `"[%f/%t] "` (note the trailing space
 to separate from the build rule). Another example of possible progress status
--- a/src/build.cc
+++ b/src/build.cc
@@ -245,6 +245,11 @@ string BuildStatus::FormatProgressStatus(
         break;
       }
 
+        // Zero-width marker (ignore).
+      case '0':
+        out += "%0";
+        break;
+
       default:
         Fatal("unknown placeholder '%%%c' in $NINJA_STATUS", *s);
         return "";
--- a/src/line_printer.cc
+++ b/src/line_printer.cc
@@ -92,7 +92,7 @@ void LinePrinter::Print(string to_print, LineType type) {
 
     have_blank_line_ = false;
   } else {
-    printf("%s\n", to_print.c_str());
+    printf("%s\n", RemoveZeroWidthMarkers(to_print).c_str());
   }
 }
 
--- a/src/util.cc
+++ b/src/util.cc
@@ -597,16 +597,84 @@ double GetLoadAverage() {
 }
 #endif // _WIN32
 
+static string RemoveZeroWidthMarkers(const string& str) {
+  string result;
+  size_t last_open = 0;
+  size_t open = 0;
+  while ((open = str.find("%0", last_open)) != string::npos) {
+    size_t zero_width_start = open + 2;
+    size_t close = str.find("%0", zero_width_start);
+    if (close != string::npos) {
+      result += str.substr(last_open, open - last_open)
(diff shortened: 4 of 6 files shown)


mathstuf:
Ping? @martine? @nico?

martine:
I'm not so familiar with this corner of terminal IO.  What is the normal syntax for zero-width markers in other similar languages?  (bash/zsh?)

mathstuf:
zsh uses `%0{` and `%}` which is messy due to the two-character escape. bash seems to use `\[` and `\]`.

nico:
The motivation here is to let eliding be correct if NINJA_STATUS contains ansi codes, right? Maybe we could have the eliding function not count ansi codes when computing widths?

mathstuf:
I could for see using it to change tmux's title using its escape codes, but right now I'd be fine with color. Does the color stripping understand 256color sequences?

nico:
Are they special? Currently it does this: https://github.com/martine/ninja/blob/master/src/util.cc#L477

mathstuf:
Here's the command which does the work of setting the title for me:

```sh
printf "\033]2;%s\033\\" "$title"
```

nico:
I meant the 256color sequences.

(Do I understand you correctly that you want your ninja status to modify your tmux title? Why?)

mathstuf:
Ah, sorry, missed the context. Well, I don't personally, but this would also support that seamlessly (possibly counting down the targets?). Here's my 256-color code:

```zsh
for color in {0..255}; do
    ZSH_COLOR[$color]=$'\033[38;5;'$color'm'
    ZSH_COLOR_PS1[$color]="%0{${ZSH_COLOR[$color]}%}"
done
```

It looks like the existing code should work.

martine:
@nico modifying your terminal title is commonly a hack for long-running processes to display status in a more visible place like alt-tab lists.  Some WMs even let you "shade" a window (collapsing it to its title bar) as a substitute for minimization.

nico:
Right, but is this something one would want to do _every time_ a ninja command updates? (I honestly don't know.)

If we end up doing #486 (which I currently do want to do, to get colors from clang diagnostics on Windows), ninja will have some knowledge about ansi codes anyhow, so it doesn't seem unreasonable for Elide() to learn about them too maybe.

Are there other use cases for zero-width markers other than ansi codes?

martine:
From a pedantic perspective it's hard to know in advance the full
possibility space of escapes, see e.g.
http://tldp.org/HOWTO/Xterm-Title-6.html

(PS: there was actually a vulnerability in one of these in the terminal
escape parser -- it allowed you to e.g. hack someone by sending them a
specially formed string via IRC if they were running their client in a
shell!)

However it might be the case that everything these days uses ANSI-style
escapes.

mathstuf:
`%0` is output literally with `-v`, so that path needs to strip things as well.

mathstuf:
Fixed the `-v` flag issue.

jhasse:
https://github.com/ninja-build/ninja/pull/1454/files#diff-005d118f141929270b3e1c4150887ae2R573 implements the stripping for `ElideMiddle`. I think we need to adjust this, so that it falls back to `result.size()` when there aren't any escapes sequences in `NINJA_STATUS`. This way there wouldn't be a slow down for users who aren't using colors in `NINJA_STATUS` (besides the one time parsing at the start, which should be negligible).

Also #1210 might be the better solution, to keep Ninja simple.

mathstuf:
Working on a patch now. It does ANSI detection by "spans" so that there's only a single `string.find` for escape-less sequences.

mathstuf:
Also, the linked patch isn't right because it detects elision properly, but doesn't split the string using the escape sequence information to not slice in the middle of one.

jhasse:
Yeah, getting this right can be tricky, as you also don't want to skip any escape sequences even when they are in the elided area. I've already implemented it in Python here: https://github.com/jhasse/ja/blob/6603076e84083ed9155404268336b5ce5a771bc4/ja/native.py#L244

mathstuf:
Closing; this is handled elsewhere (see #1491 and #1494).

PR #913, opened 2015-02-04
Turn ninja_syntax.py into a PyPI package

Maratyszcza:
The patch adds installation script and metadata to make a PyPI package out of `ninja_syntax.py` script.

Packaging `ninja_syntax.py` solves the problems discussed on `#ninja-build` mailing list [here](https://groups.google.com/forum/#!topic/ninja-build/PceLYvPJpKY).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,8 @@
 TAGS
 /build
 /build.ninja
+/misc/MANIFEST
+/misc/dist/
 /ninja
 /build_log_perftest
 /canon_perftest
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ TAGS
 /build
 /build.ninja
 /misc/MANIFEST
+/misc/version.py
 /misc/dist/
 /ninja
 /build_log_perftest
--- a/RELEASING
+++ b/RELEASING
@@ -7,13 +7,12 @@ Push new release branch:
 3. git checkout release; git merge master
 4. fix version number in src/version.cc (it will likely conflict in the above)
 5. fix version in doc/manual.asciidoc
-6. fix version number in misc/setup.py
-7. commit, tag, push (don't forget to push --tags)
+6. commit, tag, push (don't forget to push --tags)
        git commit -a -m v1.5.0; git push origin release
        git tag v1.5.0; git push --tags
        # Push the 1.5.0.git change on master too:
        git checkout master; git push origin master
-8. construct release notes from prior notes
+7. construct release notes from prior notes
    credits: git shortlog -s --no-merges REV..
 
 Release on github:
--- a/RELEASING
+++ b/RELEASING
@@ -7,17 +7,23 @@ Push new release branch:
 3. git checkout release; git merge master
 4. fix version number in src/version.cc (it will likely conflict in the above)
 5. fix version in doc/manual.asciidoc
-6. commit, tag, push (don't forget to push --tags)
+6. fix version number in misc/setup.py
+7. commit, tag, push (don't forget to push --tags)
        git commit -a -m v1.5.0; git push origin release
        git tag v1.5.0; git push --tags
        # Push the 1.5.0.git change on master too:
        git checkout master; git push origin master
-7. construct release notes from prior notes
+8. construct release notes from prior notes
    credits: git shortlog -s --no-merges REV..
 
 Release on github:
 1. https://github.com/blog/1547-release-your-software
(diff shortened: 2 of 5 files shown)


nico:
Thanks for the patch! Do you think it's possible to keep the pypi package config outside of ninja's tree? As far as I understand, there's several python package archives, and ninja_syntax.py doesn't change all that often, so manually pushing a new version some time after a release might be good enough.

(My motivation for asking this is that I'd selfishly prefer to not add more steps to RELEASING :-) ).

Maratyszcza:
Well, I see at least two reasons to keep Python packaging files in Ninja tree:

1. If we keep the Python package separate from Ninja, there is good chance that it will be lost or abandoned.
2. If the Python package is maintained in a separate repository, it would need a copy of `ninja_syntax.py`. This copy would from time to time get out-of-sync with Ninja's version, and confuse users.

Also, in the future it might be worth to distribute Ninja binaries with the Python package: unlike `apt` and `rpm` packages, PyPI packages can be installed locally without root access.

nico:
I wouldn't worry too much about it being lost or abandoned. Either the pip package is useful to some folks, then someone will be willing to keep it updated, or it's not useful to anyone and then it doesn't matter too much if it becomes out of date.

The config file readme could just say "to push a new release, copy ninja_syntax.py from the ninja release tag, and run `some_command`", it wouldn't have to check in the ninja_syntax file.

Maratyszcza:
What if I make the Python package parse the version from `src/version.cc`?

nico:
I mostly don't want to add more steps to the releasing document. (If it's easy to not duplicate the version number, that'd be good too of course.)

kirbyfan64:
Why not just use a separate tree with a Git submodule that points to Ninja?

The obvious issue to me is whose name it'll be under on PyPy.

Maratyszcza:
@kirbyfan64 The module's `__init__.py` should be in the same directory as `ninja_syntax.py`

kirbyfan64:
@Maratyszcza Which is where shutil comes in. You just copy the file to the current directory before `setup` gets called.

Anyway...why exactly does `setup.py` reference a nonexistent `ninja` package?

Maratyszcza:
@kirbyfan64 `setup.py` installs `ninja_syntax.py` + `__init__.py` as `ninja` package

kirbyfan64:
@Maratyszcza That's the thing; if there's nothing in `__init__.py`, why not just make it a standalone module?

Maratyszcza:
@kirbyfan64 What do you mean by standalone module?

kirbyfan64:
@Maratyszcza Isn't `ninja_syntax.py`the only file that's going to be distributed?

Maratyszcza:
@kirbyfan64 It is, but PyPI package also needs `setup.py`

kirbyfan64:
@Maratyszcza Yeah...but isn't `__init__.py` going to be empty? Why not just upload `ninja_syntax.py` without putting it in a package?

Maratyszcza:
@kirbyfan64 `setup.py` would install it into directory `ninja` (same as package name). However, Python wouldn't be able to import `ninja_syntax.py` from there because without `__init__.py` it won't recognize `ninja` as a Python package.

kirbyfan64:
Yeah...can't the package be omitted and ninja_syntax.py be uploaded to PyPI by itself?

Maratyszcza:
I don't know for sure, but I didn't see any single-file PyPI package

kirbyfan64:
@Maratyszcza I've made plenty. PyExt, clamav-python, etc.

Maratyszcza:
@kirbyfan64 Thanks, I will take a look

Maratyszcza:
@nico PTAL. I updated the package to automatically parse version from `src/version.cc` and install only `ninja_syntax.py`. Also, now the package is called `ninja-syntax` (Python module is imported with `import ninja_syntax`).

nico:
I think the conclusion here was to keep pypi configs out of tree if someone wants to upload this to python package managers. The script is also small enough that it can just be copied around.

remexre:
To what degree is this open to revisiting now that releases are done through GitHub workflows (or at least it [looks like they are](https://github.com/ninja-build/ninja/blob/8a3af294fd278a7ed378b36db677c46716583ff3/.github/workflows/linux.yml#L63); RELEASING seems [not to have been updated for this](https://github.com/ninja-build/ninja/blob/a37da20ae74c81703b1c811182fc154d95ed46fe/RELEASING#L21) though)?

PR #917, opened 2015-02-28
subprocess_test: gracefully handle rlim.rlim_cur < kNumProcs

pinotree:
Instead of expecting that the number of open files is well above `kNumProcs`, simply "skip" the test in that case, still printing the message about the test limit (adding the current system limit too).

Theoretically, the `SubprocessTest.SetWithLots` test could be re-enabled on Mac OS X (whose soft file limit is less than 1025, according to the comment), but I don't have a Mac OS X system to test.

### DIFF
--- a/src/subprocess_test.cc
+++ b/src/subprocess_test.cc
@@ -177,8 +177,8 @@ TEST_F(SubprocessTest, SetWithLots) {
   // Make sure [ulimit -n] isn't going to stop us from working.
   rlimit rlim;
   ASSERT_EQ(0, getrlimit(RLIMIT_NOFILE, &rlim));
-  if (!EXPECT_GT(rlim.rlim_cur, kNumProcs)) {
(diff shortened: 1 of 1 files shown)


pinotree:
Ah I see you CI passes `--gtest_filter=-SubprocessTest.SetWithLots` to `ninja_test`, which I guess could be dropped once this patch is in...

martine:
You can try modifying .travis.yml to remove that flag in this branch (I'm not sure if it'll work!).

nico:
Current, running tests doesn't produce any output (except on Windows), which I think is a nice property. So I wouldn't reenable the test on OS X, and also wouldn't change .travis.yml.

(But this change looks fine to me as-is, I can merge it as-is if you want.)

pinotree:
> (But this change looks fine to me as-is, I can merge it as-is if you want.)

Yes please.

nico:
Done, thanks for the patch.

PR #918, opened 2015-02-28
configure: recognize Hurd as platform

pinotree:
Add the 'gnu' platform string, matching the 'gnuN' Python platform which is used on Hurd.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -27,6 +27,7 @@ import pipes
 import string
 import subprocess
 import sys
+import re
 
 sys.path.insert(0, 'misc')
 import ninja_syntax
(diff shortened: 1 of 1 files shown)


martine:
Is this necessary to make things build?  The reason to put platforms into this script is so we can test for them elsewhere in the script and vary our build behavior.

nico:
Sounds like things should build fine as-is on hurd.

pinotree:
Sorry, apparently I didn't get the first reply.

> Is this necessary to make things build?

No, it is not necessary, ninja builds and has tests passing on Hurd. On the other hand ...

> The reason to put platforms into this script is so we can test for them elsewhere in the script and vary our build behavior.

... exactly. For now there's no Hurd-specific tweaking required, but having the platform string for it would be nicer, so can be used later if needed.

nico:
It can be added once it's needed. Else, the file will end up listing all platforms over time, instead of just the platforms that are difficult to deal with :-)

PR #921, opened 2015-03-01
Allow scoping rules through subninja

mohamed:
Ninja didn't support scoping rules through subninja and assumed
a unique rule name in the whole namespace. With this change, this
behavior is changed to allow scoping rules. Two rules can have the
same name if they belong to two different scopes. However, two
rules can NOT have the same name in the same scope.

### DIFF
--- a/src/manifest_parser_test.cc
+++ b/src/manifest_parser_test.cc
@@ -60,8 +60,8 @@ TEST_F(ParserTest, Rules) {
 "\n"
 "build result: cat in_1.cc in-2.O\n"));
 
-  ASSERT_EQ(3u, state.rules_.size());
-  const Rule* rule = state.rules_.begin()->second;
+  ASSERT_EQ(3u, state.bindings_.GetRules().size());
+  const Rule* rule = state.bindings_.GetRules().begin()->second;
   EXPECT_EQ("cat", rule->name());
   EXPECT_EQ("[cat ][$in][ > ][$out]",
             rule->GetBinding("command")->Serialize());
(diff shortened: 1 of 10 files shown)


nico:
Let's do this. I'm not super happy that Rule is moving into EvalEnv, but I can try to move it around later if I manage to figure out a better way of doing this. (Chances are I won't either.) Also, this needed some minor changes to build with libc++ for me, see https://github.com/nico/ninja/commit/f6461f06abdb0a1160f99da2e3e4351c915e3869 . I'll apply that myself after merging.

Thanks for the patch!

src/manifest_parser_test.cc
```diff
@@ -823,18 +823,27 @@ TEST_F(ParserTest, MissingSubNinja) {
...
+  // Test that rules are scoped to subninjas even with includes.
+  files_["rules.ninja"] = "rule cat\n"
+                         "  command = cat\n";
+  files_["test.ninja"] = "include rules.ninja\n"
+                         "build x : cat\n";
+  ManifestParser parser(&state, this);
+  string err;
+  EXPECT_TRUE(parser.ParseTest("include rules.ninja\n"
+                                "subninja test.ninja\n"
+                                "build y : cat\n", &err));
```
nico: This triggers an assert in debug builds (`./configure --debug`):

    [166/198] ParserTest.DuplicateRuleInDifferentSubninjasAssertion failed: (LookupRule(rule->name()) == NULL),   
    function AddRule, file src/eval_env.cc, line 33.

I guess AddRule needs to assert with LookupRuleCurrentScope now?

mohamed: Yes. Sorry for that!

PR #923, opened 2015-03-07
Directly pass the string instead of char * to Truncate util function. It...

HeisSpiter:
... will prevent useless conversions.


nico:
Thanks!

PR #925, opened 2015-03-09
Test for Clang by checking --version

kirbyfan64:
This is for platforms where `clang++` is aliased as `c++`, such as FreeBSD and OSX.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -317,8 +317,12 @@ else:
         cflags.remove('-fno-rtti')  # Needed for above pedanticness.
     else:
         cflags += ['-O2', '-DNDEBUG']
-    if 'clang' in os.path.basename(CXX):
-        cflags += ['-fcolor-diagnostics']
+    try:
+        proc = subprocess.Popen([CXX, '--version'], stdout=subprocess.PIPE)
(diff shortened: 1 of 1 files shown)


configure.py
```diff
@@ -317,8 +317,12 @@ def binary(name):
         cflags.remove('-fno-rtti')  # Needed for above pedanticness.
     else:
         cflags += ['-O2', '-DNDEBUG']
-    if 'clang' in os.path.basename(CXX):
-        cflags += ['-fcolor-diagnostics']
+    try:
+        proc = subprocess.Popen([CXX, '--version'], stdout=subprocess.PIPE)
+        if 'clang' in proc.communicate()[0].decode('utf-8'):
+            cflags += ['-fcolor-diagnostics']
```
nico: I wonder if we should spell this `-fdiagnostics-color` (clang understands both spellings) and also add it for gcc 4.9+. Probably good enough to do that once someone submits a patch for gcc.

nico:
Thanks!

nico:
I tweaked this a bit in #927, it should now work with both clang and newer gccs.

PR #928, opened 2015-03-10
Remove option no longer needed now that we don't use gtest.

nico:


nico:
@sgraham This was here for gtest, right?

sgraham:
Yeah, I don't think the ninja code itself uses any crazy template
instantiations, so lgtm.

PR #929, opened 2015-03-12
hash based input checking

metti:
when a rule specifies "hash_input = 1" in addition to the
modification time of its inputs the hashed content of the
inputs are taken into account. hence a target is only dirty
if at least one of the incoming edges (actually leafes) are
dirty content-wise.

the purpose of the change is addressing workflows with a lot of
branch back-and-forth switches. when quickly changing to a
completely different code line, fixing a small issue and switching
back, often a lot of files are touched due to the branch switching.
but as they were not built in between, their artifacts are still
up-to-date but considered dirty by ninja based on the timestamp.
the same holds for other merge or rebase operations.

this is implemented with a hash log that stores filename, mtime
and murmur2 hash of a file into a binary file '.ninja_hashes'.
in addition a combined hash is stored for each successfully biuld target
to reflect the contents of the inputs at the time the output was produced.
this is necessary to also behave correctly for partial builds.

to distinguish between the two different variants a variant discriminator
is persisted with every hash log entry.

the file path is stored (rather than a hash of it) to circumvent
hash collisions which are really problematic as ninja would
not behave deterministic anymore. this problem really exists :-)

the mtime of the hashed file is stored to reduce the number of
hash lookups of a touched but unchanged file. it is only checked once
and subsequent calls compare first of all the mtime of the hashed file.

the syntax is slightly extended, but the existing code-path is unchanged
(besides the two ifs) in case the feature is not enabled. hence no
performance impact on the existing behaviour is expected.

Signed-off-by: Matthias Maennich matthias@maennich.net


jimon:
I've planned to implement this feature, but really happy that it's already implemented !

nico:
(I think it's possible to do this outside of ninja by letting your command write its output to a temporary place and then move it to the final output only if it's different from the existing output, and then making your edge have `restat = 1`.)

maximuska:
(the problem with doing that externally is that you need to know the full
list of dependencies, including headers which is only revealed as part of
the compilation and processing of deps files).

I haven't checked the patch.

src/graph.cc
```diff
@@ -163,12 +163,28 @@ bool DependencyScan::RecomputeOutputDirty(Edge* edge,
...
+                "(%d vs %d)",
+                used_restat ? "restat of " : "", output->path().c_str(),
+                most_recent_input->path().c_str(),
+                output_mtime, most_recent_input->mtime());
+        return true;
+      } else {
+        EXPLAIN("output %s older than most recent input, but hashed "
+                "contents of all inputs are unchanged",
+              output->path().c_str());
+      }
```
ilyapopov: Exiting the function without return?

metti: the 'return true' statements are early exits in case the edge is already identified dirty.
if not yet determined we have to fall through to the next criteria. in this case the next criteria is the command line (changed/unchanged). if we would exit with false here, any subsequent check would be ignored.

please note the function end is further down in line 192/208 and not in line 173/189

ilyapopov: Oh, sorry. I mistakenly thought that the function ends here.

domidimi:
Great ideas and work, hope to see that in!

fbenkstein:
I've posted an update and addressed the issues raised above.

kaspar030:
Any news on this?

fbenkstein:
I believe this is pretty much dead since upstream seems uninterested.

ArseniyShestakov:
That's sad. Working on slow laptop right now and rebasing become a real pain.

fbenkstein:
@ArseniyShestakov ccache provides a pretty good workaround.

nico:
This speeds up builds after branch checking at the cost of slowing down builds when work needs to be done. Ninja is an mtime-based system for speed reasons, and for cases where content compares it's possible to do something like this outside of ninja as mentioned in https://github.com/ninja-build/ninja/pull/929#issuecomment-78483374 (which has the advantage that multiple hashes can be computed in parallel; this patch adds the hash computation to the serial ninja mail loop).

So I think this patch doesn't fit ninja's priorities well.

Hello1024:
@nico Can we reconsider this?

The 'slowing down builds when work needs to be done' is a very rare case, because it requires a file to have been modified, and the file size remain the same.    File sizes remaining the same for large but different files gets pretty unlikely, with the uncommon exception of fixed size things like disk images, but thats why it's configurable behind an option.

The speedup in the common case of a file mtime being updated with no change in content is substantially larger.  There are lots of incremental build cases where certain changes in an input file won't change the output.


Hence I believe on-balance this patch is beneficial to most users, for both the 'git branches' usecase and the case of intermediate files which don't change when one input has minor changes.

jonesmz:
@Hello1024 

Maybe rebase the pull request, and then add a command line option to enable the new functionality, so that it is disabled by default?

greened:
I'd really love to see this accepted.  It's one way to address #1459.

PathogenDavid:
@greened Have you tried using [sccache](https://github.com/mozilla/sccache)? It's been working well for me.

burdiyan:
I’d really love to see this functionality in Ninja!

I’m currently reinventing the wheel outside of Ninja to get similar behavior, and it’s not trivial to get it right (like avoiding to hash if file was not touched, and so on) and a bit unwieldy. 

Having this as an option should not introduce any performance penalties, and people deciding to opt-in would consider the performance hit it brings.

The argument of ccache is irrelevant for builds of unsupported languages. The proposed solution here is environment-agnostic which is really nice.

The argument of Ninja’s speed and different priorities seems unfair, taking into account that this solution could improve build speeds for many people who switch branches often.

Leaving this as an opt-in seems like the best of both worlds.

I’m not entirely sure if this exact solution is the best one, but discarding this functionality completely is unfortunate.

jonesmz:
Actually, instead of using hash-based input checking, could input checking be optionally farmed out to some other invoked process? This would allow for arbitrary rules to be used to determine if a current file needs to be re-generated, instead of relying on baked-in functionality.

One such invoked process might be to compare a file hash in a database / manifest.

Another such invoked process might read meta-data stored alongside the built file and compare it to something.

rulatir:
> @ArseniyShestakov ccache provides a pretty good workaround.

... for one specific compiler of one specific programming language.

rupertks:
@fbenkstein "I believe this is pretty much dead since upstream seems uninterested"

Is this still the case? (Is there another thread somewhere discussing this?)

We use Ninja for AOSP builds (https://source.android.com/setup/build) and would love it if Ninja supported hashing

domidimi:
@metti and @fbenkstein do not follow on the topic anymore as far as I know. We internally use that version with some fixes, but due to no interest here we do not push updates.

Atemu:
I have a use-case involving the AOSP where mtimes are always fixed, no matter the file content. Hashing the source files is vital here as mtimes are 100% useless.

I don't see a reason why it shouldn't be an *option* (off by default) to use hashes instead. Everyone else in unaffected and those of us who need hashing can get it. The code doesn't seem all that complex or maintenance-heavy either.

radubacaran:
@domidimi @metti would you be willing to push somehwere the latest changes you have on this topics. According to @domidimi there are some unpublished bug fixes.
I am planning on using this version since in our workflow we have a lot of codegenerators that don't care about incremental build so they write the generated files each time even if the content doesn't change.

gracicot:
As I commented in #1459, using hash based input checking can dramatically speed up C++20 modules compilation. This is because many things in an interface may not change the BMI output. This is especially true with latest clang where it supports [reduced BMI](https://clang.llvm.org/docs/StandardCPlusPlusModules.html#reduced-bmi) and [non cascading changes](https://clang.llvm.org/docs/StandardCPlusPlusModules.html#experimental-non-cascading-changes).

PR #930, opened 2015-03-12
On unexpected output in a .d file, rebuild instead erroring.

### LINKED ISSUE #417: Mark output dirty if .d file doesn't match
I can generate ninja files with either fully resolved or relative paths. Both builds work with a clean build, but if I switch between generator types and don't clean, then I get a fatal error :
ninja: error: expected depfile 'obj/release/lvm.o.d' to mention 'obj/release/lvm.o', got '/spare/local/rgeary/inet/premake-dev-rgeary/obj/release/lvm.o'

If the first line of the .d file is not the expected file, ninja should always rebuild the file.

(This behaviour is observed on linux with the latest ninja/master code)

nico:
Fixes #417.  Also make .d files without a `:` an error (before, invalid depfiles were found by the output check that's now being removed.)

### DIFF
--- a/src/build_test.cc
+++ b/src/build_test.cc
@@ -816,8 +816,7 @@ TEST_F(BuildTest, DepFileParseError) {
   fs_.Create("foo.c", "");
   fs_.Create("foo.o.d", "randomtext\n");
   EXPECT_FALSE(builder_.AddTarget("foo.o", &err));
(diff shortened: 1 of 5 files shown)


nico:
Unless someone objects, I will merge this tomorrow.

Issue #931, opened 2015-03-14
Turn duplicate rules for the same target into an error


nico:
If multiple build edges generate the same output, ninja currently warns with

```
ninja: warning: multiple rules generate $name. build will not be correct; continuing anyway
```

Internally, ninja just ignores all but the last rule. In practice, this usually happens if people accidentally repeat a rule in their meta-buildsystem. But for rules with multiple targets, ninja's current handling of this can lead to inconsistent build graphs that can lead to crashes (see e.g. #867). Also, some people mistake this diagnostic for a bug in ninja instead of a bug in their build files (e.g. #543 last comment). Arguably, ninja could try to check if the two rules are exactly the same and not complain in that case, but
1. That's something the generator could do
2. It's not always obvious – the commands could be `cd ../.. && foo/bar.sh` and `cd ../../.. && baz/foo/bar.sh`, or they could set slightly different env vars (which happen to not make a difference in practice, but ninja can't know that)

So this should be an error, not a warning. However, this is something that happens in practice every now and then (examples: http://reviews.llvm.org/D6166#109619 http://crbug.com/351052), so the rollout for this should look something like this:
1. [x] Add a flag to turn this warning into an error, but don't set it by default (`--dupeedge=[warn|error]`?), make a release with this. (Mention the flag name in the warning text.)
2. [x] File bugs with all known generators to reject duplicate rules for the same target (up to generators to allow this if the two rules effectively are the same – generators are in a better position to make that decision. If they opt to do that, they still should only write one of these rules). Wait for this to be implemented everywhere, and to make it into releases.
3. [x] Flip the flag default to error, but keep the flag around to switch back to warning, keep mentioning the flag in the error message, and in --help output. Make another release.
4. [x] Remove flag from error message and help output, but keep honoring it. Make another release.
5. [ ] Eventually, remove the flag.

Not sure if we ever actually want to do 5, but the rest all seem like Obviously Good steps.

nico:
#940 was another bug in this area.

nico:
RfC to make this err by default from a while ago (no concerns): https://groups.google.com/forum/#!topic/ninja-build/ob5vhaGvvTc

PR #944, opened 2015-03-24
Add an opt-in flag to make duplicate edges an error (`-w dupbuild=err`).

nico:
This is step 1 on #931.  Duplicated edges will become an error by default in
the future.

(Also swap the order of -k and -l in the --help output.)


src/ninja.cc
```diff
@@ -792,6 +796,32 @@ bool DebugEnable(const string& name) {
...
+    return false;
+  } else if (name == "dupbuild=err") {
+    options->dupe_edges_should_err = true;
+    return true;
+  } else if (name == "dupbuild=warn") {
+    options->dupe_edges_should_err = false;
+    return true;
+  } else {
+    const char* suggestion =
+        SpellcheckString(name.c_str(), "dupbuild=err", "dupbuild=warn", NULL);
```
martine: Should "list" be in here?

src/ninja.cc
```diff
@@ -792,6 +796,32 @@ bool DebugEnable(const string& name) {
...
 }
 
+/// Set a warning flag.  Returns false if Ninja should exit instead  of
+/// continuing.
+bool WarningEnable(const string& name, Options* options) {
+  if (name == "list") {
+    printf("warning flags:\n"
+"  dupbuild=[err|warn]  multiple build lines for one target\n");
+    return false;
+  } else if (name == "dupbuild=err") {
```
martine: Regarding the name of the flag, I think the bikeshed should be Anthrazit.

src/ninja.cc
nico: Hm, probably. -d and -t also don't include "list" in the spelling suggestions though. Maybe I'll change all 3, thanks.

PR #947, opened 2015-03-27
Cleanup: Rename islatinalpha to IsLatinAlpha.

tfarina:
This matches better with all the other functions in this file, which
are all named using CamelCase.

R=thakis@chromium.org


nico:
i think it's supposed to look similar to isalpha(), islower(), isalnum(), etc

nico:
(I'll gladly merge the formatting changes – the deleted newline and the two { reformattings – if you want to send a pull request for that)

PR #951, opened 2015-04-01
Don't get stuck on cyclic graphs where one edge has multiple outputs.

### LINKED ISSUE #934: "[stuck]" without duplicate rules
Here's a "stuck" repro without duplicate rules (!):

    rule cg
       command = touch $out; touch b;
    
    build a b: cg c
    build c: cg a
    
    $ ninja 
    ninja: build stopped: stuck [this is a bug].

(Originally reported in a comment in #867. https://github.com/martine/ninja/issues/867#issuecomment-65918850 there suggests a fix, but it's possible that it's not a good fix.)

nico:
Fixes #934. Plan::AddSubTarget() tracks in want_ if each edge has been
visited and visits every edge only once.  But Plan::CheckDependencyCycle()
worked on nodes however, so if a cycle was entered through an edge with
multiple outputs, ninja would fail to detect that cycle.

Move cycle detection to look for duplicate edges instead of nodes to fix
this.  The extra jump makes CheckDependencyCycle() a bit slower: for a
synthetic build file with 10000 serial build steps, `ninja -n` now takes
0.32s instead of 0.26s before on my laptop.  In practice, projects have
a dependency change length on the order of 50, so there shouldn't be a
noticeable slowdown in practice.  (If this does end up being a problem:
CheckDependencyCycle() currently does O(n) work and is called O(m) times
from AddSubTarget(), so I think cycle checking is O(n^2) in the build
depth.  So instead of worrying about constant overhead, we could use
a set<> instead of a stack<>.  But it doesn't seem to be a problem in
practice.)

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -318,16 +318,33 @@ bool Plan::AddSubTarget(Node* node, vector<Node*>* stack, string* err) {
 
 bool Plan::CheckDependencyCycle(Node* node, const vector<Node*>& stack,
                                 string* err) {
-  vector<Node*>::const_iterator start = find(stack.begin(), stack.end(), node);
+  vector<Node*>::const_iterator start = stack.begin();
+  while (start != stack.end() && (*start)->in_edge() != node->in_edge())
+    ++start;
   if (start == stack.end())
     return false;
 
-  *err = "dependency cycle: ";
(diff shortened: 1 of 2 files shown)


nico:
(One random flaky failure on appveyor in 7 builds so far…not looking so great :-/)

PR #953, opened 2015-04-07
Fix backslashes in graphviz causing incorrect rendering on windows.

Sushisource:
Simple fix to replace backslashes in node paths with forward slashes so the generated graphs render properly on windows instead of having tabs and newlines and whatnot all over the place.

### DIFF
--- a/src/graphviz.cc
+++ b/src/graphviz.cc
@@ -15,6 +15,7 @@
 #include "graphviz.h"
 
 #include <stdio.h>
+#include <algorithm>
 
 #include "graph.h"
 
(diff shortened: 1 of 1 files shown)


src/graphviz.cc
```diff
@@ -22,7 +23,9 @@ void GraphViz::AddTarget(Node* node) {
   if (visited_nodes_.find(node) != visited_nodes_.end())
     return;
 
-  printf("\"%p\" [label=\"%s\"]\n", node, node->path().c_str());
+  std::string pathstr = node->path();
```
nico: Can you remove the std::s? ninja has lots of `using namespace std`s.

nico:
Thanks for the patch!

Sushisource:
Removed the stds (heh, stds).

nico:
Thanks for the patch!

It's strange that this is needed; the graphviz docs claim """In quoted strings in DOT, the only escaped character is double-quote ("). That is, in quoted strings, the dyad \" is converted to "; all other characters are left unchanged. In particular, \\ remains \\. Layout engines may apply additional escape sequences."""

I suppose the layout engine screws things up?

Sushisource:
Yeah, almost certainly just a case of failing to conform to the spec.

PR #958, opened 2015-04-17
Run more than 34 processes on Win32 if we have 32+ cores.

rui314:
For a compatiblity reason, dwNumberOfProcessors in Win32 is capped at 32.
So even if your machine has more than 32 cores, Ninja spawns at most 34
subprocesses if compiled as 32 bit application. This patch fixes the issue by
using GetNativeSystemInfo, which returns the system info from Wow64 point
of view, instead of GetSystemInfo.

GetNativeSystemInfo is supported since Windows XP or Windows Server 2003.

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -500,7 +500,7 @@ string StripAnsiEscapeCodes(const string& in) {
 int GetProcessorCount() {
 #ifdef _WIN32
   SYSTEM_INFO info;
-  GetSystemInfo(&info);
+  GetNativeSystemInfo(&info);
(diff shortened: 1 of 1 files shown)


sgraham:
Clearly I need to get a new computer to test this properly. :)

rui314:
I recently got a beefy machine with 20 physical cores (40 logical cores). I confirmed that Ninja reports it had 32 processors. With this patch it correctly recognized 40 cores.

nico:
Thanks! (Might've deserved a comment, but fine as is too.)

Issue #959, opened 2015-04-17
Pool limit not being respected


sgraham:
In https://code.google.com/p/chromium/issues/detail?id=476053#c19 it seems that the depth of a pool is not being respected.

I don't have a repro though. I'm giving the brute force method a try by adding a crash for the case that seems to be happening and running under afl-fuzz.

sgraham:
afl-fuzz didn't find it (yet?) unfortunately.

With https://github.com/martine/ninja/compare/master...sgraham:pool-2 and the input of https://dl.dropboxusercontent.com/u/116514/pool-crash.zip `ninja -C pool-crash -j1 -n` will abort() reliably for me.

sgraham:
Oh wait, that repro isn't useful because it needs all the input .cc of course. Bleh.

sgraham:
Got back to this a bit more. I still don't have a small repro and I feel like it should be easy to make one, but I haven't been able to. Here's my log of debugging a failing chrome case.

Get to a `stamp` with 3 inputs:
    gen/components/components_resources_new_stamp.d.stamp
    gen/components/grit/components_resources.h
    gen/components/components_resources.pak
outputting:
    obj/components/resources/components_resources_grit.stamp
wanted=true

set outputs_ready_ = true

Then something unrelated gets scheduled for the newly freed pool slot.

Then, NodeFinished for
    obj/components/resources/components_resources_grit.stamp
has 58 output edges, 55 stamps and 3 phony

2nd out_edge of the node has inputs of
    obj/components/resources/components_resource_grit.stamp
    obj/tools/grit/grit_sources.stamp
outputting:
    obj/components/resources/components_resources.inputdeps.stamp

both of those inputs pass AllInputsReady(), but want_e->second = false
-- means obj/components/resources/components_resources.inputdeps.stamp doesn't need to be built, but something it's an input for does need to be built.

Because of that, instead of calling Plan::ScheduleWork for the edge, we immediately call Plan::EdgeFinished. This then calls edge->pool()->EdgeFinished. But that edge has never been scheduled as far as the pool is concerned, so this breaks the pool use_count (can get way < 0).

Later uses then don't have a correct count so later edges that are otherwise scheduled correctly exceed the pool limit.

cc @riannucci @dpranke

nico:
This is fixed on trunk, right?

sgraham:
Yes, I forgot to put a magic xref in 962.

PR #962, opened 2015-04-23
Fix pool use count going unbalanced

sgraham:
When a Node was finished, but its output edge was not directly needed, the pool use count would be decremented for that edge without ever having been incremented. This could result in the pool use count going below 0, so that future edges scheduled for that pool would not respect the pool's limit. This is a fix for #959.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -374,24 +374,19 @@ void Plan::ScheduleWork(Edge* edge) {
   }
 }
 
-void Plan::ResumeDelayedJobs(Edge* edge) {
-  edge->pool()->EdgeFinished(*edge);
-  edge->pool()->RetrieveReadyEdges(&ready_);
-}
-
-void Plan::EdgeFinished(Edge* edge, bool directly_wanted) {
+void Plan::EdgeFinished(Edge* edge) {
   map<Edge*, bool>::iterator e = want_.find(edge);
   assert(e != want_.end());
-  if (e->second)
+  bool directly_wanted = e->second;
+  if (directly_wanted)
     --wanted_edges_;
   want_.erase(e);
   edge->outputs_ready_ = true;
 
-  // See if this job frees up any delayed jobs
+  // See if this job frees up any delayed jobs.
   if (directly_wanted)
-    ResumeDelayedJobs(edge);
-  else
-    edge->pool()->RetrieveReadyEdges(&ready_);
+    edge->pool()->EdgeFinished(*edge);
+  edge->pool()->RetrieveReadyEdges(&ready_);
 
   // Check off any nodes we were waiting for with this edge.
   for (vector<Node*>::iterator o = edge->outputs_.begin();
@@ -415,7 +410,7 @@ void Plan::NodeFinished(Node* node) {
       } else {
         // We do not need to build this edge, but we might need to build one of
         // its dependents.
-        EdgeFinished(*oe, want_e->second);
+        EdgeFinished(*oe);
       }
     }
   }
--- a/src/build.cc
+++ b/src/build.cc
@@ -379,7 +379,7 @@ void Plan::ResumeDelayedJobs(Edge* edge) {
   edge->pool()->RetrieveReadyEdges(&ready_);
 }
 
-void Plan::EdgeFinished(Edge* edge) {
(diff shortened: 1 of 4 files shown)


nico:
Makes sense to me. @riannucci, comments?

src/build_test.cc
```diff
@@ -1092,6 +1095,31 @@ TEST_F(BuildTest, SwallowFailuresLimit) {
...
+    "build C.stamp: touch B.d.stamp\n"
+    "build final.stamp: touch || C.stamp\n";
+
+  RebuildTarget("final.stamp", manifest);
+
+  fs_.RemoveFile("B.d.stamp");
+
+  State save_state;
+  RebuildTarget("final.stamp", manifest, NULL, NULL, &save_state);
+  EXPECT_GE(save_state.LookupPool("some_pool")->current_use(), 0);
```
riannucci: Shouldn't it always be exactly 0 at this point? Or does GE mean something else?

riannucci:
Yeah this seems fine to me, though it's a bit hacky since if EdgeScheduled ever gained a check/assertion for _current_use <= _depth then this could g'splode. I would probably add a comment to Pool noting that _current_use could potentially exceed _depth (for a brief, non-race-prone) period of time, since it's generally taken as an invariant elsewhere.

nico:
Maybe EdgeFinished should only call ResumeDelayedJobs if e->second then?

riannucci:
That I'm not sure... I don't remember what e->second is :) (it's been a little while...)

sgraham:
For the current patch, other than the temporary invariant violation, I also
wasn't entirely certain that it couldn't cause scheduling something that's
already been scheduled (i.e. causing the opposite problem). Any thoughts on
that?

I think if avoiding calling ResumeDelayedJobs when !e->second, we might
still need to call pool()->RetrieveReadyEdges. EdgeFinished is called after
pulling something out of ready_ by FindWork(), so possibly it could be left
empty otherwise?

Either way passes existing+new tests.

src/build_test.cc
sgraham: Yes it means >= and, yes it should be zero as the build's done. I just intended to verify the invariant that was otherwise not being checked and failing.

sgraham:
I convinced myself that it's OK to pool()->EdgeScheduled there, so I added a comment about the hackiness of incrementing the counter only to decrement it right away, per your suggestion.

riannucci:
lgtm

nico:
As hinted at above, rather than adding and subtracting again, is it possible to not subtract for edges we don't want? (The "Maybe EdgeFinished should only call ResumeDelayedJobs if e->second then?" bit)

sgraham:
It seemed a bit ugly. I don't think it's right to not call RetrieveReadyEdges, so it has pass down a bool and do that instead of ResumeDelayedJobs. Pushed in case you prefer that way.

src/build.cc
```diff
@@ -388,7 +388,10 @@ void Plan::EdgeFinished(Edge* edge) {
   edge->outputs_ready_ = true;
 
   // See if this job frees up any delayed jobs
-  ResumeDelayedJobs(edge);
+  if (directly_wanted)
+    ResumeDelayedJobs(edge);
```
nico: ResumeDelayedJobs is only two lines and only called here. If you inline the function, this becomes

  if (directly_wanted)
    edge->pool()->EdgeFinished(*edge);
  edge->pool()->RetrieveReadyEdges(&ready_);

And directly_wanted is identical to e->second in this scope, right? …oh, I guess it isn't as we recurse down.

If we recurse, do we want the state of where we started recursing to decide if we should call EdgeFinished(), or the local state of that edge? Probably the latter, right? So not passing in a bool is the more correct version too as far as I understand?

sgraham: I was thinking it was different because of the recursion, but I think you're right. Updated.

nico:
lgtm. Want me to merge this, or do you want to do some testing and make sure I haven't talked you into something that doesn't actually work?

sgraham:
Yeah, let me use it for a couple days locally and see if I have any
problems first. I'll ping back again if it seems ok.

sgraham:
I haven't had any problems (running with a couple extra check+aborts added) so I think it's safe to merge this.

PR #963, opened 2015-04-28
Extract MurmurHash functions into their own source files.

tfarina:
The implementation were copied from
https://code.google.com/p/smhasher/source/browse/trunk/MurmurHash2.*


nico:
There's no real need for this at this point.

PR #964, opened 2015-04-29
Cleanup build on SIGHUP.

nicolasdespres:
SIGHUP is sent when the connection hang up (i.e. when the terminal
window is closed or the ssh connection is closed).


nico:
Thanks!

Issue #965, opened 2015-05-05
Build actions with pool=console sometimes fail to terminate


pcc:
To reproduce:

1. Check out LLVM and llgo (http://llvm.org/svn/llvm-project/llgo/trunk/README.TXT)
2. Apply http://reviews.llvm.org/D9495
3. Run cmake (>= 3.2) with ninja generator.
4. ninja check-libgo

Step 4 will not terminate.

axw:
I've found that it does terminate, but takes a very long time to do so; around 4 minutes longer than it should. The child /bin/sh process is a zombie, and Ninja is stuck in ppoll (same happens with pselect).

axw:
Turns out there was an orphaned process hanging onto the pipe's write end that gets created in Subprocess::Start in subprocess-posix.cc. Seems that the write end should just be closed in the console case. I'll send a PR, and someone can tell me if it's braindead.

nico:
Does this still repro on trunk? #962 fixed a bug in pools, maybe that's related?

axw:
@nico Yes, still repros. I pulled master yesterday.

nico:
See https://github.com/martine/ninja/pull/967#issuecomment-116945038

PR #966, opened 2015-05-06
Set close-on-exec for pipe writer in console mode

axw:
The pipe isn't used by subprocesses in console mode,
so close it. This avoids the possibility of the FD
being leaked into nested subprocesses, causing Ninja
to hang like in #965.

### DIFF
--- a/src/subprocess-posix.cc
+++ b/src/subprocess-posix.cc
@@ -55,8 +55,6 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
     Fatal("fork: %s", strerror(errno));
 
   if (pid_ == 0) {
-    close(output_pipe[0]);
-
     // Track which fd we use to report errors on.
     int error_pipe = output_pipe[1];
     do {
@@ -90,9 +88,12 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
         // Now can use stderr for errors.
         error_pipe = 2;
         close(output_pipe[1]);
+      } else {
+        // In the console case, the pipe is not needed by the child
+        // process at all. Close it in the child, but keep it open
+        // in the parent in case execl fails.
+        SetCloseOnExec(output_pipe[1]);
       }
-      // In the console case, output_pipe is still inherited by the child and
-      // closed when the subprocess finishes, which then notifies ninja.
 
       execl("/bin/sh", "/bin/sh", "-c", command.c_str(), (char *) NULL);
     } while (false);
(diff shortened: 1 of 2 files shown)


src/subprocess-posix.cc
```diff
@@ -55,8 +55,6 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
     Fatal("fork: %s", strerror(errno));
 
   if (pid_ == 0) {
-    close(output_pipe[0]);
```
pcc: Why is this needed? This is the parent's side of the pipe so we ought to close it, right?

src/subprocess-posix.cc
```diff
@@ -90,9 +88,12 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
         // Now can use stderr for errors.
         error_pipe = 2;
         close(output_pipe[1]);
+      } else {
+        // In the console case, the pipe is not needed by the child
+        // process at all. Close it in the child, but keep it open
+        // in the parent in case execl fails.
+        SetCloseOnExec(output_pipe[1]);
```
pcc: I think this will mean that Ninja will be unable to start tasks in parallel while the console pool task is running, as it will be stuck in the `waitpid` call in `Subprocess::Finish`. Maybe we should be watching for SIGCHLD from pool=console child processes instead of waiting on the pipe.

src/subprocess-posix.cc
axw: This is a drive-by; it's already set O_CLOEXEC, immediately before the fork.

src/subprocess-posix.cc
axw: Ah, I think I see. I'll paraphrase below, please let me know if I misunderstand.

DoWork/OnPipeReady is using the pipe to determine whether or not the child is done. If there were a console pool task, it would immediately go do Done, and then Finish would block until it's really done. While blocking, other tasks' buffers would fill up, potentially blocking them from doing work.

For *only* console pool tasks, DoWork could instead use waitpid. That would have the same effect as ppoll on a non-console task that never sends anything to stdout (i.e. they'd both wait until the subprocess dies).

axw:
@pcc Please see latest commit. It should better handle parallel tasks now. See commit message for a description, if necessary.

maximuska:
- Couldn't understand what makes _console_ case special with respect to
process termination detection to trigger the issue. If we understand the
problem, is it possible to create a small repro, like something running
some bash one-line script?
- @andrew: wouldn't setting sigchld to wakeup ppoll and doing a loop of
waitpid(-1, &stat, WNOHANG) in DoWork() on EINTR be a simpler change?

axw:
@maximuska You're right. There's nothing console-specific about it, and my change doesn't cater for non-console. Repro build.ninja:

```
rule test
    command = sleep 10s&
    pool = console # or without
build test: test
default test
```

I will investigate your suggestion.

PR #967, opened 2015-05-07
Wake ppoll/pselect on SIGCHLD

### LINKED ISSUE #965: Build actions with pool=console sometimes fail to terminate

axw:
This change introduces a sigaction for
SIGCHLD, which causes ppoll and pselect
to return -1/EINTR if a child dies before
a pipe becomes readable. We then check
for dead children, leaving them in a
waitable state for Finish.

Fixes #965
Obsoletes #966

### DIFF
--- a/src/subprocess-posix.cc
+++ b/src/subprocess-posix.cc
@@ -25,6 +25,22 @@
 
 #include "util.h"
 
+// DrainPipe reads from fd until it would block. On EOF, the fd is
+// closed and set to -1. The fd passed in must be non-blocking.
+static void DrainPipe(int *fd, std::string *out);
+
+static void CreatePipe(int fds[2]) {
+  if (pipe(fds) < 0)
+    Fatal("pipe: %s", strerror(errno));
+#if !defined(USE_PPOLL)
+  // If available, we use ppoll in DoWork(); otherwise we use pselect
+  // and so must avoid overly-large FDs.
+  if (fds[0] >= static_cast<int>(FD_SETSIZE))
+    Fatal("pipe: %s", strerror(EMFILE));
+#endif  // !USE_PPOLL
+}
+
+
 Subprocess::Subprocess(bool use_console) : fd_(-1), pid_(-1), status_(-1),
                                            use_console_(use_console) {
 }
@@ -38,16 +54,10 @@ Subprocess::~Subprocess() {
 
 bool Subprocess::Start(SubprocessSet* set, const string& command) {
   int output_pipe[2];
-  if (pipe(output_pipe) < 0)
-    Fatal("pipe: %s", strerror(errno));
+  CreatePipe(output_pipe);
   fd_ = output_pipe[0];
-#if !defined(USE_PPOLL)
-  // If available, we use ppoll in DoWork(); otherwise we use pselect
-  // and so must avoid overly-large FDs.
-  if (fd_ >= static_cast<int>(FD_SETSIZE))
-    Fatal("pipe: %s", strerror(EMFILE));
-#endif  // !USE_PPOLL
   SetCloseOnExec(fd_);
+  SetNonBlocking(output_pipe[0]);
 
   pid_ = fork();
   if (pid_ < 0)
@@ -112,17 +122,27 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
   return true;
 }
 
-void Subprocess::OnPipeReady() {
+void DrainPipe(int *fd, std::string *out) {
   char buf[4 << 10];
-  ssize_t len = read(fd_, buf, sizeof(buf));
-  if (len > 0) {
-    buf_.append(buf, len);
-  } else {
-    if (len < 0)
-      Fatal("read: %s", strerror(errno));
-    close(fd_);
-    fd_ = -1;
+  while (*fd != -1) {
+    ssize_t len = read(*fd, buf, sizeof(buf));
+    if (len > 0) {
+      if (out)
+        out->append(buf, len);
+    } else {
+      if (len < 0) {
+        if (errno == EAGAIN)
+          return;
+        Fatal("read: %s", strerror(errno));
+      }
+      close(*fd);
+      *fd = -1;
+    }
   }
+}
+
+void Subprocess::OnPipeReady() {
+  DrainPipe(&fd_, &buf_);
 }
 
 ExitStatus Subprocess::Finish() {
@@ -155,6 +175,7 @@ int SubprocessSet::interrupted_;
 
 void SubprocessSet::SetInterruptedFlag(int signum) {
   interrupted_ = signum;
+  SignalHandled();
 }
 
 void SubprocessSet::HandlePendingInterruption() {
@@ -171,9 +192,20 @@ void SubprocessSet::HandlePendingInterruption() {
 }
 
 bool SubprocessSet::child_exited_;
+#if !defined(USE_PPOLL)
+int SubprocessSet::self_pipe_[2];
+#endif
+
+void SubprocessSet::SignalHandled() {
+#if !defined(USE_PPOLL)
+  // Wake up select. See http://cr.yp.to/docs/selfpipe.html.
+  if (write(self_pipe_[1], "", 1)) {}
+#endif
+}
 
 void SubprocessSet::SetChildExited(int signum) {
   child_exited_ = true;
+  SignalHandled();
 }
 
 void SubprocessSet::HandleChildExit() {
@@ -197,17 +229,10 @@ void SubprocessSet::HandleChildExit() {
         continue;
       // Drain pipe by reading until we would block, as
       // finished processes' pipes are not read by DoWork.
-      while (subprocess->fd_ != -1) {
-        pollfd pfd = { subprocess->fd_, POLLIN | POLLPRI, 0 };
-        int n = poll(&pfd, 1, 0);
-        if (n == -1)
-          Fatal("poll: %s", strerror(errno));
-        if (n == 0) {
-          close(subprocess->fd_);
-          subprocess->fd_ = -1;
-        } else {
-          subprocess->OnPipeReady();
-        }
+      subprocess->OnPipeReady();
+      if (subprocess->fd_ != -1) {
+        close(subprocess->fd_);
+        subprocess->fd_ = -1;
       }
       subprocess->pid_ = -1;
       subprocess->status_ = status;
@@ -219,6 +244,14 @@ void SubprocessSet::HandleChildExit() {
 }
 
 SubprocessSet::SubprocessSet() {
+#if !defined(USE_PPOLL)
+  CreatePipe(self_pipe_);
+  SetCloseOnExec(self_pipe_[0]);
+  SetCloseOnExec(self_pipe_[1]);
+  SetNonBlocking(self_pipe_[0]);
+  SetNonBlocking(self_pipe_[1]);
+#endif
+
   sigset_t set;
   sigemptyset(&set);
   sigaddset(&set, SIGINT);
@@ -251,6 +284,11 @@ SubprocessSet::~SubprocessSet() {
     Fatal("sigaction: %s", strerror(errno));
   if (sigprocmask(SIG_SETMASK, &old_mask_, 0) < 0)
     Fatal("sigprocmask: %s", strerror(errno));
+
+#if !defined(USE_PPOLL)
+  close(self_pipe_[0]);
+  close(self_pipe_[1]);
+#endif
 }
 
 Subprocess *SubprocessSet::Add(const string& command, bool use_console) {
@@ -305,8 +343,12 @@ bool SubprocessSet::DoWork() {
 #else  // !defined(USE_PPOLL)
 bool SubprocessSet::DoWork() {
   fd_set set;
-  int nfds = 0;
   FD_ZERO(&set);
+
+  // We always select on the self-pipe, in case a signal is
+  // delivered. This protects us from races in pselect.
+  FD_SET(self_pipe_[0], &set);
+  int nfds = self_pipe_[0]+1;
 
   for (vector<Subprocess*>::iterator i = running_.begin();
        i != running_.end(); ++i) {
@@ -329,6 +371,7 @@ bool SubprocessSet::DoWork() {
     return IsInterrupted();
   }
 
+  HandleChildExit();
   HandlePendingInterruption();
   if (IsInterrupted())
     return true;
@@ -339,6 +382,8 @@ bool SubprocessSet::DoWork() {
     if (fd >= 0 && FD_ISSET(fd, &set))
       (*i)->OnPipeReady();
   }
+  if (FD_ISSET(self_pipe_[0], &set))
+    DrainPipe(&self_pipe_[0], NULL);
 
   return IsInterrupted();
 }
--- a/src/subprocess-posix.cc
+++ b/src/subprocess-posix.cc
@@ -25,7 +25,7 @@
 
 #include "util.h"
 
-Subprocess::Subprocess(bool use_console) : fd_(-1), pid_(-1),
+Subprocess::Subprocess(bool use_console) : fd_(-1), pid_(-1), status_(-1),
                                            use_console_(use_console) {
 }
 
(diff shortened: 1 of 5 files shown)


axw:
@pcc @maximuska PTAL, hopefully more sane.

For posterity, minimal repro:

```
rule test
    command = sleep 10s&
    pool = console # or without
build test: test
```
(should not hang Ninja)

src/subprocess-posix.cc
```diff
@@ -169,11 +171,34 @@ void SubprocessSet::HandlePendingInterruption() {
     interrupted_ = SIGTERM;
 }
 
+void SubprocessSet::HandleChildDeath() {
+  if (interrupted_ != SIGCHLD)
+    return;
+  interrupted_ = false;
```
martine: interrupted_ is an int.  Maybe this should be 0?

src/subprocess-posix.cc
```diff
@@ -184,6 +209,8 @@ SubprocessSet::SubprocessSet() {
     Fatal("sigaction: %s", strerror(errno));
   if (sigaction(SIGTERM, &act, &old_term_act_) < 0)
     Fatal("sigaction: %s", strerror(errno));
+  if (sigaction(SIGCHLD, &act, &old_chld_act_) < 0)
```
maximuska: I think setting a shared handler with sigterm and sigint is dangerous, as SIGCHLD gets an opportunity to override value of interrupted_ global (which could be set to e.g., SIGINT), preventing ninja from exiting promptly.

src/subprocess-posix.cc
```diff
@@ -169,11 +171,34 @@ void SubprocessSet::HandlePendingInterruption() {
...
 
+void SubprocessSet::HandleChildDeath() {
+  if (interrupted_ != SIGCHLD)
+    return;
+  interrupted_ = false;
+  int options = WEXITED | WNOWAIT | WNOHANG;
+  for (vector<Subprocess*>::iterator i = running_.begin();
+       i != running_.end(); ) {
+    siginfo_t info = {0};
+    int res = waitid(P_PID, (id_t)(*i)->pid_, &info, options);
```
martine: Is this loop the same as a single call to `waitpid(-1, ...)`?

maximuska: Minor: I think that if waitid(-1, ..) would be used here instead, you could loop over all finished processes in the outer loop and iterating running_ in the inner, making number of waitid() syscalls ~finished processes, which would be marginally more efficient.

src/subprocess-posix.cc
```diff
@@ -169,11 +171,34 @@ void SubprocessSet::HandlePendingInterruption() {
...
+    if (res == -1)
+      Fatal("waitid(%d): %s", (*i)->pid_, strerror(errno));
+    if (info.si_pid == 0) {
+      ++i;
+    } else {
+      close((*i)->fd_);
```
maximuska: If HandleChildDeath is called before we read all active fds as it is now, doing close() could discard several last lines of process output, I think.

src/subprocess-posix.cc
```diff
@@ -169,11 +171,34 @@ void SubprocessSet::HandlePendingInterruption() {
     interrupted_ = SIGTERM;
 }
 
+void SubprocessSet::HandleChildDeath() {
```
maximuska: May be it's better to change the name to something more "politically correct". Or that's just me? ;)

src/subprocess-posix.cc
maximuska: @martine - consequent SIGCHLD are not queued, you have to waitXX() in loop to ensure that all zobies were cleaned.

nico:
Also, is it possible to add a test for this? See `subprocess_test.cc` for examples.

maximuska:
I think ultimately, we should also remove the code marking process as 'done' when pipe read indicates EOF, which would address hypothetical tools closing their stdout and stderr w/o terminating.

src/subprocess-posix.cc
axw: Sorry about that; no longer relevant with latest diff.

src/subprocess-posix.cc
axw: Fixed.

src/subprocess-posix.cc
axw: I've changed it around a bit, so it does a loop of that sort. Now using waitpid (so no WNOWAIT), and Finish is no longer solely responsible for waiting on subprocesses.

src/subprocess-posix.cc
axw: The pipe is now drained (read until read would block) after waitpid.

axw:
@nico Added a test for the original issue, i.e. backgrounded task doesn't block observing subprocess exiting. It's using time, which I'd prefer not to do, but couldn't think of a better way to test such behaviour. Suggestions welcome.

@maximuska I have reworked things a bit, so that pipe closure is no longer used to determine whether or not a subprocess is done.

PTAL.

src/subprocess-posix.cc
axw: Renamed.

src/subprocess_test.cc
```diff
@@ -261,3 +261,17 @@ TEST_F(SubprocessTest, ReadStdin) {
   ASSERT_EQ(1u, subprocs_.finished_.size());
 }
 #endif  // _WIN32
+
+#ifndef _WIN32
+TEST_F(SubprocessTest, Background) {
+  time_t before = time(NULL);
+  Subprocess* subproc = subprocs_.Add("echo -n 123; sleep 10s&");
```
nico: Thanks for adding a test! Will this take 10s to run? If so, is it possible to write a test that doesn't take 10s? (For comparison, running all of ninja's tests takes < 0.5s at the moment and I'd like to keep the test suite fast.)

axw: The test will only take 10s if it fails, i.e. without the rest of the patch. The immediate subprocess will exit after backgrounding the sleep task, and so with my patch the exit will be observed and the test will complete immediately.

src/subprocess-posix.cc
```diff
@@ -169,11 +170,61 @@ void SubprocessSet::HandlePendingInterruption() {
...
+  for (int i = running_.size(); i > 0; --i)
+    WaitChildExit(false);
+}
+
+void SubprocessSet::WaitChildExit(bool block) {
+  int status = 0;
+  int options = block ? 0 : WNOHANG;
+  pid_t pid = waitpid(-1, &status, options);
+  if (pid == -1)
+    Fatal("waitpid: %s", strerror(errno));
```
maximuska: In 'blocking' mode, pressing Ctrl-C will interrupt waitpid()
You should probably return gracefully here if errno set to EINTR to let ninja exit cleanly

src/subprocess-posix.cc
```diff
@@ -169,11 +170,61 @@ void SubprocessSet::HandlePendingInterruption() {
...
+  if (pid == 0)
+    return;
+  for (vector<Subprocess*>::iterator i = running_.begin();
```
maximuska: This loops kinda duplicates code in DoWork() (the one right after 'nfds_t cur_nfd = 0;')
Wouldn't it be simpler to not have this additional loop here at all, rather to call HandleChildExit() just before the return from DoWork(), after all pipes are drained?

src/subprocess-posix.cc
```diff
@@ -222,6 +279,11 @@ bool SubprocessSet::DoWork() {
     ++nfds;
   }
 
+  if (nfds == 0) {
+    WaitChildExit(true);
```
maximuska: I am wondering if it would be possible to avoid handling this special case (nfds == 0) all together. I would assume that if no children have finished, DoWork() will retry and get blocked in ppoll() (with 0 fds in the set, waiting for a signal to interrupt it, including any child to finish which is what we need.
That would void the need in having two modes for WaitChildExit().

src/subprocess-posix.cc
maximuska: See below a comment proposing to get rid of the blocking wait

src/subprocess-posix.cc
```diff
@@ -169,11 +170,61 @@ void SubprocessSet::HandlePendingInterruption() {
...
+  for (vector<Subprocess*>::iterator i = running_.begin();
+       i != running_.end(); ++i) {
+    Subprocess *subprocess = *i;
+    if (pid != subprocess->pid_)
+      continue;
+    // Drain pipe by reading until we would block, as
+    // finished processes' pipes are not read by DoWork.
+    while (subprocess->fd_ != -1) {
+      pollfd pfd = { subprocess->fd_, POLLIN | POLLPRI, 0 };
+      int n = poll(&pfd, 1, 0);
```
maximuska: .. if this draining loop can be removed, that would void the need to add #ifdef USE_PPOLL / USE_SELECT here for consistency with the rest of the code. Otherwise, we probably need to #ifdef to use either select or poll here.

src/subprocess-posix.cc
```diff
@@ -273,6 +329,11 @@ bool SubprocessSet::DoWork() {
     }
   }
 
+  if (nfds == 0) {
```
maximuska: as above

src/subprocess-posix.cc
```diff
@@ -125,25 +126,25 @@ void Subprocess::OnPipeReady() {
...
 ExitStatus Subprocess::Finish() {
-  assert(pid_ != -1);
-  int status;
-  if (waitpid(pid_, &status, 0) < 0)
-    Fatal("waitpid(%d): %s", pid_, strerror(errno));
-  pid_ = -1;
-
-  if (WIFEXITED(status)) {
-    int exit = WEXITSTATUS(status);
+  if (pid_ != -1) {
```
maximuska: Can this happen with the new code? I think previously waitpid() was required here to collect zombies, now we shouldn't have these when Finish() is called.

src/subprocess-posix.cc
```diff
@@ -169,11 +170,61 @@ void SubprocessSet::HandlePendingInterruption() {
...
+
+void SubprocessSet::SetChildExited(int signum) {
+  child_exited_ = true;
+}
+
+void SubprocessSet::HandleChildExit() {
+  if (!child_exited_)
+    return;
+  child_exited_ = false;
+  for (int i = running_.size(); i > 0; --i)
```
maximuska: You are iterating on running_ in the outer loop, which is strange, as you are also doing waitpid(-1) and iterating on running_ after that in the inner function. I think it should be a 'while waipid() has not reported no-mo-processes' conditioned loop instead.

src/subprocess-posix.cc
axw: Right, this is just for the destructor ("// Reap child if forgotten").

src/subprocess-posix.cc
axw: The idea behind that was that we know the maximum number of waitable children, so we could make that many waitpid calls without having it return with errno==ECHILD. I've changed it anyway, hopefully it's clearer now.

src/subprocess-posix.cc
axw: I have removed the blocking waitpid. Thanks, much neater.

src/subprocess-posix.cc
axw: ppoll is Linux-specific, but plain old poll is in POSIX.1-2001. We use poll here instead of ppoll because we're only doing non-blocking polls. I don't think there's any reason to make it configurable.

src/subprocess-posix.cc
axw: Done.

src/subprocess-posix.cc
axw: > This loops kinda duplicates code in DoWork() (the one right after 'nfds_t cur_nfd = 0;')

Kinda, but not exactly. The one in DoWork is responding to known readability, whereas the one in HandleChildExit just blindly reads until there's nothing left. For DoWork to handle both cases, I think it would have to change considerably.

> Wouldn't it be simpler to not have this additional loop here at all, rather to call HandleChildExit() just before the return from DoWork(), after all pipes are drained?

How do we know which pipes to drain if we haven't yet handled child exit? Just do a non-blocking read loop on all FDs after ppoll/pselect returns? That doesn't sound very efficient. And we can't just wait for a read event from ppoll, because then the original issue would not be solved; i.e. the pipe might remain open, but not become readable.

axw:
@maximuska Thank you for bearing with me. I've addressed everything other than the poll/read loop; please see my response to your comments there.

src/subprocess_test.cc
nico: Ah, if this only happens on the failure case that's much better. But shouldn't 1s still be more than enough on even very slow systems?

maximuska:
The change is fine now as far as I can see, I agree with your responses.
If obsessed, I would set O_NONBLOCK on the pipe (typically a good idea when using poll()), instead of using poll() when draining the fd.

src/subprocess_test.cc
axw: I'm probably just being paranoid. The project I work on the most has had a lot of timing-related failures in the past, but they've tended to be more heavyweight than this.

I've reduced it to 2s. I'd make it 1s, but then the second might reasonably tick over between calls to time(). I could go and use gettimeofday for better resolution, but this is only matters for the theoretical failure case, so doesn't seem worthwhile complicating the test.

axw:
@maximuska Thanks. I thought about doing that too, but it doesn't make things much cleaner. OnPipeReady then ends up needing to loop and handle EAGAIN. It would be neater if it read into a buffer the same size as the pipe capacity, but I don't think there's a portable way to determine that.

@nico @martine rebased, please let me know if there's anything else of concern.

maximuska:
@axw - for the sake of the discussion, I think that you don't need to loop (or add any other change) in OnPipeReady() even with O_NONBLOCK set. The read will return the largest chunk it can give you w/o blocking you, pretty much as it does in the current code. While the pipe is blocking now, the read in OnPipeReady() never get blocked waiting for more data as it is invoked when poll reports that the fd is already readable (it just returns less than the requested 4K).

axw:
@maximuska I meant to handle the drain-on-exit case only, and only if poll were removed. If *all* we were to change is to set the FD non-blocking, then no further change would be needed. That's not enough to remove the poll loop, though.

If the exited process wrote/buffered 4K+1 bytes, one non-blocking read (with 4K buffer as it stands) would read as much as possible: only 4K. We won't come back around to that pipe, because the process is no longer running. The options I see to get the remaining 1 byte: change OnPipeReady to loop (or duplicate in HandleChildExit and loop there), or somehow find the pipe's capacity and set the user-space buffer to the same size. The latter would be nice, but I don't know a portable way to do it (there's fcntl/F_GETPIPE_SZ, but it's Linux-specific).

nico:
I started looking at this. So far I noticed that tests now hang on OS X, so it's not good to go as-is. I'm afk for the next few hours, I'll read and debug some more later today.

axw:
@nico Thanks; bootstrap seems to do the same on FreeBSD. I'll look into that.

axw:
@nico With a little debugging, I found that on FreeBSD it can occur that SIGCHLD is delivered *and* pselect returns successfully (i.e. without EINTR). Turns out that on BSD and OS X, pselect is just a user-space wrapper around sigprocmask+select. Basically, it's broken there.

So now there's a problem: select might be passed no FDs at all, in which case it'll block forever. There's a work-around for that: http://cr.yp.to/docs/selfpipe.html. Alternatively, use kqueue on BSD/OS X. That still leaves out other OSes, like Solaris. Not sure you'd want to go down the road of recreating libev(ent), or pulling that in as a dependency, so I'll look at implementing the self-pipe approach.

axw:
@nico PTAL. Tested on Linux and FreeBSD. I don't have access to OS X, but I expect it's the same issue.

axw:
Ignore last message; I screwed up while rewriting the fix on another machine. I'll ping again when it's actually ready.

axw:
@nico PTAL. Added another test case that would have picked up my last mistake.

nico:
Since this patch does make ninja a bit more complicated. we discussed this a bit on IRC. Supporting `command = foo&` doesn't seem very interesting. For the use case of making gotest work (a script that spawns `sleep $long_time` in the background), it sounds like it's possible to make that work by piping that sleep call's handles to /dev/null (or similar) to work around this. So it sounds like this change won't be needed for now.

Issue #905, opened 2015-01-23
Add command line option similar to GNU make --question


sgraham:
(I feel like we had this discussion before, but I can't find it now)

http://www.gnu.org/software/make/manual/make.html#Running

-q seems like the wrong name, as it could be "quiet" as well. But to confirm no-op builds in Chromium, that project would like to have ninja emit a status code when the build is dirty.

sgraham:
See also https://code.google.com/p/chromium/issues/detail?id=451499

dpranke:
+1. I also think this would be a nice switch to have.

dpranke:
You could also add a '-t up-to-date' or something to avoid having to add a new top-level argument.

sgraham:
(I actually think log parsing is just fine, but opened this per bikeshedding over there.)

nico:
I too think log parsing is fine, the "no work to do" change isn't ever going to change anyway.

But this seems like a reasonable feature request anyhow.

riannucci:
Bikeshedder jumping in to defend himself: a well-intentioned addition of internationalization to ninja would break your assumptions :stuck_out_tongue:

riannucci:
Alternately: as long as you promise that I won't be on the hook for maintaining the log parsing ever, then I'm fine with it ^_^

sgraham:
It's true. (We also have to emit "ninja: Entering directory `out\Release'" to match something in Emacs. English or bust!)

martine:
I feel bad about the Emacs thing but it's soooo convenient.

BTW this is what Emacs has to say about the default value of the pattern:
```
compilation-directory-matcher's value is ("\\(?:Entering\\|Leavin\\(g\\)\\) directory `\\(.+\\)'$"
 (2 . 1))
```

dseomn:
Here's a hacky workaround that works without needing to parse messages that could change, run `NINJA_STATUS='no-op expected, but ninja is doing: ' ninja --verbose`, and test if the custom `NINJA_STATUS` is in the output. It would be really nice to have a simpler way though, to catch bugs like #2641 faster.

Issue #746, opened 2014-04-25
Root bug for outstanding display/output issues/requests


fischman:
As of now these are the open ninja issues that have to do with status display or ninja output.  Listing them in hopes of a cohesive plan forming.

curses-based rule progress indicator #717 
Support color in NINJA_STATUS #713
print "still waiting" for long running commands if we serialize on them #678 
Address FIXME: Add --no-strip-ansi-escapes option #672
Throttle rate of output messages #663
Use ANSI escapes to output to Cygwin-based terminals #659
Proposed/still running #629
Add NINJA_SMART_TERMINAL environment variable, with on, off, and hide_newline options #581
Use SetConsoleTextAttribute on Windows to simulate color #486
Add a `-q` "quiet" option. #480
Automatically use pager for failed output #413
"smart" output (e.g. overprinting) doesn't work under cygwin #374
Slow down on big tool output #326
Feature request -- print elapsed time #267
Show progress when waiting #111

nico:
#916 asks for what #672 is implementing.

nicolasdespres:
Ninja output into dumb terminal is not machine parsable #1010

nicolasdespres:
Print successful command along with their output. #658

nicolasdespres:
Support table status a la Buck build system: #1020

colincross:
Print output file on failure #1033

martine:
Make width of failure printouts configurable. #606

nicolasdespres:
Print commands when they start: #1158

jhasse:
I think a nice way to fix all these is by adding support for external frontends. Check out the PR by colincross: https://github.com/ninja-build/ninja/pull/1210

The PR implements the point @nicolasdespres said:

> To complete this list I think Ninja should provide enough control on its
output so that an external program, such as GUI for instance, could easily
turn Ninja's textual output into a more structured graphical display.

I've extended Colin's Python command line frontend which you can use with the PR which already fixes some of the other points (e.g. color) raised here. https://bixense.com/ja/

nicolasdespres:
I would like to renew the proposition I have made above in this thread on June, 5th 2014 to use only a limited set of environment variables to configure Ninja's output. I think this is enough and have a small impact on Ninja's codebase while giving all the power to external scripts to do what they need to build nice frontend. That would preserve Ninja's philosophy to do one thing well and remain fast by delegating the work.

fischman:
[...]

Issue #480, opened 2013-01-06
Add a `-q` "quiet" option.


chisophugis:
In some scenarios, it would be really nice to have ninja just print nothing.

For example, I'm strace'ing ninja (and its progeny) to get information about what commands are executed during a build (not just by ninja, but also by processes it creates). The workaround I'm currently using is something ugly like:

`strace -f -e 'abbrev=none' -e 'trace=execve' sh -c 'ninja 2>&1 >/dev/null'`

jimon:
+1

cdyson37:
The lack of a --quiet option is painful when you consider that ninja merges stdout and stderr (for sensible reasons). If you want to see only the tool output (generally just the errors), you have to do something a bit crazy like set NINJA_STATUS=SILENCE_IS_GOLDEN and then grep -v SILENCE_IS_GOLDEN on the output.

moroten:
At work, we use `NINJA_STATUS_SLEEP=2000` in #1501 to throttle the output, but setting it to `NINJA_STATUS_SLEEP=inf` would make the build quiet.

jonesmz:
Related to, but not a duplicate of, https://github.com/ninja-build/ninja/issues/420

megabyde:
+1

PR #968, opened 2015-05-07
Added quiet mode

jimon:
Very simple fix, was requested in https://github.com/martine/ninja/issues/480 and https://github.com/martine/ninja/issues/420.
This fix is useful when we run ninja from cmake from windows cmd, in this combination terminal print is very slow and ninja is able to finish work before cmd is able to print output to user.

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -205,6 +205,7 @@ void Usage(const BuildConfig& config) {
 "  -l N     do not start new jobs if the load average is greater than N\n"
 "  -n       dry run (don't run commands but act like they succeeded)\n"
 "  -v       show all command lines while building\n"
+"  -q       hide all command lines while building\n"
 "\n"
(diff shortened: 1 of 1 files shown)


nico:
I'd like to think about output flags in general (see #746)

sgraham:
-q is also used for `--question` in `make`, see #905. I don't know if that's worth caring about.

nico:
Another idea: We could let ninja print stuff only ten times a second in smart mode, then terminals that don't implement throttling themselves (like apparently cmd) would transparently just work better (?)

sgraham:
I think this might be non smart mode -- we replace the contents of the console window in smart mode?

nico:
No, if this is about non smart mode, my idea won't help.

jonesmz:
@jimon Would you be able to rebase your pull request against the latest master branch?

jhasse:
Closing due to inactivity. For quiet mode discussion see #480.

PR #971, opened 2015-05-27
Bootstrap out of source

hackhaslam:
This allows ninja to bootstrap out of source. It works the expected way:

mkdir my_build_dir
cd my_build_dir
/full/path/to/ninja/configure.py --bootstrap

It's split into two commits for clarity. The first adds a $sourcedir variable to the ninja script and makes all source files relative to it. The second fixes inclusion of the generated header.

Is there any interest in this?

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -28,7 +28,8 @@ import string
 import subprocess
 import sys
 
-sys.path.insert(0, 'misc')
+sourcedir = os.path.dirname(os.path.realpath(__file__))
+sys.path.insert(0, os.path.join(sourcedir, 'misc'))
 import ninja_syntax
 
 
@@ -251,11 +252,11 @@ if platform.is_msvc():
     objext = '.obj'
 
 def src(filename):
-    return os.path.join('src', filename)
+    return os.path.join('$sourcedir', 'src', filename)
 def built(filename):
     return os.path.join('$builddir', filename)
 def doc(filename):
-    return os.path.join('doc', filename)
+    return os.path.join('$sourcedir', 'doc', filename)
 def cc(name, **kwargs):
     return n.build(built(name + objext), 'cxx', src(name + '.c'), **kwargs)
 def cxx(name, **kwargs):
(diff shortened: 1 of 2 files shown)


martine:
I think I understand the mechanics of the change but I don't understand the motivation.  Could you explain?

hackhaslam:
The motivation is that we want to integrate ninja into our build system. We have a ninja submodule that builds before configuring ninja builds with cmake. We generally try to avoid building anything in source. It's mostly for building different configurations and enabling completely clean builds by removing the build dir. Granted, the first point isn't terribly important for us since we don't plan on building different configurations of ninja.

I realize that it's pretty weak motivation, but I figured that it would be nice to have and might benefit somebody else.

moroten:
Consider doing
(cd my_build_dir && /full/path/to/ninja/configure.py --bootstrap)
(cd my_build_dir2 && /full/path/to/ninja/configure.py --bootstrap)

The sourcedir variable will be different and therefore also the generated build.ninja will differ and therefore result in a full rebuild (command line change). A few months ago I tried to work around such problems by using absolute paths but, as I wrote in https://groups.google.com/forum/#!topic/ninja-build/tIcKNLWWKzo, ninja does not support absolute paths.

Is there a need to consider the full-rebuild-problem due to changed output directory?

hackhaslam:
Rebuilding everything is expected and desired. The point of doing out-of-source builds is to not share any build products between build directories. Otherwise it wouldn't be possible to build different configurations.

moroten:
When considering it again you are correct. My own situation was different with the same build directory but mixed working directory.

nico:
Sorry about the delay.

hackhaslam:
Thanks!

bradking:
This is a good feature.  However, commit 78b1e52a5df429d336d4a08dd0d77fc74bb477dc breaks even in-source builds when there is a space in the path to the source tree.  Previously this worked.

nico:
Thanks for the report. I'll revert this for now.

hackhaslam:
Sorry about that. I just opened #979 to try to fix it. It has the whole branch rebased on top of the revert.

PR #973, opened 2015-06-08
Fix crash in attempting to canonicalize paths longer than _MAX_PATH

### LINKED ISSUE #972: Ninja crashes due to Intel Compiler warnings
When trying to compile a C++ program using the Intel Compiler under Windows ninja will crash if the compiler outputs warnings like the following:

    warning #31001: The dll for reading and writing the pdb (for example, mspdb110.dll) could  not be found on your path. This is usually a configuration error. Compilation will continue using /Z7 instead of /Zi, but expect a similar error when you link your program.

When I disable this warning on the compiler command line the compile runs through without problems. I had a similar problem with the optimization diagnostic messages from the Intel Compiler which I also need to turn off.

I'm currently using a recent ninja version from git (Reported Version number 1.5.3)

If you need the ninja.build files to debug this problem, please let me know.

sgraham:
To fix #972.

### DIFF
--- a/src/includes_normalize-win32.cc
+++ b/src/includes_normalize-win32.cc
@@ -96,8 +96,12 @@ string IncludesNormalize::Relativize(StringPiece path, const string& start) {
 
 string IncludesNormalize::Normalize(const string& input,
                                     const char* relative_to) {
-  char copy[_MAX_PATH];
+  char copy[_MAX_PATH + 1];
   size_t len = input.size();
+  if (len > _MAX_PATH) {
+    Warning("path too long '%s'\n", input.c_str());
+    return input;
+  }
   strncpy(copy, input.c_str(), input.size() + 1);
   string err;
   unsigned int slash_bits;
--- a/src/includes_normalize-win32.cc
+++ b/src/includes_normalize-win32.cc
@@ -94,19 +94,18 @@ string IncludesNormalize::Relativize(StringPiece path, const string& start) {
   return Join(rel_list, '/');
 }
 
-string IncludesNormalize::Normalize(const string& input,
-                                    const char* relative_to) {
+bool IncludesNormalize::Normalize(const string& input, const char* relative_to,
+                                  string* result, string* err) {
   char copy[_MAX_PATH + 1];
   size_t len = input.size();
   if (len > _MAX_PATH) {
-    Warning("path too long '%s'\n", input.c_str());
-    return input;
+    *err = "path too long";
+    return false;
   }
   strncpy(copy, input.c_str(), input.size() + 1);
-  string err;
   unsigned int slash_bits;
-  if (!CanonicalizePath(copy, &len, &slash_bits, &err))
-    Warning("couldn't canonicalize '%s': %s\n", input.c_str(), err.c_str());
+  if (!CanonicalizePath(copy, &len, &slash_bits, err))
+    return false;
   StringPiece partially_fixed(copy, len);
 
   string curdir;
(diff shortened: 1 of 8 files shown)


src/includes_normalize-win32.cc
```diff
@@ -96,8 +96,12 @@ string IncludesNormalize::Relativize(StringPiece path, const string& start) {
 
 string IncludesNormalize::Normalize(const string& input,
                                     const char* relative_to) {
-  char copy[_MAX_PATH];
+  char copy[_MAX_PATH + 1];
   size_t len = input.size();
+  if (len > _MAX_PATH) {
+    Warning("path too long '%s'\n", input.c_str());
+    return input;
```
martine: Should we abort here?

src/includes_normalize-win32.cc
```diff
@@ -96,8 +96,12 @@ string IncludesNormalize::Relativize(StringPiece path, const string& start) {
 
 string IncludesNormalize::Normalize(const string& input,
                                     const char* relative_to) {
-  char copy[_MAX_PATH];
+  char copy[_MAX_PATH + 1];
   size_t len = input.size();
+  if (len > _MAX_PATH) {
```
martine: >= maybe?

src/includes_normalize-win32.cc
```diff
@@ -96,8 +96,12 @@ string IncludesNormalize::Relativize(StringPiece path, const string& start) {
 
 string IncludesNormalize::Normalize(const string& input,
                                     const char* relative_to) {
-  char copy[_MAX_PATH];
+  char copy[_MAX_PATH + 1];
```
martine: There's other use of MAX_PATH in this file.  Maybe it'd be good to wrap GetFullPathName as a helper function so that none of the other code is manually managing buffers?

src/includes_normalize-win32.cc
sgraham: Do you mean using >= in the comparison? I think it's correct as is, I made the buffer _MAX_PATH + 1 so that it can handle _MAX_PATH-lengthed paths which is as large as real paths should be.

martine: Whoops, you are right.  I was thinking >= len or something.  Two problems in computer science I guess.

src/includes_normalize-win32.cc
sgraham: That seems fine also, up to you/Nico. One drawback is that I don't think we can do "death tests" with the non-gtest-test.h, but I could promote to an error return and print the message in the parent if that's preferable.

martine: I think the rule is: if we will do the wrong thing from this point on it's good to quit loudly.  But if this instead has a mild consequence, like that we will always rebuild the file or something, then it's probably fine to warn.

sgraham: I think there's potentially bad side-effects. It also indicates something pretty screwed up with the build, so I switched it to be fatal.

src/msvc_helper-win32.cc
```diff
@@ -95,9 +95,15 @@ string CLParser::Parse(const string& output, const string& deps_prefix) {
...
     if (!include.empty()) {
-      include = IncludesNormalize::Normalize(include, NULL);
-      if (!IsSystemInclude(include))
-        includes_.insert(include);
+      string normalized;
+      string err;
+      if (!IncludesNormalize::Normalize(include, NULL, &normalized, &err)) {
+        Error("failed to normalize path: %s: %s\n", include.c_str(),
+              err.c_str());
+        exit(1);
```
nico: Does Fatal() not work in this file?

sgraham: Done.

nico:
Thanks, looks fine. I personally prefer if random code wouldn't call exit (or Fatal) and I'd thread the error return through two more layers (to MSVCHelperMain / Builder::ExtractDeps) where there's existing exit hatches, but since this'll likely ~never happen in practice I suppose this is ok too.

nico:
Thanks!

nico:
Wait, did I just close this and not merge?

nico:
…yeah, looks like it :-(

sgraham:
Doh! I didn't notice in the barrage of spam that github sends.

PR #974, opened 2015-06-16
Add missing member initializer in Edge constructor

bradking:
In commit v1.1.0^2~15^2~20 (stub out an api and de-constify Pool,
2012-10-03) the Edge "pool_" member was added as a raw pointer but the
initializer was accidentally left out of the constructor.  Add it now.


nico:
Thanks!

In practice, i think all Edge objects are created by State::AddEdge() which initializes this field so it shouldn't make a difference in practice, but it's nicer.

PR #979, opened 2015-06-22
Bootstrap out of source

hackhaslam:
Reopening #971. The fix involves emulating the same kind of shell escaping as ninja in the configure script bootstrap compile commands. Only checks for the bare minimum to fix the case of paths containing spaces. Ninja checks for a bunch of other special characters. The inline.sh command also needed to be quoted in the generated build script.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -28,7 +28,8 @@ import string
 import subprocess
 import sys
 
-sys.path.insert(0, 'misc')
+sourcedir = os.path.dirname(os.path.realpath(__file__))
+sys.path.insert(0, os.path.join(sourcedir, 'misc'))
 import ninja_syntax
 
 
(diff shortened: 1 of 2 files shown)


bradking:
I've confirmed that commit 5d4473d548ec839ee2c5acc006db682afad0bd15 resolves the problem I reported in #971.  With that version I can bootstrap in-source or out-of-source with spaces in both paths.  Thanks!

nico:
Does anyone have opinions on if we should take this for 1.6? Seems pretty simple, on the other hand I sent out the 1.6 call for testing already.

@hackhaslam, @bradking: Would it make your life much easier if this was in 1.6?

hackhaslam:
No, it wouldn't matter much to me either way.

nico:
Alright, with 1.6 out let's try this again :-)

martine:
I notice this makes all the source paths in the tree effectively absolute paths.

```
sigint:~/projects/ninja$ ./ninja src/build.cc^
ninja: error: 'src/build.cc' has no out edge
sigint:~/projects/ninja$ ./ninja `pwd`/src/build.cc^
ninja: no work to do.
```

I think this is minor but kinda undesirable, especially since this configure script is sort of a demonstration of best ninja principles.  I will think about how to best fix it.

martine:
I think if we make `sourcedir = .` happen in the common (non-out-of-source-tree) case it fixes it, because ninja knows to path-normalize `./src/build.cc` into just `src/build.cc`.

hackhaslam:
Yeah, I agree that it would be nicer to detect when bootstrapping in-source and use . instead.

Issue #987, opened 2015-07-09
`restat` is broken when rebuilding build.ninja in 1.6.0


jimporter:
I updated to ninja 1.6.0 and found that `restat` commands produce an infinite loop. I didn't see anything in the release notes to suggest that `restat` had changed, but if I'm doing something wrong, let me know.

Below is a sample build.ninja that reproduces the issue. I don't think the `PHONY` part is actually necessary, but I included it just in case, since that's necessary for the real-world scenario this is used in.

PS: Sorry for missing the call for testing before 1.6.0 was released!

```
rule rescan
  command = echo rescanning...
  generator = 1
  restat = 1

rule regenerate
  command = echo regenerating...
  generator = 1

build PHONY: phony
build .bfg_watch: rescan | PHONY
build build.ninja: regenerate | .bfg_watch
```

nico:
I'd guess this is "generators can now run more often than twice if needed (#908)" from the release notes. Your generator probably doesn't write everything it promises to write, and ninja now runs generator rules until they're clean (see that issue and the mail thread linked from there) this bug is exposed.

I'm guessing you can fix things by tweaking your generator.

martine:
Perhaps we should implicitly restat the generator outputs and die if the command didn't regenerate them or something?

jimporter:
I tried removing the `generator` attribute from the rescan rule, but it didn't change anything. In 1.5.3, what would happen is it would run the rescan rule once, see that nothing had changed, and then continue on with the build. Specifically, this means that I was never (to my knowledge) making the generators run twice.

The generator I'm using for this is my own project, [bfg9000](https://github.com/jimporter/bfg9000). The general idea is that the rescan rule checks to see if any glob patterns in the source build file have changed, and if so, to force the regenerate rule to run and rebuild the build.ninja file. Normally, none of the globs have changed, so it should run the rescan and then recognize that it doesn't need to do anything else.

jimporter:
Oh, and the main reason I was using `generator = 1` for the rescan rule (which doesn't actually generate a build file itself) was because I wanted to ensure that it didn't get removed when running `ninja -t clean`. It's possible I don't need that at all, but perhaps it would be useful to have a way to say "don't clean things made by this rule".

jimporter:
> Specifically, this means that I was never (to my knowledge) making the generators run twice.

[...]

PR #989, opened 2015-07-13
Add support for build statement implicit outputs

bradking:
Some build rules produce outputs that are not mentioned on the command
line but that should be part of the build graph.  Such outputs should
not be named in the `$out` variable.  Extend the build statement syntax
to support specification of implicit outputs using the syntax
`| out1 out2` after the explicit outputs and before the `:`.

For example, compilation of a Fortran source file `foo.f90` that defines
`MODULE FOO` may now be specified as:

    rule fc
      command = f95 -c $in -o $out
    build foo.o | foo.mod: fc foo.f90

The `foo.mod` file is an implicit output generated by the compiler based
on the content of the source file and not mentioned on the command line.

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -689,6 +689,10 @@ A file is a series of declarations.  A declaration can be one of:
    Order-only dependencies may be tacked on the end with +||
    _dependency1_ _dependency2_+.  (See <<ref_dependencies,the reference on
    dependency types>>.)
++
+Implicit outputs _(available since Ninja 1.7)_ may be added before
+the `:` with +| _output1_ _output2_+ and do not appear in `$out`.
+(See <<ref_outputs,the reference on output types>>.)
 
 3. Variable declarations, which look like +_variable_ = _value_+.
 
@@ -872,6 +876,27 @@ interpretation of the command (such as the use of `&&` to chain
 multiple commands), make the command execute the Windows shell by
 prefixing the command with `cmd /c`.
 
+[[ref_outputs]]
+Build outputs
+~~~~~~~~~~~~~
+
+There are two types of build outputs which are subtly different.
+
+1. _Explicit outputs_, as listed in a build line.  These are
+   available as the `$out` variable in the rule.
++
+This is the standard form of output to be used for e.g. the
+object file of a compile command.
+
+2. _Implicit outputs_, as listed in a build line with the syntax +|
+   _out1_ _out2_+ + before the `:` of a build line _(available since
+   Ninja 1.7)_.  The semantics are identical to explicit outputs,
+  the only difference is that implicit outputs don't show up in the
+  `$out` variable.
++
+This is for expressing outputs that don't show up on the
+command line of the command.
+
 [[ref_dependencies]]
 Build dependencies
 ~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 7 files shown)


bradking:
For reference, this was previously discussed on the mailing list in two threads:

* [Any interest in an 'implicit outputs' feature?](https://groups.google.com/forum/#!topic/ninja-build/dNrDg8KiNmE)
* [Adding Ninja features needed for Fortran](https://groups.google.com/forum/#!topic/ninja-build/tPOcu5EWXio)

While Fortran support is a motivating example, this is a general-purpose feature useful on its own.

jimon:
Also could be useful when we link .dll on windows and .lib magically appears after link :)

nicolasdespres:
We could add an option to the 'clean' tool to include/exclude the implicit
outputs from the list of files to remove. This could help in use case such
as the one mentioned in #987.

The patch looks more simple than I would have though. That's great!

On Tue, Jul 14, 2015 at 10:44 AM, Dmitry Ivanov <notifications@github.com>
wrote:

> Also could be useful when we link .dll on windows and .lib magically
> appears after link :)
>
> —
> <https://github.com/martine/ninja/pull/989#issuecomment-121168903>.
>



-- 
Nicolas Desprès

bradking:
@nicolasdespres was your comment meant for me to address with an extension to this topic or as a proposal for further work after this is integrated?

nicolasdespres:
A proposal for further work after this is integrated.

martine:
You can implement this by
1) always list all outputs in the "build" line
2) using something other than $out in your command line

Doing something like that is a relatively standard pattern in Ninja.
It's a little annoying here because I guess you can't override $out in your rule.  It'd be nice if you could do
```
build foo bar: thinger baz
  out = foo
```
So that the `thinger` rule could always use `$out`.

nico:
Since it's brought up every now and then and the patch is small, merging this might be a good idea just because people seem to expect to be able to do this. If we were to start from scratch, I think the conclusion is we shouldn't have implicit inputs either since they can be emulated like Evan described above.

bradking:
Adding implicit inputs here is a preliminary step toward the [Fortran support](#issuecomment-121039791) I'm working on.  That will involve dynamically adding implicit outputs during the build (e.g. after discovering a Fortran source compilation will produce `foo.mod`).  Once that happens then `$out` will not be a predictable or usable value if the dynamically-added outputs are not implicit.

> we shouldn't have implicit inputs either

IIUC the purpose of implicit inputs is so that a depfile can add them without affecting `$in`.  Here we're building the equivalent for `$out`.

nico:
Thanks for the patch! This looks good to me. Since I dragged my feet on this so long, it has conflicts by now -- I'll resolve them tomorrow night unless you beat me to it. Very sorry about the delay.

bradking:
Rebased on `master` to resolve conflicts.  I also added more test cases to `manifest_parser_test.cc`.

nico:
Thanks for the patch!

(To remind myself in the future: The reason this was merged instead of just recommending setting "my_custom_out = blah" on an edge if this is needed is because of -t clean, and because people asked for it fairly often – probably because it's consistent with implicit inputs, which technically also aren't needed and could have a "my_custom_in" instead)

Issue #995, opened 2015-07-27
Fails to build with MSVC 2015


krf:
```
python configure.py --bootstrap
bootstrapping ninja...
warning: A compatible version of re2c (>= 0.11.3) was not found; changes to src/*.in.cc will not affect your build.
build.cc
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE\iosfwd(343): error C2220: Warnung wird als Fehler interpretiert, es wurde keine object-Datei generiert.
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE\iosfwd(343): warning C4577: 'noexcept' used with no exception handling mode specified; termination on exception is not guaranteed. Specify /EHsc
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE\iosfwd(343): note: to simplify migration, consider the temporary use of /Wv:18 flag with the version of the compiler with which you used to build without warnings
when running:  cl /nologo /Zi /W4 /WX /wd4530 /wd4100 /wd4706 /wd4512 /wd4800 /wd4702 /wd4819 /wd4355 /wd4091 /GR- /wd4267 /DNOMINMAX /D_CRT_SECURE_NO_WARNINGS /DNINJA_PYTHON="python.exe" /FS /Ox /DNDEBUG /GL -c src\build.cc /Fobuild\build.obj
```

(Sorry for some of the output being in German, but you get the idea.)

The warning causes an error b/c you've set /Wx in `configure.py`.

nico:
…wait, this is in system headers. Is it no longer possible to compile one's code without warnings without /EHsc in new MSVCs?

krf:
Sorry, didn't post the full log initially. Fixed that now.

But yes, it's a warning originating from system headers. Nothing you can fix without setting additional compiler flags.

kirbyfan64:
Exception handling modes passed to MSVC should usually be explicit anyway.

nico:
@kirbyfan64 What do you mean with that?

kirbyfan64:
@nico I'm surprised thia didn't come up before. Every time I use MSVC, it whines and whines until I pass some `/EH*` argument.

kirbyfan64:
s/thia/this/

PR #996, opened 2015-07-27
Don't set /Wx for MSVC

krf:
Make the code future proof: If a new compiler / Windows SDK pulls in a new warnings, we
don't automatically want to make the build fail.

A good explanation can be found here: https://blog.flameeyes.eu/2009/02/future-proof-your-code-dont-use-werror#gsc.tab=0

This option was causing https://github.com/martine/ninja/issues/995

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -285,7 +285,6 @@ if platform.is_msvc():
               '/nologo',  # Don't print startup banner.
               '/Zi',  # Create pdb with debug info.
               '/W4',  # Highest warning level.
-              '/WX',  # Warnings as errors.
               '/wd4530', '/wd4100', '/wd4706',
               '/wd4512', '/wd4800', '/wd4702', '/wd4819',
               # Disable warnings about passing "this" during initialization.


nico:
New toolchains don't get released that often. I think it's preferable to fix warnings over disabling warnings-as-errors (#995 might not have been reported without this flag).

krf:
Well, without warning-as-errors (`-Werror`) I would have been happy to begin with and the code would've just compiled. :)

More seriously, please consider the arguments in the referenced blog post: `-Werror` is for developers (feel free to turn it on personally, or for CI), but it really doesn't make sense in released code.

Just my two cents.

Feel free to abandon this change if you're not convinced.

martine:
One pattern I've seen is to enable it in debug builds and disable in release, since end users aren't building debug builds.  Not sure if that works here.

paulsapps:
I think leaving it on makes sense, new tool chains are rare, and the warnings should probably be fixed. Or at least looked at and possibly disabled (such as ones telling you that the behavior of the compiler has changed, this could be an issue).

krf:
Again, a hard-coded `-Werror` in your build system is just an annoyance for users trying to compile your project. You can of course insist that code compiles *without* errors on platforms you test on regularly (i.e. what the CI covers / your dev machine => thus set it there), but you can't assume that it works for every compiler/platform combination out there.

krf:
(Feel free to abandon, I don't have strong feelings about this)

martine:
To restate krf's point: at least on non-Windows, I've seen -Werror be a real pain for end users because there's so much variance in compilers.  I vaguely remember there were even cases where the warning itself was a compiler bug that was fixed in later versions of the compiler so we didn't want to change the Chrome code to work around it.  So the build instructions have all this info about how to work around it: https://code.google.com/p/chromium/wiki/LinuxBuildInstructions#Sample_configurations .

paulsapps:
Having the option to turn it off with it on by default seems like the best option, then as you say the CI and tested platforms/compilers have zero warnings. And anyone who just wants to get it compiled can flick the switch to ignore them.

nico:
I think so far having warnings on errors has done more good than harm: It caused us to quickly fix this warning, and else we wouldn't have noticed the warning for a long time and there would have been build noise. We can revisit this if the compiler landscape on Windows becomes more diverse (e.g. if clang-cl becomes very popular).

PR #999, opened 2015-07-31
Print status when edge finishes on dumb terminals

colincross:
On smart terminals ninja prints the status line both before
and after running a command, reusing the same line if possible.
On a dumb terminal that doesn't support reusing the line, it
only prints the status before starting the command, but prints
the output of the command when the command finishes, by which
point other commands may have started and printed their status
line.  This makes it impossible to determine what command
produced a line of output.

Modify BuildEdgeStarted to only print the status line if the
command is going to lock the console, or if ninja is running
on a smart terminal.  Modify BuildEdgeFinished to always
print the status line unless the command locked the console,
in which case the status was already printed and no other
command can have printed any lines.

The end result will be dumb terminal output that much more
closely matches smart terminal output.  One disadvantage is
that dumb terminals won't show anything when starting a
command, making it harder to tell what commands are currently
running, but I expect most interactive uses of ninja will use
a smart terminal.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -96,7 +96,8 @@ void BuildStatus::BuildEdgeStarted(Edge* edge) {
   running_edges_.insert(make_pair(edge, start_time));
   ++started_edges_;
 
-  PrintStatus(edge);
+  if (edge->use_console() || printer_.is_smart_terminal())
+    PrintStatus(edge);
 
   if (edge->use_console())
     printer_.SetConsoleLocked(true);
@@ -121,7 +122,7 @@ void BuildStatus::BuildEdgeFinished(Edge* edge,
   if (config_.verbosity == BuildConfig::QUIET)
     return;
 
-  if (!edge->use_console() && printer_.is_smart_terminal())
+  if (!edge->use_console())
     PrintStatus(edge);
 
   // Print the command that is spewing before printing its output.


nico:
I'm tentatively merging this as discussed on the "[rfc] changing ninja's progress output order when writing to a file?" thread.

sgraham:
I think this might have regressed output on the Chrome bots. Dirk noticed un-nice output on https://build.chromium.org/p/chromium.linux/builders/Linux%20Builder%20%28dbg%29%2832%29/builds/52253/steps/compile/logs/stdio
where the failure happens and then a whole bunch of "completed" lines push the actual failure way back in the log. (Note that they're running -j50.)

Maybe because the jobs are finishing at that point? Seems like we might need a special case for failure if we want to keep this ordering of output.

sgraham:
(Should have been @dpranke above.)

colincross:
Before this patch output from commands that were running concurrently with failed command would appear after the output of the failed command, with no status line to separate the outputs of the different commands.  Chromium probably never ran into this issue since your rules generally have no output.  After this patch every command output is preceded by its status line, but that means that failures are followed by status lines for all concurrent commands.

Hiding the status lines for concurrent commands that have no output doesn't seem right, since those rules ran and completed.  We could delay printing the error until all concurrent commands have finished, but then if you had an error concurrent with a long command you might get the error much later than you otherwise could have seen it.

On Android our commands are so noisy that identifying the build output from a failing command is impossible without this patch.

dpranke:
I think delaying the error is probably okay; I don't know how often you really need to see the failures in real time or the delay would be long enough to matter. I agree that the change is generally an improvement and I would like to avoid hiding the output from completed edges as well.

sgraham:
I think it'd be OK to delay the error, but only in dumb-mode; that would be annoying locally in smart-terminal mode where there's often a long unrelated link finishing.

dpranke:
sure, only in not-a-tty-mode is fine by me.

wjakob:
I just ran into this when 'ninja -v' did not print started jobs anymore (which run for many minutes, so there is basically no output at all). This is pretty annoying -- are there any plans to offer to revert to the previous behavior as an option?

nico:
FWIW I find this annoying in -v mode too, so I agree it'd be nice to do something there.

PR #1007, opened 2015-08-19
Support for AIX

mikesep:
These changes add support for the AIX platform.

In our AIX 7 environment, the `SubprocessTest.InterruptChild` test does not pass due to the `/bin/sh` shell being a really old version of `ksh`. I'll address the issue of using a different shell in a separate pull request.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -505,6 +505,9 @@ if platform.is_msvc():
 else:
     libs.append('-lninja')
 
+if platform.is_aix():
+    libs.append('-lperfstat')
+
 all_targets = []
 
 n.comment('Main executable is library plus main() function.')
--- a/configure.py
+++ b/configure.py
@@ -494,6 +494,8 @@ if platform.is_windows():
     objs += cc('getopt')
 else:
     objs += cxx('subprocess-posix')
+if platform.is_aix():
+    objs += cc('getopt')
 if platform.is_msvc():
     ninja_lib = n.build(built('ninja.lib'), 'ar', objs)
 else:
--- a/configure.py
+++ b/configure.py
@@ -58,11 +58,13 @@ class Platform(object):
             self._platform = 'bitrig'
         elif self._platform.startswith('netbsd'):
             self._platform = 'netbsd'
+        elif self._platform.startswith('aix'):
+            self._platform = 'aix'
 
     @staticmethod
     def known_platforms():
       return ['linux', 'darwin', 'freebsd', 'openbsd', 'solaris', 'sunos5',
-              'mingw', 'msvc', 'gnukfreebsd', 'bitrig', 'netbsd']
+              'mingw', 'msvc', 'gnukfreebsd', 'bitrig', 'netbsd', 'aix']
 
     def platform(self):
         return self._platform
@@ -89,6 +91,9 @@ class Platform(object):
     def is_solaris(self):
         return self._platform == 'solaris'
 
+    def is_aix(self):
+        return self._platform == 'aix'
+
     def uses_usr_local(self):
         return self._platform in ('freebsd', 'openbsd', 'bitrig')
 
@@ -96,7 +101,9 @@ class Platform(object):
         return self._platform in ('linux', 'openbsd', 'bitrig')
 
     def supports_ninja_browse(self):
-        return not self.is_windows() and not self.is_solaris()
+        return (not self.is_windows()
+                and not self.is_solaris()
+                and not self.is_aix())
 
 
 class Bootstrap:
@@ -344,6 +351,8 @@ if platform.is_mingw():
     cflags.remove('-fvisibility=hidden');
     ldflags.append('-static')
 elif platform.is_solaris():
+    cflags.remove('-fvisibility=hidden')
+elif platform.is_aix():
     cflags.remove('-fvisibility=hidden')
 elif platform.is_msvc():
     pass
--- a/configure.py
+++ b/configure.py
@@ -105,6 +105,8 @@ class Platform(object):
                 and not self.is_solaris()
                 and not self.is_aix())
 
+    def can_rebuild_in_place(self):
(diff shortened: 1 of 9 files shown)


src/build_log.cc
```diff
@@ -12,6 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// On AIX, inttypes.h gets indirectly included by build_log.h.
+// It's easiest just to ask for the printf format macros right away.
+#ifndef _WIN32
```
nico: Should this check for _WIN32 or _MSC_VER? Would MinGW want the __STDC_FORMAT_MACROS define?

configure.py
```diff
@@ -633,9 +647,11 @@ def has_re2c():
...
-        subprocess.check_call('ninja.bootstrap.exe%s' % verbose, shell=True)
+        final_exe = 'ninja.exe'
     else:
-        subprocess.check_call('./ninja%s' % verbose, shell=True)
+        bootstrap_exe = './ninja_bootstrap'
+        final_exe = './ninja'
+    if os.path.exists(bootstrap_exe):
+        os.unlink(bootstrap_exe)
+    os.rename(final_exe, bootstrap_exe)
+    subprocess.check_call(bootstrap_exe + verbose, shell=True)
```
nico: The idea on the lhs is that Windows can't write to a binary that's in use, so we can't use ninja.exe to build ninja.exe and we need to copy it. On other platforms this isn't true, so we don't want to do it on other platforms. (If AIX needs it, I guess we need to do it there too, but we don't want to do it on non-Win non-AIX in that case.)

src/build_log.cc
mikesep: I simply moved the `#define` up from lower down the file. It was previously guarded by `#ifndef _WIN32`.

configure.py
mikesep: I can split this into if-windows-else-if-aix-else, but I felt that was getting overly platform-specific for little reason. Is there a particular benefit for overwriting the `ninja` executable in place? It's nice that it works on most platforms, but then it's hard to compare the results of the bootstrapped vs. self-hosted builds (for example).

nico: Yes, it has the benefit that you're manually testing the ninja binary you're working on all the time.

nico:
Can you investigate a bit why the segfault happens?

configure.py
mikesep: Maybe it would help me understand better if you described your workflow as a ninja developer -- maybe it's different from how I've been working. I have a script that runs `./configure.py --bootstrap`, followed by `./ninja ninja_test` and `./ninja_test`. To me, the impact of not overwriting the bootstrap first-pass result is just another file on disk. With the rename of the first executable as `ninja_bootstrap` and the self-built result as `ninja`, no one will run the first one by accident.

It's looking like the overwrite is definitely a problem on AIX, and my teammates suggest that isn't a great idea other platforms, so I'd like to understand what your usage pattern is.

mikesep:
Sorry for the lack of response here. I investigated the "segfault" a bit more, and it turns out that that was a bad description. In our AIX environment, I see different behaviors depending on the filesystem where the executable file runs. On NFS, I get `Illegal instruction (core dumped)` when I overwrite an executable file that's running. On non-NFS (SAN), AIX stops me from overwriting the file:
```
$ cp looper.cpp /tmp/looper
cp: try to overwrite '/tmp/looper', overriding mode 0755 (rwxr-xr-x)? y
cp: cannot create regular file '/tmp/looper': Text file busy
```

configure.py
martine: Running configure --bootstrap builds ninja twice.  Once you have the first ninja binary you can just run it to have it rebuild itself (at least on non-Windows, non-AIX platforms).

mikesep: Sure, but what is the disadvantage to renaming the first binary before running it the second time to build the final binary? Why do we need to special-case this logic on Windows and AIX -- why not just make that the norm?

martine: I see your point.  I think I agree.  The one question I have is what to do with the ninja_bootstrap binary after bootstrapping.  On non-Win/AIX, we don't need it after bootstrapping, not sure if we should clean up or not.

mikesep: Is it useful to keep both outputs for possible comparison? Have you ever seen that the bootstrapped exe behaves differently from the final exe?

If not, then I don't think it's any more useful on Win/AIX to keep `ninja_bootstrap` around, since it is always recreated each time you run the bootstrap script. We could auto-delete it on all platforms by adding `os.unlink(bootstrap_exe)` after `subprocess.check_call`, since `check_call` will raise an exception if the process fails.

mikesep:
Hi @nico and @martine, I'd appreciate if you could take another look here. I'm still pretty certain that most of these changes are good.

Per my [last comment](https://github.com/martine/ninja/pull/1007#discussion_r39198952) in the unlink subthread, are you okay with that approach, and should I add an `os.unlink(bootstrap_exe)` line?

configure.py
nico: I usually run the bootstrap script once on Windows, and then build using `ninja_bootstrap.exe` all the time while developing. (Every now and then, I then copy `ninja.exe` over `ninja_bootstrap.exe`.)

mikesep: Ah, that makes sense, since the bootstrapped ninja will be able to do a smarter build than the bootstrap script, which just builds naïvely. As far as I can tell, that would be an argument for accepting the code as proposed in the PR, which doesn't unlink `bootstrap_exe` after it has been executed.

mikesep:
Force-pushed an update to the last commit to avoid the extra `ninja.bootstrap` where not needed.

nico:
Thanks for sticking with us, very sorry about the delay, and thanks for tirelessly iterating on this.

mikesep:
Thanks for merging! :smile:

PR #1008, opened 2015-08-27
Add null.o to .gitignore

Qix-:
This file seems to be popping up across all builds of Ninja, dirtying up submodules and such.

On Mac Yosemite, clang info

```
Apple LLVM version 6.1.0 (clang-602.0.53) (based on LLVM 3.6.0svn)
Target: x86_64-apple-darwin14.1.0
Thread model: posix
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ TAGS
 /doc/doxygen
 /gtest-1.6.0
 *.patch
+/null.o
 
 # Eclipse project files
 .project


sgraham:
Seems to be because of
https://github.com/martine/ninja/blob/master/configure.py#L328

And indeed:

[scottmg:~/work/ninja ((30d1d7b...))]$ rm null.o
rm: null.o: No such file or directory
[scottmg:~/work/ninja ((30d1d7b...))]$ g++ -fdiagnostics-color -c -x c++
/dev/null
[scottmg:~/work/ninja ((30d1d7b...))]$ ls -l null.o
-rw-r--r--  1 scottmg  5000  184 Aug 26 19:28 null.o
[scottmg:~/work/ninja ((30d1d7b...))]$

Qix-:
Does opening to `/dev/null` in python create `null.o`? O.o

sgraham:
No, I think os.devnull works fine in python, clang is guessing a file name
for output and so makes an empty .o file.

Qix-:
Oh I see, interesting.

PR #1009, opened 2015-08-27
Don't output null.o during configure

sgraham:
Fixes #1008.


nico:
Thanks!

Qix-:
Thanks! :dancer:

PR #1011, opened 2015-08-29
Added EBNF grammar

jimon:
It's based on [grako](https://pypi.python.org/pypi/grako) parser syntax, but could be useful for everyone :)


martine:
This is neat, and I appreciate your submission, but I don't think we want to commit to maintaining this (and verifying it is an accurate representation of the syntax).  So I think it'd be best to either put this on the wiki or on a blog somewhere that an interested person could find it via Google.

PR #1014, opened 2015-09-03
Optimize StringPiece hash map

moroten:
Replace strncmp with memcmp to improve runtime performance.

Sorry for not having any figures to prove it, but I tried this ages ago.


martine:
Wow, I am surprised we didn't already do this!

nico:
Thanks for the patch! I don't remember this showing up in profiles, and I can't measure any perf change with this (I'm doing empty chromium builds on OS X to test), but I guess merging this might stop someone from wondering if this change might help in the future.

PR #1015, opened 2015-09-03
Describe why to use relative paths

moroten:
This commit updates the documentation to avoid rookie mistakes like I did with relative and absolute paths.

https://groups.google.com/d/msg/ninja-build/tIcKNLWWKzo/ia5cYsXk8lcJ
https://groups.google.com/d/msg/ninja-build/4ErJT6vKR2M/psrYggkQCgAJ


nico:
Thanks for the patch!

PR #1016, opened 2015-09-03
Describe how to make a phony rule always up to date

moroten:
A phony rule with now input is always out of date. Describe how to make a
rule always up to date.


doc/manual.asciidoc
```diff
@@ -445,6 +445,14 @@ without any dependencies, the target will be considered out of date if
 it does not exist.  Without a phony build statement, Ninja will report
 an error if the file does not exist and is required by the build.
 
+To create a rule that never rebuilds, make use a build rule without any input:
```
martine: Typo: "make use"

PR #1017, opened 2015-09-04
Fix releasing pools for failing edges

moroten:
When an edge finishes building, it should be release from its pool. Make sure that this also is the case when an edge fails to build.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -374,14 +374,16 @@ void Plan::ScheduleWork(Edge* edge) {
   }
 }
 
-void Plan::EdgeFinished(Edge* edge) {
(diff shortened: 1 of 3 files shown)


nico:
Thanks for the patch! Can you explain what goes wrong without this change?

moroten:
With a pool of size X, then 'ninja -k 0' exits after X failing jobs from that pool. The correct behaviour is to continue to execute until nothing more is possible to schedule.

bradking:
See #1126 which reconciles this with #1023.

nico:
Closing in favor of #1126 (which is based on this).

PR #1019, opened 2015-09-10
Improve const correctness

per-gron:
I'm toying around with the idea of making a fork of Ninja that does (distributed) caching, exploiting the fact that Ninja knows about the full dependency graph so it should be possible to cache executables and libraries as well as object files.

As part of that, I'm trying to understand the Ninja code base. One thing that makes the code a little bit hard to follow for me is how mutable `Node*`s and `Edge*`s are passed around everywhere, and that these objects contain a mixture of the immutable state that is parsed from the manifest file, dirty flags, dependency log ids and so on.

This PR is an attempt to make the code slightly easier to follow by explicitly marking some of the places that do nothing but read from the build graph.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -91,7 +91,7 @@ void BuildStatus::PlanHasTotalEdges(int total) {
   total_edges_ = total;
 }
 
-void BuildStatus::BuildEdgeStarted(Edge* edge) {
+void BuildStatus::BuildEdgeStarted(const Edge* edge) {
   int start_time = (int)(GetTimeMillis() - start_time_millis_);
   running_edges_.insert(make_pair(edge, start_time));
   ++started_edges_;
@@ -102,7 +102,7 @@ void BuildStatus::BuildEdgeStarted(Edge* edge) {
     printer_.SetConsoleLocked(true);
 }
 
-void BuildStatus::BuildEdgeFinished(Edge* edge,
+void BuildStatus::BuildEdgeFinished(const Edge* edge,
                                     bool success,
                                     const string& output,
                                     int* start_time,
@@ -238,7 +238,7 @@ string BuildStatus::FormatProgressStatus(
   return out;
 }
 
-void BuildStatus::PrintStatus(Edge* edge) {
+void BuildStatus::PrintStatus(const Edge* edge) {
   if (config_.verbosity == BuildConfig::QUIET)
     return;
 
@@ -261,11 +261,11 @@ void BuildStatus::PrintStatus(Edge* edge) {
 Plan::Plan() : command_edges_(0), wanted_edges_(0) {}
 
 bool Plan::AddTarget(Node* node, string* err) {
-  vector<Node*> stack;
+  vector<const Node*> stack;
   return AddSubTarget(node, &stack, err);
 }
 
-bool Plan::AddSubTarget(Node* node, vector<Node*>* stack, string* err) {
+bool Plan::AddSubTarget(Node* node, vector<const Node*>* stack, string* err) {
   Edge* edge = node->in_edge();
   if (!edge) {  // Leaf node.
     if (node->dirty()) {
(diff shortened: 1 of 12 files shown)


ariccio:
Nobody with commit privs has looked at this? I'm surprised...

martine:
In general, I like const correctness.  Normally when passing a non-mutable value we've tried to use const references to make that explicit.  Another way of saying that is that we always either have a plain pointer or a const ref, but never a const pointer or a plain ref.  So I wonder if some of these ought to be changed to references as well.  But I do know we maybe rely on pointer equality in some places around Node handling so maybe it's an exception.

ariccio:
I totally get the *no non-`const` references* ("plain ref"s) - Titus Winters made a great case foe that. What's the reasoning for *no non-`const` pointers*?

pereckerdal:
I don't really mind using const refs instead of pointers to const values* but I agree with @ariccio that it seems like a slightly odd principle. When I write my own C++ code I tend to almost never use raw pointers so I haven't really thought about it a lot though.

*: It's not possible to store references in STL containers. In that case pointers to const values are an option that is close at hand.

Let me know if you would like me to adjust the PR and if so how

src/build.cc
```diff
@@ -316,16 +316,16 @@ bool Plan::AddSubTarget(Node* node, vector<Node*>* stack, string* err) {
   return true;
 }
 
-bool Plan::CheckDependencyCycle(Node* node, const vector<Node*>& stack,
+bool Plan::CheckDependencyCycle(const Node* node, const vector<const Node*>& stack,
                                 string* err) {
-  vector<Node*>::const_iterator start = stack.begin();
+  vector<const Node*>::const_iterator start = stack.begin();
```
ariccio: Instead, would `auto start = stack.cbegin();` be nor succinct?

src/build.cc
```diff
@@ -340,7 +340,7 @@ bool Plan::CheckDependencyCycle(Node* node, const vector<Node*>& stack,
   }
 
   *err = "dependency cycle: ";
-  for (vector<Node*>::const_iterator i = cycle.begin(); i != cycle.end(); ++i) {
+  for (vector<const Node*>::const_iterator i = cycle.begin(); i != cycle.end(); ++i) {
```
ariccio: Same as above.

src/build_log.cc
```diff
@@ -139,11 +139,11 @@ bool BuildLog::OpenForWrite(const string& path, const BuildLogUser& user,
   return true;
 }
 
-bool BuildLog::RecordCommand(Edge* edge, int start_time, int end_time,
+bool BuildLog::RecordCommand(const Edge* edge, int start_time, int end_time,
                              TimeStamp restat_mtime) {
   string command = edge->EvaluateCommand(true);
   uint64_t command_hash = LogEntry::HashCommand(command);
-  for (vector<Node*>::iterator out = edge->outputs_.begin();
+  for (vector<Node*>::const_iterator out = edge->outputs_.begin();
```
ariccio: Same as above.

src/clean.cc
```diff
@@ -112,7 +112,7 @@ void Cleaner::PrintFooter() {
 int Cleaner::CleanAll(bool generator) {
   Reset();
   PrintHeader();
-  for (vector<Edge*>::iterator e = state_->edges_.begin();
+  for (vector<Edge*>::const_iterator e = state_->edges_.begin();
```
ariccio: Same as above.

src/clean.cc
```diff
@@ -120,7 +120,7 @@ int Cleaner::CleanAll(bool generator) {
     // Do not remove generator's files unless generator specified.
     if (!generator && (*e)->GetBindingBool("generator"))
       continue;
-    for (vector<Node*>::iterator out_node = (*e)->outputs_.begin();
+    for (vector<Node*>::const_iterator out_node = (*e)->outputs_.begin();
```
ariccio: Same as above.

pereckerdal:
@ariccio last time I checked Ninja did not use C++11. I agree auto would be nice but with my understanding that is out of scope for this PR

src/build.cc
jhasse: IIRC `cbegin()` returns `vector<Node* const>::…` though, not `vector<const Node*>::…`.

src/clean.cc
jhasse: Note to myself: Nope, `cbegin` is C++11.

jhasse:
Thanks for the PR, sorry it's been such a long time. Would you be willing to rebase it on current master? I would merge it then :)

jhasse:
In order to reduce the number of open PRs I'm closing this one. Feel free to reopen it.

PR #1021, opened 2015-09-15
Set output mtime of phony edges to the latest inputs

moroten:
This commit fixes issue #478.

Observed:
Real edges depending on a phony edge will not be marked as dirty or
rebuilt if the phony's (real) inputs are updated.

Expected:
An edge should always be rebuilt if its inputs or transitive inputs are
newer than the output's mtime.

Change:
Node::mtime_ was overloaded, 0 represented "does not exist". This change
disambiguates it by adding Node::exists_. Then to fix the observed
behaviour, Node::UpdatePhonyMtime was added to update the mtime if the
node does not exist.

Add tests BuildTest.PhonyUseCase# GraphTest.PhonyDepsMtimes.
Unit tests will also test for always-dirty behaviour if a phony rule has
no inputs.

---

This patch is a rebase of #634 onto the master branch of 2015-09-16.

### DIFF
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -28,7 +28,10 @@
 #include "util.h"
 
 bool Node::Stat(DiskInterface* disk_interface, string* err) {
-  return (mtime_ = disk_interface->Stat(path_, err)) != -1;
+  METRIC_RECORD("node stat");
+  mtime_ = disk_interface->Stat(path_, err);
+  exists_ = (mtime_ > 0) ? ExistanceStatusExists : ExistanceStatusMissing;
+  return mtime_ != -1;
 }
 
 bool DependencyScan::RecomputeDirty(Node* node, string* err) {
@@ -234,6 +237,13 @@ bool DependencyScan::RecomputeOutputDirty(Edge* edge,
               output->path().c_str());
       return true;
     }
+
+    // Update the mtime with the newest input. Dependents can thus call
+    //  mtime() on the fake node and get the latest mtime of the dependencies
+    if (most_recent_input)
+      output->UpdatePhonyMtime(most_recent_input->mtime());
+
+    // Phony edges are clean, nothing to do
     return false;
   }
 
--- a/src/graph.h
+++ b/src/graph.h
@@ -15,6 +15,7 @@
 #ifndef NINJA_GRAPH_H_
 #define NINJA_GRAPH_H_
 
+#include <algorithm>
 #include <string>
 #include <vector>
 using namespace std;
@@ -40,6 +41,7 @@ struct Node {
       : path_(path),
         slash_bits_(slash_bits),
         mtime_(-1),
+        exists_(ExistanceStatusUnknown),
         dirty_(false),
         dyndep_pending_(false),
         in_edge_(NULL),
@@ -47,6 +49,13 @@ struct Node {
 
   /// Return false on error.
   bool Stat(DiskInterface* disk_interface, string* err);
+
+  /// If the file doesn't exist, set the mtime_ from its dependencies
+  void UpdatePhonyMtime(const TimeStamp mtime) {
+    if (!exists()) {
+      mtime_ = std::max(mtime_, mtime);
+    }
+  }
 
   /// Return false on error.
   bool StatIfNecessary(DiskInterface* disk_interface, string* err) {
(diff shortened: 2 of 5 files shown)


clacke:
Is there something holding this PR back? The worries in the final comment on [PR #634](https://github.com/ninja-build/ninja/pull/634#issuecomment-26146626)?

nocnokneo:
Any updates on this PR?

lucasmeijer:
since @nico mentioned on the ninja mailinglist: "Pull requests that have lots of "that would be useful to me too" comments are more likely to be merged",   let this be my "useful for me too!"    the current situation requires my ninjagenerator to manually "unfold" all phony targets everywhere they are used, which makes parsing (and generating) the ninja file much slower than it needs to be.

davidgiven:
I've just discovered this while trying to figure out why my ninjafile isn't rebuilding my target when I touch a dependent file. I've been grouping them together with phonies. [sad face]

src/graph.h
```diff
@@ -55,20 +64,24 @@ struct Node {
...
   void ResetState() {
     mtime_ = -1;
+    exists_ = -1;
     dirty_ = false;
   }
 
   /// Mark the Node as already-stat()ed and missing.
   void MarkMissing() {
-    mtime_ = 0;
+    if (mtime_ < 0) {
```
jhasse: Please change to `== -1` for clarity.

src/graph.h
```diff
@@ -38,14 +39,22 @@ struct Node {
...
 
-  /// Return false on error.
+  /// If the file doesn't exist, set the mtime_ from its dependencies
+  void UpdatePhonyMtime(const TimeStamp mtime) {
+    if (!exists()) {
+      mtime_ = std::max(mtime_, mtime);
+    }
+  }
+
+  /// Return true if we needed to stat.
```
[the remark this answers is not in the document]
moroten: I agree, can't understand or remember why it was changed.

src/graph.h
moroten: Done

src/graph_test.cc
```diff
@@ -308,7 +308,7 @@ TEST_F(GraphTest, DepfileOverrideParent) {
 // Verify that building a nested phony rule prints "no work to do"
 TEST_F(GraphTest, NestedPhonyPrintsDone) {
   AssertParse(&state_,
-"build n1: phony \n"
+"build n1: phony\n"
```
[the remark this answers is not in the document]
moroten: No. Reverted.

jhasse:
This means we can drop https://github.com/ninja-build/ninja/blob/a65f0d42dc8820cc4d80561ddcd4b693c0adc892/src/disk_interface.cc#L200, right?

Can you change `exists_` to an enum? I would like to avoid all those magic numbers.

moroten:
When I have some time, next week or so, I'll change `exists_` to an enum. It also sounds reasonable to remove the special handling of `mtime=0`.

src/graph.cc
```diff
@@ -28,7 +28,10 @@
 #include "util.h"
 
 bool Node::Stat(DiskInterface* disk_interface, string* err) {
-  return (mtime_ = disk_interface->Stat(path_, err)) != -1;
+  METRIC_RECORD("node stat");
+  mtime_ = disk_interface->Stat(path_, err);
+  exists_ = (mtime_ > 0) ? ExistanceStatusExists : ExistanceStatusMissing;
```
jhasse: wait - this will be incorrect now that the Flatpak-hack is gone, right?

moroten: True. I've reverted that to the original `disk_interface.cc` code to have a branch that is working correctly. It is a too big change for me to refactor the interface of `DiskInterface::Stat()`, `git grep 'Stat(' | wc` gives 115 hits. I also think it should be done in a separate commit anyway.

src/graph.h
```diff
@@ -113,9 +126,19 @@ struct Node {
   /// Possible values of mtime_:
   ///   -1: file hasn't been examined
   ///   0:  we looked, and file doesn't exist
-  ///   >0: actual file's mtime
+  ///   >0: actual file's mtime, or the latest mtime of its dependencies if it doesn't exist
```
jhasse: I'm scared that this might result in a regression somewhere `mtime()` is used.

src/graph.cc
```diff
@@ -234,6 +237,13 @@ bool DependencyScan::RecomputeOutputDirty(Edge* edge,
               output->path().c_str());
       return true;
     }
+
+    // Update the mtime with the newest input. Dependents can thus call
+    //  mtime() on the fake node and get the latest mtime of the dependencies
+    if (most_recent_input)
```
jhasse: Please add {} for this if.

src/graph.h
```diff
@@ -48,6 +50,13 @@ struct Node {
   /// Return false on error.
   bool Stat(DiskInterface* disk_interface, string* err);
 
+  /// If the file doesn't exist, set the mtime_ from its dependencies
+  void UpdatePhonyMtime(const TimeStamp mtime) {
```
jhasse: Can you move the implementation to graph.cc?

jhasse:
Closing in favor of #1964.

PR #1023, opened 2015-09-20
Fix pool->current_use_ not being decremented when edges fail

zdave:
eg With this build file...

  pool failpool
    depth = 1
  rule fail
    command = fail
    pool = failpool
  build out1: fail
  build out2: fail
  build out3: fail
  build final: phony out1 out2 out3

...and "-k 0", we should run out1..3 sequentially before failing. Previously we
would fail after just running out1.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -374,24 +374,28 @@ void Plan::ScheduleWork(Edge* edge) {
   }
 }
 
-void Plan::EdgeFinished(Edge* edge) {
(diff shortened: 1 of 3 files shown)


sgraham:
Also in #1017 I think.

zdave:
Ah indeed. Should I just close this?

bradking:
See #1126 which reconciles this with #1017.

nico:
Closing in favor of #1126 (which has your test too).

PR #1025, opened 2015-09-21
Resurrect the 'rules' tool.

### LINKED ISSUE #1024: Feature request: more flexible rule name matching in the compdb tool
Currently, to use the compdb tool you need to know the exact name of the rules which you want to query. This works OK provided these names are predictable, but this not necessarily the case for some generators.

Most notably, in version 3.2 (I think) CMake's ninja generator has started generating rules of the form `<lang>_COMPILER__<directory>`. This means that in order to query the compilation database, you need to first scan the project directory tree, or resort to other hacks. CMake prior to 3.2 generates rules of the form `<lang>_COMPILER` which made them possible to query from generic tools - without specific knowledge of the project.

Hence this feature request is for more flexible matching of rule names against command line arguments. A minimally-useful implementation would allow prefix-only matching. Alternatively regex matching would also be useful.

nicolasdespres:
This tool is useful for writing shell completion script for tools
expecting a rule name as argument.

The tool was dropped by 34b46f28c.

Fix #1024.

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -283,6 +283,9 @@ target, show just the target's dependencies. _Available since Ninja 1.4._
 
 `recompact`:: recompact the `.ninja_deps` file. _Available since Ninja 1.4._
 
+`rules`:: output the list of all rules (eventually with their description
+if they have one).  It can be used to know which rule name to pass to
++ninja -t targets rule _name_+ or +ninja -t compdb+.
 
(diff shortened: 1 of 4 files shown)


jhasse:
Very useful IMHO. Do you want to look at the merge conflict or should I?

nicolasdespres:
Done fixing merge conflict.

jhasse:
Thanks!

PR #1026, opened 2015-09-22
Print finish status for interrupted edges

colincross:
Stay in the event loop when ninja receives SIGINT or SIGTERM so that
running commands that were interrupted can have their status printed
along with "interrupted by user".  Allows the end user to see what was
running at the time ninja was interrupted.


colincross:
Rebased to fix merge conflict

jonesmz:
@colincross Could you rebase this pull request against the latest master branch?

jhasse:
Part of this would be fixed by #2516, but you also print the output-so-far of commands that haven't finished, right?

PR #1028, opened 2015-09-26
Generate a shell script to debug failed commands

nicolasdespres:
As discussed [here](https://groups.google.com/forum/#!topic/ninja-build/ahcZQuMOs2w), it may be useful when investigating what is wrong in a long command line to have all the failed commands of the last run logged in a ready to edit and run shell script.

This PR generates such a script. Tests and documentation are included.


nico:
1.7 was released since that thread, and it changed output so that "FAILED:" is now on a separate line from the failed command, and the command is just a triple-click middle-click away…is this still needed?

nicolasdespres:
Well, the idea was to avoid the copy/paste step and to be able to edit the long commands in a true text editor rather than the shell readline editor. Maybe I should change my shell (zsh) and/or my terminal (iTerm2) because my triple-click does not select command lines that span over multiple lines in the terminal :-).

We can still put this feature in stand-by and wait to see if more people vote for it. I haven't recently worked on a project with long command lines.

jhasse:
I think it's your terminal, I doubt the shell has much influence over this.

Hope you're okay with me closing this, I think the tripple-click solution is sufficient (you can paste in an editor, too).

PR #1030, opened 2015-09-26
Builder no longer leaks BuildStatus object.

nicolasdespres:


nico:
Thanks for the patch!

I think ninja leaks most things at the moment – cleaning up everything orderly slows down runtime measurably and doesn't add much as long as ninja isn't used as a library (and if we ever do this, then we should only delete stuff in library mode), and cleaning up just some things makes it look like the remaining leaks are accidents. Maybe we should get rid of all of all deletes that aren't in loops.

nicolasdespres:
ok

PR #1033, opened 2015-10-05
Print output file on failure

colincross:
Modify the FAILED: output to provide the output files that failed to
build, followed by the failed command on the next line.  This makes the
failure much easier to read, as you can immediately see much shorter
name of the file that failed instead of trying to parse a very long
command line.  It also makes manually re-running the failed command much
easier because you can copy the whole line without ending up with the
FAILED: prefix.


martine:
I like it!

nico:
This seems like a good change to me. It's slightly backwards-incompatible, but it feels fairly low risk and like a good improvement. Thanks for the patch!

PR #1036, opened 2015-10-13
Reserve size of string before appending to it

itensionanders:
Pre-allocate a string for the entire file length at the start, to avoid having to grow the string in 64K buffers.

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -372,6 +372,9 @@ int ReadFile(const string& path, string* contents, string* err) {
   ::CloseHandle(f);
   return 0;
 #else
+  struct stat st;
+  if (stat(path.c_str(), &st) == 0)
+    contents->reserve(contents->length() + st.st_size);
   FILE* f = fopen(path.c_str(), "rb");
   if (!f) {
     err->assign(strerror(errno));


nico:
IIRC when I tried this I couldn't measure any performance difference due to this. Are you seeing a performance improvement due to this on your system?

ariccio:
Pre-allocating space for a standard library container, when you know in advance how much you'll need, is just good practice.

itensionanders:
I suppose it depends on the size of the files, and on the string implementation in the std lib. But it seemed that it in the worst case it would also copy those strings around when growing it and better off avoided ?
In some places it even seemed that it would be even better to *not* load it all into memory at once, but do the scanning and parsing directly from the disk buffer. But will need to do some actual benchmarks first...

martine:
This is adds an extra system call as well.  In the absence of data (about whether this improves performance or not) we should always prefer less code over more code.

martine:
(I also seem to recall that `stat` is extra-expensive on Windows, because to fill in the data structure it needs to retrieve data from multiple different places.  Elsewhere in Ninja when we want file metadata we use a different API in the Windows codepath.)

ariccio:
> I also seem to recall that  stat  is extra-expensive on Windows, because to fill in all the fields in the Unixy data structure it needs to retrieve data from multiple different places. 

Eh? This is a non-Windows change?

Benchmarks would be good :)

Side note: That function for hitting a single file's info is going to be unnecessarily slow, as you're retrieving the 8.3 filename, which is comically slow.

itensionanders:
Okay, will be back with some before/after timings... As noted you will need a rather big and massively parallel build in order to see any difference. And it sounds like most of what you do on Windows is slow ?

itensionanders:
Think that some of the concerns were valid... Even though it does help (a bit) with large files, it does do a bunch of extra stats for the *.d dep files (especially) - and most of those are indeed below 64K anyway.
So it would probably need some other approach, like a "hint" parameter - or maybe just reserving the input string since we append to it anyway. But just doing an unconditional stat is probably not the best ?

Issue #876, opened 2014-12-03
depfile parser could be smarter about linebreaks


nico:
```
rule cgt
   depfile = c.d
   command = touch a; touch b; 
      echo "a: /usr/bin/ls" > c.d; $
      echo "b: /usr/bin/ls" >> c.d

build a b: cgt
```

This generates a depfile looking like

```
a: /usr/bin/ls
b: /usr/bin/ls
```

Ninja currently thinks that `a` is the output, and `/usr/bin/ls`, `b` (ninja strips the trailing ':' for some reason), and `/usr/bin/ls` are inputs. Instead, it should only read things on the first line as inputs for `a` (with `"\\\n"` being a line continuation character).

One issue with this is that `gcc -MMD -MP -MF foo.d` output will then count as having multiple outputs in the depfile. If that's an issue in practice, maybe outputs with 0 inputs could be silently skipped.

PR #1039, opened 2015-10-19
Fix depfile parser handling of multiple rules 

bradking:
Currently we handle Makefile rules of the form:

```
out: in1 in2 in3
```

and the form:

```
out: in1 \
     in2 \
     in3
```

Teach the depfile parser to handle the additional form:

```
out: in1
out: in2
out: in3
```

This is also valid Makefile syntax and is the depfile format
generated by the Intel Compiler for Windows.


ilyapopov:
Also related: #704 and #876

bradking:
Discussion in #704 raises the possibility of indented rules.  Makefile syntax allows the outputs of a rule to be indented, e.g.

    # indented outputs
     out: in1
     out: in2
     out: in3

I extended the test suite to cover this case too.

bradking:
Discussion in #704 and #876 identified that phony rules generated by `-MP` need to be tolerated.  I've revised the change accordingly and added a test case.

bradking:
Rebased to resolve conflict.

bradking:
Rebased on `master`.

bradking:
Rebased on `master` after 1.7.1 release.

mikesep:
These changes are very similar to ones I was about to propose to handle depfiles from the Sun Studio C++ compiler, which generates depfiles that follow the pattern Brad mentioned in the opening post of this PR.

mikesep:
@nico @bradking Is there anything I can do to help get this merged?

bradking:
@mikesep I'd like to see this merged too, but I don't have access.

@nico IIRC you had concerns about drawing a line for how much arbitrary Makefile syntax will be accepted by the depfile parser.  While I agree supporting variable replacements and such is on the other side of such a line, I don't think splitting dependencies into multiple rules is too far.  There are real compilers that generate this syntax and Ninja currently does not support them.

bradking:
Rebased on `master` after 1.7.2 release.

poelmanc:
@nico @bradking Several folks at my company have been using Brad's modified version of ninja that includes this feature as well as #1111 for many months, on projects with several hundred libraries (mostly C++, but with a little Fortran and Javascript), without any problems. But others are reluctant to adopt ninja until these capabilities appear in an official version. We'd love to see these changes merged so we can migrate to ninja company-wide.

janusw:
@bradking Could you please comment on the relevance of this feature for ninja's Fortran support (#1265)?

Is it only relevant for ifort, or also for other Fortran compilers like gfortran?
Is the recent merge of #1111 enough to make cmake+ninja work with Fortran projects (using gfortran)?

Thanks for all your great work!

bradking:
@janusw #1111 was only one step along the way.  The next step is #1294, and there will be several more to follow.  I haven't yet had time to rebase our branch on that and update our readme.  #1039 is necessary only to support `ifort`, but that is an important player.

janusw:
Thanks for the clarification!

bradking:
Rebased on `master`.

niecore:
meanwhile pre build binarys for windows can be found here: 
https://github.com/niecore/ninja-advDepFileSupport-prebuild

If there are any legal issues, contact me and the repo will be removed.

Meinersbur:
This seems to be a straightforward extension required to support some compilers, how does it take that long to merge?

Thanks @bradking for the great work.

emmenlau:
Sorry to bump up this thread, but I am under the impression that this PR is one of the most suitable solutions to fix support for the Intel Compiler on MSVC. Currently the use of `/showIncludes` can randomly break builds with icl on Windows. See for a more detailed discussion https://gitlab.kitware.com/cmake/cmake/issues/17890 or https://github.com/ninja-build/ninja/issues/972

Could this PR be merged, or are there any blockers against meging? Thanks a lot!

christensson:
This pull-request also fixes builds using the TI cl6x compiler which outputs the dependency file format described above with multiple rules.

Any status

christensson:
This is a good PR since it adds support for more compilers.

What is stopping this MR?
Is ninja dead?

bradking:
For reference, I maintain a ninja branch with binary releases that include this PR.  See [Kitware/ninja releases](https://github.com/Kitware/ninja/releases/).

christensson:
@bradking Do you know what is currently stopping this MR, anything I can do to help?

bradking:
@christensson see my [earlier note](https://github.com/ninja-build/ninja/pull/1039#issuecomment-260445395) on the upstream concerns.

christensson:
OK, I understand the concerns for implementing a full-fledged makefile parser. However I think that this PR adds real value and since it allows more compilers to be used with ninja.

@nico Any chance you would reconsider merging this PR?

bradking:
It was discovered [here](https://bugs.gentoo.org/667952) that there are old `rustc` versions that write

```Makefile
out: in1 in2 in3
out.d: in1 in2 in3
```

This is equivalent to

```Makefile
out out.d: in1 in2 in3
```

which ninja rejects with "depfile has multiple output paths".  With the change originally proposed in this PR, the two-line version above is also rejected.  Previously we silently tolerated it and incorrectly recorded `out.d` as a dependency of `out`.

In order to retain compatibility with such buggy tools, I've updated this PR to also add a `-w depfilemulti={err,warn}` option so that we can merely warn about this case by default.

nico:
Alright, let's get this in. Do you think we should default to err or warn for that new flag?

bradking:
I've updated the branch to rebase on `master` to resolve conflicts.

> Do you think we should default to err or warn for that new flag?

I think it should be `warn` by default just `dupbuild` was originally.  That will give people a chance to find/fix offending depfile generators.

nico:
Thanks for the patch!

vmatveic:
@bradking @emmenlau  sorry for disturbance, could you please advice how should I call "cmake -G Ninja" (should I specify some additional option?) to avoid including of /showincludes to the generated ninja project (we are using icl compiler on Windows)? As far as I understand it should be possible to do that after this PR merged, right? Really appreciate your help!

emmenlau:
Dear @vmatveic I found no other solution than to patch the ninja build files. My current solution is that I have everywhere
```
cmake -G Ninja ... && \
perl -pi -e "s@\s*/showIncludes@@g" rules.ninja
```
Its ugly but works like a charm, no problems ever since. Of course this is just a workaround until I have a release with the patch included. But it kept me going for quite some time now.

vmatveic:
@emmenlau , thanks a lot for your answer!!! Yes, we are also doing it in this way. Just last question. Do I understand correctly that we will NOT need this patch applied when we get fixed ninja version? And in this case /showIncludes will still be included to the rules but NO crash should happen?

bradking:
@vmatveic @emmenlau CMake will need to be taught that the new Ninja version (when released) can support the depfiles produced by the Intel Compiler for Windows.  I've opened [CMake issue 18855](https://gitlab.kitware.com/cmake/cmake/issues/18855) for further discussion and tracking.

Issue #1041, opened 2015-10-23
Add option to ninja_syntax.Writer.comment to prevent line breaks


avallee:
When outputting some comments I don't want some filenames to get broken onto several lines.

    >>> w = ninja_syntax.Writer(sys.stderr)
    >>> w.comment('Generated by /usr/local/home/avallee/averylong-hyphenated-pathnameindeeddontyouthink/itsveryincredible.py')
    # Generated by /usr/local/home/avallee/averylong-hyphenated-
    # pathnameindeeddontyouthink/itsveryincredible.py

NiklasRosenstein:
Oh, I see just now that the problem is only about breaking filenames. And I didn't see there was already a PR merged (why is the issue not closes?). My PR might be irrelevant :)

PR #1042, opened 2015-10-23
Fix wrapping file paths in Writer.comment

avallee:
Long file names, especially with hyphens will get incorrectly wrapped by
the comment method. Pass has_path=True to prevent this type of wrapping.

This is mainly so that longer path names can show up in comments on
their on line without breaking them up.

Should fix issue #1041

### DIFF
--- a/misc/ninja_syntax.py
+++ b/misc/ninja_syntax.py
@@ -21,8 +21,11 @@ class Writer(object):
     def newline(self):
         self.output.write('\n')
 
-    def comment(self, text):
-        for line in textwrap.wrap(text, self.width - 2):
+    def comment(self, text, has_path=False):
+        args = {}
+        if has_path:
+          args['break_long_words'] = args['break_on_hyphens'] = False
+        for line in textwrap.wrap(text, self.width - 2, **args):
             self.output.write('# ' + line + '\n')
 
     def variable(self, key, value, indent=0):
--- a/misc/ninja_syntax.py
+++ b/misc/ninja_syntax.py
@@ -22,10 +22,8 @@ class Writer(object):
         self.output.write('\n')
 
     def comment(self, text, has_path=False):
-        args = {}
-        if has_path:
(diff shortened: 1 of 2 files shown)


misc/ninja_syntax.py
```diff
@@ -21,8 +21,11 @@ def __init__(self, output, width=78):
     def newline(self):
         self.output.write('\n')
 
-    def comment(self, text):
-        for line in textwrap.wrap(text, self.width - 2):
+    def comment(self, text, has_path=False):
+        args = {}
+        if has_path:
+          args['break_long_words'] = args['break_on_hyphens'] = False
```
martine: Four space tabs, please!

martine:
This looks good if you could just fix the tabbing!

misc/ninja_syntax.py
avallee: Fixed.

misc/ninja_syntax.py
```diff
@@ -21,8 +21,11 @@ def __init__(self, output, width=78):
     def newline(self):
         self.output.write('\n')
 
-    def comment(self, text):
-        for line in textwrap.wrap(text, self.width - 2):
+    def comment(self, text, has_path=False):
+        args = {}
+        if has_path:
```
nico: Should this just be done unconditionally? i.e. always pass break_long_words=False, break_on_hyphens=False ? I'm not sure, but I can't think of any case where you wouldn't want this, and is most callers will end up passing True for has_path over time, and the ones that don't will likely be incorrect.

martine: I think I agree with Nico -- it's fine if a comment with a very long word goes over the width, because the only reason it would is if it has a long word which is unlikely to be splittable.

avallee: Agreed. Made it just unconditional.

avallee:
PTAL

PR #1045, opened 2015-10-28
Extend format status output - Allow to change also order of description in NINJA_STATUS

pali:
**Align all edge numbers to same length when formatting output**

So when updating status output on terminal, make sure that all numbers will
be always aligned at same length. Even if progress variables change number
of characters.

___

**Allow to change also order of description in NINJA_STATUS**

This commit adds new format string %d for printing description. When %d is
not specified in NINJA_STATUS, then description is automatically printed at
the end of line.

This allows to add additional strings after description, for example
changing colors on ANSI terminals via NINJA_STATUS variable.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -188,6 +188,9 @@ string BuildStatus::FormatProgressStatus(
   string out;
   char buf[32];
   int percent;
+  char total_edges_str[32];
+  snprintf(total_edges_str, sizeof(total_edges_str), "%d", total_edges_);
+  int total_edges_len = strlen(total_edges_str);
   for (const char* s = progress_status_format; *s != '\0'; ++s) {
     if (*s == '%') {
       ++s;
@@ -198,14 +201,13 @@ string BuildStatus::FormatProgressStatus(
 
         // Started edges.
       case 's':
-        snprintf(buf, sizeof(buf), "%d", started_edges_);
+        snprintf(buf, sizeof(buf), "%*d", total_edges_len, started_edges_);
         out += buf;
         break;
 
         // Total edges.
       case 't':
-        snprintf(buf, sizeof(buf), "%d", total_edges_);
-        out += buf;
+        out += total_edges_str;
         break;
 
         // Running edges.
@@ -214,20 +216,20 @@ string BuildStatus::FormatProgressStatus(
         // count the edge that just finished as a running edge
         if (status == kEdgeFinished)
           running_edges++;
-        snprintf(buf, sizeof(buf), "%d", running_edges);
+        snprintf(buf, sizeof(buf), "%*d", total_edges_len, running_edges);
         out += buf;
         break;
       }
 
         // Unstarted edges.
       case 'u':
-        snprintf(buf, sizeof(buf), "%d", total_edges_ - started_edges_);
+        snprintf(buf, sizeof(buf), "%*d", total_edges_len, total_edges_ - started_edges_);
         out += buf;
         break;
 
         // Finished edges.
       case 'f':
-        snprintf(buf, sizeof(buf), "%d", finished_edges_);
+        snprintf(buf, sizeof(buf), "%*d", total_edges_len, finished_edges_);
         out += buf;
         break;
 
--- a/src/build.cc
+++ b/src/build.cc
@@ -184,10 +184,11 @@ void BuildStatus::BuildFinished() {
 }
 
 string BuildStatus::FormatProgressStatus(
-    const char* progress_status_format, EdgeStatus status) const {
+    const char* progress_status_format, EdgeStatus status, const string to_print) const {
   string out;
   char buf[32];
   int percent;
+  bool has_desc = false;
   char total_edges_str[32];
   snprintf(total_edges_str, sizeof(total_edges_str), "%d", total_edges_);
   int total_edges_len = strlen(total_edges_str);
@@ -261,6 +262,12 @@ string BuildStatus::FormatProgressStatus(
         break;
       }
 
+        // Description (to_print)
+      case 'd':
+        out += to_print;
+        has_desc = true;
+        break;
+
       default:
         Fatal("unknown placeholder '%%%c' in $NINJA_STATUS", *s);
         return "";
@@ -269,6 +276,9 @@ string BuildStatus::FormatProgressStatus(
       out.push_back(*s);
     }
   }
+
+  if (!has_desc)
+    out += to_print;
 
   return out;
 }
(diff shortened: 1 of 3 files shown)


pali:
Hi! Can you review my patches?

tfarina:
@pali can you collapse all these patches into one? I believe that will make it easier to be reviewed.

pali:
Hm.. there are more patches as before... and some not mine. Have you did some git push --force? Looks like there are more changes since I created this pull request in Oct 2015.

pali:
I cherry-picked those two my commits on top of current master.

pali:
Is there anything wrong with my patch?

pali:
@tfarina so it is OK now? Ready for merge?

tfarina:
@pali, sorry, it is not my call. The decision makers are @nico and/or @evmar.

pali:
@nico, @evmar. any objects?

src/build.cc
```diff
@@ -168,6 +168,9 @@ string BuildStatus::FormatProgressStatus(
   string out;
   char buf[32];
   int percent;
+  char total_edges_str[32];
+  snprintf(total_edges_str, sizeof(total_edges_str), "%d", total_edges_);
```
evmar: Note: total_edges_ can change over the course of a build, so this isn't guaranteed.  The most common thing in fact is for the total edge count to be overestimated and then reduced in the case where a command doesn't modify its outputs.

src/build.cc
```diff
@@ -164,10 +164,12 @@ void BuildStatus::BuildFinished() {
 }
 
 string BuildStatus::FormatProgressStatus(
-    const char* progress_status_format) const {
+    const char* progress_status_format,
+    const string to_print) const {
```
evmar: Passing to_print by value here makes a copy.  It's surely harmless but from a style perspective you should use `const string&` like elsewhere in Ninja.

evmar:
Sorry for not looking at this earlier.  The single pull request with a vague description makes it easy to overlook!  I think the %d change is pretty reasonable but I think the alignment one may not work as you expect.

src/build.cc
pali: I think that this is not a problem. It is just used as an estimate for length alignment. It just provide better output for next print...

src/build.cc
pali: I can change it to const string&. I do not think any differences as string passed to to_string is not too long as it could cause some performance changes...

pali:
Alignment in all snprintf calls make sure that all "%*d" formats will be aligned to same length. As during one FormatProgressStatus call is total_edges_ the highest number which can "%*d" format.

Also that number total_edges_ is not directly used. But its logarithm, so if number of edges do not start exponentially increasing or decreasing then output is pretty constant and good aligned also when doing progress.

pali:
@evmar Let me know if you are going accept this patch in some way and if something is needed to rework...

pali:
@evmar I updated github pull request description from commit messages (sorry for vague description on github, moreover it is github fault, that it hides git commits itself!)

@evmar can you look at above comments ↑↑↑?

pali:
@evmar: PING

pali:
@evmar gentle PING

jhasse:
This looks quite useful, especially if we add support for colors in NINJA_STATUS. Would you be willing to rebase this on current master?

pali:
I already did it more times. But nobody replied.

pali:
So is still somebody interesting in this patch after rebasing? And can review it?

jhasse:
Yes, I would do it.

pali:
@jhasse Done!

jhasse:
Thanks :) Please have a look at the CI failures.

pali:
Done!

jhasse:
Thanks. This will come in handy when #1494 is merged.

Code looks good, though I don't like the need for `%*d`. Maybe use `%D` for the description instead?

pali:
> I don't like the need for `%*d`

printf's `%*d` is one of the main points of this pull request. It aligns all numbers to the same length when formatting.

> Maybe use %D for the description instead?

I used lowercase `d` for `NINJA_STATUS` format as all other `NINJA_STATUS` format characters are lowercase.

jhasse:
Ah, I thought that was because it would clash with %d from the description otherwise. Sorry, I didn't read it more clearly.

Can you split the PR in "Align all edge numbers to same length when formatting output" (not sure if we want this by default, needs discussion) and "Allow to change also order of description in NINJA_STATUS"?

pali:
Reason for aligning is that output format does not change so drastically when printing it in-place on terminal. Before this change, when updating `1/100` to `10/100` and then to `100/100` it changes 3 times width of text. So if you have custom NINJA_STATUS, text in that status move on that line... After this you would see `  1/100`, ` 10/100` and `100/100`, always this text consist of 7 characters and with custom NINJA_STATUS text, it would not move and stay on same position.

Second commit allows to put description at any position and to have it on stable position, that alignment from first commit is really needed.

jhasse:
I understand the reasoning. Users may rely on the old behavior though and it changes Ninja's default output. A way to change the behavior, maybe `%S` instead of `%s`, might still be nice to have.

> I used lowercase d for NINJA_STATUS format as all other NINJA_STATUS format characters are lowercase.

Yeah, keep that. I was completely wrong there, nvm ;)

jhasse:
No, this is independent of #1494.

I'm still not sure about the "Align all edge numbers to same length when formatting output" part. What do you think about

> A way to change the behavior, maybe %S instead of %s, might still be nice to have.

?

(btw: As for all frontend stuff, I'm a fan of #1210 instead of extending Ninja itself)

pali:
And what with %u, %r and %f then?

evmar:
BTW, one of my design regrets in Ninja is that we have this mechanism for generating command line strings internally (via "$foo ${other}") and then we didn't reuse it for status.  Not sure if it's worth thinking about here though.

jhasse:
> And what with %u, %r and %f then?

Good point. Hm ... maybe it would indeed be better to go with #1210 / #1600 for such changes.

Challanger524:
Is this relatable?: [idea] Align build output for `build step` and `object name`: https://github.com/ninja-build/ninja/issues/2564

jhasse:
> BTW, one of my design regrets in Ninja is that we have this mechanism for generating command line strings internally (via "$foo ${other}") and then we didn't reuse it for status. Not sure if it's worth thinking about here though.

Fixed by #2772 :)

I've introduced a new --status flag in that PR and created #2774 to allow changing where the description appears. This way we stay backwards compatible as we don't change anything for NINJA_STATUS.

PR #1046, opened 2015-10-30
Speed up zsh target completion

zed0:
Using sed instead of a read/echo/cut loop makes it significantly faster to tab complete targets.


martine:
This change also came by, it's funny how similar they are
https://github.com/martine/ninja/pull/1046

nico:
Thanks!

PR #1047, opened 2015-10-31
Speed-up bash target auto-complete.

pykello:
Bash auto-complete was too slow when I used it for Chromium, which had 40k targets. The following speed-ups reduce the time to auto-complete from 2038ms to 570ms.

* Let compgen do the command substitution. Similar to
  https://lists.gnu.org/archive/html/bug-bash/2012-03/msg00115.html
* Use "cut" instead of "awk" for separating fields.

### DIFF
--- a/misc/bash-completion
+++ b/misc/bash-completion
@@ -49,9 +49,8 @@ _ninja_target() {
 		C) eval dir="$OPTARG" ;;
 	    esac
 	done;
-	targets_command="eval ninja -C \"${dir}\" -t targets all"
-	targets=$(${targets_command} 2>/dev/null | awk -F: '{print $1}')
-	COMPREPLY=($(compgen -W "$targets" -- "$cur"))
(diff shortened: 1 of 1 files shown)


martine:
This change also came by, it's funny how similar they are
https://github.com/martine/ninja/pull/1046

pykello:
Yeah, I was working on it yesterday and when wanted to submit it saw the other change and was surprised someone else having a similar problem in the same day.

BTW, when I use "grep" instead of "compgen" I get some more speed-up:

```
COMPREPLY=($(${targets_command} | grep "^$cur"))
```

For example, ```ninja -C out/Debug ex[TAB]```, with compgen the script takes 420ms, with grep it takes 280ms.

I'm not sure if something can go wrong if I don't use compgen. I need to double-check this one.

nico:
Thanks! Sounds like this is a big win already, if grep is even better then that can be done in a follow-up.

PR #1051, opened 2015-11-14
Change struct to namespace

compnerd:
The struct contained only static members.  Convert the functions to freestanding
functions in a namespace.  This provides identical functionality in a more
idiomatic manner.


nico:
Seems like change just for the sake of it, adds two lines of code, introduces a line > 80 columns, and has a typo in the commit message…

compnerd:
Agreed on the first and third point.  I don't see any line that is >80 col that has been introduced in this patch though.

nico:
Ok, in that case let's not merge this :-)

(Sorry about the 80 columns; I was confused.)

PR #1054, opened 2015-11-18
Allow overriding POSIX subprocess shell at configure time

mikesep:
This is useful on systems where `/bin/sh` is not well-behaved.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -209,6 +209,9 @@ parser.add_option('--with-gtest', metavar='PATH', help='ignored')
 parser.add_option('--with-python', metavar='EXE',
                   help='use EXE as the Python interpreter',
                   default=os.path.basename(sys.executable))
+parser.add_option('--with-subprocess-shell', metavar='EXE',
+                  help='use EXE as the shell for POSIX subprocesses',
+                  default='/bin/sh')
 parser.add_option('--force-pselect', action='store_true',
                   help='ppoll() is used by default where available, '
                        'but some platforms may need to use pselect instead',)
@@ -331,7 +334,8 @@ else:
               '-fno-rtti',
               '-fno-exceptions',
               '-fvisibility=hidden', '-pipe',
-              '-DNINJA_PYTHON="%s"' % options.with_python]
+              '-DNINJA_PYTHON="%s"' % options.with_python,
+              '-DNINJA_SUBPROCESS_SHELL="%s"' % options.with_subprocess_shell]
     if options.debug:
         cflags += ['-D_GLIBCXX_DEBUG', '-D_GLIBCXX_DEBUG_PEDANTIC']
         cflags.remove('-fno-rtti')  # Needed for above pedanticness.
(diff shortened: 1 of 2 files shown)


itensionanders:
Interesting. Guess one shouldn't complain too much about the bashism in `/bin/sh` on some Linux distributions then... Suppose a configure option is slightly better for you than just having to patch it.

nico:
Ninja's tests aren't supposed to be bash-specific. If they fail in a recent posix-compatible sh, that's a bug in the test.

I don't know if we should workaround in ninja for shells that aren't posix-compatible. See #827 for the reasoning.

mikesep:
Hi @nico, I don't think it's the case that Ninja's tests are not posix-compatible &ndash; I entirely expect the old, busted `/bin/sh` on my systems to be the issue. I just want to be able to point Ninja to use a non-busted shell. You had provided feedback against using `$SHELL` in [this Groups thread][1], so I went for the configure-time approach for the least amount of surprise for the user at run-time.

(Now that I think about it, it would be good to have a way to discover which shell path was embedded into Ninja. Are there any precedents for showing configure-time settings, such as the value of `NINJA_PYTHON`? I guess I could add them to the `--help` output.)

I'm open to alternatives. It's not particularly hard for me for to apply a custom patch to change the shell path in our local build of Ninja, but I figured I'd put this out there in case others needed it too.

[1]: https://groups.google.com/d/msg/ninja-build/eByS91Slbwc/xzjkugrcl48J

gsauthof:
Another example: Solaris 10 is a POSIX conforming system where `/bin/sh` isn't POSIX.

The thing is that POSIX doesn't specify that `/bin/sh` has to be conforming - it is just specified that there has to be a conforming `sh` available 'somewhere' on the system.

With Solaris 11, Sun/Oracle changed `/bin/sh` to a POSIX conforming one, surprisingly.

The POSIX way to detect a POSIX conforming `sh` is to do something like this:

    PATH=$(getconf PATH) command -v sh

This prints `/usr/xpg4/bin/sh` on Solaris 10 and `/usr/bin/sh` on Fedora 23.

Perhaps this is something the ninja configure script could do.

nico:
Hm, since /bin/sh is POSIX almost everywhere (from this bug, it sounds like it isn't in Solaris before 11 – Solaris 11 was released 4.5 years ago; and it isn't in AIX), I think it's ok to say "if your system's /bin/sh isn't POSIX, you need to patch up ninja locally".

(#827 is somewhat related to this issue here; it's about allowing different shells in general)

dje4g:
As a thought, suppose one wants to provide the option to run the command with some kind of wrapper.
E.g., run the command in a docker container. [this is just one example, I think this could be generally useful]
One *could* avoid enhancing ninja and encode this in the command itself, but that can lead to complexities like having to have two copies of the command, one for the normal method of /bin/sh -c and another for running via the wrapper. E.g., the command is just a string, and if it is being passed to another command, say my-wrapper, which will in turn be passed as a string to /bin/sh -c, then one has to deal with quoting quotes in the original command.
OTOH, if one could just tell ninja "For this invocation please use my-wrapper instead of /bin/sh -c." then that greatly simplifies the design and implementation.

mikesep:
I'm going to close this out based on @nico's [last comment](https://github.com/ninja-build/ninja/pull/1054#issuecomment-179483358).

Issue #1056, opened 2015-11-22
-k 0 is not documented


orlp:
Current documentation is the following:

    -k N     keep going until N jobs fail [default=1]

I suggest this instead:

    -k N     keep going until N jobs fail (0 is unbounded) [default=1]

Although the exact wording doesn't matter, a small documentation of this feature would be nice.

tfarina:
To me this does not seem to improve much. What unbounded means? Or, passing zero will result in which behavior?

orlp:
I'm not a native speaker, so I'm not sure whether I used the right word.

Unbounded in this context means that `ninja` will not stop after any finite number of jobs fail - only when it can no longer make any progress (all leaves in the build dependency tree have failed or completed).

nicolasdespres:
What about "unlimited"?

orlp:
@nicolasdespres I fail to see how that is an improvement over unbounded. I feel that if one is going to be unclear then so is the other.

jimporter:
As a native English speaker, "unlimited" is a bit clearer (and certainly more natural sounding).

nico:
dupe of #1006

PR #1057, opened 2015-11-22
Documented -k 0.

orlp:
Solves https://github.com/ninja-build/ninja/issues/1056.

### DIFF
--- a/misc/zsh-completion
+++ b/misc/zsh-completion
@@ -59,7 +59,7 @@ _arguments \
   '-f+[Specify input build file (default=build.ninja)]:files:_files' \
   '-j+[Run N jobs in parallel (default=number of CPUs available)]:number of jobs' \
   '-l+[Do not start new jobs if the load average is greater than N]:number of jobs' \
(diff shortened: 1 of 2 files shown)


nico:
Thanks for the patch. I haven't merged it yet because "(0 is unbounded)" sounds a bit awkward to me. I can't think of anything better either though. Maybe someone else out there is good with words and can suggest a wording?

nicolasdespres:
Some of us did suggest alternatives in the comments about #1056.

nico:
That was just the same message instead of "unlimited" instead of "unbounded". That sounds a bit better, but still strange to me. "0 is unlimited" sounds like a statement about the integer 0 to me. "0 means unlimited" works, but isn't really true as the build stops when it can't proceed any more. "keep going; if N > 0 stop after N jobs failed" is maybe ok?

nico:
(has anyone ever used a k parameter that isn't 0? i actually don't know why this gets a parameter. not that i want to change it at this point.)

jhasse:
Superseded by #1373.

PR #1058, opened 2015-11-30
Add keys to choose shell and shellflags

itensionanders:
This can be used for instance to choose /bin/dash, or to run the scripts with -ec.
There's a lot of other uses, as can be seen by make's $(SHELL) and .SHELLFLAGS.


nico:
Let's close this as a dupe of #827.

PR #1059, opened 2015-12-01
Avoid double-scheduling build edges in another case

bradking:
The change in commit v1.2.0~3^2~3^2~3 (Fix duplicate edge Pool crash in
the minimally invasive way, 2013-03-18) avoids double-scheduling in a
case involving duplicate out edges.  However, double-scheduling may also
occur on a consistent graph when an edge and one of its dependencies
share an order-only input:

    $ cat build.ninja
    ...
    build c: touch
    build b: touch || c
    build a: touch | b || c
    $ ninja a
    $ rm a c
    $ ninja a

In this case 'c' will build first.  When NodeFinished('c') loops over
the out edges it will find AllInputsReady is true for 'b' and call
EdgeFinished('b') since it is not wanted (up to date).  This will
call NodeFinished('b') which will loop over its out edges, find
AllInputsReady is true for 'a', and call ScheduleEdge('a').  When
we eventually return to the loop in NodeFinished('c') it will move
on to its second output and find that AllInputsReady is true for
'a' and call ScheduleEdge('a') again.

Teach ScheduleEdge to tolerate duplicate calls for an edge that has
already been scheduled.  Avoid calling EdgeScheduled more than once
for the same edge.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -366,19 +366,22 @@ Edge* Plan::FindWork() {
 }
 
 void Plan::ScheduleWork(Edge* edge) {
+  set<Edge*>::iterator e = ready_.lower_bound(edge);
+  if (e != ready_.end() && !ready_.key_comp()(edge, *e)) {
+    // This edge has already been scheduled.  We can get here again if an edge
+    // and one of its dependencies share an order-only input, or if a node
+    // duplicates an out edge (see https://github.com/ninja-build/ninja/pull/519).
+    // Avoid scheduling the work again.
+    return;
+  }
+
   Pool* pool = edge->pool();
   if (pool->ShouldDelayEdge()) {
-    // The graph is not completely clean. Some Nodes have duplicate Out edges.
-    // We need to explicitly ignore these here, otherwise their work will get
-    // scheduled twice (see https://github.com/ninja-build/ninja/pull/519)
-    if (ready_.count(edge)) {
-      return;
-    }
     pool->DelayEdge(edge);
     pool->RetrieveReadyEdges(&ready_);
   } else {
     pool->EdgeScheduled(*edge);
-    ready_.insert(edge);
+    ready_.insert(e, edge);
   }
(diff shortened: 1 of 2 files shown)


bradking:
Rebased on `master`.

tfarina:
@bradking does this improve performance? Have you measured it? It would be interesting if you could share some numbers (perhaps from ninja -C out/Release chrome). WDYT?

bradking:
@tfarina I just ran timings on a project that takes 7m30s to build 7500 sources locally.  There is no measurable difference in performance.  This is not surprising because the change does not affect the number of build statements that are actually executed, and the build time is dominated by the tools and not Ninja internals.

This change is not about performance, but internal correctness.  ScheduleEdge was originally written assuming that it would be called at most once per edge.  Then pool support was added and depends on this assumption.  Later a problem was discovered where ScheduleEdge could be called more than once on an inconsistent graph.  The check for repeat scheduling was added to the pool code path because it could not tolerate repeat calls.  I do not know why it wasn't added for both code paths at the time because it is only by chance that the other code path is not affected (but it could be with future changes to e.g. EdgeScheduled).

Now I've discovered that even on correct graphs it is possible to get a repeat call to ScheduleEdge for one edge.  It caused a lot of confusion while debugging other topics I've been developing.  Therefore I've moved the check to the beginning of the function where it belongs.  The use of `lower_bound` and an `insert` hint should produce equivalent performance.

nico:
Thanks for the good explanation of the patch. I noticed that the new test passes without the code change -- is the test change complete?

bradking:
> the new test passes without the code change -- is the test change complete?

Yes.  No behavior difference can currently be observed.  The goal is to avoid calling EdgeScheduled more than once for the same edge to protect future changes to it.  The fix in commit v1.2.0~3^2~3^2~3 was only for the case that a repeat call would break.  This change generalizes it.

PR #1064, opened 2015-12-06
add "msvc_deps_prefix" to Rule::IsReservedBinding()

### LINKED ISSUE #1043: Can msvc_deps_prefix be set on a per-rule basis?
I'm working on a meta build system. Although very unlikely, it might be possible that a build script makes use of two different versions of the MSVC compiler which could theoretically be in different languages. Is it possible to set the `msvc_deps_prefix` on a per-rule basis?

NiklasRosenstein:
fixes the error about an unexpected variable for a rule that
declares the msvc_deps_prefix. The manual suggests that this
should work since Ninja 1.5
(https://ninja-build.org/manual.html#ref_rule).

Closes #1043

### DIFF
--- a/src/eval_env.cc
+++ b/src/eval_env.cc
@@ -71,7 +71,8 @@ bool Rule::IsReservedBinding(const string& var) {
       var == "pool" ||
       var == "restat" ||
       var == "rspfile" ||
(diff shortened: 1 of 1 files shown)


martine:
LGTM

PR #1067, opened 2015-12-08
use escape_path() in Writer.default(), close #990

### LINKED ISSUE #990: ninja_syntax.py: Writer.default() doesn't escape spaces
The `Writer.default()` function doesn't escape spaces in path names and threats a single target with spaces as multiple targets. I suggest changing

```python
    def default(self, paths):
        self._line('default %s' % ' '.join(as_list(paths)))
```

to 

```python
    def default(self, paths):
        transformed = []
        for path in paths:
            transformed.append(re.sub('(\s)', '$\\1', path))
        self._line('default %s' % ' '.join(transformed))
```

NiklasRosenstein:
see #990


jhasse:
Can you rebase this on master and reopen the PR? Thanks!

Issue #1070, opened 2015-12-10
add cwd variable to specify working directory of rule/build


NiklasRosenstein:
I want to suggest supporting a `cwd` variable to specify the working directory for a rule or a single build. `build` will have to translate relative `$in` and `$out` filenames respectively. Meta build systems will know the behaviour of specifying `cwd` and can make sure to transform relative paths used in `command` respectively.

__Example__

```
rule echo:
  command = echo $in --> $out

build resources/some_res.txt: echo obj/some_res.o
  cwd = resources
```

I would expect the following output

    some_res.txt --> ../obj/some_res.o

__Possible Application__

Using `objcopy` to create an object file which embeds resources and binary data will create a symbol based on the input filename, including the directory! Eg. the file `resources/some_res.txt` would get the symbol `_resources_some_res_txt_start`. Switching to the `resources` directory would allow us to pass only the filename and get `_some_res_txt_start` as the symbol.

```
rule embed:
  command = objcopy -I binary -O elf32-i386 -B i386 $in $out'

build resources/some_res.txt: embed obj/some_res.o
  cwd = resources
```

nicolasdespres:
Hi. You could generate something like that:
```
rule embed:
  command = cd ${cwd} && objcopy -I binary -O elf32-i386 -B i386 $input $output

build resources/some_res.txt: embed obj/some_res.o
  input = some_res.txt
  output = obj/some_res.o
  cwd = resources
```

NiklasRosenstein:
```
rule dtschump.CImg.run_edge_explorer2d
  command = cd c:\users\niklas\desktop\cimg\examples && c:\users\niklas\desktop\cimg\build\dtschump.cimg\edge_explorer2d.exe
  pool = console
build dtschump.CImg.run_edge_explorer2d: dtschump.CImg.run_edge_explorer2d c$:\users\niklas\desktop\cimg\build\dtschump.cimg\edge_explorer2d.exe
```

That doesn't seem to work, at least on Windows. :(

martine:
In what way does it fail?

NiklasRosenstein:
```
[1/1] cd C:\Users\niklas\Desktop\CImg\examples && c:\u...dtschump.cimg\build\dtschump.cimg.examples\captcha.exeFAILED: cd C:\Users\niklas\Desktop\CImg\examples && c:\users\niklas\desktop\dtschump.cimg\build\dtschump.cimg.examples\captcha.exe
[...]

PR #1074, opened 2015-12-16
add a section to the manual discussion the command= variable

martine:
This includes a mention of using cmd /c on Windows.
This would have helped on issue #1070 for example.


nico:
Thanks!

Issue #1075, opened 2015-12-16
Can't depend on "."


colincross:
When handling globs, Blueprint emits dependencies to regenerate build.ninja if a directory is modified, in case the modification was to add a missing path.  For example, when globbing for dir/*, it will add a dependency on "dir" if it exists (to handle the case where a file is added or delete from dir), or "." if a does not exist (to handle the case where dir is added).  The first case works fine, but the second case fails with:
`ninja: error: rebuilding 'build.ninja': path canonicalizes to the empty path`

"." seems like a perfectly reasonable path, and should canonicalize to ".".

martine:
I can't quite remember why we have this check, but I think it's to prevent accidentally incorrect paths, such as when generators accidentally produce the empty string.  E.g.

```
outfile = foobar.so
build $outfilex: ...
# oops, typoed the variable name
```

The other accident is paths like `foo/..` which I don't think people intend to provide.

This is all a long way of asking: would it still work for you if we explicitly allowed the string "." as a path but disallowed the other cases?

colincross:
Allowing "." would be fine for my issue, although I think disallowing `foo/..` is unnecessary - depending on `foo/..` is clearly depending on a directory, not a file, and is not much different than `foo/bar/..`.

The empty string case is already caught by the first check check for *len == 0 in CanonicalizePath.

colincross:
The "canonicalizes to the empty path" error comes from e9087f5cdf2c8485b5b0a6858782bb02dbffbcdd which was preventing *len being set to -1, and not actually trying to prevent ".".

nico:
fixed by #1076

PR #1076, opened 2015-12-16
Canonicalize "." to "."

colincross:
"." is a legal path, if the string is empty after canonicalization
return ".".

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -226,8 +226,8 @@ bool CanonicalizePath(char* path, size_t* len, unsigned int* slash_bits,
   }
 
   if (dst == start) {
-    *err = "path canonicalizes to the empty path";
-    return false;
+    *dst++ = '.';
+    *dst++ = '\0';
   }
 
   *len = dst - start - 1;
--- a/src/util_test.cc
+++ b/src/util_test.cc
@@ -91,6 +91,22 @@ TEST(CanonicalizePath, PathSamples) {
   path = "/";
   EXPECT_TRUE(CanonicalizePath(&path, &err));
   EXPECT_EQ("", path);
+
+  path = "/foo/..";
+  EXPECT_TRUE(CanonicalizePath(&path, &err));
+  EXPECT_EQ("", path);
+
+  path = ".";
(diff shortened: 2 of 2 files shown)


colincross:
Fixes #1075

nico:
Is it good enough for your purposes to detect "." early on and return immediately in that case?

Do you have your build files in your checkout? Or do you need to glob in your build outputs directory? Why do you have a dependency on "."?

colincross:
Blueprint should canonicalize paths, but I could imagine a situation where a ninja variable contains `..`, so `a/$var` would expand to `a/..` and canonicalize to `.`, and that canonicalization can't happen in the generator.

The dependency on `.` is caused by support for optional directories.  Ninja is run from the top-level of the source tree (for legacy reasons), with the outputs placed in another directory.  In a simplified case, Blueprint will attempt to parse `a/Blueprint`, but if `a/` doesn't exist it adds a dependency on `.` to the rule that regenerates `build.ninja`.  When a new file is created in `.` the generator re-runs to parse the newly-created file.  In practice we add an extra restat rule between the `.` dependency and the generator that reruns the glob that the generator cares about, if the resulting file list doesn't change the generator is not rerun.

sgraham:
Depending on a directory doesn't mean anything useful on Windows fwiw, but maybe Blueprint doesn't care.

colincross:
Blueprint doesn't run on windows, but why wouldn't dependencies on a directory work in windows?

sgraham:
Maybe I misinterpreted what the dependency is for, but if a file changes in
a directory on Windows the mtime of the directory isn't necessarily
modified. So it doesn't tend to be overly useful.

sgraham:
Actually, I can't find a reference for that now. It may have only applied
to FAT. Sorry for the non-sequitur noise.

colincross:
The behavior on linux (and hopefully the same on windows) is for the directory mtime to change if an entry is added to or removed from the directory, but not if a file in the directory is changed.

nico:
Sounds like nobody remembers why this is not allowed, and it fixes a valid use-case. Let's merge this.

nico:
Oh, and thanks for the patch!

PR #1078, opened 2015-12-20
Print 'Leaving directory'.

nicolasdespres:
Ninja already prints 'Entering directory' so Emacs can properly
identify that the cwd has changed for subsequent commands. When ninja
is called from another ninja process (for instance when building a
sub-project) it must notify that it leaves the directory.


nico:
Since this is too is for recursive ninjas, I'm closing this one as well.

Issue #922, opened 2015-03-02
Environment variable controlling max number of jobs


scheib:
I'd like to propose an environment variable NINJA_JOBS as an alternative to the command line argument -j.

This would allow systems where the default configuration isn't appropriate to have a single setup step to correct for it but simplify all future command line executions. E.g. environments where distributed build tools such as distcc or goma are used.

Happy to code patch if there's agreement.

nico:
dupe of #816, related to #797

(if you want this, can't you just `alias pninja=ninja -j $FOO`?)

scheib:
A different name for ninja requires all other aliases, cut-n-paste command lines, and manual entry to conform. I'd rather set an environment variable once and not think about it again.

A same name alias mostly works, ```alias ninja=`which ninja` -j 123```, but trips up if a -j is ever entered. This is (almost) my current work around, but the environment variable seems cleaner.

(Ok, my actual work around has some platform detection
```
alias ninja='`ninja-detect-goma`'
alias ninja-detect-goma='if [ -e ~/goma ]; then if [ `uname` == "Darwin" ]; then echo ninja -j 100; else echo ninja -j 1024; fi; else echo ninja; fi'
```
)

nico:
Let's close this as dupe of #816, which we think we don't want to do.

PR #1079, opened 2015-12-20
Forward flags to sub processes.

### LINKED ISSUE #797: request: new environment variable NINJA_OPTIONS (or NINJA_FLAGS)
This would allow global user-override parameters such as specifying -j#  

e.g.

export NINJA_OPTIONS=-j6

would force all builds to be run with -j6 instead of the default. (unless overridden on the commandline of a particular execution of ninja).

nicolasdespres:
When ninja invokes ninja (for instance when building a sub-project) it
is useful to have a way to control the flags used by the sub-ninja
process. If I pass '-d explain' or '-v' when starting
ninja from the command line on the top project, I expect these flags
to be forwarded to the sub-ninja process.

Note that this patch has limited parsing capabilities of the
NINJA_FLAGS environment variables but that should be enough for our
use case.

Fixes #797, #922 and #816.

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -993,7 +993,6 @@ int ExceptionFilter(unsigned int code, struct _EXCEPTION_POINTERS *ep) {
 /// Returns an exit code, or -1 if Ninja should continue.
 int ReadFlags(int* argc, char*** argv,
               Options* options, BuildConfig* config) {
-  config->parallelism = GuessParallelism();
 
   enum { OPT_VERSION = 1 };
   const option kLongOptions[] = {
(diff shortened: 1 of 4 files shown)


nico:
As mentioned on the mailing list, I don't think we should make recursive ninja a thing. I don't think this is a good change.

nocnokneo:
@nico, Could you link to the mailing list discussion you're referring to?

nico:
https://groups.google.com/forum/#!searchin/ninja-build/%22recursive$20ninja%22 finds a few threads, and I think there are several threads on a jobserver where it's also been discussed.

nocnokneo:
Okay, I'd found some of those. In particular these two (for anyone else stumbling across this PR):

https://groups.google.com/forum/#!searchin/ninja-build/"recursive$20ninja"/ninja-build/niYOkWjkPW8/LSgArOH_ZZoJ
https://groups.google.com/forum/#!searchin/ninja-build/jobserver/ninja-build/PUlsr7-jpI0/Ga19TOg1c14J

But I guess I was expecting to find a thread that ended in some kind of definitive solution or consensus.

For the parallel jobs problem it seems like the current stance from the Ninja team can be summed up as: ALL the world should migrate to Ninja together and start using subninja. Changes like making Ninja jobserver-aware to ease the transition from a Makefile-dominated world are unacceptable because it wouldn't showcase all of Ninjas awesome performance optimizations.

In the case where both parent and child project are CMake-based the proposed solution to teach CMake to use the subninja command definitely makes sense and that work has begun here:

http://thread.gmane.org/gmane.comp.programming.tools.cmake.devel/15364

PR #1082, opened 2015-12-24
Teach zsh to complete intermediary targets.

nicolasdespres:
Bash completion script uses "-t targets all" to list the target which
is faster than "-t targets" and reports intermediary targets
(see the manual entry for the 'targets' tool).

See commit fc135c45.

### DIFF
--- a/misc/zsh-completion
+++ b/misc/zsh-completion
@@ -22,8 +22,8 @@ __get_targets() {
   then
     eval dir="${opt_args[-C]}"
   fi
-  targets_command="ninja -C \"${dir}\" -t targets"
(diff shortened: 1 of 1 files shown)


nico:
Thanks! The commit description seems off, this just makes things faster and doesn't teach zsh completion new tricks, right? Also, the `cut` change isn't mentioned. But merged anyways :-)

nicolasdespres:
Yes it only makes things faster and include intermediary targets in the completion list. The `cut` change is a side effect since the output of `ninja -t targets all` is different than `ninja -t targets`.

It don't understand why github cut the commit description.

Thanks for merging it anyway :)

PR #1003, opened 2015-08-10
Fix compile on VS2015

sgraham:
This fixes issue #995. Defining _HAS_EXCEPTIONS=0 is the important part.

Unfortunately, it seems set_terminate and terminate_handler have declarations both in and out of `std::` now (maybe only when exceptions are disabled? I didn't stare too hard at the headers), so `using namespace std;` causes some problems (below). I hacked around these by removing/adding a couple `std::` but maybe it's time to get rid of `using std`. :(

graphviz.cc
C:\Program Files (x86)\Windows Kits\10\include\10.0.10150.0\ucrt\corecrt_terminate.h(33): error C2872: 'terminate_handler': ambiguous symbol
C:\Program Files (x86)\Windows Kits\10\include\10.0.10150.0\ucrt\corecrt_terminate.h(18): note: could be 'void (__cdecl *__cdecl terminate_handler)(void)'
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE\exception(191): note: or       'std::terminate_handler'
C:\Program Files (x86)\Windows Kits\10\include\10.0.10150.0\ucrt\corecrt_terminate.h(34): error C2872: 'terminate_handler': ambiguous symbol
C:\Program Files (x86)\Windows Kits\10\include\10.0.10150.0\ucrt\corecrt_terminate.h(18): note: could be 'void (__cdecl *__cdecl terminate_handler)(void)'
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE\exception(191): note: or       'std::terminate_handler'
C:\Program Files (x86)\Windows Kits\10\include\10.0.10150.0\ucrt\corecrt_terminate.h(37): error C2872: 'terminate_handler': ambiguous symbol
C:\Program Files (x86)\Windows Kits\10\include\10.0.10150.0\ucrt\corecrt_terminate.h(18): note: could be 'void (__cdecl *__cdecl terminate_handler)(void)'
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE\exception(191): note: or       'std::terminate_handler'
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE\eh.h(35): error C2872: 'unexpected_handler': ambiguous symbol
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE\eh.h(19): note: could be 'void (__cdecl *__cdecl unexpected_handler)(void)'
[...]

PR #1085, opened 2016-01-10
Remove unnecessary `std::`

tetsuok:


src/ninja.cc
```diff
@@ -1163,7 +1163,7 @@ int main(int argc, char** argv) {
 #if defined(_MSC_VER)
   // Set a handler to catch crashes not caught by the __try..__except
   // block (e.g. an exception in a stack-unwind-block).
-  std::set_terminate(TerminateHandler);
+  set_terminate(TerminateHandler);
```
nico: This one was actually necessary. It was added in #1003 to fix MSVS2015, and #1110 now added it back again.

PR #1086, opened 2016-01-10
Use container-based infrastructure for travis-ci.

tetsuok:
To start up faster. I think we want build results as soon as possible.
See https://docs.travis-ci.com/user/workers/container-based-infrastructure/ for the details.


martine:
I just set this up for a different project, thanks!

PR #1087, opened 2016-01-11
Delete a newed status object when it is not needed any more.

tetsuok:
`Builder` allocates a `BuildStatus` object in its constructor by `new`, but the status object is not released by `delete`. I believe it is a good practice to free the allocated object when it is not needed any more.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -552,6 +552,8 @@ Builder::Builder(State* state, const BuildConfig& config,
 
 Builder::~Builder() {
   Cleanup();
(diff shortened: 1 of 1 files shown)


nicolasdespres:
That's a dupe of #1030.

tetsuok:
Oh, thanks @nicolasdespres for letting me know! Sounds like not-deleting the leaking object is for performance reasons if I understand correctly Nico's comment in #1030. Will close this PR and #1088. Thanks again.

PR #1088, opened 2016-01-11
Delete a newed DiskInterface object if it is owned.

tetsuok:
`Cleaner` has two type of constructors: (1) allocates a `DiskInterface` object by `new` (i.e., owns the object); (2) takes a pointer to a `DiskInterface` object (AFAIK, the class doesn't own the object).

If the first type of the constructor is called, it is the Cleaner's responsibility to free the `DiskInterface` object when the object is not needed any more.


tetsuok:
Closing as I mentioned in #1088

PR #1093, opened 2016-01-19
Introduce binary manifest format

shinh:
You can translate build.ninja to build.ninja.bin by

$ ninja -f build.ninja -tcompile build.ninja.bin

and use it by

$ ninja -b build.ninja.bin

This change makes NULL build more than 3x faster for Android
build. See src/serializer.cc for the detail for the binary
manifest format.


jhasse:
Another solution to improve noop build times would be #1389.

jhasse:
Closing in favor of #1438.

PR #1095, opened 2016-01-27
Make dupbuild=err work in subninja

sgraham:
Was only an error in top level .ninja previously.

### DIFF
--- a/src/manifest_parser_test.cc
+++ b/src/manifest_parser_test.cc
@@ -24,7 +24,7 @@
 struct ParserTest : public testing::Test,
                     public ManifestParser::FileReader {
   void AssertParse(const char* input) {
-    ManifestParser parser(&state, this);
+    ManifestParser parser(&state, this, false);
     string err;
     EXPECT_TRUE(parser.ParseTest(input, &err));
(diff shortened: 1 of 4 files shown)


src/manifest_parser_test.cc
```diff
@@ -24,7 +24,7 @@
 struct ParserTest : public testing::Test,
                     public ManifestParser::FileReader {
   void AssertParse(const char* input) {
-    ManifestParser parser(&state, this);
+    ManifestParser parser(&state, this, false);
```
nico: nit: s:false:/*dupe_edge_shoud_err=*/false

nico:
Thank you very much for the fix!

src/manifest_parser_test.cc
sgraham: Did you want me to send a followup for all the |false|s? It seemed a bit noisy to put it in every test.

nico: I probably should've made it an enum parameter, not a bool parameter. If you want, you can either add all the comments or change it to enum; or you can do nothing and I'll do the enum thing eventually.

PR #1096, opened 2016-01-27
dupe_edge_should_err from bool to enum

sgraham:
Follow up to #1095.

(`enum class` seems to be disabled by build settings.)


nico:
Thanks! Yes, no c++11 for ninja – else it'll require c++11-compatible standard libraries, and I feel it's nice if build tools can run on older systems (unless it gets in the way too much).

bradking:
This broke compilation of some perftest sources.  I've included a fix in #1060.

PR #1097, opened 2016-01-28
Use vfork() instead of fork() to invoke subprocess

tzik:
This CL introduces "execute" as a new helper tool and replaces fork()
with vfork(). Ninja calls exec() to invoke the execute tool soon after
vfork() is called, and the execute tool simply plumps pipes and exec
a give command.

fork() performance depends on the memory usage of the parent process.
If fork() is called on the beginning of the main(), it takes less than
100us on a powerful machine, Z620. On the other hand, after it reads and
parse 33MB build.ninja and 44MB .ninja_deps, the cost raises up to
2400us.

In a warm cached Chrome full build, previously, fork() takes 55s in
total according to the builtin metric, while whole build process takes
75s in total for ~20000 targets. After this CL, the replacement,
vfork() + the execute tool exec(), takes ~600us per task and
the build process takes 53s in total.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -100,6 +100,9 @@ class Platform(object):
     def supports_ppoll(self):
         return self._platform in ('linux', 'openbsd', 'bitrig')
 
+    def supports_vfork(self):
+        return self._platform in ('linux')
+
     def supports_ninja_browse(self):
         return (not self.is_windows()
                 and not self.is_solaris()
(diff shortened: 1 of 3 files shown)


nico:
I wanted to do it like so https://github.com/nico/ninja/commits/posix_spawn which I think is a bit simpler. Getting the vfork on linux requires the ugly POSIX_SPAWN_USEVFORK thing, and since I couldn't really measure a huge speedup (53s vs 55s – meh) I wanted to probably not even use that flag on Linux. But if you think it's important, we could use that flag I suppose. But basing this on posix_spawn seems nicer. The only drawback is that https://github.com/ninja-build/ninja/pull/910 needs to be reverted, but @nicolasdespres thinks that's a good thing anyways :-) Does that sound ok to you?

tzik:
Your posix_spawn version looks simpler and is faster than mine! If setsid() is considered less important recerntly, it's nice to me to have your CL in the mainline.

PR #1098, opened 2016-01-28
Add -d keepdepfile to preserve depfiles

shinh:
This is useful when you are developing a tool which generates
GCC-style depfiles.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -892,9 +892,11 @@ bool Builder::ExtractDeps(CommandRunner::Result* result,
       deps_nodes->push_back(state_->GetNode(*i, slash_bits));
     }
 
-    if (disk_interface_->RemoveFile(depfile) < 0) {
-      *err = string("deleting depfile: ") + strerror(errno) + string("\n");
-      return false;
+    if (!g_keep_depfile) {
+      if (disk_interface_->RemoveFile(depfile) < 0) {
+        *err = string("deleting depfile: ") + strerror(errno) + string("\n");
+        return false;
+      }
     }
   } else {
     Fatal("unknown deps type '%s'", deps_type.c_str());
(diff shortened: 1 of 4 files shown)


lunakid:
Nice to see this feature! But is this GCC-only? Or is this something else than keeping the _textual_ ("GCC-style") dependency files?

MSVC here, and `-d keepdepfiles` seems to make no difference, no matter the dance routine. (The `.ninja_deps` binary is always there, and no extra dependency-related file is ever there, AFAICS.)

I only have `deps = msvc` in my (trivial example) script that's related:
```
rule cl
    deps = msvc
    command = cl -showIncludes -c $in -Fo$out

build test.obj: cl test.c

rule link
    command = link $in -out:test.exe

build test.exe: link test.obj

default test.exe

```

```
Q:\tmp\build\ninja>ninja --version
1.9.0

Q:\tmp\build\ninja>ninja -d list
debugging modes:
  stats        print operation counts/timing info
  explain      explain what caused a command to execute
  keepdepfile  don't delete depfiles after they're read by ninja
  keeprsp      don't delete @response files on success
  nostatcache  don't batch stat() calls per directory and cache them
multiple modes can be enabled via -d FOO -d BAR

Q:\tmp\build\ninja>ninja -t clean
Cleaning... 5 files.

Q:\tmp\build\ninja>ninja -d keepdepfile
[5/5] link -nologo -dll main.obj x1.obj x2.obj xx.obj -out:ninja.dll

```
And the dir contents, whether or not I use the `-d` option:
```
.ninja_deps
.ninja_log
build.ninja
test.exe
test.obj
test.c
```

_(BTW, I wouldn't consider this a "debug" feature. if this is, indeed, about keeping a file with readily make-includable rules, this could make ninja also an instant "mkdepend for CL" tool, which is a huge use-case to conquer. (I just found myself in the need for it, that's how I got here, with high hopes, after days of "NMAKE despair").)_

PR #1099, opened 2016-02-03
perftest: fix compilation after dupe_edge_should_err change

bradking:
Fix some ManifestParser constructor calls missed by commit 56bab441b7
(dupe_edge_should_err from bool to enum, 2016-01-27).


bradking:
@nico I split this out of #1060 because that one may take a few iterations and this fix is important because `master` fails to build with `ninja all` whenever I rebase other topics on it now.

nico:
thanks!

PR #1102, opened 2016-02-04
Remove StatIfNecessary call that is never necessary

bradking:
The call to StatIfNecessary in DependencyScan::RecomputeOutputsDirty was
added by commit 41c6258ec1928cbcfc5642504cb9c2b3659fa897 (Share more code between CleanNode() and
RecomputeDirty(), 2013-09-02) while consolidating code paths.  However,
it was needed only when called from RecomputeDirty because prior to
refactoring the CleanNode code path did not call it.

Later commit 015c818cbe85093316115c5a510274eccdb4180b (Let DependencyScan::RecomputeDirty() work
correclty with cyclic graphs, 2014-12-07) added back to RecomputeDirty a
loop over outputs that calls StatIfNecessary.  Therefore
RecomputeOutputsDirty no longer needs to call StatIfNecessary for either
of its own callers.

### DIFF
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -37,6 +37,7 @@ bool DependencyScan::RecomputeDirty(Edge* edge, string* err) {
   edge->outputs_ready_ = true;
   edge->deps_missing_ = false;
 
+  // Load output mtimes so we can compare them to the most recent input below.
   // RecomputeDirty() recursively walks the graph following the input nodes
   // of |edge| and the in_edges of these nodes.  It uses the stat state of each
   // node to mark nodes as visited and doesn't traverse across nodes that have
@@ -126,8 +127,6 @@ bool DependencyScan::RecomputeOutputsDirty(Edge* edge, Node* most_recent_input,
   string command = edge->EvaluateCommand(/*incl_rsp_file=*/true);
   for (vector<Node*>::iterator o = edge->outputs_.begin();
        o != edge->outputs_.end(); ++o) {
-    if (!(*o)->StatIfNecessary(disk_interface_, err))
-      return false;
     if (RecomputeOutputDirty(edge, most_recent_input, command, *o)) {
       *outputs_dirty = true;
       return true;


nico:
The way I understood 41c6258 was to add that StatIfNecessary() call to the CleanNode() call chain, since having up-to-date mtimes is a precondition for RecomputeOutputDirty() (note singular). So it seems like things are easier to reason about if RecomputeOutputsDirty() calls it: If the call is indeed not needed, then StatIfNecessary() will do nothing and be fast, and if it is needed then not calling this is a bug.

(If it's a precondition, why doesn't RecomputeOutputDirty() call StatIfNecessary() itself? I'm not sure, maybe it should?)

It's possible that in practice this call is never necessary though. What's the motivation for this patch? (I'm not opposed to merging it, I just don't quite understand it yet.)

@maximuska

bradking:
I revised the comment to avoid focusing on RecomputeOutputsDirty.  It is really that RecomputeDirty is responsible for initializing mtimes on nodes it encounters.

CleanNode is walking forward through a portion of the graph that was previously walked by RecomputeDirty during the build planning phase.  StatIfNecessary is an initialization step done when a node is first encountered by RecomputeDirty, and its completion is an invariant established by RecomputeDirty.

The only reason commit 41c6258ec1928cbcfc5642504cb9c2b3659fa897 brought along a call to StatIfNecessary is because it sat in a code path used by RecomputeDirty to call StatIfNecessary at the time.  Since commit 015c818cbe85093316115c5a510274eccdb4180b RecomputeDirty handles it earlier.  Had the two changes been made in the other order we may have ended up at the state proposed in this PR originally.

The motivation for this patch is to clean up some existing logic to make it easier to reason about changes in a future PR that will refactor RecomputeDirty further.

bradking:
Rebased on master.  I also opened #1111 which starts with this change and provides further context for its purpose.

nico:
Thanks for explaining!

Issue #1103, opened 2016-02-05
initial MSVC build via --bootstrap doesn't record dependencies


fbenkstein:
When building initially with MSVC via ``python configure.py --bootstrap`` cl.exe is executed without the /showIncludes option.  Unfortunately the option is also missing from the build.ninja that is generated.  This in turn means that header dependencies are incorrect until configure.py is run again:

```
C:\Users\...\Projects\ninja>python configure.py --bootstrap
bootstrapping ninja...
warning: A compatible version of re2c (>= 0.11.3) was not found; changes to src/*.in.cc will not affect your build.
build.cc
build_log.cc
clean.cc
debug_flags.cc
depfile_parser.cc
deps_log.cc
disk_interface.cc
edit_distance.cc
eval_env.cc
graph.cc
graphviz.cc
lexer.cc
line_printer.cc
manifest_parser.cc
metrics.cc
state.cc
util.cc
version.cc
subprocess-win32.cc
includes_normalize-win32.cc
msvc_helper-win32.cc
msvc_helper_main-win32.cc
minidump-win32.cc
getopt.c
ninja.cc
Generating code
Finished generating code
wrote build.ninja.
bootstrap complete.  rebuilding...
[27/27] LINK ninja.exe
Generating code
Finished generating code

C:\Users\...\Projects\ninja>ninja -t deps
build/getopt.obj: #deps 0, deps mtime 476382870 (VALID)

build/util.obj: #deps 0, deps mtime 476382872 (VALID)

build/minidump-win32.obj: #deps 0, deps mtime 476382872 (VALID)

build/version.obj: #deps 0, deps mtime 476382873 (VALID)

build/includes_normalize-win32.obj: #deps 0, deps mtime 476382874 (VALID)

build/subprocess-win32.obj: #deps 0, deps mtime 476382874 (VALID)

build/msvc_helper_main-win32.obj: #deps 0, deps mtime 476382874 (VALID)

build/debug_flags.obj: #deps 0, deps mtime 476382874 (VALID)

build/msvc_helper-win32.obj: #deps 0, deps mtime 476382874 (VALID)

build/depfile_parser.obj: #deps 0, deps mtime 476382875 (VALID)

build/ninja.obj: #deps 0, deps mtime 476382875 (VALID)

build/edit_distance.obj: #deps 0, deps mtime 476382876 (VALID)

build/build_log.obj: #deps 0, deps mtime 476382877 (VALID)

[...]

PR #1104, opened 2016-02-05
try to fix msvc bootstrap

fbenkstein:
Try to address issue https://github.com/ninja-build/ninja/issues/1103.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -527,11 +527,46 @@ ninja = n.build(binary('ninja'), 'link', objs, implicit=ninja_lib,
 n.newline()
 all_targets += ninja
 
+ninja_writer.comment('Regenerate build files if build script changes.')
+ninja_writer.rule('configure',
+                  command='${configure_env}%s $root/configure.py $configure_args' %
+                  options.with_python,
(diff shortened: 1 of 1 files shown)


nico:
Whoa, thanks for noticing that! #1106 might be a better fix though -- your patch makes it so that the test targets aren't written in bootstrap mode from what I understand. (Also, I didn't understand why the MinGW check is no longer necessary.) Does that make sense?

fbenkstein:
https://github.com/ninja-build/ninja/pull/1106 much better than this.  Please have a look at https://github.com/ninja-build/ninja/pull/1105 though.  Solaris has the same issue with ``-H`` being the equivalent of ``/showIncludes``.  I tried generalizing this but couldn't find a good solution.  I'll look into the MinGW issue, maybe it's no longer relevant.  WRT to the test targets my change makes it so that ``build.ninja`` is always rewritten by the bootstrap ninja rerunning ``configure.py`` as the first step - that means the test targets are not necessary at this stage.

nico:
ok, closing this in favor of #1106 then.

PR #1105, opened 2016-02-05
enable building ninja with Oracle Solaris Studio

fbenkstein:
This adds support for compiling with Solaris Studio on Solaris.  I've only tested Solaris 11.3 x86 so far.  The Solaris C++ compiler supports printing include files when given the `-H` option so I've add a small parser for that format.  It is very similar to MSVC's /showIncludes.  If requested I can move this parser into a separate PR.  Other than that only very few changes to the code were necessary:  STLPort has a hash_map implementation that needs to be used and `src/test.cc` uses `mkdtemp` without including `stdlib.h` (I suppose on other platforms it is already included through `unistd.h`).  And lastly `src/build.h` includes `cstdio` instead of `stdio.h` the former does not declare `snprintf` only `sprintf` on Solaris.  ~~I`ve also fixed the warnings issued by the compiler as they mostly seemed sensible.  I can move that into a separate PR as well.~~

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -488,6 +488,7 @@ for name in ['build',
              'manifest_parser',
              'metrics',
              'state',
+             'unixcc_parser',
              'util',
              'version']:
     objs += cxx(name)
@@ -586,6 +587,7 @@ for name in ['build_log_test',
              'state_test',
              'subprocess_test',
              'test',
+             'unixcc_parser_test',
              'util_test']:
     objs += cxx(name)
 if platform.is_windows():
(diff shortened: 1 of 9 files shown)


fbenkstein:
I took out the warnings change an will work on that separately.

nico:
The minor changes (hash_map, the additional include) are fine.

I don't want to add more deps modes, please see #721 -- how does solaris studio work with make?

https://groups.google.com/forum/?fromgroups#!topic/ninja-build/eByS91Slbwc suggests that it does support writing regular depfiles -- maybe that could be a path forward?

jhasse:
In order to reduce the number of open PRs I'm closing this one. If you still think this should be merged, feel free to reopen and add a comment.

PR #1109, opened 2016-02-15
enhance write_fake_manifest.py

fbenkstein:
Add command line options to write_fake_manifest.py to change the number of
targets being generated and the seed.  More importantly, add an option to
generate sources as well so the manifest generated can actually be executed
successfully.

### DIFF
--- a/misc/write_fake_manifests.py
+++ b/misc/write_fake_manifests.py
@@ -50,9 +50,10 @@ def moar(avg_options, p_suffix):
 
 
 class GenRandom(object):
-    def __init__(self):
+    def __init__(self, src_dir):
         self.seen_names = set([None])
         self.seen_defines = set([None])
+        self.src_dir = src_dir
 
     def _unique_string(self, seen, avg_options=1.3, p_suffix=0.1):
         s = None
@@ -76,7 +77,7 @@ class GenRandom(object):
 
     def src_obj_pairs(self, path, name):
         num_sources = paretoint(55, alpha=2) + 1
-        return [(os.path.join('..', '..', path, s + '.cc'),
+        return [(os.path.join(self.src_dir, path, s + '.cc'),
                  os.path.join('obj', path, '%s.%s.o' % (name, s)))
                 for s in self._n_unique_strings(num_sources)]
 
@@ -103,12 +104,8 @@ class Target(object):
         self.kind = kind
         self.has_compile_depends = random.random() < 0.4
 
-    @property
-    def includes(self):
-        return ['-I' + dep.dir_path for dep in self.deps]
 
-
-def write_target_ninja(ninja, target):
+def write_target_ninja(ninja, target, src_dir):
     compile_depends = None
     if target.has_compile_depends:
       compile_depends = os.path.join(
@@ -117,8 +114,7 @@ def write_target_ninja(ninja, target):
       ninja.newline()
 
     ninja.variable('defines', target.defines)
-    if target.deps:
-        ninja.variable('includes', target.includes)
+    ninja.variable('includes', '-I' + src_dir)
     ninja.variable('cflags', ['-Wall', '-fno-rtti', '-fno-exceptions'])
     ninja.newline()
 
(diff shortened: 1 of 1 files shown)


misc/write_fake_manifests.py
```diff
@@ -129,13 +128,49 @@ def write_target_ninja(ninja, target):
     deps = [dep.output for dep in target.deps]
     libs = [dep.output for dep in target.deps if dep.kind == LIB]
     if target.kind == EXE:
-        ninja.variable('ldflags', '-Wl,pie')
+        # ninja.variable('ldflags', '-Wl,-pie')
```
nico: ?

misc/write_fake_manifests.py
```diff
@@ -129,13 +128,49 @@ def write_target_ninja(ninja, target):
...
     ninja.build(target.output, link, [obj for _, obj in target.src_obj_pairs],
                 implicit=deps)
 
 
+def write_sources(target, root_dir):
+    need_main = target.kind == EXE
+
+    includes = []
+
+    # include siblings
```
nico: nit: Please use sentences for comments (i.e. `# Include siblings.`)

misc/write_fake_manifests.py
```diff
@@ -129,13 +128,49 @@ def write_target_ninja(ninja, target):
...
+    for cc_filename, _ in target.src_obj_pairs:
+        includes.append(os.path.basename(os.path.splitext(cc_filename)[0] + '.h'))
+
+    # include deps
+    for dep in target.deps:
+        for cc_filename, _ in dep.src_obj_pairs:
+            includes.append("%s/%s" % (dep.dir_path, os.path.basename(os.path.splitext(cc_filename)[0] + '.h')))
```
nico: please wrap at 80 columns

misc/write_fake_manifests.py
```diff
@@ -149,7 +184,7 @@ def write_master_ninja(master_ninja, targets):
       command='$cxx -MMD -MF $out.d $defines $includes $cflags -c $in -o $out',
       depfile='$out.d', deps='gcc')
     master_ninja.rule('alink', description='LIBTOOL-STATIC $out',
-      command='rm -f $out && libtool -static -o $out $in')
+      command='rm -f $out && ar rcs -o $out $in')
```
nico: I suppose this needs `libtools -static` on OS X and `ar rcs` on other POSIXen (check sys.platform)

nico:
Cool, looks like a nice change, thanks :-)

fbenkstein:
I have made the requested changes and made sure the generated manifest compiles on Linux and OSX.

misc/write_fake_manifests.py
```diff
@@ -199,17 +243,23 @@ def random_targets():
 
 def main():
     parser = argparse.ArgumentParser()
+    parser.add_argument('-s', '--sources', action='store_true', help='write source files', default=False)
```
nico: nit: please wrap at 80 columns (also below)

misc/write_fake_manifests.py
```diff
@@ -76,7 +76,7 @@ def path(self):
 
     def src_obj_pairs(self, path, name):
         num_sources = paretoint(55, alpha=2) + 1
-        return [(os.path.join('..', '..', path, s + '.cc'),
+        return [(os.path.join('src', path, s + '.cc'),
```
nico: Hm, having the build directory as a child of the src directory and the cwd during compilations is fairly common (it's how cmake, gyp, gn all work), so if at all possible I think it'd be good to keep the "../.." prefix on source file references.

fbenkstein: I made this change because I didn't want to write outside the target directory. What if I change the `-s, --sources` option from a boolean to taking a source directory argument?

nico: That sounds good.

nico: (It's my impression that this isn't blocked on you waiting for me; is that correct?)

fbenkstein: Yes, I'm just busy at the moment and suspend my ninja related activities.
Will pick this up again in a few days hopefully.

fbenkstein:
Should have all comments addressed now.

nico:
Thanks, sorry about the super slow turn-around time.

PR #1110, opened 2016-02-17
Fix ambiguous call to set_terminate on Windows platform

G4m4:
On Windows set_terminate() could either be the standard C++ one or (actually the same one but in the global namespace) the CRT one declared in corecrt_terminate.h
Hence this ambiguity - this patch solves it.

Signed-off-by: g4m4 <misept.dieseneuf@gmail.com>

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -1161,7 +1161,7 @@ int main(int argc, char** argv) {
 #if defined(_MSC_VER)
   // Set a handler to catch crashes not caught by the __try..__except
   // block (e.g. an exception in a stack-unwind-block).
-  set_terminate(TerminateHandler);
+  std::set_terminate(TerminateHandler);
   __try {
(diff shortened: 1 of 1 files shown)


G4m4:
Yes it is, sorry about the lack of details, I was considering opening an issue right now...In a nutshell, when trying to build Ninja I got the following error:
_.\src\ninja.cc(1164): error C2668: 'set_terminate': ambiguous call to overloaded
 function_

The two available overloads being the aforementioned:
- _C:\Program Files (x86)\Windows Kits\10\include\10.0.10240.0\ucrt\corecrt_terminate.h(33)_
- _C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE\exception(196)_

I'm using msvc 14 (VS2015).

nico:
Thanks for the patch!

PR #1111, opened 2016-02-25
Detect build graph cycles as early as possible

bradking:
RecomputeDirty is the earliest traversal of the build graph complete with depfile-loaded dependencies.  Teach it to detect cycles and fail immediately.  This avoids the need to tolerate cycles in RecomputeDirty only to diagnose them later.  This also simplifies Plan logic because it can now assume DAG input.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -640,9 +640,10 @@ Node* Builder::AddTarget(const string& name, string* err) {
 }
 
 bool Builder::AddTarget(Node* node, string* err) {
+  if (!scan_.RecomputeDirty(node, err))
+    return false;
+
   if (Edge* in_edge = node->in_edge()) {
-    if (!scan_.RecomputeDirty(in_edge, err))
-      return false;
     if (in_edge->outputs_ready())
       return true;  // Nothing to do.
   }
--- a/src/build.cc
+++ b/src/build.cc
@@ -298,25 +298,21 @@ void Plan::Reset() {
 }
 
 bool Plan::AddTarget(Node* node, string* err) {
-  vector<Node*> stack;
-  return AddSubTarget(node, &stack, err);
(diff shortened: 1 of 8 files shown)


bradking:
Rebased on `master`.

bradking:
Rebased on `master` after 1.7.1 release.

bradking:
Rebased on `master` to resolve conflicts.

FWIW I've been running with this branch for months with no issues.

bradking:
Rebased on `master` after 1.7.2 release.

bradking:
@nico this is the next incremental step toward my work on Fortran support.  It is also a nice cleanup independent of that work.

bradking:
Rebased on `master`.

nico:
Thanks for the patch!

bradking:
Great, thanks for merging!

atetubou:
This change does not run correctly with cmake 2.8.12.
https://bugs.chromium.org/p/chromium/issues/detail?id=762407
Can you fix this?

bradking:
CMake versions 2.8.12.x and 3.0.x incorrectly produce `build testCCompiler.c: phony testCCompiler.c` in `build.ninja`.  Ninja didn't previously detect this cycle.  We'll have to teach Ninja to tolerate it again.  I'll work on a fix.

For reference, this was a bug in CMake introduced by commit `v2.8.12~248^2` (Ninja: Custom Command file depends don't need to exist before building, 2013-06-07) and fixed by commit `v3.1.0-rc1~353^2~2` (Ninja: Consider only custom commands deps as side-effects, 2014-06-27).  See [CMake Issue 14972](https://gitlab.kitware.com/cmake/cmake/issues/14972).

bradking:
I've opened issue #1322 for the problem this causes for CMake 2.8.12 and 3.0.

PR #1112, opened 2016-02-25
browse tool: Allow customization of port number, Py3 fix

Lekensteyn:
Note: the environment variable is named `PORT`, but could also be renamed if you'd like.

I use this tool also remotely over SSH where I set the `BROWSER=true` environment variable to avoid lynx from popping up. Is that information worth adding to the manual (it is part of the Python webbrowser package)?

### DIFF
--- a/src/browse.py
+++ b/src/browse.py
@@ -182,7 +182,7 @@ class RequestHandler(httpserver.BaseHTTPRequestHandler):
     def log_message(self, format, *args):
         pass  # Swallow console spam.
 
-port = 8000
+port = int(os.getenv("PORT", '8000'))
 httpd = httpserver.HTTPServer(('',port), RequestHandler)
 try:
     hostname = socket.gethostname()
--- a/src/browse.py
+++ b/src/browse.py
@@ -31,7 +31,10 @@ import socket
 import subprocess
 import sys
 import webbrowser
-import urllib2
+try:
+    from urllib.request import unquote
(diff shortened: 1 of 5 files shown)


src/browse.py
```diff
@@ -31,7 +31,10 @@
 import subprocess
 import sys
 import webbrowser
-import urllib2
+try:
+    from urllib2 import unquote
```
nico: Isn't the recommendation "try py3 variant, use py2 as fallback"? (I'm not sure, I don't do much py3.)

nico:
Thanks! What do you think about making the port a command-line parameter for the tool instead of an env var?

src/browse.py
Lekensteyn: Yes it is, fixed in updated patch.

Lekensteyn:
The tool is normally invoked via "ninja -t browse [target name to investigate]". I would be ok with (in fact, prefer) adding something like `ninja -t browse -p 1234`, is this something that aligns with the ninja command options?

While at it, I would add a `--no-browser` option as well (inspired by LLVM's `scan-view` tool).

Lekensteyn:
I went ahead and replaced the PORT env var with --port and also added a --no-browser option. The initial command is also made optional (using `all` instead as default, matches the behavior of invoking `ninja` without options).

nico:
Looks great, thanks!

PR #1113, opened 2016-03-05
Revert #910.

nico:
The change caused some issues (it makes it impossible ot use
posix_spawn() and makes it harder to suspend children on ctrl-z).  After
discussing with jln: Since it fixes a corner case that can be fixed by
explicitly running commands that need it in a wrapper that setsid()s
them, let's try reverting it for a while.  Please shout if this is a
problem for you.

See also #1097.

### DIFF
--- a/src/subprocess-posix.cc
+++ b/src/subprocess-posix.cc
@@ -70,11 +70,8 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
         break;
 
       if (!use_console_) {
-        // Put the child in its own session and process group. It will be
-        // detached from the current terminal and ctrl-c won't reach it.
-        // Since this process was just forked, it is not a process group leader
-        // and setsid() will succeed.
-        if (setsid() < 0)
+        // Put the child in its own process group, so ctrl-c won't reach it.
+        if (setpgid(0, 0) < 0)
           break;
 
         // Open /dev/null over stdin.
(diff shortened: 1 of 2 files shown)


nicolasdespres:
I think that if you revert #910, we should also remove the line:
```c
// Put the child in its own process group, so ctrl-c won't reach it.
if (setpgid(0, 0) < 0)
    break;
```
and some other code needed to forward signals received by ninja to its children.

#910 was a fix because they were corner case with `setpgid()`. Putting children in their own process group have the same issue regarding signal handling than putting them in their own session: require to forward signals which is not always easy like with ctrl+z. 

If you want I could prepare a patch that do all that at once, so that users can easily revert it if they need to.

nico:
I was wondering why that call is there :-) I tried to write a test that shows where this line makes a difference (right now all tests pass both with and without it) yesterday. I noticed that without this line, many of the subprocess tests that do "send signal to parent, then sleep" now take one second (the length of the sleep) instead of being instant. I didn't try very long and didn't check blame output yet for why that line is there.

nicolasdespres:
The idea was to be able to interrupt subsets of running commands without stopping ninja completely. I think nobody use this ability at the moment, so we can get rid of it and if a relevant use case come up later, we could still bring it back.

It was introduced by 85ff781fa30fff63c0 and you can find a full discussion about the rational behind that line [here](https://groups.google.com/forum/#!searchin/ninja-build/Rational$20behind$20putting$20child$20processes$20in$20their$20own$20session/ninja-build/Uq6pTV3FC2Y/HAlaNRP2NpcJ).

Issue #373, opened 2012-07-27
ninja -t browse needs to pass ninja flags/environment to subprocess


evmar:
See https://github.com/martine/ninja/pull/141 for example.  Any build-time flags like -C and -f need to be forwarded to the subprocess.

PR #1116, opened 2016-03-07
Pass location of build.ninja file to browse script

colincross:
Pass the values of -f and -C arguments to the browse python script so
they can be passed back to ninja -t query.

### DIFF
--- a/src/browse.cc
+++ b/src/browse.cc
@@ -22,7 +22,7 @@
 #include "build/browse_py.h"
 
 void RunBrowsePython(State* state, const char* ninja_command,
-                     int argc, char* argv[]) {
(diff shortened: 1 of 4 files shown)


colincross:
Fixes #373

nico:
`ninja -C out/Release/ -t browse base` seems to work fine for me already. ninja processes -C and then the subninja running in browse already has that as it's cwd. Why is that part needed?

colincross:
I'll remove it.

colincross:
-C argument removed, and tested that it still works.

nico:
Looks good after resolving conflicts, thanks!

colincross:
Rebased

nico:
Thanks!

PR #1125, opened 2016-04-05
Make deps=msvc experimentally available on non-Windows.

nico:
This makes it possible to run most of the clparser tests on non-Windows,
and is potentially useful for cross-compiling on non-Windows hosts.
Also, the manual didn't document this as Windows-only previously.

If you use this on non-Windows, please let me know, else I might undo
this change again in the future.


NiklasRosenstein:
What tools are there for cross compiling for Windows (with the CL command line interface)? I only know about clang-cl. Though I haven't tried using it on a non-Windows machine (I'm loosely assuming that it's possible to do so), I'd like to try it in the future.

nico:
In addition to clang-cl, cl.exe might work under wine. (I'm using this with clang-cl.)

martine:
https://github.com/ninja-build/ninja/blob/master/HACKING.md#using-microsoft-compilers-on-linux-extremely-flaky

PR #1126, opened 2016-04-06
Release the pool slot held by an edge whether it succeeds or fails

bradking:
This reconciles #1017 and #1023 in a combined PR.  I resolved the conflicts, took tests from both changes, and simplified the logic updates slightly by using ideas from both approaches.

Both changes add a `success` parameter to `Plan::EdgeFinished` and update the implementation in a similar way.  One used a default argument to avoid updating call sites and the other made it required and updated call sites.  I chose the non-default approach because a later PR I will submit for the Fortran work adds another argument to `EdgeFinished` and updates call sites too.  It is better for all callers to be explicit about their intended value for the `success` argument.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -382,19 +382,24 @@ void Plan::ScheduleWork(Edge* edge) {
   }
 }
 
-void Plan::EdgeFinished(Edge* edge) {
+void Plan::EdgeFinished(Edge* edge, bool success) {
   map<Edge*, bool>::iterator e = want_.find(edge);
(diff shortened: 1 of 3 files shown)


nico:
Thanks!

I might replace the bool parameter with an enum going forward, I find `EdgeFinished(edge, kEdgeSucceeded)` easier to read than `EdgeFinished(edge, true)`.

bradking:
> replace the bool parameter with an enum going forward

Good idea.  Done in #1131.

PR #1130, opened 2016-04-18
ninja_test: Fix Visual C++ 2015 warnings/errors about name hiding.

JohnstonJ:
Visual C++ 2015 warns if a local variable hides visibility of another variable
in a higher scope.  Since this project declares warnings as errors, ninja_test
simply won't build on Visual C++ 2015.

The variables have been renamed and scope limited as appropriate, so that
ninja_test will build without error now on Visual C++ 2015.


nico:
Thanks!

PR #1131, opened 2016-04-19
Improve Plan::EdgeFinished signature

bradking:
Use an enumeration instead of a boolean to clarify the purpose of
arguments at call sites.


nico:
Thanks :-)

PR #1137, opened 2016-04-27
Fix crash with duplicate implicit outputs.

### LINKED ISSUE #1136: Crash with multiple duplicate implicit outputs
```
ninja/tmp$ cat build.ninja
rule b
  command = touch $out

build asdf | asdf asdf: b a
ninja/tmp$ touch a
ninja/tmp$ ../ninja
ninja: warning: multiple rules generate asdf. builds involving this target will not be correct; continuing anyway [-w dupbuild=warn]
ninja: warning: multiple rules generate asdf. builds involving this target will not be correct; continuing anyway [-w dupbuild=warn]
Segmentation fault
```

Found by afl-fuzz. @bradking, do you want to try and fix this for 1.7?

(Sorry, should've started fuzzing earlier)

nico:
Sadly, duplicate outputs aren't an error by default in Ninja (see also
a new edge has no effect.  Remember to decrement the "number of
implicit outputs" counter for the new edge when this happens.

Fixes #1136.

### DIFF
--- a/src/manifest_parser.cc
+++ b/src/manifest_parser.cc
@@ -339,8 +339,8 @@ bool ManifestParser::ParseEdge(string* err) {
   }
 
   edge->outputs_.reserve(outs.size());
-  for (vector<EvalString>::iterator i = outs.begin(); i != outs.end(); ++i) {
-    string path = i->Evaluate(env);
+  for (size_t i = 0, e = outs.size(); i != e; ++i) {
+    string path = outs[i].Evaluate(env);
     string path_err;
     unsigned int slash_bits;
     if (!CanonicalizePath(&path, &slash_bits, &path_err))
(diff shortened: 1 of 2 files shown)


nico:
@bradking, does this make sense to you?

bradking:
Yes, this looks good, except for the unclosed "(see also" in the commit message.

Good catch in finding this, and thanks for fixing it.  I saw the issue just as I had to leave work yesterday and was going to fix it first thing this morning.  Now I find it's already done :)

bradking:
> Sadly, duplicate outputs aren't an error by default in Ninja

Is this only for compatibility with `build.ninja` files that have duplicate outputs that were not originally diagnosed?  We could consider making an *implicit* output that duplicates any other output be an error because such `build.ninja` files cannot exist yet since the feature has never been released.  We have a chance now to do this before 1.7.

bradking:
> consider making an implicit output that duplicates any other output be an error 

See #1141.

nico:
I thought about that, but I thought it's weird to have different behavior for the two output types.

PR #1138, opened 2016-04-27
ignore

nico:


jhasse:
I think this was a test for CI, right?

nico:
yes :-)

Issue #1139, opened 2016-04-27
Add GNU make jobserver client support


stefanb2:
As long as ninja is the only build execution tool, the current `ninja -jN` implementation works fine.

But when you try to convert parts of an existing recursive GNU make based SW build system to ninja, then you have the following situation:
- top-level GNU Make (with -jX, acts as job server)
- M instances of GNU make (with -j, act as job server clients)
- N instances of ninja (don't know anything about job server)

Simply calling `ninja -jY' isn't enough, because then the ninja instances will try to run Y*N jobs, plus the X jobs from the GNU make instances, causing the build host to overload. Relying on -lZ to fix this issue is sub-optimal, because load average is sometimes too slow to reflect the actual situation on the build host.

It would be nice if GNU make jobserver client support could be added to Ninja. Then the N ninja instances would cooperate with the M GNU make instances and on the build host only X jobs would be executed at one time.

stefanb2:
I have tested this implementation over the last few weeks in two different recursive GNU make based build systems that originally had M+1 GNU make instances:
* use case A: top-level GNU make, 1 ninja instance, M-1 GNU make instances
* use case B: top-level GNU make, N ninja instances, M-N GNU make instances

FYI: google/kati was used to convert existing single makefile GNU make parts to Ninja build file.

nico:
Thanks for the patch!

We've discussed this on the mailing list a few times (e.g. here https://groups.google.com/forum/#!searchin/ninja-build/jobserver/ninja-build/PUlsr7-jpI0/Ga19TOg1c14J). Ninja works best if it knows about the whole build. Now that kati exists, one can convert those to ninja files and munge them up to have a single build manifest (that's Android's transition strategy from Make to Ninja -- they use kati to get everything converted to Ninja files, and then they're incrementally converting directories to use something-not-make -- and then kati produces parts of their Ninja files and the new thing produces parts of the ninja files.)

Is your use case that you have recursive makefiles?

stefanb2:
I could have guessed that this has been discussed before, because I'm surely not the first person facing such a situation.

Here are *my* reasons for requesting this:

1. **recursion:** kati currently can't translate recursive GNU make based build systems, like Linux kernel kbuild. IMHO a major effort and unfortunately I can't wait for kati to provide this, hence the such sub-component builds will have to stay with GNU make for the time being. 
[...]

PR #1140, opened 2016-04-27
Add GNU make jobserver client support

stefanb2:
- add new TokenPool interface
- GNU make implementation for TokenPool parses and verifies the magic
  information from the MAKEFLAGS environment variable
- RealCommandRunner tries to acquire TokenPool
  - if no token pool is available then there is no change in behaviour
- When a token pool is available then RealCommandRunner behaviour
  changes as follows
  - CanRunMore() only returns true if TokenPool::Acquire() returns true
  - StartCommand() calls TokenPool::Reserve()
  - WaitForCommand() calls TokenPool::Release()

Fixes https://github.com/ninja-build/ninja/issues/1139


stefanb2:
For easier review I provided a PR based on the squashed commit with the final code.

If you are interested in the development history with all the exploration and mistakes then please look at branch https://github.com/stefanb2/ninja/tree/topic-issue-1139-full

stefanb2:
I see what you mean. That probably explains why ninja instances are less aggressive than GNU make instances in acquiring tokens. In real life that doesn't really matter, because except of the tail end of the build the tokens are always used by somebody else and most build steps are short, i.e. if a token should be available the ninja instance will become aware of it soon.

Unfortunately there is no "token interrupt" available. I guess we would need to add the TokenPool to the SubProcessSet::DoWork() API and then add an internal SubProcessSet <-> TokenPool interface where TokenPool can provide poll/select information (if available) to SubProcessSet's waiting mechanism.

Could this be handled as future improvement request or do you see this as a blocker for getting this change merged?

src/tokenpool-gnu-make.cc
```diff
@@ -0,0 +1,148 @@
...
+#else
+  fd_set set;
+  struct timeval timeout = { 0, 0 };
+  FD_ZERO(&set);
+  FD_SET(rfd_, &set);
+  int ret = select(rfd_ + 1, &set, NULL, NULL, &timeout);
+#endif
+  if (ret > 0) {
+    char buf;
+    int ret = read(rfd_, &buf, 1);
```
colincross: The select above provides no guarantee that the read here will not block.  If another job takes the last token between the select and the read, ninja will be stuck here until a make job finishes.  That means that ninja can not process of its children exiting and start any new processes until make finishes a job, which could be an arbitrarily long time.

http://make.mad-scientist.net/papers/jobserver-implementation/ describes how difficult this is to get right, and involves dup'ing the fd so that it can be closed in a SIGCHLD signal handler to abort the read.

A non-blocking fd would have made this all trivial, but the make authors decided that select wasn't portable enough.  Since all jobserver clients share the same pipe, you can't make the fd non-blocking without affecting all of them.

src/tokenpool-gnu-make.cc
```diff
@@ -0,0 +1,148 @@
...
+    if (ret > 0) {
```
colincross: Handle EINTR, and log on error?  If something happens to the pipe ninja will silently stop running jobs.

src/tokenpool-gnu-make.cc
```diff
@@ -0,0 +1,148 @@
...
+}
+
+void GNUmakeTokenPool::Reserve() {
+  available_--;
+  used_++;
+}
+
+void GNUmakeTokenPool::Return() {
+  const char buf = '+';
+  if (write(wfd_, &buf, 1) > 0)
```
colincross: Handle EINTR?

src/tokenpool-gnu-make.cc
stefanb2: Please remember: this code is only called from build.cc when ninja has at least one active child [`subprocs_.running_.empty() == false`].

If read() gets interrupted it just means we didn't get a token, i.e. we return `false`. The code then returns to the main loop in build.cc for normal processing, e.g. reaping a terminated child process.

As I wrote already in another comment: this means ninja is a less aggressive jobserver client. Increasing the aggressiveness should be handled by a future change.

src/tokenpool-gnu-make.cc
stefanb2: Please remember: this code is only called from build.cc when ninja has at least one active child [`subprocs_.running_.empty() == false`].

In the case we do enter the read() and it blocks then the following 4 things can happen:

1. it returns successfully. That's what we wanted anyway and return `true`.
2. a child process terminates, interrupting the read which then returns -1. We didn't get a token and return `false`. The code returns to the main loop in build.cc for normal processing, e.g. reaping the child.
3. a child terminates between the poll/select() call and entering read(). The read returns immediately with -1 -> same as (2).
4. any other signal -> same as (2) or (3)

Please correct me if I'm wrong but in every case ninja returns to the main loop and never stalls.

src/tokenpool-gnu-make.cc
stefanb2: Makes sense. Implemented in https://github.com/stefanb2/ninja/commit/c139f319a6b509dfce538183baf4349683bda3bc

src/tokenpool-gnu-make.cc
colincross: Ninja doesn't enable SIGCHLD, and it is ignored by default, so it won't interrupt the read.

colincross: Also, statement #3 is wrong.  Even if SIGCHLD is enabled, the read will not return -1 if the SIGCHLD happens after the poll/select but before the read.  That race condition is the reason make dups the fd before reading it - the SIGCHLD handler closes the dup'd copy of the fd, so the read returns -EBADF.

stefanb2: Addressed concern by wrapping read() in alarm(1)/alarm(0) + installing corresponding signal handler.

IMHO this simple solution is good enough, because the race condition is rare. From the two different, large test builds I have seen the perror() only once in one build log.

stefanb2: FYI: I made a test change that added a dummy SIGCHLD handler that does nothing. I.e. it just unblocks the token read() when an active child exited. But that negatively impacted ninja performance. In my test build the build time with ~50K build steps it increased ~30%. Therefore I will not push that change.

The current patch set should be good to go then.

colincross:
I think the race conditions in both directions (ninja not reaping and starting new jobs because it is stuck on the read in Acquire(), and ninja not noticing an available token because it is stuck in DoWork waiting for jobs to finish) are both pretty fundamental problems.  If they are not fixable I don't think this patch should be merged.

Ninja doesn't use threads right now, but one option would be to have a separate thread that managed acquiring tokens and signalling the main thread when they are available.  The main thread would signal the jobserver thread when it wanted a token, the jobserver thread would call a blocking read on the jobserver pipe and write to a signalling pipe to interrupt DoWork when it returned.  That's a lot of complexity to add to ninja though, for a feature where the ideal solution is still to convert everything to ninja.

maximuska:
I think using timer signal makes read essentially non blocking, solving the
issue, except it still leaves a theoretical window for alarm signal to be
delivered before read is called. A read call which atomically unblocks
signals would be needed to make this rock solid, somewhat like pselect().

The comment about sigchld performance impact made me curious. Have you
replaced alarm signal or used both signals in your experiment?

On May 26, 2016 11:12 PM, "colincross" <notifications@github.com> wrote:
>
> I think the race conditions in both directions (ninja not reaping and
starting new jobs because it is stuck on the read in Acquire(), and ninja
not noticing an available token because it is stuck in DoWork waiting for
jobs to finish) are both pretty fundamental problems. If they are not
fixable I don't think this patch should be merged.
>
> Ninja doesn't use threads right now, but one option would be to have a
separate thread that managed acquiring tokens and signalling the main
thread when they are available. The main thread would signal the jobserver
thread when it wanted a token, the jobserver thread would call a blocking
read on the jobserver pipe and write to a signalling pipe to interrupt
DoWork when it returned. That's a lot of complexity to add to ninja though,
for a feature where the ideal solution is still to convert everything to
ninja.
>
> —
> You are receiving this because you are subscribed to this thread.
<https://github.com/ninja-build/ninja/pull/1140#issuecomment-221980890>

src/tokenpool-gnu-make.cc
```diff
@@ -0,0 +1,198 @@
...
+  struct timeval timeout = { 0, 0 };
+  FD_ZERO(&set);
+  FD_SET(rfd_, &set);
+  int ret = select(rfd_ + 1, &set, NULL, NULL, &timeout);
+#endif
+  if (ret > 0) {
+    char buf;
+
+    interrupted_ = false;
+    alarm(1);
```
colincross: There is still a race here, although an extremely unlikely one.  If the process were preempted after alarm(1) but before read the signal could be delivered before the read started, and the read would block indefinitely.

See https://android.googlesource.com/platform/build/+/master/tools/makeparallel/makeparallel.cpp#166 for what I believe is a race-free implementation of this read, based on the ideas from the make jobserver document.

stefanb2: > and ninja not noticing an available token because it is stuck in DoWork waiting for jobs to finish) are both pretty fundamental problems

Not really:
- ninja will not stall, i.e. it will always have at least one job running
- tokens will not be wasted, other clients will be using them

> That's a lot of complexity to add to ninja though, for a feature where the ideal solution is still to convert everything to ninja.

See https://github.com/ninja-build/ninja/issues/1139#issuecomment-215160371 why that is not possible.

stefanb2: Implemented something similar which also installs the same signal handler temporarily also for SIGCHLD.

Performance is about the same in my testbuild as with the last patchset.

stefanb2:
I've added a proposal for a token monitor in https://github.com/stefanb2/ninja/commit/9832d656e9d80814f07f9c3805b78e5765b8e9ef. SubprocessSet::DoWork() will now exit when token read file descriptor becomes readable.

stefanb2:
I hope the latest patch set addresses all reviewer comments.

dcolascione:
What are the remaining reasons not to merge this code?

fabio-porcedda:
I would like too have this feature merged, i simply cannot convert all projects to ninja-build because i'm not allowed to do that.

@stefanb2 Thanks a lot for your work

hwti:
Unlike make, this will use the jobserver even if `-j` is specified on the command line.
Perhaps the jobserver should be ignored in these cases, so a `make -j2` could launch a `ninja -j4` for example.

stefanb2:
@hwti: this is unnecessary, because when you use ninja in a makefile rule you must *explicitly* enable the jobserver magic. As documented in the GNU make manual [here](https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html#MAKE-Variable) and [here](https://www.gnu.org/software/make/manual/html_node/POSIX-Jobserver.html#POSIX-Jobserver) the job server magic will only be activated for a job if

- the command line contains the string `$(MAKE)` *before* it is expanded, or
- the start of the command line is tagged with the `+` token *after* it is expanded.

I.e. if your makefile has

```
target: ....
    ninja -j4 ...
```

ninja will never use the jobserver but always run 4 jobs in parallel, no matter if -jN is given on the GNU make command line or not.

In the case that the ninja invocation is hidden inside a script that is executed by GNU make with active jobserver magic you can simply use

```
export MAKEFLAGS=
...
ninja -j4 ...
```
or

```
MAKEFLAGS= ninja -j4 ...
```

hwti:
@stefanb2: What I meant is this is a case where make and ninja behavior differs, so it can be confusing.

For example, if the top-level makefile has
```
target:
     +Tools/Scripts/build-webkit --gtk ...
```
The build-webkit script uses cmake with either ninja (if available) or make, and adds -jNCPUs by default.
Ninja would use the jobserver, even if make doesn't, effectively hiding the issue that the script should check MAKEFLAGS before adding the argument.

Moreover, if a script wanted to disable the jobserver, it would have to carefully filter MAKEFLAGS, or to do it only in ninja case, as the variable also contains other flags and variable overrides.

stefanb2:
@hwti I guess you meant

> ...uses cmake with either ninja ..., and adds -jNCPUs to ninja command line by default.

That assumption is no longer correct when ninja supports job server, hence cmake behaviour should be updated or at least be made configurable.

But even without that update I don't see why this is a problem. If you update your ninja to a version that supports jobserver, then the ninja instance in the build will simply honour the jobserver instead of running with hard-coded `-jNCPUs`. If the build system would fail due to "incorrect" value of NCPUs then this would be a bug in the build system, not ninja.

BTW: if my guess was wrong and you meant that cmake *always* adds -jNCPUs no matter if make or ninja, then the `+` in your makefile rule is incorrect, because build-webkit is supposed to ignore jobserver.

hwti:
@stefanb2 
> That assumption is no longer correct when ninja supports job server, hence cmake behaviour should be updated or at least be made configurable.

It's build-webkit which calls `cmake --build ... -- -jNCPUs`, which gives the argument to either ninja or make.
If ninja honors the jobserver in this case, the build could work, instead of triggering an OOM due to too many g++ processes with make and the current official ninja.

So the real bug is in the script, which should check `MAKEFLAGS`, and only consider adding the argument if the top-level makefile was called without parallelism (or if there is no `+`, then it would be a bad buildsystem integration).

In this particular case, the behavior is "known", since it isn't new.
But you can imagine someone starting a new project, making the same kind of mistake, and not seeing it. Then only other people which only have make, or an older/unpatched ninja version would have issues.

xim:
+1 for the feature, would like to see this merged.

I've been experimenting with using these patches for a project which has a bunch of subprojects. There's a top-level python script that generates the ninja files, then calls them (each with a set of different env params / command line switches). Currently the ninja files are run in serial, which means plenty of CPU idle during linking. I want to use a jobserver approach to improve build times on systems with many CPUs.

I made a proof-of-concept wrapping Makefile that runs the python script inside a jobserver, and it works.

Had to pass close_fds=False to the subprocess.call(...) in my python script to get this to work. If I didn't, things appeared to work but ninja ate all the CPU it could. I suspect ninja was in a tight loop trying to access the jobserver fds. Minor bug?

xim:
I see my earlier assumption on why ninja was eating all my CPU may have been wrong. I was running a massive build process on a build server with many parallel ninjas on a build server, and the ninjas appear to run into some kind of trouble after a while if I used `-l` to scale jobs.

I am running on a 20-core build server (+ hyper-threading), 42 ninja-based projects to be built. If I use many parallel ninjas, and pass `-l50` for load based task scaling, intermittently all the ninjas will use 100% CPU. As the number of ninjas increases, the risk of them getting stuck eating all the CPU increases.

Without `-l` there is no problem.

stefanb2:
@xim: -l is only necessary when you run multiple independent(!) build jobs, each with its own top-level make (== own jobserver), on the same machine.

If you have only one build job per machine, with one top-level make (== 1 jobserver), then the total amount of parallel executed build steps will never pass the -jN limit **if** all children adhere to the jobserver protocol. If you have children that do not adhere to the protocol, i.e. that execute more than one build step in parallel without requesting a token, e.g. hard-coded `make -jN` commands, then your build will run into overload situations.

stefanb2:
@hwti sorry for taking so long to follow up on your question.

Do I understand you correctly that you want `nina -jN` to emulate the *exact* GNU make behaviour? I.e. 

1. print a warning that jobserver is ignored, and
2. execute N jobs in parallel

I wrote a small `.ninja` file and tested the following situations with `make -jN -f Makefile.top` that calls

case | command | result
------|------------|----------
1 | `+ninja` | will execute max. N jobs
2 | `+ninja -j1` | will execute max. 1 job
3 | `+ninja -jM` | with M <= N: will execute max. M jobs
4 | `+ninja -jM` | with M > N: will execute max. N jobs

* case 2 already works correctly.
* case 3 is mainly correct, but ninja may execute less than the desired M jobs.
* case 4 will execute less than the desired M jobs.

So in cases 3 & 4 `ninja -jN` currently behaves differently than `make -jN`.

stefanb2:
@xim in one respect you are correct though: the current implementation does not honor the `-lN` value that GNU make passes down in `MAKEFLAGS`. I.e. with `make -jN -lM` calling `+ninja`, ninja will execute max N jobs, but not check that load average is below M when starting a parallel job.

I will have to check the source if this improvement can be implemented.

src/build.cc
```diff
@@ -508,21 +518,27 @@ vector<Edge*> RealCommandRunner::GetActiveEdges() {
...
 bool RealCommandRunner::CanRunMore() {
   size_t subproc_number =
       subprocs_.running_.size() + subprocs_.finished_.size();
   return (int)subproc_number < config_.parallelism
-    && ((subprocs_.running_.empty() || config_.max_load_average <= 0.0f)
-        || GetLoadAverage() < config_.max_load_average);
+    && (subprocs_.running_.empty() ||
+        ((config_.max_load_average <= 0.0f ||
+          GetLoadAverage() < config_.max_load_average)
+      && (!tokens_ || tokens_->Acquire())));
```
michaelolbrich: I don't think this is correct. The configured maximum number of jobs should be ignored when the jobserver is used. With the current code, the number of jobs is always limited to GuessParallelism() even if there are tokens available.

stefanb2:
@hwti your concerns should be addressed with the latest patch set
@xim can you try your scenario with `make -jN -lM` calling `ninja` again with the latest patch set?

src/build.cc
stefanb2: Fixed in the latest patch set

stefanb2: @michaelolbrich actually back then this was *intentional* and made sense with my proposed change to the [AOSP build system](https://android-review.googlesource.com/231561) as it always passes `-jN` down to ninja.

I'm not sure any more why I didn't realize that this isn't the correct behaviour for `+ninja` when running tests on a non-Android build system. And since then I forgot this little detail...

xim:
I can, and I did, and there's still a bug! If I add -l50 -j50 and the load hits 50, the ninjas sping out of control and eat lots of CPU. The CPU used by ninja further increases the load, so we don't drop below a load of 50 too easily. I think this should be easily reproducible. I use a Makefile that invokes a bunch of parallel ninjas, each doing tons of small jobs.

Another issue I stumbled upon: ninja -j and ninja -j0 should be supported. Make allows `-j` and it's interpreted as "as many jobs as you like" or thereabouts.

By the way, IMHO the message that we're using a make jobserver should not be printed by default.

stefanb2:
```
$ ./ninja -j build.ninja 
ninja: fatal: invalid -j parameter
$ ./ninja -j0 build.ninja 
ninja: fatal: invalid -j parameter
```

@xim: that is default ninja behaviour. Changing this is beyond the scope of this PR.

xim:
Unsure if I agree, @stefanb2, as `-j` is part of the default `$MAKEFLAGS` sent from Make < 4.2. But sure, I can see wanting to do that in its own pull request. Will you, or do you want me to give it a go? ^_^

stefanb2:
I fail to see what the contents of `$MAKEFLAGS` have to do with the ninja command line. All this PR does is interpret & validate the contents of `$MAKEFLAGS` and setting up a mechanism with which ninja can act as a client to a GNU make jobserver. All of this happens *after* ninja has parsed the command line.

Maybe I should repeat this. To replace GNU make in a top-level makefile rule like
```
subtarget: ...
       $(MAKE) -f Makefile.subtarget ...
```

with ninja, you should use
```
subtarget:
       +ninja -f build-subtarget.ninja ...
```
There should be no `-j` nor `-l` option on the command line.

stefanb2:
> the message that we're using a make jobserver should not be printed by default.

@xim: unfortunately ninja `-d` does not offer a flag for internal debugging messages. The closest one that could be used would be `-d explaining` but IMHO that would be bending the meaning of that flag.

stefanb2:
@xim OK, I think I have been able to reproduce your `-lN` issue with a single ninja that is started from a top-level GNU make.

The problem does not occur when I the run the same file with `ninja -jN -lN`, i.e. without jobserver.

xim:
> There should be no -j nor -l option on the command line.

Agree that `-l` isn't required. but we should still fix the bug. Do you want me to set up a minimal reproduction?

xim:
> unfortunately ninja -d does not offer a flag for internal debugging messages

Maybe just drop the message? IMHO it's great for debugging, but annoying when that's the expected behavior.

stefanb2:
@xim IMHO the message is valuable, because it indicates a change in behaviour for ninja. Without it you'll never notice if some integration is broken and you are actually **not** running with jobserver although you think you are.

stefanb2:
@xim: no need. I think I already identified the busy loop:

* at the start of the main loop `::CanRunMore()` is called. It will always return false when there are pending jobs and the load average limit has been reached.
* the main loop then goes to the next step and calls `::WaitForCommand()`
* if a token is available that method returns immediately and the main loop is started all over again
* but as nobody consumes the available token until the load is under the limit again this goes on and on...

Without the jobserver `::WaitForCommand()` is always blocking and will in normal case only return when `SIGCHLD` has been received.

I think one solution could be to pass the return value of `::CanRunMore()` to `::WaitForCommand()` as signal to ignore available tokens. But I have to think this through before making any changes.

stefanb2:
@xim easy reproducer is `make -jN -l1`, N >> 1 and a single ninja started from the top-level makefile. That will create the situation that always many tokens are available while the load limit prevents them form being used.

stefanb2:
@xim the -lN busy loop issue has been addressed in the latest patch set. I can't reproduce it any more with my test case.

Please note: as it was a bug in the TokenPool monitoring the fix has been squahsed into that commit. The non-squashed version can be found here https://github.com/stefanb2/ninja/commit/e1fc476fde3bf5c28ba2350210fd665e7e22b02f

stefanb2:
@xim I changed the printf() to LinePrinter calls. The GNU make server message is now only printed for verbose builds. I hope that this choice fulfils your request.

yann-morin-1998:
@stefanb2 @xim What is missing for this to get more attention and a chnace to get merged?
Is more tested needed? If so, what kind of report do you expect?

stefanb2:
@yann-morin-1998 testing isn't the problem here, although I agree it would be nice to have some test cases for the new functionality in the ninja test suite. Any suggestions are welcome.

Judging by https://github.com/ninja-build/ninja/issues/1139#issuecomment-215087721 the functionality simply isn't wanted by the project/maintainers. IMHO a lot of different people, beside myself, have given good arguments why this PR is valuable, but that doesn't seem to be enough.

yann-morin-1998:
@stefanb2 Thanks for pointing to the issue tracker. I've further replied there as well.

Thanks for your work on this, it's much appreciated. :-)

stefanb2:
@yann-morin-1998 you mention Buildroot in your comment. I have been contemplating submitting this PR as patch to Buildroot. Especially now that `meson`, which requires ninja, has been integrated it would make a lot of sense.

Maybe getting this patch into downstream projects can increase the pressure on upstream to merge it.

yann-morin-1998:
@stefanb2 We try to avoid carrying feature patches in Buildroot. Note
also that we currently have no ninja-based packages in Buildroot.

Besides, we're not yet ready for top-level parallel builds anyway, so
we do not really have an official way of testign this. Officially, at least! ;-)

So, I've been briefly testing your patches on top of current master, and
I did at least not notice any regression in our use-case so far. I'll report
further when I've done more testing (which will concentrate on the single
package we're currently converting to meson/ninja: systemd).

Thanks! :-)

src/build.h
```diff
@@ -116,15 +120,16 @@ struct CommandRunner {
     bool success() const { return status == ExitSuccess; }
   };
   /// Wait for a command to complete, or return false if interrupted.
-  virtual bool WaitForCommand(Result* result) = 0;
+  virtual bool WaitForCommand(Result* result, bool more_ready) = 0;
```
nico: document what more_ready means

nico:
As said on #1139, let's try and move forward with this.

This is a large-ish test with 0 new test coverage. Is it possible to add some test coverage?

@colincross You had some concerns about the basic approach. Are they addressed now?

src/tokenpool-gnu-make.cc
```diff
@@ -0,0 +1,241 @@
...
+      struct sigaction act, old_act;
+      int ret = 0;
+
+      memset(&act, 0, sizeof(act));
+      act.sa_handler = CloseDupRfd;
+      if (sigaction(SIGCHLD, &act, &old_act) == 0) {
+        char buf;
+
+        // block until token read, child exits or timeout
+        alarm(1);
```
colincross: 1 second is a pretty long time for ninja to not reap or start any jobs, setitimer would allow a finer grained timeout.

src/tokenpool-gnu-make.cc
```diff
@@ -0,0 +1,241 @@
...
+  int ret = poll(pollfds, 1, 0);
+#else
+  fd_set set;
+  struct timeval timeout = { 0, 0 };
+  FD_ZERO(&set);
+  FD_SET(rfd_, &set);
+  int ret = select(rfd_ + 1, &set, NULL, NULL, &timeout);
+#endif
+  if (ret > 0) {
+    dup_rfd_ = dup(rfd_);
```
colincross: The dup/alarm/read/close logic deserves a long comment explaining the problems it is solving.

colincross:
My main concerns have all been addressed, left some minor comments.

src/tokenpool-gnu-make.cc
stefanb2: If a token is available then Acquire() exits immediately with true. Hence for this case the length of the timeout is not relevant.

If no token is available but a child exits during the timeout then Acquire() exits immediately. As the read() will not have read anything, Acquire() will return false, which causes the main loop to skip starting a new edge and move on to the child reaping step. Hence for this case the length of the timeout is not relevant.

If no token is available and no child exits then reaching this timeout means that we would have more work available but are not allowed to start a new edge (no other jobserver client has released a token). Hence the length of the timeout just controls the interval when ninja checks for user interrupts. IMHO for that use case 1 second is small enough.

stefanb2:
@nico I agree some test coverage would be nice.

The juicy bits that would need testing are between `RealCommandRunner` / `Subprocess(Set)` and `GNUmakeTokenPool`. But ninja test use `FakeCommandRunner` which doesn't have a `TokenPool` at all.

I'm open to be proven wrong on this statement if someone has some ideas what tests to add.

src/tokenpool-gnu-make.cc
```diff
@@ -0,0 +1,261 @@
...
+      if (sigaction(SIGCHLD, &act, &old_act) == 0) {
+        char buf;
+
+        alarm(1);
+
+        // Now try to read() from dup_rfd_. Return values from read():
+        //
+        // 1. token read                               -> 1
+        // 2. pipe closed                              -> 0
+        // 3. alarm expires, i.e. CloseDupRfd() called -> 0
```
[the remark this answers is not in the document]
stefanb2: My original git commit (see https://github.com/stefanb2/ninja/tree/topic-issue-1139-full for the full history) had this. While writing this yesterday I thought that was incorrect and the close() would cause read() to return 0 instead, 

I've now re-read the original block post on which the code is based and realized my mistake.

DONE

bradking:
> But ninja test use `FakeCommandRunner` which doesn't have a `TokenPool` at all.

If anyone works on that, please see #1424 which extends `FakeCommandRunner` in another way.  It's part of my work toward Fortran support.

stefanb2:
Finally have some free cycles to look into writing tests.

AppVeyor build tests on Windows still seem to be broken, not caused by the one test my changes add.

stefanb2:
@nico I've noew added tests for the new functionality. I hope this removes the last obstacle for getting this PR merged.

BTW: the full history is visible at https://github.com/stefanb2/ninja/tree/topic-issue-1139-full. This might be preferable over the squashed version (associated with this PR) to retain the history of the development.

As mentioned before tests in AppVeyor build are still failing for my branch, but not in my code or my tests.

stefanb2:
Latest commit simplifies tests for build module a bit.

Seems like AppVeyor CI fails/passes depending on the phase of the moon....

stefanb2:
Ping? Anybody? Can we get closure on this PR?

stefanb2:
Latest commit adds a Win32 implementation for GNUmakeTokenPool. I have tested in a WIn10 VM with mingw64 make and it looks like it works as expected. It would be nice if someone could test it with a heavy workload.

The current code compiles fine with mingw64 GCC, but VS C++ barfs on a sscanf() feature.

jcfr:
Look like `%m` is a POSIX extension to `scanf`, explicitly allocation of `auth` and using a different approach would address the compilation problem. See https://stackoverflow.com/questions/11926883/scanf-dynamic-allocation#11927148

stefanb2:
Replaced sscanf() with a manual string parser.

Test failure on AppVeyor not caused by my changes.

stefanb2:
@jhasse 

1. no, it is the public API to access the token pool
2. yes, after adding win32 implementation it is unused. But I left it in to make it easy to comment out token pool functionality if necessary
3. tokenpool_test tests the real implementation. Do you mean adding more tests to subprocess_test?
4. done

src/tokenpool-gnu-make.cc
```diff
@@ -0,0 +1,107 @@
...
+
+bool GNUmakeTokenPool::Acquire() {
+  if (available_ > 0)
+    return true;
+
+  if (AcquireToken()) {
+    // token acquired
+    available_++;
+    return true;
+  } else {
```
jhasse: You can drop the `else` here.

src/tokenpool-gnu-make.h
```diff
@@ -0,0 +1,40 @@
...
+
+  // token pool implementation
+  virtual bool Acquire();
+  virtual void Reserve();
+  virtual void Release();
+  virtual void Clear();
+  virtual bool Setup(bool ignore, bool verbose, double& max_load_average);
+
+  // platform specific implementation
+  virtual const char *GetEnv(const char *name) = 0;
```
jhasse: Please align the `*` with the type (or just run clang-format on the file).

src/tokenpool-gnu-make.cc
```diff
@@ -0,0 +1,107 @@
...
+}
+
+GNUmakeTokenPool::~GNUmakeTokenPool() {
+}
+
+bool GNUmakeTokenPool::Setup(bool ignore,
+                             bool verbose,
+                             double& max_load_average) {
+  const char *value = GetEnv("MAKEFLAGS");
+  if (value) {
```
jhasse: Use `if (!value) { return false }` to reduce the indentation of the rest of the function.

src/tokenpool-gnu-make-win32.cc
```diff
@@ -0,0 +1,237 @@
...
+    while ((c = *++end) != '\0')
+      if (!(isalnum(c) || (c == '_')))
+        break;
+    len = end - start; // includes string terminator in count
+
+    if ((len > 1) && ((auth = (char *)malloc(len)) != NULL)) {
+      strncpy(auth, start + 1, len - 1);
+      auth[len - 1] = '\0';
+
+      if ((semaphore_jobserver_ = OpenSemaphore(SEMAPHORE_ALL_ACCESS, /* Semaphore access setting */
```
jhasse: Try to fit lines into 80 columns.

src/tokenpool-gnu-make-posix.cc
```diff
@@ -0,0 +1,203 @@
...
+    if (ret > 0)
+      return true;
+    if ((ret != -1) || (errno != EINTR))
+      return false;
+    // write got interrupted - retry
+  }
+}
+
+int GNUmakeTokenPoolPosix::GetMonitorFd() {
+  return(rfd_);
```
jhasse: no need for `()`.

src/subprocess_test.cc
```diff
@@ -151,9 +184,11 @@ TEST_F(SubprocessTest, Console) {
         subprocs_.Add("test -t 0 -a -t 1 -a -t 2", /*use_console=*/true);
     ASSERT_NE((Subprocess*)0, subproc);
 
+    subprocs_.ResetTokenAvailable();
     while (!subproc->Done()) {
-      subprocs_.DoWork();
+      subprocs_.DoWork(NULL);
     }
+    ASSERT_EQ(false, subprocs_.IsTokenAvailable());
```
jhasse: Use `ASSERT_FALSE` instead.

src/tokenpool-gnu-make-posix.cc
```diff
@@ -0,0 +1,203 @@
...
+
+int GNUmakeTokenPoolPosix::dup_rfd_ = -1;
+
+void GNUmakeTokenPoolPosix::CloseDupRfd(int signum) {
+  close(dup_rfd_);
+  dup_rfd_ = -1;
+}
+
+bool GNUmakeTokenPoolPosix::SetAlarmHandler() {
+  struct sigaction act;
```
jhasse: Remove `struct`

src/tokenpool-gnu-make-posix.cc
```diff
@@ -0,0 +1,203 @@
...
+  memset(&act, 0, sizeof(act));
+  act.sa_handler = CloseDupRfd;
+  if (sigaction(SIGALRM, &act, &old_act_) < 0) {
+    perror("sigaction:");
+    return(false);
```
jhasse: No need for `()`.

src/tokenpool-gnu-make-posix.cc
```diff
@@ -0,0 +1,203 @@
...
+  } else {
```
jhasse: `else` not needed.

src/tokenpool-gnu-make-win32.cc
```diff
@@ -0,0 +1,237 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "tokenpool-gnu-make.h"
+
+// always include first to make sure other headers do the correct thing...
```
jhasse: What is "the correct thing"?

src/tokenpool-gnu-make-win32.cc
```diff
@@ -0,0 +1,237 @@
...
+
+    // wait for child thread to exit
+    WaitForObject(child_);
+    CloseHandle(child_);
+    child_ = NULL;
+  }
+
+  if (semaphore_restart_) {
+    CloseHandle(semaphore_restart_);
+    semaphore_restart_ = NULL;
```
jhasse: Why do you set it to `NULL` here? We're about to be destructed anyway.

src/tokenpool-gnu-make-win32.cc
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  if (semaphore_enter_wait_) {
+    CloseHandle(semaphore_enter_wait_);
+    semaphore_enter_wait_ = NULL;
+  }
+}
+
+const char *GNUmakeTokenPoolWin32::GetEnv(const char *name) {
+  // getenv() does not work correctly together with tokenpool_tests.cc
```
jhasse: Why not?

src/tokenpool-gnu-make-win32.cc
```diff
@@ -0,0 +1,237 @@
...
+  static char buffer[MAX_PATH + 1];
+  if (GetEnvironmentVariable("MAKEFLAGS", buffer, sizeof(buffer)) == 0)
+    return NULL;
+  return(buffer);
```
jhasse: Drop `()`

src/tokenpool-gnu-make.h
```diff
@@ -0,0 +1,40 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "tokenpool.h"
+
+// interface to GNU make token pool
+struct GNUmakeTokenPool : public TokenPool {
+  GNUmakeTokenPool();
+  virtual ~GNUmakeTokenPool();
```
jhasse: No need, as `~TokenPool` is already virtual.

src/tokenpool-gnu-make-win32.cc
stefanb2: I tried to use getenv(), but it was not working or unusable. At least with non-commercial Visual C version I was using in the Windows 10 VM to develop & test the win32 implementation

stefanb2:
@jhasse 

1. the idea is that TokenPool API can be used for other implementations than GNUmake jobserver protocol.
2. I will make a change to configure.py

jhasse:
> I will make a change to configure.py

Not sure if that is really what we want. More options means more stuff we need to test.

src/tokenpool-gnu-make-posix.cc
stefanb2: That `struct` is mandatory as this is a C data type.

src/tokenpool-gnu-make-win32.cc
stefanb2: Reworded the comment

src/tokenpool-gnu-make-win32.cc
stefanb2: Old habit: better safe than sorry.

src/build_test.cc
```diff
@@ -2308,3 +2314,361 @@ TEST_F(BuildTest, Console) {
...
+    int value = va_arg(ap, int);
+    booleans.push_back(!!value); // force bool
+  }
+}
+
+TEST_F(BuildTokenTest, CompleteNoWork) {
+  // plan should not execute anything
+  string err;
+
+  EXPECT_TRUE(builder_.Build(&err));
```
bradking: With assertions enabled (`python ./configure.py --debug`) I get:

```
BuildTokenTest.CompleteNoWork
ninja_test: src/build.cc:674: bool Builder::Build(std::__cxx11::string*): Assertion `!AlreadyUpToDate()' failed.
```

stefanb2: Thanks. It seems the test is pointless then, as we should not call Builder::Build() with an empty plan.

bonzini:
Any news on this pull request? As mentioned by many others in #1139, it's not always possible (probably a euphemism) to convert "superbuilds" to ninja.

jhasse:
> the idea is that TokenPool API can be used for other implementations than GNUmake jobserver protocol.

As long as we don't have one (and I'm not sure if that's something we want to include into Ninja at all), I would prefer to remove the interface to keep the code change as small as possible.

Same goes with the `--without-token-pool` flag. Is there a need for it? (Sorry, that I was the one who suggested that as a possibility in the first place)

bradking:
@stefanb2 this PR's work has been part of [Kitware's branch](https://github.com/Kitware/ninja/tree/kitware-staged-features) that I've been maintaining in combination with the `dyndep` work from #1521 so I'm experienced in resolving the conflicts between the branches.  With the `dyndep` work merged to `master`, the conflicts now show up here.  I've rebased this PR's branch on `master`, resolved the conflicts, and temporarily published it as a [`jobserver` branch in my fork](https://github.com/bradking/ninja/tree/jobserver).  That serves as a reference to help you update this PR to resolve the conflicts.

bonzini:
@bradking Maybe you could open a new pull request if @stefanb2 does not have time to follow up right now?

bradking:
@bonzini while I resolved the conflicts I'm not familiar enough with the changes to be in a position to respond to further feedback.  Anyone else interested in taking over the work is welcome to fetch my branch and start a new PR from there.  Or, a maintainer may be able to push to this PR's source branch.

stefanb2:
@bradking @bonzini this PR represents the squashed version of the full development history [stefanb2/ninja@topic-issue-1139-full](https://github.com/stefanb2/ninja/tree/topic-issue-1139-full). I want to retain that valuable information, hence rebasing takes a little bit more effort.

Currently this PR is low priority for me, so I don't rebase it onto every change on master. If the maintainers indicate that they are willing to merge this PR, I'll of course raise the priority.

nox:
@jhasse You mentioned this PR to me in the issue, but no one is taking care of landing this or even reviewing it, could we get a quick summary of what's happening?

jhasse:
I've reviewed it briefly and https://github.com/ninja-build/ninja/pull/1140#issuecomment-473898963 still stands.

StefanBruens:
There already *is* another TokenTool implementation, the one used by the tests. So IMHO the comment does not apply.

https://github.com/ninja-build/ninja/blob/f711000c26a3889640ae46f2f455f26e87055a9d/src/subprocess_test.cc#L36

avikivity:
☹

luser:
FYI if you're interested in looking at another jobserver implementation for reference, [the Rust `jobserver` crate](https://github.com/alexcrichton/jobserver-rs/) is used in both `cargo` and `rustc`. Alex originally wrote it in support of the Firefox build calling `cargo` from its GNU make build system, but even in a standard Rust build nowadays `rustc` supports parallel compilation internally so it will use cargo's jobserver to play nice across multiple compiles.

nox:
> There already _is_ another TokenTool implementation, the one used by the tests. So IMHO the comment does not apply.

It sounds a bit circular to say that we need a public `TokenPool` interface because it is used by a test of the public `TokenPool` interface.

jhasse:
> It sounds a bit circular to say that we need a public `TokenPool` interface because it is used by a test of the public `TokenPool` interface.

The test is for Subprocess and having that interface makes it easier to mock the token pool.

(`TokenPoolTest` should be called `TestTokenPool` to be in line with `TestDiskInterface` and to avoid confusion with `TokenPoolTest` in tokenpool_test.cc)

nox:
> The test is for Subprocess and having that interface makes it easier to mock the token pool.

Now I'm confused. :) In a previous comment you said you don't want a public `TokenPool` interface, right? So which changes do you want exactly for this interface?

jhasse:
I've missed the mock @StefanBruens pointed out when I wrote that comment. Now I'm not sure if it's a good idea or not. It could be removed by overriding `GNUmakeTokenPool` directly for the mock.

stefanb2:
> (`TokenPoolTest` should be called `TestTokenPool` though to be in line with `TestDiskInterface` and to avoid confusion with `TokenPoolTest` in tokenpool_test.cc)

Fixed, thanks. I also removed the unused no-op TokenPool implementation.

avikivity:
It would be wonderful if this can be moved forward.

neheb:
https://github.com/neheb/ninja/tree/mangix-master rebased version of this PR.

vilkov:
> Is there any way to help moving forward with this?

Up. The same wish goes for me )

peterwaller-arm:
Ping, this is a great feature. Also a feature request: It would be very handy to have a way to tell if the ninja server supports being a job-client or not. At the moment I'm making scripts which assume that ninja does support it, which will lead to a bad time™️ if your ninja build doesn't have this feature. So having a way to detect and bail out or fall back to other less good methods (such as significantly under-committing) would be a bonus.

stefanb2:
> So having a way to detect and bail out or fall back to other less good methods (such as significantly under-committing) would be a bonus.

After this feature I have queued code to support the master mode. If that would be merged too you could probably use the existence of the `--tokenpool-master` (or short `-m`) as an indicator in scripts.

peterwaller-arm:
Ping @nico @jhasse -- I used this PR for the elfshaker manyclangs builds in order to do many parallel builds of clang on one machine spanning >2 years worth of clang commits without having overcommit problems, and making full utilization of all the cores available. This makes it possible to build ~2k commits in a few hours on one machine for example. I did not spot any issues, and this did constitute a pretty significant long running stress test of the machinery. It would be great to see this feature merged. Several people have asked if there is anything they can do to help, I guess this is bottlenecked on review, correct?

@stefanb2 it looks like this again needs a rebase, but obviously it would be great to have buy-in or direction from someone who can merge the feature first!

Thanks to anyone who invested time and support in this.

neheb:
I'll note that Kitware has a fork of Ninja with only this PR applied: https://github.com/Kitware/ninja

I noticed this because ninja-cmake is installable with scoop on Windows.

haampie:
Any chance to get this merged?

ePirat:
@jhasse Is there a fundamental problem with this, causing it to not be merged? Or just lack of time to review from maintainers? It is a relatively important feature in complex build scenarios, unfortunately seems to be stuck in review for a very long time already for some reason without any input from a maintainer.

src/tokenpool-gnu-make-posix.cc
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  // read() would block, i.e. no token available,
+  // cases 2-6 from above list or
+  // select() / poll() / dup() / sigaction() / setitimer() failed
+  return false;
+}
+
+bool GNUmakeTokenPoolPosix::ReturnToken() {
+  const char buf = '+';
```
jeremybettis, line 183: The jobserver client is supposed to write the same byte that it read. See https://www.gnu.org/software/make/manual/html_node/POSIX-Jobserver.html

PolynomialDivision:
This branch does not apply anymore. Can you rebase?

eli-schwartz:
> @bradking @bonzini this PR represents the squashed version of the full development history [stefanb2/ninja@topic-issue-1139-full](https://github.com/stefanb2/ninja/tree/topic-issue-1139-full). I want to retain that valuable information, hence rebasing takes a little bit more effort.
> 
> Currently this PR is low priority for me, so I don't rebase it onto every change on master. If the maintainers indicate that they are willing to merge this PR, I'll of course raise the priority.

This PR has now been sitting open and getting periodically rebased for 6.5 years now, so that strikes me as a pretty reasonable approach by the PR author.

I think asking for this PR to be rebased is probably a waste of time. If you just want to give the PR a spin to see how it works, simply build a slightly older version of ninja. If you want to review it, most review comments should apply either way. If you happened to be a ninja maintainer and want to click the merge button then that would be a different story.... :)

neheb:
Kitware had a fork of ninja with this patchset applied. So track that.

jameshilliard:
> Kitware had a fork of ninja with this patchset applied. So track that.

It's missing [v1.11.1](https://github.com/ninja-build/ninja/releases/tag/v1.11.1) at the moment.

> This branch does not apply anymore. Can you rebase?

I went ahead and [rebased the kitware fork on v1.11.1](https://github.com/Kitware/ninja/pull/1), haven't tested extensively yet but built in tests appear to pass.

bradking:
See the [v1.11.1.g95dee.kitware.jobserver-1](https://github.com/Kitware/ninja/releases/tag/v1.11.1.g95dee.kitware.jobserver-1) release.

src/tokenpool-gnu-make-posix.cc
stefanb2: Thanks

htot:
@stefanb2 thanks for creating this. I created 2 patches on top of Yocto Kirkstone (https://github.com/htot/poky/tree/kirkstone) to add shared make and ninja jobserver support working. As you may know by default Yocto spawns 4 tasks that each could be compiling n threads. So in particular many core build machines may suffer unless stuffed with sufficient memory. However increasing the number of parallel makes /ninjas higher then the number of cores brings no benefits. The only viable alternative to the jobserver I can think of would be a process supervisor that stops compile jobs when disk thrashing starts and continues them when memory becomes available. Still that is way less efficient then the jobserver approach.

The one patch creates a jobserver named pipe and makes sure make is run as a submake. The other if to pull ninja from your repo and run as a subninja. Seems to be building my image well with 16 parallel threads and 16GB RAM running various combinations of of make and ninja (some called from cmake). I have seen no issues like the mentioned races in this discussion.

If you would ask me: congrats and time to merge this PR.

ivyl:
Thanks for working on this. I hope it will eventually get merged :-)

In https://github.com/ValveSoftware/Proton we build multiple sub-projects that use autotools, meson and cmake among others. Having ninja being able to talk to a job server and effectively parallelize saves us a significant amount of time each build. We no longer have to chose between stalling due to overzealous build ordering or DOSing the CPU and RAM with all the processes spawned by all parallel ninja invocations in addition to make.

neheb:
@ivyl if you're impatient, install ninja from pip. It includes these patches.

src/tokenpool.h
```diff
@@ -0,0 +1,26 @@
...
+// limitations under the License.
+
+// interface to token pool
+struct TokenPool {
+  virtual ~TokenPool() {}
+
+  virtual bool Acquire() = 0;
+  virtual void Reserve() = 0;
+  virtual void Release() = 0;
+  virtual void Clear() = 0;
```
digit-google, line 26: Please document this interface to explain when these methods are supposed to be called (and under which preconditions).

src/tokenpool.h
```diff
@@ -0,0 +1,26 @@
...
+
+  // returns NULL if token pool is not available
+  static struct TokenPool *Get(void);
```
digit-google: nit: There is no need for `struct` here, please remote.

src/tokenpool-none.cc
```diff
@@ -0,0 +1,27 @@
...
+
+#include <fcntl.h>
+#include <poll.h>
+#include <unistd.h>
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+
+// No-op TokenPool implementation
+struct TokenPool *TokenPool::Get(void) {
```
digit-google: nit: please remote the unnecessary `struct` here (see comment below).

src/tokenpool-none.cc
```diff
@@ -0,0 +1,27 @@
...
+// limitations under the License.
+
+#include "tokenpool.h"
+
+#include <fcntl.h>
+#include <poll.h>
+#include <unistd.h>
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
```
digit-google: nit: Please remove un-needed header includes. Introduce them with the code that actually requires them instead.

src/tokenpool-gnu-make.cc
```diff
@@ -0,0 +1,211 @@
...
+void GNUmakeTokenPool::CloseDupRfd(int signum) {
+  close(dup_rfd_);
+  dup_rfd_ = -1;
+}
+
+bool GNUmakeTokenPool::SetAlarmHandler() {
+  struct sigaction act;
+  memset(&act, 0, sizeof(act));
+  act.sa_handler = CloseDupRfd;
+  if (sigaction(SIGALRM, &act, &old_act_) < 0) {
```
digit-google: Please move all signal-handling code to subprocess-posix.cc instead, since it is _considerably_ easier to understand what's going on when all signal-related code is in the same source file, and exposed through a sane API for the rest of the code base to use.

src/tokenpool-gnu-make.cc
```diff
@@ -0,0 +1,211 @@
...
+#ifdef _WIN32
+  // @TODO
+#else
+  int rfd_;
+  int wfd_;
+
+  struct sigaction old_act_;
+  bool restore_;
+
+  static int dup_rfd_;
```
digit-google: nit: Please clarify that this is a static variable, for example by using a prefix like `s_dup_rfd_`.

src/tokenpool-gnu-make.cc
```diff
@@ -0,0 +1,211 @@
...
+
+GNUmakeTokenPool::~GNUmakeTokenPool() {
+  Clear();
+  if (restore_)
+    sigaction(SIGALRM, &old_act_, NULL);
+}
+
+bool GNUmakeTokenPool::CheckFd(int fd) {
+  if (fd < 0)
+    return false;
```
digit-google, line 36: nit: the `fd < 0` check is redundant with the `fcntl()` one and can be removed.

src/tokenpool-gnu-make.cc
```diff
@@ -0,0 +1,211 @@
...
+  int ret = fcntl(fd, F_GETFD);
+  if (ret < 0)
+    return false;
+  return true;
```
digit-google: nit: simplify with `return fcntl(fd, F_GETFD) != -1;`

configure.py
```diff
@@ -543,6 +543,7 @@ def has_re2c():
     objs += cxx(name, variables=cxxvariables)
 if platform.is_windows():
     for name in ['subprocess-win32',
+                 'tokenpool-none',
```
digit-google: Please keep configure.py consistent with CMakeLists.txt in all patches.

src/tokenpool-gnu-make.cc
```diff
@@ -0,0 +1,211 @@
...
+int GNUmakeTokenPool::dup_rfd_ = -1;
+
+void GNUmakeTokenPool::CloseDupRfd(int signum) {
+  close(dup_rfd_);
+  dup_rfd_ = -1;
+}
+
+bool GNUmakeTokenPool::SetAlarmHandler() {
+  struct sigaction act;
+  memset(&act, 0, sizeof(act));
```
digit-google: nit: simplify with `struct sigaction act = {};`

src/tokenpool-gnu-make.cc
```diff
@@ -0,0 +1,211 @@
...
+      act.sa_handler = CloseDupRfd;
+      if (sigaction(SIGCHLD, &act, &old_act) == 0) {
+        char buf;
+
+        // block until token read, child exits or timeout
+        alarm(1);
+        ret = read(dup_rfd_, &buf, 1);
+        alarm(0);
+
+        sigaction(SIGCHLD, &old_act, NULL);
```
digit-google: As noted in a previous comment, this logic interferes with the one in subprocess-posix.cc in really brittle ways. It should be moved there instead to keep the code manageable and testable.

carterols:
Super hopeful this can get merged. This has been a long time coming!

jhasse:
see #2450

jhasse:
closing in favor of #2506

PR #1141, opened 2016-04-27
Reject duplicate outputs involving an implicit output

### LINKED ISSUE #1136: Crash with multiple duplicate implicit outputs

bradking:
For compatibility with older versions of Ninja we do not reject
duplicate outputs by default.  However, we can reject them when at least
one of the outputs is implicit because older versions of Ninja do not
support implicit output syntax anyway.

Fixes #1136.

### DIFF
--- a/src/manifest_parser.cc
+++ b/src/manifest_parser.cc
@@ -339,15 +339,19 @@ bool ManifestParser::ParseEdge(string* err) {
   }
 
   edge->outputs_.reserve(outs.size());
-  for (vector<EvalString>::iterator i = outs.begin(); i != outs.end(); ++i) {
-    string path = i->Evaluate(env);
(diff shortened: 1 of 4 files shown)


bradking:
Closing in favor of #1137 to keep the behavior between the output types consistent.

PR #1144, opened 2016-04-28
Update %p to use finished edges after #1143.

nico:


colincross:
This needs a documentation update too

nico:
thanks filed #1145

PR #1147, opened 2016-04-28
Fix NINJA_STATUS %e and %r on dumb terminals

### LINKED ISSUE #1146: Regression from #999: NINJA_STATUS %e on dumb terminals starts from epoch
#999 made dumb terminals only output when edges finish.  BuildStatus::overall_rate_ stopwatch is only initialized to the current time when PrintStatus is called with finished_edges_ == 0, but on a dumb terminal it will be called for the first time when finished_edge_ = 1, which results in very long elapsed times:
NINJA_STATUS="[%r processes, %f/%t @ %o/s : %es ] "
[0 processes, 2/2 @ 0.0/s : 1461869902.367s ]

### LINKED ISSUE #1148: Regression from #999: NINJA_STATUS %r on dumb terminals reports 1 fewer process
PR #999 made dumb terminals only output when edges finish.  PrintStatus is called after finished_edges_ is incremented, which means the calculation for running edges will always return 1 less than the real number of running processes.  This happens on smart terminals too, but ninja will immediately print the status for the next edge with starting_edges_ incremented, so the incorrect value is never visible.

colincross:
PR #999 on dumb terminals broke NINJA_STATUS messages with %e and slightly changed the behavior of %r.

Fixes #1146
Fixes #1148

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -97,7 +97,7 @@ void BuildStatus::BuildEdgeStarted(Edge* edge) {
   ++started_edges_;
 
   if (edge->use_console() || printer_.is_smart_terminal())
-    PrintStatus(edge);
+    PrintStatus(edge, kEdgeStarted);
 
   if (edge->use_console())
     printer_.SetConsoleLocked(true);
@@ -129,7 +129,7 @@ void BuildStatus::BuildEdgeFinished(Edge* edge,
     return;
 
   if (!edge->use_console())
-    PrintStatus(edge);
+    PrintStatus(edge, kEdgeFinished);
 
   // Print the command that is spewing before printing its output.
   if (!success) {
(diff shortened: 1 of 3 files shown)


src/build.cc
```diff
@@ -97,7 +97,7 @@ void BuildStatus::BuildEdgeStarted(Edge* edge) {
   ++started_edges_;
 
   if (edge->use_console() || printer_.is_smart_terminal())
-    PrintStatus(edge);
+    PrintStatus(edge, false);
```
nico: I always find bare bool parameters like this mysterious. What does `false` mean here? Either write `PrintStatus(edge, /*edge_finished=*/false);` here, or add `enum EdgeStatus { kEdgeStarted, kEdgeFinished };` above the PrintStatus() prototype and write `PrintStatus(edge, kEdgeStarted);` here.

nico:
Thanks for the patch! Just a cosmetic comment.

colincross:
Fixed

nico:
Thanks!

A test would've been nice, but I don't remember if this code is all stubbed out for testing.

colincross:
Yeah, I think there should be tests for the expected behaviors of %r, %s, %f, and %p when starting and finishing builds in both dumb and smart terminal modes, but that seemed like a much bigger project.

PR #1150, opened 2016-05-03
Make misc/measure.py compatible with python3

KiYugadgeter:
Make misc/measure.py compatible with python3
fixed point: use of `print function` in source

### DIFF
--- a/misc/measure.py
+++ b/misc/measure.py
@@ -17,6 +17,8 @@
 """measure the runtime of a command by repeatedly running it.
 """
 
+from __future__ import print_function
+
 import time
 import subprocess
 import sys
@@ -24,7 +26,7 @@ import sys
 devnull = open('/dev/null', 'w')
 
 def run(cmd, repeat=10):
-    print 'sampling:',
+    print('sampling:', end=' ')
     sys.stdout.flush()
 
     samples = []
@@ -33,10 +35,10 @@ def run(cmd, repeat=10):
         subprocess.call(cmd, stdout=devnull, stderr=devnull)
         end = time.time()
         dt = (end - start) * 1000
-        print '%dms' % int(dt),
+        print('%dms' % int(dt), end=' ')
         sys.stdout.flush()
         samples.append(dt)
-    print
+    print()
 
     # We're interested in the 'pure' runtime of the code, which is
     # conceptually the smallest time we'd see if we ran it enough times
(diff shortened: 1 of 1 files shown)


misc/measure.py
```diff
@@ -17,26 +17,26 @@
...
 import time
 import subprocess
 import sys
 
 devnull = open('/dev/null', 'w')
 
 def run(cmd, repeat=10):
-    print 'sampling:',
-    sys.stdout.flush()
+    print('sampling:', sys.stdout.flush())
```
NiklasRosenstein: This is not an equivalent tranformation. 
```python
print('sampling:', end='')
sys.stdout.flush()
```
would be correct.

KiYugadgeter:
Thank you for reviewing.
fixed to equivalent tranformation.

NiklasRosenstein:
I did a mistake there myself, you need to pass a single whitespace as string instead of an empty string for the `end` parameter, like

    print(text, end=' ')

misc/measure.py
```diff
@@ -26,15 +26,17 @@
...
+    sys.stdout.flush()
 
     samples = []
     for _ in range(repeat):
         start = time.time()
         subprocess.call(cmd, stdout=devnull, stderr=devnull)
         end = time.time()
         dt = (end - start) * 1000
-        print('%dms' % int(dt), sys.stdout.flush())
+        print('%dms' % int(dt))
```
NiklasRosenstein: Don't want to be nit-picking, and I haven't used `measure.py` myself so I don't know how important it is. But, I think you're missing the `end=' '` argument in this line still. :) The old code looked like this:

```python
        print '%dms' % int(dt),
        sys.stdout.flush()
```

PS: And as I mentioned in a comment for this merge request, I did a mistake myself there and used `end=''` instead of the correct `end=' '` (note that string with a single whitespace character).

KiYugadgeter:
fix a argument of print function to space

nico:
Thanks! If it's not too much trouble, could you squash your commits? (I thought github now had a "merge and squash" button on pull requests, but even though that's enabled in ninja's settings it doesn't show up for me.)

KiYugadgeter:
I squashed commits

nico:
Thanks for the patch!

(Looks like github's "commit and squash" UI appears after clicking "merge pull request", so next time I'll know how to do this without having to ask you to squash.)

PR #1154, opened 2016-05-12
Extract Replace() function into util.cc, add tests, fix edge case

cdbennett:
Extract string search-and-replace function `Replace()` from anonymous
namespace in `msvc_helper-win32.cc` into `util.cc` in the global
namespace, so it can be used by other modules.

Add unit tests to validate the function.

Fix infinite loop if `Replace()` called with empty string as `find`
argument.

### DIFF
--- a/src/msvc_helper-win32.cc
+++ b/src/msvc_helper-win32.cc
@@ -18,20 +18,6 @@
 
 #include "util.h"
 
-namespace {
-
-string Replace(const string& input, const string& find, const string& replace) {
-  string result = input;
-  size_t start_pos = 0;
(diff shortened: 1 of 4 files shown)


jhasse:
In which other module do you need `Replace`?

cdbennett:
I don't need to use it in another module, but it was a very general-purpose function so it seemed logical to put it in the `util.cc` module.  It could just as well be left in the `msvc_helper-win32.cc` file if desired.

jhasse:
It's only used `EscapeForDepfile` as far as I can tell. The endless loop isn't triggered due to a hardcoded find-string.

Let's keep it there. Sorry for closing this: We are very conservative with changes and want to reduce the number of open PRs.

Issue #773, opened 2014-06-04
Ninja adds extra CR to subprocess output on Windows yielding CR CR LF


cdbennett:
Ninja 1.4.0 on Windows is adding an extra CR character to any CRLF line endings in output produced by a subprocess, so that lines are terminated with CR CR LF.

This is causing Eclipse (CodeWarrior 10.5, to be precise) to have trouble parsing the compiler warning and error messages.

This test case has two .c programs designed to be compiled with MinGW gcc which print "Hello, World" with a newline. But one program changes the mode of stdout to binary so that the MinGW stdio library does not convert the \n to a \r\n.

[Test case: ninja_extra_cr_test](https://github.com/cdbennett/ninja_extra_cr_test)
A self-contained test case for Ninja which uses MinGW gcc to build two programs, one which produces LF and one CRLF line endings output, and shows how Ninja is adding an extra CR to the CRLF output:

Optiligence:
This is also an issue here: https://bugreports.qt.io/browse/QTCREATORBUG-14177.

cdbennett:
@nico I pushed two branches with pull requests that both solve the problem.  One by stripping extra CR characters, and the other based on your above patch that writes to `stdout` in binary mode on Windows.

I can see some possible advantages to each option.

nico:
Thanks! I kind of like the "set output to binary" version best, I think. What do you think?

martine:
I also like "set output to binary".  Ninja shouldn't even need to care about the newline format of the text output by its subprocesses -- it just wants to buffer and pass the bytes through.  So the fewer conversions/text processing the better.

nico:
#1156 should've fixed this.

PR #1155, opened 2016-05-12
Strip subprocess extra CR character before output

cdbennett:
Fix for #773 option 2: normalize the line endings of text before writing it to `stdout`.

(See also option 1 in Pull Request #1156, which writes to `stdout` in binary mode instead.)

All `CR LF` sequences get replaced with `LF`, since the `fwrite()` call will subsequently write out a `CR LF` for the `LF` itself.


cdbennett:
We only need either #1155 or #1156 but I would like to share both, and we can discuss the merits of each.  The one that is not accepted can be closed ultimately.

nico:
I merged the other one.

PR #1156, opened 2016-05-12
Write subprocess output to stdout in binary mode

cdbennett:
Fix for #773 option 1: write subprocess output to `stdout` in binary mode.

(See also option 2 in Pull Request #1155, which normalizes the newlines before writing out.)

Set stdout to binary mode while writing subprocess output, so that the
CR in a CR LF sequence is not replaced with CR LF itself, which would
result in CR CR LF.

Based on patch posted by nico in issue #773 comment.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -20,6 +20,11 @@
 #include <stdlib.h>
 #include <functional>
 
+#ifdef _WIN32
+#include <fcntl.h>
+#include <io.h>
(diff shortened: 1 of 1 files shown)


nocnokneo:
Looking forward to seeing this PR merged! Are there any blocking issues?

nico:
Alright, let's see how it goes.

PR #1157, opened 2016-05-20
Escape ninja output inserted into HTML

colincross:
Ninja query or error output may contain characters that need to be
escaped when being inserted into HTML.  Replace &, ", <, and > with
their & equivalent.

### DIFF
--- a/src/browse.py
+++ b/src/browse.py
@@ -27,6 +27,7 @@ try:
 except ImportError:
     import BaseHTTPServer as httpserver
 import argparse
+import cgi
 import os
 import socket
 import subprocess
@@ -57,6 +58,9 @@ def match_strip(line, prefix):
     if not line.startswith(prefix):
         return (False, line)
     return (True, line[len(prefix):])
+
+def html_escape(text):
+    return cgi.escape(text, quote=True)
 
 def parse(text):
     lines = iter(text.split('\n'))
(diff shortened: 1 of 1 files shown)


src/browse.py
```diff
@@ -58,6 +58,17 @@ def match_strip(line, prefix):
...
+html_escape_table = {
+    "&": "&amp;",
+    '"': "&quot;",
+    "'": "&apos;",
+    ">": "&gt;",
+    "<": "&lt;",
+    }
+
+def html_escape(text):
+    return "".join(html_escape_table.get(c,c) for c in text)
```
martine: https://docs.python.org/2/library/cgi.html#cgi.escape perhaps?

colincross: cgi.escape doesn't escape a single quote.  Leaving single quotes in would be fine, unless someone introduced a single quoted HTML attribute in generate_html, in which case the links could break for any input or output file that contained a single quote.

nico:
Haha :-) Not sure what the threat model is, but this gives you an alert on click (which this patch would fix):

```
$ cat build.ninja 
rule test
  command = echo $out = $var
build "$ onclick=javascript$:alert('hi'): test
build test1: test "$ onclick="javascript$:alert('hi')
```

https://docs.python.org/2/library/cgi.html says "If the value to be quoted might include single- or double-quote characters, or both, consider using the quoteattr() function in the xml.sax.saxutils module instead." – that'd mean using different escapers for attributes and tag contents, which makes things more complicated, and depending on xml.sax seems a bit unfortunate anyways.

Since nothing currently uses single quotes, I think using cgi.escape() should be fine (and in a few decades, we can assume py 3.2 and use html.escape which does escape single quotes).

colincross:
Replaced with cgi.escape

nico:
Thanks!

PR #1159, opened 2016-05-25
Parser accepts no explicit outputs.

nicolasdespres:
There is a class of commands that take an output directory where
they create their output files. Among them are cp(1), tar(1) to name a
few. These commands have one or more implicit outputs but no explicit
output.

With this patch, Ninja's parser accepts build edge with an
empty list of explicit outputs.

Discussed here: https://groups.google.com/forum/#!topic/ninja-build/Nt8QlmA5cQA


bradking:
The implementation and tests LGTM.

nico:
Thanks!

PR #1163, opened 2016-06-08
Add single quotes to the output file path if it has a space.

stgaochromium:
In the "FAILED: output nodes", add single quotes to the output file path if it has a space. This is to make it easier to run ninja with the output nodes and also the failure output parsable by script.

`ninja -C out/gn_build_dir/ output/node/with/a/ space` won't work, but `ninja -C out/gn_build_dir/ 'output/node/with/a/ space'` works.

Before this change:
`FAILED: gen/path/to/a space.cc gen/path/to/no_space.h `
After this change:
`FAILED: 'gen/path/to/a space.cc' gen/path/to/no_space.h `

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -135,8 +135,13 @@ void BuildStatus::BuildEdgeFinished(Edge* edge,
   if (!success) {
     string outputs;
     for (vector<Node*>::const_iterator o = edge->outputs_.begin();
-         o != edge->outputs_.end(); ++o)
-      outputs += (*o)->path() + " ";
+         o != edge->outputs_.end(); ++o) {
+      if((*o)->path().find(' ') != string::npos) {
+        outputs += "'" + (*o)->path() + "' ";
+      } else {
+        outputs += (*o)->path() + " ";
+      }
+    }
 
     printer_.PrintOnNewLine("FAILED: " + outputs + "\n");
     printer_.PrintOnNewLine(edge->EvaluateCommand() + "\n");


src/build.cc
```diff
@@ -135,8 +135,13 @@ void BuildStatus::BuildEdgeFinished(Edge* edge,
   if (!success) {
     string outputs;
     for (vector<Node*>::const_iterator o = edge->outputs_.begin();
-         o != edge->outputs_.end(); ++o)
-      outputs += (*o)->path() + " ";
+         o != edge->outputs_.end(); ++o) {
+      if((*o)->path().find(' ') != string::npos) {
+        outputs += "'" + (*o)->path() + "' ";
```
sgraham: ' isn't Windows-friendly.

(Is it worth worrying about other things like \s being in there if you're planning on parsing in a script?)

nico: I think Scott's ' note means 'use "'. What if the outname contains quote chars?

stgaochromium: Given that we print all the outnames in one line, I don't think we could prevent a selected separator from being part of the outnames, unless a more complex escaping mechanism is introduced. And I guess this is not preferred.

Currently the selected separator is an empty space " ", but it conflicts with outnames containing an empty space. In this patch, a quote is introduced to resolve such conflict.

We have some options here, but I prefer option 1 or 2 below:
1. Go with a double quote (instead of a single quote in this patch). And ignore the support for outnames containing double quote chars, which seems rare. Note: a double quote can not be part of a directory name or file name on Win7 (not tested on other versions of Windows).

2. Make ninja print such information to a given file and in a parsable format like json.

3. Introduce a more complex escaping mechanism when printing in one line to the stdout.

Nico, Scott, what do you think?

jhasse: Regarding 1: Using double quotes you'll also have to escape $.

I think we should add `'` (and escape all `'` in the path) on anything non-Windows.

jhasse:
please reopen after resolving conflict

PR #1166, opened 2016-06-14
Add 'stats' tool.

nicolasdespres:
Useful to answer questions like:
- How many files are created as part of the build?
- How many jobs can run concurrently?
- What is the length of the longest chain of commands?
- How many files are produced by my build?

This is useful when trying to improve the performance of a build system.

Maybe it could interest other people than me. The idea is to provide the basic statistics needed by more elaborated tools.

If it is considered for merging, I will write tests.


src/graph_stats.cc
```diff
@@ -0,0 +1,160 @@
...
+    return height;
+  }
+
+  size_t GetMaxHeight(const vector<Node*>& nodes) {
+    size_t height = numeric_limits<size_t>::min();
+    for (vector<Node*>::const_iterator i = nodes.begin();
+         i != nodes.end(); ++i) {
+      height = max(height, (*this)(*i));
+    }
+    return height;
```
wilhelmtell: Isn't this loop equivalent to something like

    return std::max_element(nodes.begin(), nodes.end(), LesserHeightOf);

src/graph_stats.cc
```diff
@@ -0,0 +1,160 @@
...
+    }
+  }
+
+  queue<Node*> _q;
+  set<Node*> _s;
+}; // class GetGraphWidth
+
+} // anonymous namespace
+
+void GraphStats::Reset() {
```
wilhelmtell: If the default constructor and this function do the same thing, wouldn't be better to not have this function at all and just create a new object when you want the values reset?

jhasse:
Thanks for the PR! It seems nice to have, but doesn't necessarily need to be part of upstream Ninja. Since there hasn't been any comments I'm closing this one, feel free to reopen.

PR #1173, opened 2016-07-14
Explain as human readable datetime instead of epoch

doheonlee:
Ninja explains output is older than most recent input with their epoch.
Datetime is changed from epoch value to datetime human reads easily.

### DIFF
--- a/src/state.cc
+++ b/src/state.cc
@@ -159,11 +159,12 @@ bool State::AddDefault(StringPiece path, string* err) {
   return true;
 }
 
-vector<Node*> State::RootNodes(string* err) {
+vector<Node*> State::RootNodes(string* err) const {
   vector<Node*> root_nodes;
   // Search for nodes with no output.
-  for (vector<Edge*>::iterator e = edges_.begin(); e != edges_.end(); ++e) {
-    for (vector<Node*>::iterator out = (*e)->outputs_.begin();
+  for (vector<Edge*>::const_iterator e = edges_.begin();
+       e != edges_.end(); ++e) {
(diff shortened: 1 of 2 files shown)


doheonlee:
strftime can reduce code lines.

doheonlee:
this explanation lines are shown with explain option.

src/graph.cc
```diff
@@ -174,11 +174,22 @@ bool DependencyScan::RecomputeOutputDirty(Edge* edge,
...
     if (output_mtime < most_recent_input->mtime()) {
+      time_t output_t = static_cast<time_t>(output_mtime);
+      time_t most_recent_input_t = static_cast<time_t>(most_recent_input->mtime());
+
+      char *output_t_buff = ctime(&output_t);
+      char *most_recent_input_t_buff = ctime(&most_recent_input_t);
+
+      char output_t_str[100] = {0,};
+      char most_recent_input_t_str[100] = {0,};
+      strncpy(output_t_str, output_t_buff, strlen(output_t_buff) - 1);
```
wilhelmtell: Oops, this is fragile. First we hardcode the length of `output_t_str[]` to 100, then we fill it with `strlen()` characters. How sure are we that we're not overflowing here?

I don't think the C standard says anything about the length of the static the function writes into. I think you should take the minimum between 100 (or 99) and that.

doheonlee: Sorry for late comment. I did re-write the codes and make a commit.

I will request with the modificated version again.

Thanks

doheonlee:
Will be requested with another version.

PR #1174, opened 2016-07-14
Fix for not properly initialized elapsed time.

gkistanova:
There is a recently introduced bug when elapsed time does not get properly initialized in case of a dumb terminal. This is a fix for this issue.


nico:
Thanks!

nico:
(Note to self: https://github.com/ninja-build/ninja/pull/1147 fixed %e in dumb terminal mode but broke it in smart terminal mode. This here fixes it in smart terminal mode too.)

PR #1176, opened 2016-07-27
check setvbuf() return-value

wilhelmtell:
`setvbuf()` returns a non-zero integer value on error. check for that, and report any error.

### DIFF
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -127,7 +127,9 @@ bool BuildLog::OpenForWrite(const string& path, const BuildLogUser& user,
     *err = strerror(errno);
     return false;
   }
-  setvbuf(log_file_, NULL, _IOLBF, BUFSIZ);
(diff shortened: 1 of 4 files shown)


wilhelmtell:
I wasn't entirely sure how to handle the error cases here. What do you think of calling `Error()`, and of the error strings?

nico:
I think most places that check "this really shouldn't fail" function results call Fatal() instead of Error(). Are you seeing this function fail for you, or are you adding these checks mostly for hygienic reasons? If it's for hygiene, I'd make this call Fatal().

wilhelmtell:
> Are you seeing this function fail for you, or are you adding these checks mostly for hygienic reasons?

I'm not sure I know what you mean by that. I haven't experienced this triggering, no. I found a function call that can fail, and added a check for that.

wilhelmtell:
Alright, thanks @nico. I pushed an update, s/error/fatal/.

PR #1180, opened 2016-08-05
Ninja status format string from argv.

gkistanova:
Ninja supports a status format string from NINJA_STATUS env. This is a great feature to have.
Though, in an automated build environment provisioning environmental variables could be tricky.

This patch adds a support for a command line option (--status-format=<STATUS>) to set the status format string. If given, the status format string from the command line would override the one from NINJA_STATUS env.

Only a long option is supported for now. Easy to add a short form if it will be a use case for that.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -87,7 +87,12 @@ BuildStatus::BuildStatus(const BuildConfig& config)
   if (config_.verbosity != BuildConfig::NORMAL)
     printer_.set_smart_terminal(false);
 
-  progress_status_format_ = getenv("NINJA_STATUS");
+  if (config_.status_format) {
+    // Command line arg overrides env if given.
+    progress_status_format_ = config_.status_format;
+  } else {
+    progress_status_format_ = getenv("NINJA_STATUS");
+  }
   if (!progress_status_format_)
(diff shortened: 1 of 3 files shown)


nicolasdespres:
If you want to set NINJA_STATUS just for a single process invocation, you can do `env NINJA_STATUS='foo' ninja`. Note that depending on your shell, you may even omit the call to the `env` program.

gkistanova:
Right. Though, it is not always launched from a shell. Addressing that use case was the original motivation of the patch.

nico:
Thanks, that looks like a good feature.

(Note to self: Related to #746)

jhasse:
> Though, it is not always launched from a shell.

`env` doesn't require a shell though :)

jonesmz:
@gkistanova Could you rebase this PR on the latest master branch?

gkistanova:
@jonesmz The PR has been rebased on the latest master branch.

jhasse:
This is a good change, thanks!

But we should use the opportunity and also do #1045 for the --status-format flag. There we can have the description to be always part of the status as we don't have to think about backwards compatibility.

Issue #1161, opened 2016-06-04
hitting max path component limit


jhgit:
I hit this error on a non-windows (bsd) system:

ninja: fatal: path has too many components

My first question upon hitting that is - what is meant by "too many"?  Why did I hit this?

Opening the source, I see the code in CanonicalizePath in src/util.cc - so I have the answer to what is meant by "how many".  Turns out, I'm building llvm in a mount point.  The project already adds a few directory levels.  Then on top of that, my mount paths added more.  After all was said and done, I had a path with 32 components which exceeds the hard-coded kMaxPathComponents (30).

Okay, so now what?
- Accept the limitation and work in a shallower hierarchy.  Maybe it could be documented better, but that's fine for now.  I'm sure 99% of ninja uses won't hit this problem.
- But can this be improved in ninha?
  - better error mesage? - sure, easy enough
  - extend the hard-coded limit?  A little fragile due to the current algorithm (thanks a lot, Windows), but one could modify that (64?) and still retain decent performance.
  - gracefully transition to a less performant mode when long paths are encountered?  Maybe add a footnote in the docs about avoiding shooting oneself in the foot, performance-wise?

I'm willing to concoct a patch and do some benchmarking, but I'd like to know the project's feeling about this.  Where is the thinking on the spectrum from "hell, no - if you have a path that long, you're an idiot - we're not changing anything" to "yeah, we could document that a little better" to "sure, if the OS can support a longer path, why not modify ninja to allow that"?

I also haven't done the profiling/inspection necessary to understand how often paths need to be canonicalized (and whether it's necessary in all cases where it's used - i.e., why bother spending cycles canonicalizing at all?).

martine:
Thanks for researching the solutions!  You're the first person to ever run into this problem, so it's a bit of a surprise.  The intent of the error message was mostly to direct someone at the bug tracker, since we didn't expect to hit it -- I definitely should've made it clearer.

My first guess is that just upping the limit would be the easiest solution.  The only reason this is a fixed value at all is that we didn't want to spend extra time growing an array.  Unfortunately, from a glance at the code it looks like we use a 32-bit bitfield for some windows backslash trickery.  Nico can likely comment on whether making it a 64-bit bitfield would matter.

[...]

PR #1181, opened 2016-08-22
Allow more path components

DanielWeber:
- 60 instead of 30 path components
- 64 instead of 32 backslashes in a path (windows only)

Fixes Issue #1161

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -90,7 +90,7 @@ void Error(const char* msg, ...) {
   fprintf(stderr, "\n");
 }
 
-bool CanonicalizePath(string* path, unsigned int* slash_bits, string* err) {
+bool CanonicalizePath(string* path, uint64_t* slash_bits, string* err) {
   METRIC_RECORD("canonicalize str");
   size_t len = path->size();
   char* str = 0;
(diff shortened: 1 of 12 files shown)


DanielWeber:
Tests passed under windows (using mingw), travis covered the linux build for me :)

nico:
Thanks, this looks like a good change. I'll merge it soon.

TheOneRing:
The issue also occurs with the latest version of GammaRay on Windows, so it would be cool if the change could make it to the next release?

jomof:
We're also hitting this issue today in Android Studio, +1 for this PR

gunrot:
Hi nico, sorry to bother you but this is an important fix. I was happy to see Daniel make this PR, even more happy i was you liked the change and wanted to merge it soon. Albeit Happiness has decreased since ...

src/util.cc
```diff
@@ -125,7 +125,7 @@ bool CanonicalizePath(char* path, size_t* len, unsigned int* slash_bits,
     return false;
   }
 
-  const int kMaxPathComponents = 30;
+  const int kMaxPathComponents = 60;
```
sgraham: This was 32 - a bit, so maybe 62?

mbjorge:
Just ran into this issue also. Any ETA as to when the PR might get merged? Thanks!

nico:
Sorry about the delay. I checked that this doesn't hurt performance on a few platforms.

This will probably break the build with older MSVCs that don't yet have uint64_t. If someone runs into that, please send a pull request that adds a uint64_t typedef on old-enough MSVCs.

nico:
We just landed a different fix for this in #1314. Could people mentioning they were hitting this issue check if they still hit this issue with just #1314 but with this change here reverted?

nico:
@atetubou points out that some of you were on non-Windows systems, which answers my question ("yes, this is needed"). Sorry for the noise.

PR #1182, opened 2016-08-26
Use POSIX_SPAWN_USEVFORK if available

moosotc:
Existing comments were alluding to it's usage and it makes building
ninja itself go a litle bit faster (i.e. taking less wall clock time).

FWIW FreeBSD even uses vfork by default c.f.
https://www.freebsd.org/cgi/man.cgi?query=posix_spawn

### DIFF
--- a/src/subprocess-posix.cc
+++ b/src/subprocess-posix.cc
@@ -73,8 +73,6 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
   // default action in the new process image, so no explicit
   // POSIX_SPAWN_SETSIGDEF parameter is needed.
 
-  // TODO: Consider using POSIX_SPAWN_USEVFORK on Linux with glibc?
-
   if (!use_console_) {
     // Put the child in its own process group, so ctrl-c won't reach it.
     flags |= POSIX_SPAWN_SETPGROUP;
@@ -93,6 +91,9 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
     // In the console case, output_pipe is still inherited by the child and
     // closed when the subprocess finishes, which then notifies ninja.
   }
+#ifdef POSIX_SPAWN_USEVFORK
+  flags |= POSIX_SPAWN_USEVFORK;
+#endif
 
   if (posix_spawnattr_setflags(&attr, flags) != 0)
     Fatal("posix_spawnattr_setflags: %s", strerror(errno));


src/subprocess-posix.cc
```diff
@@ -93,6 +91,9 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
     // In the console case, output_pipe is still inherited by the child and
     // closed when the subprocess finishes, which then notifies ninja.
   }
+#ifdef POSIX_SPAWN_USEVFORK
```
nico: I think on linux with glibc one is to define `__USE_GNU` before including spawn.h to get this define. But then we either have to unconditionally define that, or check for glibc.

I wasn't able to measure a performance difference when I wrote the TODO, and in theory C libraries could be smart (if their are no atfork handlers, etc, then use vfork for posix_spawn automatically). On several systems, they are already smart.

So unless there's data that shows that this measurably helps somewhere, I'm a bit skeptical about this change.

moosotc:
Nico Weber <notifications@github.com> writes:

> nico commented on this pull request.
>
>
>
>> @@ -93,6 +91,9 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
>      // In the console case, output_pipe is still inherited by the child and
>      // closed when the subprocess finishes, which then notifies ninja.
>    }
> +#ifdef POSIX_SPAWN_USEVFORK
>
> I think on linux with glibc one is to define `__USE_GNU` before
> including spawn.h to get this define. But then we either have to
> unconditionally define that, or check for glibc.
>

_GNU_SOURCE more like (that will result in __USE_GNU being defined), but
anyhow, the guard inside spawn.h indeed checks for __USE_GNU. That said:

g++ -E -dD -g -Wall -Wextra -Wno-deprecated -Wno-missing-field-initializers -Wno-unused-parameter -fno-rtti -fno-exceptions -fvisibility=hidden -pipe '-DNINJA_PYTHON="python"' -O2 -DNDEBUG -fdiagnostics-color -DUSE_PPOLL -DNINJA_HAVE_BROWSE -I. -c src/subprocess-posix.cc | grep USEVF
#define POSIX_SPAWN_USEVFORK 0x40

USE_GNU is being set to 1 inside /usr/include/features.h because of
_GNU_SOURCE is defined, why it is - I've not checked

> I wasn't able to measure a performance difference when I wrote the
> TODO, and in theory C libraries could be smart (if their are no atfork
> handlers, etc, then use vfork for posix_spawn automatically). On
> several systems, they are already smart.
>

Perhaps. My only experience deadling with this particular issue is from
several years back and back then only FreeBSD's libcwas out of the box
smart.

> So unless there's data that shows that this measurably helps
> somewhere, I'm a bit skeptical about this change.

I don't have data for ninja, for my own thing the difference was
striking (when trying to spawn somethig from a process with a huge
resident set, and on a Linux system with disabled overcommit)

-- 
mailto:moosotc@gmail.com

nico:
From what I understand, how much this helps is proportional to the working set of the forked process, and ninja tends to not use a lot of ram (70MB for chromium).

Thanks for teaching me that compilers (at least gcc and clang) define _GNU_SOURCE themselves when compiling C++ code, I wasn't aware of that. That means this patch is complete, and it's pretty short, so I suppose we might as well merge it :-)

colincross:
On an internal Android build building all optional modules, ninja uses 1.3GB, so I expect this will have a significant positive impact.

danw:
I did some quick tests on our Android AOSP (smaller) builds, and for a set of 79 short commands that we run immediately after reading the manifest, this shrinks the total time spent on those commands from 511ms to 31ms. For the entire build, it seems to save about 40 minutes (out of 80) in the kernel, and saves about 7 minutes (out of 34) of our real time. I only took one sample for comparison, so there's likely to be some noise in the numbers.

PR #1185, opened 2016-09-15
FreeBSD supports ppoll(2)

gahr:
This also fixes this test error `SubprocessTest.SetWithLotsninja: fatal: pipe: Too many open files` caused by 
```cpp
if (fd_ >= static_cast<int>(FD_SETSIZE))                                      
    Fatal("pipe: %s", strerror(EMFILE)); 
```
in Subprocess::Start

### DIFF
--- a/src/subprocess-posix.cc
+++ b/src/subprocess-posix.cc
@@ -24,6 +24,13 @@
 #include <sys/wait.h>
 #include <spawn.h>
 
+#ifdef __FreeBSD__
+#  include <sys/param.h>
+#  if defined USE_PPOLL && __FreeBSD_version < 1002000
+#    undef USE_PPOLL
+#  endif
+#endif
+
 extern char** environ;
 
 #include "util.h"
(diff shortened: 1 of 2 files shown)


MagikBSD:
ppoll() is available only on FreeBSD 10.2 and up, but not on older, but still supported releases.
EOL for 9.3 and 10.1 are established to 31 Dec 2016.

gahr:
You are right. Let's keep this here until we EOL older releases.

evmar:
Is there a way to test the FreeBSD version at compile-time?  (E.g. an ifdef?)

MagikBSD:
#include <osreldate.h>
#if (__FreeBSD_version >= 1002000)

gahr:
It's in sys/pram.h

gahr:
... sys/param.h

src/subprocess-posix.cc
```diff
@@ -24,6 +24,13 @@
 #include <sys/wait.h>
 #include <spawn.h>
 
+#ifdef __FreeBSD__
+#  include <sys/param.h>
+#  if defined USE_PPOLL && __FreeBSD_version < 1002000
```
nico: fwiw we don't put spaces between # and the word after it in ninja; maybe you can zap them in a follow-up

gahr:
Done in #1187

PR #1187, opened 2016-09-20
Style cleanup

gahr:
Cleanup my commit of #1185


nico:
Sorry,  forgot about this one and moved the macros to the header (to fix the test failing on older BSDs), so now your patch no longer applies :-( If you want, you can rebase it, or I can manually do this some time soon.

gahr:
Will be quicker to just reindent the block than rebasing and merging this pull req. Feel free to proceed at your discretion ;)

Issue #1186, opened 2016-09-16
Symbolic links created as outputs are not stat'ed correctly


agrieve:
In Chromium's Android build, we have [a rule](https://cs.chromium.org/chromium/src/build/symlink.gni?q=symlink.gni&sq=package:chromium&dr&l=18) that creates a symlink. We mark it as an output in ninja, but it looks like [ninja is using](https://github.com/ninja-build/ninja/blob/master/src/disk_interface.cc#L187) `stat` rather than `lstat`, so it is actually following the symlink when querying its mtime rather than using the output's actual mtime.

I think there are cases where you would actually want to follow the symlink when requesting its mtime. Namely - when there's a non-generated input (or basically any time where you're not actually dealing with a symlink that's an output.

nico:
Copying over from the pull request:

I found a repro!

Let's say you're trying to build an executable, and you also want a symlink to that executable under a different name (think e.g. clang and clang-cl). In gn, you might express this as a target that creates the symlink, and then make the actual executable link depend on the symlink target (the symlink's dest does after all not have to exist when the symlink's created). gn will then make a build rule that creates a symlink, then a stamp file that depends on the symlink for the symlink action, and then a link rule for the executable, order-only depending on the stamp, like so:

```
rule symlink
  command = ln -sf lld lld-link
  restat = 1
build lld-link: symlink

rule stamp
  command = touch $out
  restat = 1
build lld-link.stamp: stamp lld-link

rule link
  command = sleep 2 && touch $out
  restat = 1
build lld: link || lld-link.stamp
```

With that setup:

```
Nicos-MacBook-Pro:foo thakis$ ninja -d explain | cat 
ninja explain: output lld-link doesn't exist
ninja explain: lld-link is dirty
ninja explain: output lld doesn't exist
ninja explain: output lld-link.stamp doesn't exist
[1/3] ln -sf lld lld-link
[2/3] touch lld-link.stamp
[3/3] sleep 2 && touch lld
Nicos-MacBook-Pro:foo thakis$ ninja -d explain | cat 
ninja explain: restat of output lld-link.stamp older than most recent input lld-link (1522896947 vs 1522896949)
[1/1] touch lld-link.stamp
Nicos-MacBook-Pro:foo thakis$ ninja -d explain | cat 
ninja: no work to do.
```

nico:
[...]

PR #1188, opened 2016-09-21
Be smarter about mtimes when stat()'ing symlinks.

agrieve:
Before this change, symlinks were always followed when retrieving
mtimes. This can cause ninja to think targets are stale when outputs are
symlinks.

This patch introduces the following stat() logic for symlinks:
1. When Stat()ing inputs, use the newer of the symlink or what it points to.
- By looking at both, it maintains the pre-existing behaviour, where
  you could use a symlink as an input and ninja would know to rebuild
  when the thing it points to changes.
  1. When Stat()ing outputs, always use the timestamp of the symlink.
- If it's listed as an output, then it contains the only mtime ninja
  should care about.

Fixes Bug #1186

### DIFF
--- a/src/disk_interface.cc
+++ b/src/disk_interface.cc
@@ -131,7 +131,7 @@ bool DiskInterface::MakeDirs(const string& path) {
   if (dir.empty())
     return true;  // Reached root; assume it's there.
   string err;
-  TimeStamp mtime = Stat(dir, &err);
+  TimeStamp mtime = Stat(dir, kDontFollow, &err);
   if (mtime < 0) {
     Error("%s", err.c_str());
     return false;
@@ -148,7 +148,10 @@ bool DiskInterface::MakeDirs(const string& path) {
 
 // RealDiskInterface -----------------------------------------------------------
 
-TimeStamp RealDiskInterface::Stat(const string& path, string* err) const {
+TimeStamp RealDiskInterface::Stat(
+    const string& path,
+    DiskInterface::SymlinkTreatment symlink_treatment,
+    string* err) const {
   METRIC_RECORD("node stat");
 #ifdef _WIN32
   // MSDN: "Naming Files, Paths, and Namespaces"
@@ -181,7 +184,22 @@ TimeStamp RealDiskInterface::Stat(const string& path, string* err) const {
   return di != ci->second.end() ? di->second : 0;
 #else
   struct stat st;
-  if (stat(path.c_str(), &st) < 0) {
+  int result;
+  switch (symlink_treatment) {
+  case kDontFollow:
+    result = lstat(path.c_str(), &st);
+    break;
+  case kTakeNewest:
+  default:
+    result = lstat(path.c_str(), &st);
+    if (result == 0 && S_ISLNK(st.st_mode)) {
+      struct stat st2;
+      if (stat(path.c_str(), &st2) == 0) {
+        st.st_mtime = max(st.st_mtime, st2.st_mtime);
+      }
+    }
+  }
+  if (result < 0) {
     if (errno == ENOENT || errno == ENOTDIR)
       return 0;
     *err = "stat(" + path + "): " + strerror(errno);
--- a/src/disk_interface.h
+++ b/src/disk_interface.h
@@ -44,9 +44,24 @@ struct FileReader {
 /// Abstract so it can be mocked out for tests.  The real implementation
 /// is RealDiskInterface.
 struct DiskInterface: public FileReader {
+  /// How to treat symlinks when calling Stat().
+  enum SymlinkTreatment {
+    // Used for outputs. Even though outputs are often used as inputs as well, a
+    // newly created symlink should always have an mtime newer than the thing it
+    // points to at the time of its creation.
+    kDontFollow,
+    // Takes the newest mtime of the symlink or the file it points to. Does not
+    // follow symlink chains. Used when computing input mtimes, as it will cause
+    // a target to be dirty when either the symlink changes, or the thing it
+    // points to changes.
+    kTakeNewest,
+  };
+
   /// stat() a file, returning the mtime, or 0 if missing and -1 on
   /// other errors.
-  virtual TimeStamp Stat(const string& path, string* err) const = 0;
+  virtual TimeStamp Stat(const string& path,
+                         SymlinkTreatment symlink_treatment,
+                         string* err) const = 0;
 
   /// Create a directory, returning false on failure.
   virtual bool MakeDir(const string& path) = 0;
@@ -75,7 +90,8 @@ struct RealDiskInterface : public DiskInterface {
 #endif
                       {}
   virtual ~RealDiskInterface() {}
-  virtual TimeStamp Stat(const string& path, string* err) const;
+  virtual TimeStamp Stat(const string& path, SymlinkTreatment symlink_treatment,
+                         string* err) const;
   virtual bool MakeDir(const string& path);
   virtual bool WriteFile(const string& path, const string& contents);
   virtual Status ReadFile(const string& path, string* contents, string* err);
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -27,8 +27,11 @@
 #include "state.h"
 #include "util.h"
 
-bool Node::Stat(DiskInterface* disk_interface, string* err) {
-  return (mtime_ = disk_interface->Stat(path_, err)) != -1;
+bool Node::Stat(DiskInterface* disk_interface,
+                DiskInterface::SymlinkTreatment symlink_treatment,
+                string* err) {
+  mtime_ = disk_interface->Stat(path_, DiskInterface::kDontFollow, err);
+  return mtime_ != -1;
 }
 
 bool DependencyScan::RecomputeDirty(Node* node, string* err) {
@@ -44,7 +47,8 @@ bool DependencyScan::RecomputeDirty(Node* node, vector<Node*>* stack,
     if (node->status_known())
       return true;
     // This node has no in-edge; it is dirty if it is missing.
-    if (!node->StatIfNecessary(disk_interface_, err))
+    if (!node->StatIfNecessary(disk_interface_, DiskInterface::kTakeNewest,
+                               err))
       return false;
     if (!node->exists())
       EXPLAIN("%s has no in-edge and is missing", node->path().c_str());
@@ -71,7 +75,8 @@ bool DependencyScan::RecomputeDirty(Node* node, vector<Node*>* stack,
   // Load output mtimes so we can compare them to the most recent input below.
   for (vector<Node*>::iterator o = edge->outputs_.begin();
        o != edge->outputs_.end(); ++o) {
-    if (!(*o)->StatIfNecessary(disk_interface_, err))
+    if (!(*o)->StatIfNecessary(disk_interface_, DiskInterface::kDontFollow,
+                               err))
       return false;
   }
 
--- a/src/build.cc
+++ b/src/build.cc
@@ -577,7 +577,8 @@ void Builder::Cleanup() {
         // mentioned in a depfile, and the command touches its depfile
         // but is interrupted before it touches its output file.)
         string err;
-        TimeStamp new_mtime = disk_interface_->Stat((*o)->path(), &err);
+        TimeStamp new_mtime = disk_interface_->Stat(
+            (*o)->path(), DiskInterface::kDontFollow, &err);
         if (new_mtime == -1)  // Log and ignore Stat() errors.
           Error("%s", err.c_str());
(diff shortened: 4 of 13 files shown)


agrieve:
Note: I confirmed this fixes the Chromium Cronet Mips bot's up-to-date check by scp'ing a copy of ninja with this patch:

```
chrome-bot@slave114-c1:(Linux 14.04):~$ /b/build/scripts/slave/.recipe_deps/depot_tools/ninja -w dupbuild=err -C /b/build/slave/Android_Cronet_MIPS_Builder/build/src/out/Release cronet_package cronet_perf_test_apk cronet_sample_test_apk cronet_test_instrumentation_apk cronet_unittests cronet_unittests_apk net_unittests net_unittests_apk -j50 -d explain -n
ninja: Entering directory `/b/build/slave/Android_Cronet_MIPS_Builder/build/src/out/Release'
ninja explain: restat of output md5sum_bin_host older than most recent input obj/tools/android/md5sum/md5sum_bin_host.inputdeps.stamp (0 vs 1474466541)
ninja explain: md5sum_bin_host is dirty
ninja explain: obj/tools/android/md5sum/md5sum_bin_host.stamp is dirty
ninja explain: restat of output host_forwarder older than most recent input obj/tools/android/forwarder2/host_forwarder.inputdeps.stamp (0 vs 1474466541)
ninja explain: host_forwarder is dirty
[5/5] STAMP obj/tools/android/md5sum/md5sum.stamp
chrome-bot@slave114-c1:(Linux 14.04):~$ /b/build/scripts/slave/.recipe_deps/depot_tools/ninja -w dupbuild=err -C /b/build/slave/Android_Cronet_MIPS_Builder/build/src/out/Release cronet_package cronet_perf_test_apk cronet_sample_test_apk cronet_test_instrumentation_apk cronet_unittests cronet_unittests_apk net_unittests net_unittests_apk -j50 -d explain -n
ninja: Entering directory `/b/build/slave/Android_Cronet_MIPS_Builder/build/src/out/Release'
ninja explain: restat of output md5sum_bin_host older than most recent input obj/tools/android/md5sum/md5sum_bin_host.inputdeps.stamp (0 vs 1474466541)
ninja explain: md5sum_bin_host is dirty
ninja explain: obj/tools/android/md5sum/md5sum_bin_host.stamp is dirty
ninja explain: restat of output host_forwarder older than most recent input obj/tools/android/forwarder2/host_forwarder.inputdeps.stamp (0 vs 1474466541)
ninja explain: host_forwarder is dirty
[5/5] STAMP obj/tools/android/md5sum/md5sum.stamp
chrome-bot@slave114-c1:(Linux 14.04):~$ ./ninja -w dupbuild=err -C /b/build/slave/Android_Cronet_MIPS_Builder/build/src/out/Release cronet_package cronet_perf_test_apk cronet_sample_test_apk cronet_test_instrumentation_apk cronet_unittests cronet_unittests_apk net_unittests net_unittests_apk -j50 -d explain -n
ninja: Entering directory `/b/build/slave/Android_Cronet_MIPS_Builder/build/src/out/Release'
ninja: no work to do.
```

maximuska:
I haven't looked at the patch.

But as for the described logic, IMHO it would be simpler to allow
explicitly specifying in the manifest whether stat or lstat should be used
for each specific target.

How are you proposing handling broken symlinks?

agrieve:
I've added a unit test for broken links, but basically I'm proposing that we use the symlink's timestamp for this case.

If there's a case where explicitly specifying the whether to follow the symlink when checking mtimes is useful, then I'd be fine going that route as well. I don't think it's necessarily simpler though to add another thing the .ninja author needs to worry about.

I think what I've proposed here is a better default behavior at least. Specifically - it never makes sense to follow an output symlink when retrieve mtimes.

nico:
Hm, that's a pretty large patch for a fairly edge-case-y behavior that I wouldn't be surprised someone relied on in some surprising way. Can you make your build not use symlinks instead?

agrieve:
I could switch the Android build to generate wrapper scripts rather than symlinks, but in the spirit of fixing the root cause of things, I arrived here (plus I also discovered iOS creates symlinks as well). The patch is actually quite small if you ignore updating every call site of `Stat()`.

Here's why I believe this is a clear bug:

Image you have a ninja rule with 2 inputs and an output symlink to input1. If input2 has a timestamp newer than input1, ninja will forever think that the output is stale, because it's reporting input1's timestamp rather than the output symlink's timestamp.

agrieve:
Thought of an even simpler example:

* A rule with only one input: makesymlink.py
* An output symlink that points to a checked-in file (not listed as an input)

If the checked-in file the symlink points to is not newer than the checked-in makesymlink.py, ninja will always think the target is dirty.

agrieve:
🐼?

colincross:
The android build needs symlinks in lots of places, and we aren't going to be able to avoid that.  Right now we create them as side-effects of installing the the file they link to, but that means that ninja won't recreate them if you delete one.  Using lstat on output files would be perfect for us.

agrieve:
🐹

agrieve:
@nico - Any further thoughts here? The main change here is ~30 lines, which I don't think is that big.

rsesek:
I also hit this issue in Mac framework packaging and have had to work around it (https://codereview.chromium.org/2487763002/).

agrieve:
🎁

agrieve:
Another month, another ping.

agrieve:
@nico What could I do to convince you to look at this? Are there other owners for ninja that could take a look?

agrieve:
Q: Why not switching everything to lstat()?
A: Consider an existing setup like:
1. Shared network drive with BigFile.dat on it
2. Local build checkout that expects BigFile.dat to be within the checkout
3. User symlinks BigFile.dat to point to the network copy
4. Network copy of BigFile.dat changes, you'd want ninja to consider it dirty.

Q: Okay, so why not leave it as stat()?
A: Consider this setup:
1. A target depends on foo.dat, which links to /a/foo.dat
2. User changes symlink from /a/foo.dat to /b/foo.dat
3. Both /a/foo.dat and /b/foo.dat are older than the output, but because the symlink's mtime is now newer, ninja should rebuild.

Q: Why not use stamp files for output edges, then have symlinks ninja doesn't know about:
A: If someone deletes the actual symlink and not the stamp file, ninja will not know to rebuild it. Having ninja just do the right thing would be much better than having unspecified outputs.

Q: Why not make this patch smaller using a default value for Stat()'s new parameter:
A: I could make the patch a bunch smaller by making DiskInterface::kDontFollow be the default value for Stat(), but I thought making it explicit everywhere would lead to fewer bugs.

Q: Why is this issue important?
A: Because correct builds are important.

agrieve:
Somehow I don't see it in the history here, but the relavant Android Chrome bug for this is:
https://bugs.chromium.org/p/chromium/issues/detail?id=646165

agrieve:
ping

agrieve:
ping

src/disk_interface.h
```diff
@@ -44,9 +44,18 @@ struct FileReader {
 /// Abstract so it can be mocked out for tests.  The real implementation
 /// is RealDiskInterface.
 struct DiskInterface: public FileReader {
+  /// How to treat symlinks when calling Stat().
+  enum SymlinkTreatment {
+    kFollow,      // Used when the only the file contents relevant.
```
nico: the only the

src/disk_interface.h
```diff
@@ -44,9 +44,18 @@ struct FileReader {
...
+    kDontFollow,  // Used for outputs.
+    kTakeNewest,  // Used for inputs.
```
nico: Thanks for the comments here, they're very helpful.

However, aren't most nodes both outputs and inputs, yet stat()'d only once? What happens in that case?

nico:
Alright, let's get this in. Sorry about the very long wait. Please see the one question below.

agrieve:
Thanks for taking the time on this!

Updated the comments to better explain things. Rebased, and also removed kFollow. I don't think there's ever a situation where kTakeNewest wouldn't be safer to use.

nico:
I tried reading some docs and from what I can tell posix says "For symbolic links, the st_mode member shall contain meaningful information when used with the file type macros, and the st_size member shall contain the length of the pathname contained in the symbolic link. File mode bits and the contents of the remaining members of the stat structure are unspecified." – i.e. it's up to chance what's in the timestamp for symbolic links. That maybe feels a bit brittle?

agrieve:
When I was looking at this before, that had me worried as well. I came across this though:
https://linux.die.net/man/2/lstat

According to POSIX.1-2001, lstat() on a symbolic link need return valid information only in the st_size field and the file-type component of the st_mode field of the stat structure. POSIX.-2008 tightens the specification, requiring lstat() to return valid information in all fields except the permission bits in st_mode.

I'd guess by now that assuming posix-2008 compliance is okay, but really have no experience with making these sorts of judgements. wdyt?

src/graph.cc
```diff
@@ -27,8 +27,12 @@
 #include "state.h"
 #include "util.h"
 
-bool Node::Stat(DiskInterface* disk_interface, string* err) {
-  return (mtime_ = disk_interface->Stat(path_, err)) != -1;
+bool Node::Stat(DiskInterface* disk_interface,
+                DiskInterface::SymlinkTreatment symlink_treatment,
+                string* err) {
+  METRIC_RECORD("node stat");
```
nico: Is this a mismerge? danalbert moved this METRIC_RECORD into disk_interface recently.

nico:
Ah thanks, I think that's fine then.

One tiny comment, but let's merge this and see how it goes. Is this the only thing needed to make the chrome/android build have clean rebuilds, or is there more work needed that's blocked on this?

Everyone else, I'm interested in feedback on this – so please let me know how your builds look with this merged (once it is merged).

src/graph.cc
agrieve: blarg, looks like it. Thanks for catching!

agrieve:
Fixed up the mismerge. This is the only thing required to enable the clean rebuild check on the android bots (unless a new source of uncleanliness has crept in in the meantime)

src/disk_interface.cc
```diff
@@ -181,7 +184,22 @@ TimeStamp RealDiskInterface::Stat(const string& path, string* err) const {
...
 #else
   struct stat st;
-  if (stat(path.c_str(), &st) < 0) {
+  int result;
+  switch (symlink_treatment) {
+  case kDontFollow:
+    result = lstat(path.c_str(), &st);
+    break;
+  case kTakeNewest:
+  default:
```
atetubou: need  this default label?

agrieve: Lint likes it to be there.

agrieve:
Ready for action?

nico:
agrieve and I discussed this change in person some earlier this week and mostly convinced ourselves that it'd break some existing build files. We then tried to understand what exactly goes wrong without this patch and failed to recreate that too. So for now, it's not clear if anything needs to be done in ninja here.

dpranke:
I wonder if this issue was partially a result of crbug.com/748623, where the way Chromium managed testing patches on the bots confused the mtimes of files, and running `git update-index --refresh` has maybe made the problem go away?

nico:
I found a repro!

Let's say you're trying to build an executable, and you also want a symlink to that executable under a different name (think e.g. clang and clang-cl). In gn, you might express this as a target that creates the symlink, and then make the actual executable link depend on the symlink target (the symlink's dest does after all not have to exist when the symlink's created). gn will then make a build rule that creates a symlink, then a stamp file that depends on the symlink for the symlink action, and then a link rule for the executable, order-only depending on the stamp, like so:

```
rule symlink
  command = ln -sf lld lld-link
  restat = 1
build lld-link: symlink

rule stamp
  command = touch $out
  restat = 1
build lld-link.stamp: stamp lld-link

rule link
  command = sleep 2 && touch $out
  restat = 1
build lld: link || lld-link.stamp
```

With that setup:

```
Nicos-MacBook-Pro:foo thakis$ ninja -d explain | cat 
ninja explain: output lld-link doesn't exist
ninja explain: lld-link is dirty
ninja explain: output lld doesn't exist
ninja explain: output lld-link.stamp doesn't exist
[1/3] ln -sf lld lld-link
[2/3] touch lld-link.stamp
[3/3] sleep 2 && touch lld
Nicos-MacBook-Pro:foo thakis$ ninja -d explain | cat 
ninja explain: restat of output lld-link.stamp older than most recent input lld-link (1522896947 vs 1522896949)
[1/1] touch lld-link.stamp
Nicos-MacBook-Pro:foo thakis$ ninja -d explain | cat 
ninja: no work to do.
```

nico:
Since we're closing pull requests we're not planning on merging nowadays, let's close this out given the discussion.

PR #1191, opened 2016-10-07
Add NINJA_STATUS_STARTED and _FINISHED

sgraham:
I had a thought that this might be useful. https://github.com/ninja-build/ninja/pull/999 has vaguely bugged me on Windows since now it looks like I'm waiting for the last .cc file to compile, when it's actually of course the link.

Also, with a simple further addition to the %-formats to include a job-slot-index, I think #1020 could then be implemented as a simple (?) filter.


nicolasdespres:
+1

Will be useful to address some of the issues in #746.

nico:
(Note to self: Related to #746 and #1180)

PR #1193, opened 2016-10-08
Code clean up and deletion of deprecated bootstrap.py file.

maciejczyzewski:


src/browse.h
```diff
@@ -22,7 +22,7 @@ struct State;
 /// \a args are the number of arguments to be passed to the Python script.
 /// \a argv are arguments to be passed to the Python script.
 /// This function does not return if it runs successfully.
-void RunBrowsePython(State* state, const char* ninja_command,
-                     const char* input_file, int argc, char* argv[]);
+void RunBrowsePython(State *state, const char *ninja_command,
```
nico: these moved in the wrong direction

nico:
Thanks for the patch! I don't think that clang-formatting all the code is super useful though. Some of the things it changes looks like obviously good changes, but much of it is noise. I put the clang-format file there mostly for clang-format-diff use.

PR #1194, opened 2016-10-12
Improve error message when a depfile contains a bad path

ilor:
Debugging "ninja: error: empty path" is not fun, so make the error
message mention the depfile name.


nico:
Thanks!

PR #1196, opened 2016-10-19
Optimize ReadFile allocations

danw:
Instead of continuously reallocating the output string, call `reserve()` up front to enlarge the string's buffer.

For our larger build.ninja files (~550MB), this saves about 500ms, which is more than half the time spent in this function. (Measured by adding `METRICS_RECORD` to this function during manifest_parser_perftest)

For the standard manifest_parser_perftest, this only saves ~20ms out of ~600ms for a full run.

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -380,9 +380,19 @@ int ReadFile(const string& path, string* contents, string* err) {
     return -errno;
   }
 
+  struct stat st;
+  if (fstat(fileno(f), &st) < 0) {
+    err->assign(strerror(errno));
+    fclose(f);
+    return -errno;
+  }
+
+  // +1 is for the resize in ManifestParser::Load
+  contents->reserve(st.st_size + 1);
+
   char buf[64 << 10];
   size_t len;
-  while ((len = fread(buf, 1, sizeof(buf), f)) > 0) {
+  while (!feof(f) && (len = fread(buf, 1, sizeof(buf), f)) > 0) {
     contents->append(buf, len);
   }
   if (ferror(f)) {


evmar:
Doesn't this double the number of syscalls per input file?  I guess the
number of input files is relatively small enough for it to not matter?

danw:
As far as I can tell, this function is used primarily in two places: manifest reading and depfile reads. There are few enough manifests that the overhead is minimal, and the gains for larger manifests is significant.

For depfiles, there are many more of them, and they'll generally be smaller. It will increase the number of file-related syscalls by 33% (fopen/fread/fclose) for files <64KB, but it doesn't seem to materially change the results of depfile_parser_perftest on a 32KB file on my machine.

sgraham:
Is it expensive in practice to just reserve() a "big number" instead of stat()ing?

nicolasdespres:
@sgraham In practice you can assume that one reserve() call = one malloc() call. So if it is too big and if the buffers stay around you may run out of memory.

When there is no call to reserve() the typical allocation scheme is to start with a capacity of 1, 2 or 4 and to multiply it by 2 when it must be increased. Calling reserve() up front reduce the number of malloc() calls because the capacity does not change when the string size increases.

An alternative to get rid of the stat() call could be to always call reserve(4096) (for instance). If 4096 bytes are not enough the buffer capacity will be automatically increased ; involving extra malloc() calls but less than if we had started from a capacity of 2 or 4.

If the buffer is dropped once the parsing is finished we can even use a bigger value than 4096.

Another alternative would be to mmap'ed the file. This way we get rid of the memory allocator calls and the file-related syscalls all together.

sgraham:
Yes, I understand the reason the change was proposed. It appears that ReadFile() is only used one at a time. So I was wondering if _in practice_ simply doing "contents->reserve(100*1024*1024);" (or some other number) would be faster and simpler.

danw:
In addition to all the malloc calls, every time the size needs to increase, the current buffer needs to be copied to the new buffer. So for allocators that use mmap for large allocations, that's at least two system calls (mmap, munmap) and potentially a multi-hundred megabyte copy. For smaller allocations, it can likely be handled without a system call at all.

This same function is used for <4KB files and >500MB files, so a single "big number" would not be appropriate. Especially on 32-bit machines, or machines with overcommit limitations, using too much virtual address space is a problem.

Why is fstat so much a problem here? It should be faster than stat() -- it doesn't have to do any path lookups, since the kernel already has the inode from the open file descriptor.

mmap is an option, but it didn't seem to help that much here in my limited testing. With a binary manifest format, and string tables, it could be more useful to reduce the number of allocations elsewhere, but that's a separate discussion -- #1093 or similar.

nicolasdespres:
@danw Simply calling a system function come with an overhead. This is the overhead Evan was talking about. It is independent of the complexity of the system function that is actually called.

I would try with reserve(4KB) so that only one allocation is performed for the many files <4KB and that the behavior for the not so many files >4KB is almost unchanged.

danw:
All the tests below were done on ~Ubuntu 14.04 on a 64-bit machine with 2x Xeon E5-2690 v3 @ 2.6Ghz.

Yes, there is some overhead from every syscall. But not every syscall is equal either -- fstat takes 210ns on my machine, while stat takes 500ns. It's also important to see which syscalls are actually being used (with strace):

```
fopen:
  open
  fstat
fread:
  mmap
  read
  read
fread:
  read
fclose:
  close
  munmap
```

Where this change adds another fstat/fstat call, so adding 1 to 8 current syscalls.

I also did some microbenchmarking with different file sizes:

readfile_small

readfile_medium

readfile_big

1. There's a lot of noise in the larger file size runs. Due to the longer runs, there are fewer loops (at least 20 -- smaller ones are closer to 5 million) and more variation. I added error bars to the big file graph to show the approximate range of one standard deviation away from the mean.
2. Using reserve(4096) is almost always worse than any other solution I benchmarked.
3. Using fstat+reserve really starts helping the moment we get data from fread more than once -- so >64kb files.

The feof and fstat&feof lines come from the observation above that we can remove the second call to fread in the <= 64kb cases. So in the while loop, !feof(f) is checked first, then fread is called. I've updated the code here with that optimization as well.

I tried deferring the fstat call until after the first fread, so that we didn't need to do it for these small files. I see that working in strace, but the benchmark results don't show any improvement except in the file size == 0 case.

jhasse:
Makes sense. Thanks for your extensive benchmarking :)

PR #1197, opened 2016-10-20
Explain as human readable datetime instead of epoch

doheonlee:
Ninja explains output is older than most recent input with their epoch.
Datetime is changed from epoch value to datetime human reads easily.

### DIFF
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -174,11 +174,22 @@ bool DependencyScan::RecomputeOutputDirty(Edge* edge,
     }
 
     if (output_mtime < most_recent_input->mtime()) {
+      time_t output_t = static_cast<time_t>(output_mtime);
+      time_t most_recent_input_t = static_cast<time_t>(most_recent_input->mtime());
+
+      // Convert to human readable time string instead of epoch time
+      char *output_t_str = ctime(&output_t);
+      char *most_recent_input_t_str = ctime(&most_recent_input_t);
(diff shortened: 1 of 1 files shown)


src/graph.cc
```diff
@@ -174,11 +174,22 @@ bool DependencyScan::RecomputeOutputDirty(Edge* edge,
     }
 
     if (output_mtime < most_recent_input->mtime()) {
+      time_t output_t = static_cast<time_t>(output_mtime);
+      time_t most_recent_input_t = static_cast<time_t>(most_recent_input->mtime());
```
evmar: On Windows, this value is not a time_t.  I think we need a function that converts these mtimes to human-readable times that is implemented differently on Linux and Windows.

evmar: (It'd be fine to just report the number as-is on Windows. It's just that passing it to ctime isn't right.)

doheonlee:
Dear @evmar , thank you very much for your detailed reviewing action :)
I uploaded new commit for some modification.

First, let this patch works on not WINDOWS platform and wrapped it with g_explaning condition.
EXPLAIN macro acts only g_explaining is true, thus this wrapping does not need else condition.

jhasse:
Sorry that this has been open for such a long time without a comment.

I think that having a consistent message on all platforms is more important though. Maybe we can revisit this later, closing for now.

PR #1199, opened 2016-11-03
Use ninja startup as time base for the ninja_log

danw:
Currently, the time base used for the ninja log is reset for every build, including if the build.ninja is regenerated and the build restarts. It also won't include the manifest read time - in most cases that's minimal, but ours is several seconds.

For our builds that required regen, this caused the log to look like:

```
  0  3     ...
  1  4     ...
  4  1697  ... out/build.ninja
  0  5     ...
  1  6     ...
  5  10    ...
```

With this patch, it looks more like:

```
  3337  3340  ...
  3338  3341  ...
  3341  5034  ... out/build.ninja
  9487  9492  ...
  9488  9493  ...
  9492  9497  ...
```

Then https://github.com/nico/ninjatracing can create proper traces that show the relation between the restarts, and can account for the initial setup overhead.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -73,7 +73,6 @@ bool DryRunCommandRunner::WaitForCommand(Result* result) {
 
 BuildStatus::BuildStatus(const BuildConfig& config)
     : config_(config),
-      start_time_millis_(GetTimeMillis()),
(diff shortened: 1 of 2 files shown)


jhasse:
In order to reduce the number of open PRs. If you still think this should be merged, please reopen (and rebase on master).

PR #1200, opened 2016-11-03
Add a new deps format for a newline-separate list of files.

brettw:
Simplifies integration with scripts which no longer need to write a
Makefile format. The Makefile format is confusing in the presence of
multiple outputs, and has complex escaping and quoting rules.

The new list format has no escaping rules as the format is a newline-
separated list of literal filenames.


jhasse:
Out of scope for now, sorry.

PR #1208, opened 2016-11-14
Optimize lexing and parsing

colincross:
These two patches add a few small optimizations to the lexing and parsing code that were found with vallgrind --tool=callgrind and verified with perf stat:
   - Reserve string sizes before repeated appends in EdgeEnv::MakePathList
   - Separate Lexer::ReadPath and Lexer::ReadVarValue so that ReadVarValue doesn't split on whitespace
   - Use a computed lookup table in IsKnownShellSafeCharacter

### DIFF
--- a/src/eval_env.cc
+++ b/src/eval_env.cc
@@ -107,12 +107,7 @@ string EvalString::Evaluate(Env* env) const {
 }
 
 void EvalString::AddText(StringPiece text) {
-  // Add it to the end of an existing RAW token if possible.
-  if (!parsed_.empty() && parsed_.back().second == RAW) {
-    parsed_.back().first.append(text.str_, text.len_);
-  } else {
-    parsed_.push_back(make_pair(text.AsString(), RAW));
-  }
+  parsed_.push_back(make_pair(text.AsString(), RAW));
 }
 void EvalString::AddSpecial(StringPiece text) {
   parsed_.push_back(make_pair(text.AsString(), SPECIAL));
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -272,6 +272,12 @@ string EdgeEnv::MakePathList(vector<Node*>::iterator begin,
                              vector<Node*>::iterator end,
                              char sep) {
   string result;
+  size_t len = 0;
+  for (vector<Node*>::iterator i = begin; i != end; ++i) {
+    len += (*i)->path().size() + 1;
+  }
+
+  result.reserve(len);
   for (vector<Node*>::iterator i = begin; i != end; ++i) {
     if (!result.empty())
       result.push_back(sep);
--- a/src/lexer_test.cc
+++ b/src/lexer_test.cc
@@ -33,7 +33,7 @@ TEST(Lexer, ReadEvalStringEscapes) {
   string err;
   EXPECT_TRUE(lexer.ReadVarValue(&eval, &err));
   EXPECT_EQ("", err);
-  EXPECT_EQ("[ $ab c: cde]",
+  EXPECT_EQ("[ ][$][ab c][:][ ][cde]",
             eval.Serialize());
 }
 
--- a/src/util.cc
+++ b/src/util.cc
@@ -26,6 +26,7 @@
 #include <assert.h>
 #include <errno.h>
 #include <fcntl.h>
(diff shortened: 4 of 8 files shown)


src/lexer_test.cc
```diff
@@ -33,7 +33,7 @@ TEST(Lexer, ReadEvalStringEscapes) {
   string err;
   EXPECT_TRUE(lexer.ReadVarValue(&eval, &err));
   EXPECT_EQ("", err);
-  EXPECT_EQ("[ $ab c: cde]",
+  EXPECT_EQ("[ ][$][ab c][:][ ][cde]",
```
jhasse: What is happening here? Sorry I don't really understand this part of the code.

colincross: The AddText optimization to concatenate raw string tokens was removed in favor of keeping an extra token to avoid copying the whole string.  Each escaped character (`$ `, `$$`, `$:`, `$\n`) is parsed and added separately using AddText, so they are now separate tokens in the result.

jhasse: I see, thanks for the explanation :)

jhasse:
Can you rebase on master (and regenerate lexer.cc)?

src/util.cc
```diff
@@ -289,18 +307,18 @@ void GetShellEscapedString(const string& input, string* result) {
...
 
   string::const_iterator span_begin = input.begin();
-  for (string::const_iterator it = input.begin(), end = input.end(); it != end;
-       ++it) {
-    if (*it == kQuote) {
-      result->append(span_begin, it);
-      result->append(kEscapeSequence);
-      span_begin = it;
-    }
+  string::const_iterator it = std::find(span_begin, input.end(), kQuote);
```
jonesmz: it is not used outside the scope of the for loop.

Consider putting this line as the initialization statement of the forloop.

```cpp
for (string::const_iterator it = std::find(span_begin, input.end(), kQuote);
      it != input.end();
      it = std::find(span_begin, input.end(), kQuote)) {
    result->append(span_begin, it);
    result->append(kEscapeSequence);
    span_begin = ++it;
}
```

src/graph.cc
```diff
@@ -272,6 +272,12 @@ string EdgeEnv::MakePathList(vector<Node*>::iterator begin,
                              vector<Node*>::iterator end,
                              char sep) {
   string result;
+  size_t len = 0;
+  for (vector<Node*>::iterator i = begin; i != end; ++i) {
+    len += (*i)->path().size() + 1;
```
jonesmz: Where does the +1 come from? Add a comment elaborating.

src/eval_env.cc
```diff
@@ -107,12 +107,7 @@ string EvalString::Evaluate(Env* env) const {
 }
 
 void EvalString::AddText(StringPiece text) {
-  // Add it to the end of an existing RAW token if possible.
-  if (!parsed_.empty() && parsed_.back().second == RAW) {
```
jonesmz: Why is this change safe?

I think a comment indicating the change in semantics, and justification thereof, of the AddText function would be wise here.

src/graph.cc
```diff
@@ -272,6 +272,12 @@ string EdgeEnv::MakePathList(vector<Node*>::iterator begin,
                              vector<Node*>::iterator end,
                              char sep) {
   string result;
+  size_t len = 0;
+  for (vector<Node*>::iterator i = begin; i != end; ++i) {
```
jonesmz: I feel like this should instead be a call to std::accumulate

src/util.cc
```diff
@@ -239,22 +241,38 @@ bool CanonicalizePath(char* path, size_t* len, unsigned int* slash_bits,
...
+        return false;
+    }
   }
-}
+
+  bool operator()(char ch) const {
+    return known_shell_safe_characters_[static_cast<unsigned char>(ch)];
+  }
+
+  bool known_shell_safe_characters_[UCHAR_MAX];
```
jonesmz: Can't this be made static, and then initialized once at program start?

Better even would be to conditionally compile so that when C++11 is available, the initialization of this array is constexpr, and the array is part of the read-only-data section of the binary.

PR #1209, opened 2016-11-18
add script ninja-deps to query dependencies

joez:
Ninja has already maintain a internal database to store the dependencies
for the target files, the information is generated from the compiler,
which is very useful. But Ninja has not open the API to query this database,
so that is the intention for this tool


Lekensteyn:
I was looking for this feature, but it seems missing. About the implementation, I think it should be added as `ninja -t deps-query` command. From there you can then inspect the internal database from C++.

PR #1210, opened 2016-11-21
Add support for external frontends

colincross:
There are many requests for customizations to Ninja's build output cataloged in #746.  This patch set attempts to address the underlying issue by spawning a separate frontend process and feeding serialized build status updates to it.  The frontend can be extended without complicating Ninja's core logic, or can be completely swapped out for a custom frontend like implemented in #1020.  The output can also be serialized to disk for later parsing, which will work well for buildbots.

The interface between Ninja and the frontend is documented in FRONTEND.md in 7th patch in the series, "Add frontend option for build status output"


frontend/FRONTEND.md
```diff
@@ -0,0 +1,54 @@
...
+----------------------------
+
+To run Ninja with a frontend that mimics the behavior of Ninja's normal output:
+```
+$ ./ninja --frontend=frontend/native.py
+```
+
+To save serialized output to a file:
+```
+$ ./ninja --frontend='cat /proc/self/fd/3 > ninja.pb all
```
danw: `cat <&3 >ninja.pb` would be more portable (at least Mac, which doesn't have /proc)

And you're missing a closing `'`

jhasse:
Running with `frontend/native.py` I'm getting:

```
  File "../../ninja-serialize/frontend/native.py", line 300, in <module>
    main()
  File "../../ninja-serialize/frontend/native.py", line 297, in main
    native.handle(msg)
  File "../../ninja-serialize/frontend/native.py", line 124, in handle
    if msg.message.level == msg.message.Level.INFO:
AttributeError: 'EnumTypeWrapper' object has no attribute 'INFO'
```

Anything I'm missing?

colincross:
Oops, I misunderstood the python protobuf class structure, combined with an untested path.  Pushed a quick fix for now.

jhasse:
Works now, thanks :)

frontend/frontend.py
```diff
@@ -0,0 +1,66 @@
...
+        return self
+
+    def __next__(self):
+        return self.next()
+
+    def next(self):
+        size = 0
+        shift = 0
+        while True:
+            byte = self.reader.read(1)
```
jhasse: Use `byte = bytearray(self.reader.read(1))` ...

frontend/frontend.py
```diff
@@ -0,0 +1,66 @@
...
+            if len(byte) == 0:
+                raise StopIteration()
+
+            byte = ord(byte[0])
```
jhasse: ... and remove `ord` call here: `byte = byte[0]`. It will then work in both Python 2 and 3.

frontend/frontend.py
```diff
@@ -0,0 +1,66 @@
...
+    """Generator class that parses length-delimited ninja status messages
+    through a ninja frontend interface.
+    """
+
+    def __init__(self, reader=default_reader()):
+        self.reader = reader
+        self.status_class = self.get_status_proto()
+
+    def get_status_proto(self):
+        set = google.protobuf.descriptor_pb2.FileDescriptorSet()
```
jhasse: `[pylint] W0622:Redefining built-in 'set'` Maybe fd_set instead?

frontend/frontend.py
```diff
@@ -0,0 +1,66 @@
...
+        descriptor = os.path.join(os.path.dirname(sys.argv[0]), 'frontend.pb')
```
jhasse: Not sure if `sys.argv[0]` is a good idea, maybe `__file__` instead?

frontend/frontend.py
```diff
@@ -0,0 +1,66 @@
...
+            size += (byte & 0x7f) << (shift * 7)
+            if (byte & 0x80) == 0:
+                break
+            shift += 1
+            if shift > 4:
+                raise "Expected varint32 length-delimeted message"
+
+        message = self.reader.read(size)
+        if len(message) != size:
+            raise "Unexpected EOF reading %d bytes" % size
```
jhasse: `raise EOFError("Unexpected EOF reading {} bytes".format(size))`

frontend/frontend.py
```diff
@@ -0,0 +1,66 @@
...
+            if len(byte) == 0:
+                raise StopIteration()
+
+            byte = ord(byte[0])
+            size += (byte & 0x7f) << (shift * 7)
+            if (byte & 0x80) == 0:
+                break
+            shift += 1
+            if shift > 4:
+                raise "Expected varint32 length-delimeted message"
```
jhasse: Instead of raising a str it's nicer to use something like `RuntimeError` (which inherits from `BaseException`).

jhasse:
I've used this PR to write a new frontend which adds some features I've missed: https://bixense.com/ja/ It works pretty good so far :)

frontend/FRONTEND.md
colincross: Thanks, fixed

frontend/FRONTEND.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+To run a frontend with serialized input from a file:
+```
+$ frontend/native.py 3< ninja.pb
+```
+
+The serialized output of a clean Ninja build is included in `frontend/ninja.pb`.
```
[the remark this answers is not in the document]
colincross: Removed, frontend/frontend.pb is the compiled protobuf descriptor, not serialized build output.

frontend/frontend.py
colincross: Thanks, fixed and tested with python3

frontend/frontend.py
colincross: Thanks, fixed

frontend/frontend.py
colincross: Fixed

frontend/frontend.py
colincross: Fixed, and all the other raised strings

frontend/frontend.py
colincross: Fixed

jhasse:
Is there anything I could do to help getting this merged?

Boddlnagg:
I would also really like to see this being merged. My goal is to have a frontend that logs everything to a file (similar to a simple stdout redirect with current ninja), but still provides status output. In this way, warnings and executed commands can be found in the logfile, while a status line (and output of failed commands) is shown directly on the console.

nico:
I thought a bit about this. As said on the mail thread, I'm not a fan of putting protobuf in ninja itself (even if it's a reimplementation). At the same time, I agree that lots of people want to tweak ninja's output in many ways, and this is a good approach to attack that problem.

What do you think about adding a second binary, `ninja_protoout` or `ninja_buildcore` or what have you, somewhere below contrib/, make that link to the ninja lib, put the proto code there, and make the ninja lib have an interface for selecting the output behavior (and core ninja would provide a class that works like the current output and the new binary would instead use the proto behavior)? That way, ninja itself stays the way it is, and people who want very custom output can use the new binary for that?

jhasse:
Doesn't ninja also stay the way it is when one isn't using the `--frontend` option?

nico:
The behavior stays the same, but the code still ends up in the binary.

jhasse:
IMHO that's an acceptable tradeoff. What binary size difference are we talking about here? Probably a few kilobytes.

> That way, ninja itself stays the way it is, and people who want very custom output can use the new binary for that?

So the binary gets distributed with ninja? That way the size of the *package* stays the same anyway.

If not, it would still be very inconvenient as frontends still need to ship a binary.

nico:
It's more about the conceptual "ninja doesn't depend on protobuf" imho.

Depends on what you mean by "distributed". The only distribution we do is the releases page on github; we can certainly include the binary there. All other distribution is up to the people owning the linux packages and whatnot.

jhasse:
> It's more about the conceptual "ninja doesn't depend on protobuf" imho.

Can you elaborate what you mean by this exactly? Would a compile time switch be satisfactory for example?

Btw: An argument why merging this might actually decrease `ninja`'s size in the long run: Lots of fancy interface stuff can be implemented by external frontends and doesn't need any code in ninja itself.

> All other distribution is up to the people owning the linux packages and whatnot.

Yeah that's what I meant, because I think that's what most Linux people use.

nico:
Ninja should stay small and simple. It should do builds, quickly, and not much else. Having a flexible cross-process communication protocol in there is philosophically out of scope for ninja itself.

colincross:
I'm happy to implement this if it is necessary to get this merged, but I agree with Jan that it doesn't make much sense vs. the extra complexity.  The proto stuff is all encapsulated through a subclass of the Status interface anyways (StatusSerializer in status.cc, which uses the generated frontend.pb.h file and the 264 line proto runtime in src/proto.cc and src/proto.h), so moving it to its own binary won't materially change the new code.

All the argument parsing logic and the main loop in ninja.cc would need to be in the new binary, which probably means moving ninja.cc into libninja and putting two new main.cc files that just configure the output class and call real_main().

As a measure of how big this patch is, for a local build before and after serialize, there is an 8% increase in text section size:
```
	   text	   data	    bss	    dec	    hex	filename
before:	 179193	   2312	    560	 182065	  2c731	ninja
after:	 193758	   2952	    600	 197310	  302be	ninja

```

nicolasdespres:
I think there is a simpler approach to customize ninja's output. By simpler I mean with a smaller footprint on Ninja's code base. I think I have mentioned it in the past but never got comment on it.

Following the same spirit than the NINJA_STATUS environment variable we could have an environment variable specifying the template to use when printing the start of an edge (NINJA_START_EDGE), when an edge stop (NINJA_STOP_EDGE). Users can use the protocol they like (json or ad-hoc for instance) when defining the value of these variables. Eventually, we will have to support some extra placeholders (like one printing the job id so that users can match edge stop line with edge start line. Some escaping function would maybe land in ninja code base too but that will be all. Users would finally just have to write a filter understanding their protocol and do the nice printing.

colincross:
It's certainly simpler, but I have a few concerns about using NINJA_START_EDGE/NINJA_STOP_EDGE instead of proper serialization.

- It only covers edges, this patch also covers messages from ninja and starting and ending the build.  Messages (warnings and errors) in particular would require frontends to parse various strings that ninja prints.  Protobuf also allows new messages to be added later (periodic messages containing system cpu usage or memory usage?  heartbeat?)
- There's no way to tell that edge output (which will follow a NINJA_STOP_EDGE) has finished until another edge has started or finished or the build has ended.  When using pools, or at the end of builds, there is no guarantee that an edge finish is immediately followed by an edge start, so this could be a long time.
- Printing things like the list of input/output targets will require a placeholder that produces an escaped list, which means there won't end up being that much variability in the output format.  For example, `%O` could produce: `"a.pb.cc", "a.pb.h"`.  At that point ninja might as well have just produced the whole output in JSON instead of forcing each frontend to manually smoosh NINJA_START_EDGE into something that looks like JSON.
- Separating the start/stop edge lines from edge output requires some sort of magic sequence that must never appear in an edge output, which seems hacky.

colincross:
Another use case covered by this pull request but not by NINJA_START_EDGE, and one of my main motivations for pursuing this pull request, is the ability to get ANSI escape codes (colorized logs) even when outputting to something that is not a smart terminal, like a file or a pipe.  I want to produce build breakage emails that retain the original colorized error messages, but ninja strips them.  This pull request puts all control of the output in the frontend, so ninja doesn't have to have an option for every different way someone wants the output to look.

nicolasdespres:
It is true that we cannot achieve the same level of versatility with two environment variables than with properly designed inter-process communication protocol.

* For warning/error messages, I think most IDEs already parse compiler warnings/errors without problem so we could follow this pseudo-standard (as with "Entering directory...")
* Edge's output should indeed be prefixed by some kind of edge id so that the front-end can relate one to another and enable stream-mode. In stream-mode ninja no longer buffers the whole command output but print each line prefixed by the edge id when it is received.
* People know their build system and the commands involved with the kind of output they produce. I see no problem in letting front-end designers deciding of the magic sequence to find the beginning/end of an edge output. Something like `__ninja_edge_start` will work.
* Placeholder expanded values could use the same syntax as in the ninja build file to express file list.
* Ninja should not be concerned by ANSI escape codes specially in such a batch mode (except for dependency extraction). IMHO, front-ends are in charge of parsing command's output. 

I think a compromise between versatility and code base impact should be found.

jhasse:
> I think most IDEs already parse compiler warnings/errors without problem

I can assure that they aren't doing this "without problems", I've had my own share of problems with that. More likely they are doing it because of no available alternative.

For example the Rust compiler added JSON output for IDEs, because the text parsing method was restricting them and IDE developers didn't like it either. Without it they wouldn't have been able to change the default output format for their error messages.

Keep that in mind: Frontends need a stable interface and if you let them use Ninja's normal output it will be harder to change that.

> Something like `__ninja_edge_start` will work.

What if the output of an edge contains that, too? Might happen when you recursively call Ninja.

nicolasdespres:
@jhasse
 
Emacs and others finds error/warning messages. It is not pretty and probably hacky but it works 90% of the time. That's what I meant when I said "without problems".

It is extremely unlikely that `__ninja_edge_start` is included in the output and ninja does not support recursive mode. You could add the recursion level if needed anyway.

To all:

Please do not take me wrong. I am fully aware of the benefit of well defined interface and I generally prefer principled approaches. But this pull-request diffstat is `+2,743 −400` and @nico's last comment suggests that it won't be merged unless a compromise is found. Ninja covers only 80% of build-system use cases. It is a design choice. Maintainers do not seem inclined to invest in the effort required to support the remaining 20% which may probably double the code base size. I think the same philosophy applies in the case of the front-end.

Boddlnagg:
I already stated my use case for this in https://github.com/ninja-build/ninja/pull/1210#issuecomment-375633679, and I'd like to add that this does not necessarily require a fully customizable frontend. It would be enough if ninja would (optionally) print information about started jobs in addition to finished jobs, so when I pipe the output in a different process ("custom frontend"), I can still show a status line about the currently running unfinished job(s).

Btw, this PR should be tagged with the `frontend` label.

jonesmz:
@jhasse Consider closing this PR in favor of the rebased one here: #1600

PR #1211, opened 2016-12-02
Amalgamate sources into a single .cc file when bootstrapping

haberman:
This PR needs some work still, but I wanted to get feedback before polishing/fixing it.

The ultimate goal here is to lower the barrier to entry to using Ninja. You should be able to just drop configure.py and ninja-bootstrap.cc into your project, and now your project can use Ninja without worrying about whether end-users have installed Ninja or not (or have the right version). You bootstrap in this case using simply: `./configure.py --bootstrap_from_amalgamation`.

To reduce the number of code-paths/variations, I made regular bootstrapping build from the amalgamation also. And to make things more principled/consistent, `ninja.bootstrap` is always built from `ninja-bootstrap.cc` and `ninja` is always built in the regular way, instead of spawning a ninja that will attempt to overwrite itself.

As a side benefit, this makes bootstrapping faster (since it's only compiling one file):

```
Right now:
$ time ./configure.py --bootstrap
bootstrapping ninja...
wrote build.ninja.
bootstrap complete.  rebuilding...
[27/27] LINK ninja

real    0m10.444s
user    0m20.639s
sys     0m1.579s

With my change:
$ time ./configure.py --bootstrap
bootstrapping ninja...
wrote build.ninja.
bootstrap complete.  rebuilding...
[26/26] LINK ninja

real    0m6.364s
user    0m16.623s
sys     0m1.148s
```

The `--bootstrap_from_amalgamation` flag was added for the case where you want to bootstrap directly from the amalgamation (the case I described above where you want to use Ninja in your project).

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -157,7 +157,9 @@ class Bootstrap:
         for key, val in kwargs.get('variables', []):
             local_vars[key] = ' '.join(ninja_syntax.as_list(val))
 
-        self._run_command(self._expand(cmd, local_vars))
+        if (outputs == binary('ninja.bootstrap') or
+            not options.bootstrap_from_amalgamation):
+            self._run_command(self._expand(cmd, local_vars))
 
         return self.writer.build(outputs, rule, inputs, **kwargs)
 
@@ -192,6 +194,8 @@ parser = OptionParser()
 profilers = ['gmon', 'pprof']
 parser.add_option('--bootstrap', action='store_true',
                   help='bootstrap a ninja binary from nothing')
+parser.add_option('--bootstrap_from_amalgamation', action='store_true',
+                  help='bootstrap a ninja binary from the amalgamation')
 parser.add_option('--verbose', action='store_true',
                   help='enable verbose build')
 parser.add_option('--platform',
@@ -219,6 +223,9 @@ if args:
     print('ERROR: extra unparsed command-line arguments:', args)
     sys.exit(1)
 
+if options.bootstrap_from_amalgamation:
+    options.bootstrap = True
+
 platform = Platform(options.platform)
 if options.host:
     host = Platform(options.host)
@@ -230,11 +237,12 @@ ninja_writer = ninja_syntax.Writer(open(BUILD_FILENAME, 'w'))
 n = ninja_writer
 
 if options.bootstrap:
-    # Make the build directory.
-    try:
-        os.mkdir('build')
-    except OSError:
-        pass
+    if not options.bootstrap_from_amalgamation:
+      # Make the build directory.
+      try:
+          os.mkdir('build')
+      except OSError:
+          pass
     # Wrap ninja_writer with the Bootstrapper, which also executes the
     # commands.
     print('bootstrapping ninja...')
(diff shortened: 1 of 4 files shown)


src/build_log.cc
```diff
@@ -45,9 +45,9 @@
 
 namespace {
 
-const char kFileSignature[] = "# ninja log v%d\n";
+const char kLogFileSignature[] = "# ninja log v%d\n";
 const int kOldestSupportedVersion = 4;
-const int kCurrentVersion = 5;
+const int kLogCurrentVersion = 5;
```
evmar: These changes seem unrelated to the intended change(?)

configure.py
```diff
@@ -192,6 +194,8 @@ def _run_command(self, cmdline):
 profilers = ['gmon', 'pprof']
 parser.add_option('--bootstrap', action='store_true',
                   help='bootstrap a ninja binary from nothing')
+parser.add_option('--bootstrap_from_amalgamation', action='store_true',
+                  help='bootstrap a ninja binary from the amalgamation')
```
evmar: I think if we think amalgamation is a good idea, we should just always do it and not put it behind an option.  Or decide to never do it for bootstrap.  But either way, no option.

configure.py
```diff
@@ -449,8 +482,8 @@ def shell_escape(str):
             variables=[('varname', 'kBrowsePy')])
     n.newline()
 
-    objs += cxx('browse', order_only=built('browse_py.h'))
-    n.newline()
+    srcs += [cxxsrc('browse')]
+    order_only_deps['browse'] = built('browse_py.h')
```
evmar: Another idea is to just drop this when doing amalgamation, since if the user can't be bothered to install ninja they certainly can't be bothered to install python.

src/build_log.cc
haberman: They are required because otherwise the amalgamation contains multiple conflicting definitions of `kFileSignature` and `kCurrentVersion`.

configure.py
haberman: I think the option is required, unless you are ok with checking in `ninja-bootstrap.cc` in to this repo and updating it on every checkin.

The two cases are:

1. in the Ninja repo, there is no `ninja-bootstrap.cc` checked in, so `--bootstrap` will build `ninja-bootstrap.cc` from the other source files before building `ninja.bootstrap`. In this case, `ninja-bootstrap.cc` is considered to be an artifact.

2. in other repos, people will check in *only* `ninja-bootstrap.cc` and no other sources, so attempting to build `ninja-bootstrap.cc` will fail since the other sources aren't present. So users will pass `--bootstrap_from_amalgamation` because it's bootstrapping *directly* from the amalgamation, without trying to build the amalgamation first. In this case, `ninja-bootstrap.cc` is considered authoritative.

If you wanted to avoid the two cases (and the option), you would need to check `ninja-bootstrap.cc` into this repo and make sure it is updated every time someone makes a change to one of the other source files.

configure.py
haberman: Possibly, but if we always use the amalgamation to bootstrap, it seems simpler to not special-case the Python support.

haberman:
I fixed the Python indentation.

haberman:
PTAL -- let me know what you think about the flag.

If you prefer to check in ninja-bootstrap.cc and update it on every checkin, I can try to make that work. It seems like it would be more fiddly though, since `ninja-bootstrap.cc` would be both a source and an artifact within the Ninja repo. You'd probably want an explicit "regenerate" operation to update it, and a Travis test that it is up-to-date.

haberman:
FYI the main thing I still have to fix here is: the amalgamation needs to include all `.cc` files for all platforms, since the amalgamation should build on all platforms without needing to regenerate it. This will take a little bit of `#ifdef`, but I think it should be pretty straightforward.  If you are ok with the approach, I'll make it robust across platforms.

nico:
FWIW I don't understand the motivation for this change. It seems to make things more complicated in more ways than it makes things simpler.

haberman:
I tried to explain the motivation of this change in my first message. I do personally think this can simplify things a bit, but that is not the primary goal. The primary goal is to make Ninja easier to embed directly into source trees so the user doesn't need to install Ninja themselves manually.

If the change doesn't make sense to you though, I'll just close the PR and use Ninja as a Git submodule through the regular bootstrapping process.

haberman:
(Though perhaps I would create a repo that also has a statically-build Windows binary to cover that case, since Python is not preinstalled on that platform).

PR #1215, opened 2016-12-09
add verbosity to environment of created processes

vvvrrooomm:
cmake automoc needs VERBOSE=1 in env to print its commands:

cmake generates rules of the form "cd QT && cmake -E cmake_autogen project Debug"
This call to cmake executes the QT moc and uic and rcc compilers which generate files. If any of those calls fail it is impossible to see the actual command lines even if ninja is run with "-v". cmake prints the commands if the environment variable VERBOSE is set - so add this to the current environment.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -537,7 +537,7 @@ bool RealCommandRunner::CanRunMore() {
 
 bool RealCommandRunner::StartCommand(Edge* edge) {
   string command = edge->EvaluateCommand();
-  Subprocess* subproc = subprocs_.Add(command, edge->use_console());
+  Subprocess* subproc = subprocs_.Add(command, config_.verbosity, edge->use_console());
   if (!subproc)
     return false;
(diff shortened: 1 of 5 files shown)


evmar:
This feels kinda cmake-specific.  Why not just
  VERBOSE=1 ninja ...
?

vvvrrooomm:
VERBOSE=1 ninja  would only put the cmake_generator part into debug mode, not ninja itself, essentially it leads to to mechanisms to turn on verbose output and a user would need to  "VERBOSE ninja -v" to see it all.

With ninja -v the user excpects to see all executed commands, so additionally to the proposed patch the ninja documentation should state that all commands should honor the VERBOSE environment if called from ninja.

jhasse:
Thanks for the PR, but I still think this is too CMake-specific, sorry. IMHO the correct way to fix this, is to further the integration with `cmake --build`.

PR #1217, opened 2016-12-21
Handle insecure string pickle issue, fixes #1216

### LINKED ISSUE #1216: Error bootstrapping Ninja: insecure string pickle
An unhandled exception is preventing me from bootstrapping Ninja from source. I get the following stack trace when I run `CXX=clang++ ./configure.py --bootstrap` on macOS 10.12.2 Sierra:

```python
bootstrapping ninja...
  File "./configure.py", line 462, in <module>
    if has_re2c():
  File "./configure.py", line 458, in has_re2c
    proc = subprocess.Popen(['re2c', '-V'], stdout=subprocess.PIPE)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 1334, in _execute_child
    child_exception = pickle.loads(data)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pickle.py", line 1382, in loads
    return Unpickler(file).load()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pickle.py", line 858, in load
    dispatch[key](self)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pickle.py", line 966, in load_string
    raise ValueError, "insecure string pickle"
ValueError: insecure string pickle
```

I did make an effort to see if this issue had been reported before, but I didn't see anything.

I'm not a Python person, so I'm using the stock Python on macOS Sierra, which reports itself as being version 12.7.10. I did a little research on `insecure string pickle` and discovered that this is a [known issue](http://bugs.python.org/issue26083) involving the `subprocess` module. Apparently, this issue is fixed in Python 2.7.12, which has not been incorporated into macOS yet. As a result, I can't build Ninja from source as is. The brute force fix would be installing another Python, but that is something I would rather avoid, if possible.

yesmar:
I patched `./configure.py` to catch `ValueError` and silently fail. This results in `./configure.py` reporting `warning: A compatible version of re2c (>= 0.11.3) was not found; changes to src/*.in.cc will not affect your build.` Although not perfect, this fix at least lets people build Ninja from source instead of being stumped by a bug in Python 2.7. It doesn't help people that want to make changes to the code in `src/`, but at least they get this error and know that some additional work needs to be done.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -457,6 +457,8 @@ def has_re2c():
     try:
         proc = subprocess.Popen(['re2c', '-V'], stdout=subprocess.PIPE)
         return int(proc.communicate()[0], 10) >= 1103
+    except ValueError:
+        return False
     except OSError:
         return False
 if has_re2c():


jhasse:
Will configure.py also report this if re2c is on the PATH?

jhasse:
GIven that Python 2 is now EOL and we also support building with CMake, I'll close this. Feel free to reopen if you disagree.

PR #1219, opened 2016-12-31
Use high-resolution timestamps

### LINKED ISSUE #371: Use sub-1s resolution in timestamps
There's a pull request here that implements it:
https://github.com/martine/ninja/pull/337

However, the timing changes are mixed in with other unrelated changes, so finishing this bug will involve extracting that logic.

QuLogic:
I tried to follow #337, but it was too tangled to understand and just wrote it from scratch. Only the other-OS struct fields from there were used. #337 also tried to increase the resolution of the clocks for timing; I didn't bother with this change here.

Fixes #371.

### DIFF
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -290,7 +290,7 @@ bool BuildLog::Load(const string& path, string* err) {
     if (!end)
       continue;
     *end = 0;
-    restat_mtime = atol(start);
+    restat_mtime = strtoll(start, NULL, 10);
     start = end + 1;
 
     end = (char*)memchr(start, kFieldSeparator, line_end - start);
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -353,7 +353,7 @@ BuildLog::LogEntry* BuildLog::LookupByOutput(const string& path) {
 }
 
 bool BuildLog::WriteEntry(FILE* f, const LogEntry& entry) {
-  return fprintf(f, "%d\t%d\t%d\t%s\t%" PRIx64 "\n",
+  return fprintf(f, "%d\t%d\t%" PRId64 "\t%s\t%" PRIx64 "\n",
           entry.start_time, entry.end_time, entry.mtime,
           entry.output.c_str(), entry.command_hash) > 0;
 }
--- a/src/deps_log.cc
+++ b/src/deps_log.cc
@@ -30,7 +30,7 @@
 // The version is stored as 4 bytes after the signature and also serves as a
 // byte order mark. Signature and version combined are 16 bytes long.
 const char kFileSignature[] = "# ninjadeps\n";
-const int kCurrentVersion = 3;
+const int kCurrentVersion = 4;
 
 // Record size is currently limited to less than the full 32 bit, due to
 // internal buffers having to have this size.
@@ -124,7 +124,7 @@ bool DepsLog::RecordDeps(Node* node, TimeStamp mtime,
     return true;
 
   // Update on-disk representation.
-  unsigned size = 4 * (1 + 1 + node_count);
+  unsigned size = 4 * (1 + 2 + node_count);
   if (size > kMaxRecordSize) {
     errno = ERANGE;
     return false;
@@ -135,8 +135,7 @@ bool DepsLog::RecordDeps(Node* node, TimeStamp mtime,
   int id = node->id();
   if (fwrite(&id, 4, 1, file_) < 1)
     return false;
-  int timestamp = mtime;
-  if (fwrite(&timestamp, 4, 1, file_) < 1)
+  if (fwrite(&mtime, 8, 1, file_) < 1)
     return false;
   for (int i = 0; i < node_count; ++i) {
     id = nodes[i]->id();
@@ -218,9 +217,9 @@ bool DepsLog::Load(const string& path, State* state, string* err) {
       assert(size % 4 == 0);
       int* deps_data = reinterpret_cast<int*>(buf);
       int out_id = deps_data[0];
-      int mtime = deps_data[1];
-      deps_data += 2;
-      int deps_count = (size / 4) - 2;
+      TimeStamp mtime = reinterpret_cast<TimeStamp*>(&deps_data[1])[0];
+      deps_data += 3;
+      int deps_count = (size / 4) - 3;
 
       Deps* deps = new Deps(mtime, deps_count);
       for (int i = 0; i < deps_count; ++i) {
--- a/src/deps_log.cc
+++ b/src/deps_log.cc
@@ -135,7 +135,11 @@ bool DepsLog::RecordDeps(Node* node, TimeStamp mtime,
   int id = node->id();
   if (fwrite(&id, 4, 1, file_) < 1)
     return false;
-  if (fwrite(&mtime, 8, 1, file_) < 1)
+  uint32_t mtime_part = static_cast<uint32_t>(mtime & 0xffffffff);
+  if (fwrite(&mtime_part, 4, 1, file_) < 1)
+    return false;
+  mtime_part = static_cast<uint32_t>((mtime >> 32) & 0xffffffff);
+  if (fwrite(&mtime_part, 4, 1, file_) < 1)
     return false;
   for (int i = 0; i < node_count; ++i) {
     id = nodes[i]->id();
@@ -217,7 +221,9 @@ bool DepsLog::Load(const string& path, State* state, string* err) {
       assert(size % 4 == 0);
       int* deps_data = reinterpret_cast<int*>(buf);
       int out_id = deps_data[0];
-      TimeStamp mtime = reinterpret_cast<TimeStamp*>(&deps_data[1])[0];
+      TimeStamp mtime;
+      mtime = (TimeStamp)(((uint64_t)(unsigned int)deps_data[2] << 32) |
+                          (uint64_t)(unsigned int)deps_data[1]);
       deps_data += 3;
       int deps_count = (size / 4) - 3;
 
--- a/src/deps_log.h
+++ b/src/deps_log.h
@@ -57,8 +57,9 @@ struct State;
 ///      one's complement of the expected index of the record (to detect
 ///      concurrent writes of multiple ninja processes to the log).
 ///    dependency records are an array of 4-byte integers
-///      [output path id, output path mtime (8-byte int), input path id,
-///       input path id...]
+///      [output path id,
+///       output path mtime (lower 4 bytes), output path mtime (upper 8 bytes),
+///       input path id, input path id...]
 ///      (The mtime is compared against the on-disk output path mtime
 ///      to verify the stored data is up-to-date.)
 /// If two records reference the same output the latter one in the file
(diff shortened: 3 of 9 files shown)


src/deps_log.h
```diff
@@ -75,10 +76,10 @@ struct DepsLog {
 
   // Reading (startup-time) interface.
   struct Deps {
-    Deps(int mtime, int node_count)
+    Deps(int64_t mtime, int node_count)
         : mtime(mtime), node_count(node_count), nodes(new Node*[node_count]) {}
     ~Deps() { delete [] nodes; }
-    int mtime;
+    int64_t mtime;
```
evmar: I guess this ought to use the TimeStamp type?

src/disk_interface.cc
```diff
@@ -60,11 +60,11 @@ int MakeDir(const string& path) {
 TimeStamp TimeStampFromFileTime(const FILETIME& filetime) {
   // FILETIME is in 100-nanosecond increments since the Windows epoch.
   // We don't much care about epoch correctness but we do want the
-  // resulting value to fit in an integer.
+  // resulting value to fit in a 64-bit integer.
```
evmar: https://msdn.microsoft.com/en-us/library/windows/desktop/ms724284(v=vs.85).aspx
indicates that perhaps the raw time is just a 64-bit integer, so perhaps we could just use the time directly.  Another way of saying this is that we could drop lines 66 and 67 I think (and then adjust the relevant comments).

Oh this doc is better:
https://msdn.microsoft.com/en-us/library/windows/desktop/ms724290(v=vs.85).aspx

src/timestamp.h
```diff
@@ -17,8 +17,8 @@
 
 // When considering file modification times we only care to compare
 // them against one another -- we never convert them to an absolute
-// real time.  On POSIX we use time_t (seconds since epoch) and on
-// Windows we use a different value.  Both fit in an int.
-typedef int TimeStamp;
+// real time.  On POSIX we use timespec (seconds&nanoseconds since epoch)
+// and on Windows we use a different value.  Both fit in an int64.
```
evmar: (...with the Windows change, this would just say something like "File Time":
https://msdn.microsoft.com/en-us/library/windows/desktop/ms724290(v=vs.85).aspx
)

src/util.h
```diff
@@ -18,6 +18,7 @@
 #ifdef _WIN32
 #include "win32port.h"
 #else
+#include <inttypes.h>
```
evmar: What is this for?

src/deps_log.cc
```diff
@@ -218,9 +217,9 @@ bool DepsLog::Load(const string& path, State* state, string* err) {
       assert(size % 4 == 0);
       int* deps_data = reinterpret_cast<int*>(buf);
       int out_id = deps_data[0];
-      int mtime = deps_data[1];
-      deps_data += 2;
-      int deps_count = (size / 4) - 2;
+      int64_t mtime = reinterpret_cast<int64_t*>(&deps_data[1])[0];
```
evmar: Are there potential alignment issues here?

src/disk_interface.cc
QuLogic: According to the first link, you shouldn't treat `FILETIME` as a 64-bit integer directly because of possible alignment issues.

src/deps_log.cc
QuLogic: Probably, I was trying to avoid thinking about endianness, but I'll have to see if I can implement this in some way that's not annoying.

src/util.h
QuLogic: `win32port.h` defines `int64_t`, etc., and related printf-specifiers. This header defines the print specifiers on non-Windows. However, based on the CI failing, I probably need to define something to enable those on older systems, or since this is C++ code, maybe try `cinttypes`.

src/deps_log.cc
```diff
@@ -174,14 +173,24 @@ bool DepsLog::Load(const string& path, State* state, string* err) {
...
   bool valid_header = true;
   int version = 0;
-  if (!fgets(buf, sizeof(buf), f) || fread(&version, 4, 1, f) < 1)
+  if (!fgets(buf, sizeof(buf), f)) {
     valid_header = false;
+  } else if (strcmp(buf, kFileSignature) != 0) {
+    valid_header = false;
+  } else if (fread(&version, 4, 1, f) < 1) {
+    valid_header = false;
+  } else if (version != kCurrentVersion) {
```
QuLogic: BTW, I can attempt to read version 3 if wanted; it's not a major change.

QuLogic:
Ping?

src/deps_log.cc
```diff
@@ -174,14 +173,24 @@ bool DepsLog::Load(const string& path, State* state, string* err) {
...
+    valid_header = false;
+  } else if (fseek(f, -4, SEEK_CUR) != 0) {
+    valid_header = false;
+  } else if (fread(&buf, 1, 4, f) < 4) {
+    valid_header = false;
+  }
```
maximuska: IMHO, the original code was slightly more readable, I wouldn't change it (&see below the question about endianess handling).

src/deps_log.cc
```diff
@@ -218,9 +228,16 @@ bool DepsLog::Load(const string& path, State* state, string* err) {
...
-      deps_data += 2;
-      int deps_count = (size / 4) - 2;
+      TimeStamp mtime;
+      if (little_endian) {
+        mtime = (TimeStamp)((uint64_t)deps_data[2] << 32 |
+                            (uint64_t)deps_data[1]);
+      } else {
+        mtime = (TimeStamp)((uint64_t)deps_data[1] << 32 |
+                            (uint64_t)deps_data[2]);
+      }
```
maximuska: Wouldn't casting to uint64_t* and deference, instead of having to handle endianess be good enough? Does anyone really care builds on platforms with misalignment access issues?

Also, an alternative could be writing mtime in two parts (low32, high32), so you could combine these in the same way w/o considering platform endianess.

maximuska:
+1, I've just hit timestamp issue in my environment after adding some automatic tests to our metabuild system, could be nice to put it behind w/o adding sleeps.

src/deps_log.cc
QuLogic: The Windows link above says not to do that because of alignment problems; I think that's a large enough platform to worry about.

maximuska:
@QuLogic, I believe MSDN guys had itanium (IA64) hardware in mind when writing the article.

x86/amd64 hardware handles misaligned access w/o involving OS, even performance impact vanished with the years, see e.g., [here](http://lemire.me/blog/2012/05/31/data-alignment-for-speed-myth-or-reality/).

It would be nice to use a wrapper macro `getUint64(ptr)` for misaligned access, which can expand to a simple cast & dereference on x86 hardware.

fooishbar:
@maximuska ARM (unsure about v8, but certainly v5-7) requires natural alignment, or an OS exception handle and fixup.

maximuska:
Yeah, I know.
Anyone here uses ARM build servers?

fooishbar:
@maximuska Yes, I do, partly because for smaller projects at least, Ninja-backed build systems can make it quicker to build on-device (or on an ARM build server; there are many vendors shipping blade-style servers) than to set up cross-compiling.

Maratyszcza:
@fooishbar ARM doesn't require natural alignment since ARMv6. `The ARMv6 architecture introduced the first hardware support for unaligned accesses.  ARM11 and Cortex-A/R processors can deal with unaligned accesses in hardware, removing the need for software routines.` [source](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.faqs/ka15414.html)

QuLogic:
I rebased this and changed it up a bit to use (low, high) instead of dealing with endian issues in the old manner.

QuLogic:
Turns out this delay was somewhat useful because I manager to find a bug in re-reading the saved timestamp.

Apparently, casting from a signed narrow type to an unsigned wide type does sign extension, producing a completely incorrect result out of the lower 32 bits, but it's kind of random because it depends on the 32nd bit.

nico:
Thanks for the patch! Sorry about the slow turnaround time.

Generally looks great.

Did you check how this affects empty build time? The larger log size might impact incremental build speeds, it'd be good to know by how much.

nirbheek:
As an aside, Meson has a PR that runs this patch on our testsuite on macOS, Windows (mingw, cygwin, native), and Linux (Ubuntu): https://github.com/mesonbuild/meson/pull/1931. Also contains a test for the high-res changes. It's running a full build on all platforms right now.

QuLogic:
I rebased this to fix the conflicts, and modified some bits so that it works correctly on Windows, Cygwin, and MinGW.

nirbheek:
Small caveat about macOS: if you're using HFS+, these patches will do nothing because the filesystem timestamp resolution is 1s. However, I tested on AppleFS and the patches work perfectly.

@nico 
> Did you check how this affects empty build time? The larger log size might impact incremental build speeds, it'd be good to know by how much.

So I tested full, incremental, and empty builds on Linux (Fedora 26, 64-bit) with a rotating disk and ext4 `relatime,data=ordered`

I compared ninja before and after these patches on [gst-build](https://cgit.freedesktop.org/gstreamer/gst-build/), which is a huge mostly-C project with ~3000 targets and a 7.3MB `build.ninja` file.

1. The log file size increased from 312K to 360K with these patches.

2. Hot cache:

    Full builds, incremental builds, and empty builds took the same time for both.

3. Cold cache `echo 2 > /proc/sys/vm/drop_caches` empty builds:

    I was getting strange results (it was faster with the patches than without), and after doing a bunch of testing and statistical analysis, I figured out that it was all noise. Both take the same time.

So all in all, these patches have negligible to no effect on the build times.

QuLogic:
This is rebased and I think I fixed the issue on Windows with the last commit. @nirbheek will try it out on the Meson test suite when he has a chance.

ClausKlein:
what is about merging to next?

QuLogic:
We are now using this patch for testing downstream: mesonbuild/meson#2328.

src/build_log.cc
```diff
@@ -353,7 +353,7 @@ BuildLog::LogEntry* BuildLog::LookupByOutput(const string& path) {
 }
 
 bool BuildLog::WriteEntry(FILE* f, const LogEntry& entry) {
-  return fprintf(f, "%d\t%d\t%d\t%s\t%" PRIx64 "\n",
+  return fprintf(f, "%d\t%d\t%" PRId64 "\t%s\t%" PRIx64 "\n",
```
nico: Unrelated to this change, but older ninjas will be pretty confused if they read build logs created by the new ninja. This isn't a new problem, but the `if (log_version < kOldestSupportedVersion` above should probably use `!=` instead. Not in this change, mostly talking to myself.)

nico:
Alright, it's early in the cycle, so let's merge this and see what (if anything) breaks :-) Do you want to send a short note to the list asking people to test this out? (If not, I can do it.)

nico:
...and thanks everyone who contributed to this patch and helped with measuring performance. Much appreciated, and thanks for your patience.

PR #1220, opened 2017-01-02
Fix build with uclibc

### LINKED ISSUE #985: Build fails with uclibc: src/util.cc:426:28: error: 'getloadavg' was not declared in this scope
Hi,

The ninja builds fails on Linux systems that use uclibc as the C standard library rather than glibc: 

i686-gentoo-linux-uclibc-g++ -O2 -march=i686 -pipe -Wno-deprecated -DNINJA_PYTHON="/usr/bin/python2.7" -DNI/graphviz.cc src/manifest_parser.cc src/eval_env.cc src/line_printer.cc src/subprocess-posix.cc src/debug_fc src/lexer.cc -o ninja.bootstrap
src/util.cc: In function 'double GetLoadAverage()':
src/util.cc:426:28: error: 'getloadavg' was not declared in this scope
   if (getloadavg(loadavg, 3) < 0) {
                            ^

phajdan:
Resolves #985

This is based on musl implementation,
http://git.musl-libc.org/cgit/musl/commit/?id=20cbd607759038dca57f84ef7e7b5d44a3088574
(thanks to jbergstroem@ for reference)

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -585,6 +585,13 @@ double GetLoadAverage() {
   // Calculation taken from comment in libperfstats.h
   return double(cpu_stats.loadavg[0]) / double(1 << SBITS);
 }
+#elif defined(__UCLIBC__)
+double GetLoadAverage() {
+  struct sysinfo si;
+  if (sysinfo(&si) != 0)
+    return -0.0f;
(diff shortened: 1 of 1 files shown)


nico:
Thanks!

PR #1223, opened 2017-01-09
[compdb] Expand response files inline based on a switch.

asankah:
References to response files in a clang compile_commands.json file can
be tricky to deal with when tooling expects all the command flags to be
present in the 'command' field.

This change introduces a '-x' option to '-t compdb' that will expand
@rspfile style response file invocations inline.

E.g.

```sh
$ ninja -t compdb cc
[
  {
    "directory": "/src/foo",
    "command": "cc -foo -bar @foo.obj.rsp",
    "file": "foo.cc"
  }
]

$ ninja -t compdb -x cc
[
  {
    "directory": "/src/foo",
    "command": "cc -foo -bar foo.cc",
    "file": "foo.cc"
  }
]
```

Does this change seem reasonable?

I tested this manually. The tools are all implemented in `ninja.cc` and none of them seem to have tests. I'd need to refactor that file to expose the tooling for tests. I wanted to check with folks on whether this is a reasonable change before I went ahead and did the necessary reworking of ninja.cc to expose tools for tests.

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -697,9 +697,6 @@ int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
   argc++;
   argv--;
 
-  bool first = true;
-  vector<char> cwd;
-
   EvaluateCommandMode eval_mode = ECM_NORMAL;
 
   optind = 1;
@@ -723,6 +720,9 @@ int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
   }
   argv += optind;
   argc -= optind;
+
+  bool first = true;
+  vector<char> cwd;
 
   do {
     cwd.resize(cwd.size() + 1024);
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -662,9 +662,67 @@ void EncodeJSONString(const char *str) {
   }
 }
 
-int NinjaMain::ToolCompilationDatabase(const Options* options, int argc, char* argv[]) {
+enum EvaluateCommandMode {
+  ECM_NORMAL,
+  ECM_EXPAND_RSPFILE
+};
+string EvaluateCommandWithRspfile(Edge* edge, EvaluateCommandMode mode) {
+  string command = edge->EvaluateCommand();
+  if (mode == ECM_NORMAL)
+    return command;
+
+  string rspfile = edge->GetUnescapedRspfile();
+  if (rspfile.empty())
+    return command;
+
+  size_t index = command.find(rspfile);
+  if (index == 0 || index == string::npos || command[index - 1] != '@')
+    return command;
+
+  string rspfile_content = edge->GetBinding("rspfile_content");
+  size_t newline_index = 0;
+  while ((newline_index = rspfile_content.find('\n', newline_index)) !=
+         string::npos) {
+    rspfile_content.replace(newline_index, 1, 1, ' ');
+    ++newline_index;
+  }
+  command.replace(index - 1, rspfile.length() + 1, rspfile_content);
+  return command;
+}
+
+int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
+                                       char* argv[]) {
+  // The compdb tool uses getopt, and expects argv[0] to contain the name of
+  // the tool, i.e. "compdb".
+  argc++;
+  argv--;
+
   bool first = true;
   vector<char> cwd;
+
+  EvaluateCommandMode eval_mode = ECM_NORMAL;
+
+  optind = 1;
+  int opt;
+  while ((opt = getopt(argc, argv, const_cast<char*>("hx"))) != -1) {
+    switch(opt) {
+      case 'x':
+        eval_mode = ECM_EXPAND_RSPFILE;
+        break;
+
+      case 'h':
+      default:
+        printf(
+            "usage: ninja -t compdb [options] [rules]\n"
+            "\n"
+            "options:\n"
+            "  -x     expand @rspfile style response file invocations\n"
+            );
+        return 1;
+    }
+  }
+  argv += optind;
+  argc -= optind;
 
   do {
     cwd.resize(cwd.size() + 1024);
@@ -688,7 +746,7 @@ int NinjaMain::ToolCompilationDatabase(const Options* options, int argc, char* a
         printf("\n  {\n    \"directory\": \"");
         EncodeJSONString(&cwd[0]);
         printf("\",\n    \"command\": \"");
-        EncodeJSONString((*e)->EvaluateCommand().c_str());
+        EncodeJSONString(EvaluateCommandWithRspfile(*e, eval_mode).c_str());
         printf("\",\n    \"file\": \"");
         EncodeJSONString((*e)->inputs_[0]->path().c_str());
         printf("\",\n    \"output\": \"");


evmar:
The reason for the rspfile in the first place is when command lines exceed OS limits.  Is it possible to adjust your .ninja generator to not use them in cases like these where you believe you can use the command line without an rspfile?

asankah:
Those were just examples :) The use case here is to make the `compile_commands.json` file more readily consumable.

As you noted, the response files are only useful as a workaround for shell command line restrictions. These limits don't translate to the use cases for a compilation database. The clang tooling, for example, parses the command line directly rather than pass it through the shell.

A more real world example would be Chromium. On Windows we make heavy use of response files in `clang-cl` and `cl` invocations. A typical command line looks like:

```
C:\goma/gomacc.exe ../../third_party/llvm-build/Release+Asserts/bin/clang-cl.exe /nologo /showIncludes /FC @obj/third_party/protobuf/protoc_lib/cpp_field.obj.rsp /c ../../third_party/protobuf/src/google/protobuf/compiler/cpp/cpp_field.cc /Foobj/third_party/protobuf/protoc_lib/cpp_field.obj /Fd"obj/third_party/protobuf/protoc_lib_cc.pdb"
```

Let's say we need to consume the compilation database for semantic code completion or source indexing, then we'd need to always run a full build and retain the response files. All this despite the fact that the full command line was always available to `ninja` the whole time. It would be much easier if we had something like the `-x` option to expand the full command line.

Leandros:
I second this, this is super useful!

src/ninja.cc
```diff
@@ -650,10 +650,68 @@ void EncodeJSONString(const char *str) {
...
+  argv--;
+
   bool first = true;
   vector<char> cwd;
 
+  EvaluateCommandMode eval_mode = ECM_NORMAL;
+
+  optind = 1;
+  int opt;
+  while ((opt = getopt(argc, argv, const_cast<char*>("hx"))) != -1) {
```
jhasse: According to https://linux.die.net/man/3/getopt getopt also accepts `const char*`. Am I missing something?

src/ninja.cc
```diff
@@ -650,10 +650,68 @@ void EncodeJSONString(const char *str) {
...
+
+int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
+                                       char* argv[]) {
+  // The compdb tool uses getopt, and expects argv[0] to contain the name of
+  // the tool, i.e. "compdb".
+  argc++;
+  argv--;
+
   bool first = true;
   vector<char> cwd;
```
jhasse: I would move these declarations to the do loop, as they are only needed there.

nico:
@asankah On Windows, ninja doesn't pass commands through the shell. So if an rsp file is actually needed, it will be needed for clang tooling too. In that case, building once with `-n -d keeprsp` to generate and keep the rspfiles is actually necessary.

In practice, I can believe that generators often ask for rsp files when they aren't actually needed. So in that case, this option might be useful.

I wonder what exactly fails if we tweak the Chromium build to not use rspfiles for compiles.

nico:
(Can you reply to @jhasse's questions?)

jhasse:
> So if an rsp file is actually needed, it will be needed for clang tooling too.

Not necessarily, as some tooling (e.g. clang-tidy or a language server like [cquery](https://github.com/cquery-project/cquery)) won't actually execute any command, but parse the flags directly from the JSON to get a list of flags and include directories.

asankah:
Sorry, this fell through the cracks. I'll address @jhasse 's comments in a new patchset shortly.

nico:
Re Chromium: Apparently I wondered why we use rspfiles for compiles before in https://chromium-review.googlesource.com/c/chromium/src/+/832593#message-0406d8fd9edea87bf503144ce90c918968b26e9c and the answer was "no good reason" and we removed that. So chromium compdbs no longer use rspfiles for compiles.

The patch is still useful for other projects, and the bar for `-t` tools is lower, so I'm fine merging this. But it's a close call since

1.) The `-d keeprsp -n` workaround isn't that bad.
2.) Most projects shouldn't use rspfiles for their compiles.

jhasse:
> 2.) Most projects shouldn't use rspfiles for their compiles.

What do you think about making them opt-in instead? That way projects like Meson would also not need to check for availability of the `-x` flag to fix https://github.com/mesonbuild/meson/issues/3030.

nico:
"them" meaning?

jhasse:
rspfiles

jhasse:
Sorry, I've mixed up some stuff. I meant for the compdb tool, so that instead of passing `-x`, rsp files will get expanded by default, regardless of their use for compiling.

nico:
That seems overly magic to me.

Tangentially, do you know why meson uses rspfiles for compiles?

src/ninja.cc
asankah: I kept this to be consistent with the rest of the source file. :-) I assumed this was deliberate to deal with some incompatibility somewhere.

We can update the whole file to not use the `const_cast<>` with `getopt`, which I think would be better than one call site being an anomaly.

asankah:
(Addressed the comment and updated the PR, also tested on macOS and Linux with a build file that used `rspfile`.)

jhasse:
> Tangentially, do you know why meson uses rspfiles for compiles?

Yes, to fix https://github.com/mesonbuild/meson/issues/271.

Leandros:
(I'm using the PR in my own ninja fork (https://github.com/leandros/ninja) for roughly 6 months in production without any problems.

src/ninja.cc
```diff
@@ -662,7 +662,65 @@ void EncodeJSONString(const char *str) {
...
+  if (rspfile.empty())
+    return command;
+
+  size_t index = command.find(rspfile);
+  if (index == 0 || index == string::npos || command[index - 1] != '@')
+    return command;
+
+  string rspfile_content = edge->GetBinding("rspfile_content");
+  size_t newline_index = 0;
+  while ((newline_index = rspfile_content.find('\n', newline_index)) !=
```
nico: How do you end up with rspfile_content that contains a newline? In the manifest you say

```
   rspfile_content = aasdfsdaf $
                     asdfasdf
```

and ninja replaces the '$\n' with nothing at manifest parse time as far as I remember. How do you get a newline in there?

(This is possibly a silly question; please bear with me.)

nico:
@Leandros gn doesn't use rspfiles by default for compiles any more, see my https://github.com/ninja-build/ninja/pull/1223#issuecomment-378969196 above. I commented on the meson issue.

(To be clear, I'm not arguing against this patch, I'm just saying that generators that produce ninja files containing many rspfiles for compilations likely have opportunities to improve since rsp files shouldn't be necessary for the vast majority of compiler invocations, and including them slows down your build.)

src/ninja.cc
asankah: At the time I couldn't rule out that there wouldn't be newlines, and if there were, then the resulting command line could be malformed. Also, Microsoft's tools allow newlines in response files, so I was hedging the risk of newlines being included.

I can remove the newline handling if the `rspfile_content` lookup is guaranteed to not have any newlines. I haven't looked at the code closely to make that call.

nico: Oh, I guess you can get newlines if you do `rspfile_content = $in_newline`. But other than that magic variable, I don't think it's possible. Hrm, annoying, but for that it is possible. (in_newline is only needed in practice for link.exe so it shouldn't appear in compilations either, but…).

nico:
Thanks for the patch!

nico:
And just to reiterate, users of ninjas that don't have this flag yet can run `ninja -d keeprsp -n` to generate the rsp files and then use the compdb that does point to rspfiles.

asankah:
Yay!

PR #1224, opened 2017-01-10
fix GetProcessorCount() on embedded systems like NVIDIA TK1

tp-m:
Use _SC_NPROCESSORS_CONF rather than _SC_NPROCESSORS_ONLN to
query the number of available CPUs, as on some systems, especially
embedded systems, CPUs may be marked as offline for power saving
reasons.

On an otherwise idle Nvidia TK1 system GetProcessorCount() would
always return 1 just because the other CPUs are not in use, and
then we'd only use 2 out of 4+ cores for compilation.

sched_getaffinity() + CPU_COUNT returns the same as _SC_NPROCESSORS_ONLN
in that case, so does not provide a better solution either.

/proc/cpuinfo will always list all CPUs, but parsing that seems
cumbersome and is likely also less portable.

It's conceivable that there are some cornercases where CONF
returns a higher number than CPUs actually available for use,
but arguably the marginal cost of spawning too many processes
is negligible compared to the cost of not using available cores.

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -505,7 +505,7 @@ int GetProcessorCount() {
   GetNativeSystemInfo(&info);
   return info.dwNumberOfProcessors;
 #else
-  return sysconf(_SC_NPROCESSORS_ONLN);
+  return sysconf(_SC_NPROCESSORS_CONF);
 #endif
 }
 


nico:
Hm, neither current nor new behavior seem clearly better to me. The current behavior works fine on "normal" hardware, and some more exotic hardware probably works better with what's in trunk and some better after this patch. If others have an opinion here, please chime in.

tp-m:
Fair. I think only a very small percentage of systems will be affected by this patch in any case. For me the main argument in favour is still that spawning too few jobs and not utilizing cores is much worse than spawning too many jobs, arguably.

nico:
In chromium land, we often build with -j 1000 and a distributed compiler. If you forget to enable the distributed compiler and run 1000 cc jobs locally, your machine locks up. That's an example where launching too many jobs is pretty bad too :-)

(The example is unrelated to this patch since there we pass an explicit -j, but you could imagine something similar with an implicit -j)

jhasse:
@tp-m Can you test if

```cpp
#include <iostream>
#include <thread>

int main() {
	std::cout << std::thread::hardware_concurrency() << std::endl;
}
```

would also work on the TK1?

jhasse:
Nevermind. Can you test if https://github.com/ninja-build/ninja/commit/1bcc689324bdee090eed035353724abc3fa7c909 works instead? If not, please reopen this PR :)

jhasse:
Ah sorry, just noticed that you ruled this out already:

> sched_getaffinity() + CPU_COUNT

Hm ... not sure what's the proper fix then.

Issue #1227, opened 2017-01-17
Depfile path parsing broken for some characters


maciej-pawlowski:
Paths in depfile containing `[`, `]`, `%` (and possibly some other characters) are not parsed properly.

Repro:
create `a.cc` with:
```
#include "a[1].h"
int main() {
    return 0;
}
```
touch `a[1].h`

Create a GN target:
```
executable("a") {
  sources = [
    "a.cc",
  ]
}
```

Run `ninja a`
Run `ninja a -d explain`

The target will keep getting rebuilt and `explain` will say:
```
ninja explain: output /home/mpawlowski/dev/work/temp/a of phony edge with no inputs doesn't exist
ninja explain: /home/mpawlowski/dev/work/temp/a is dirty
ninja explain: output 1 of phony edge with no inputs doesn't exist
ninja explain: 1 is dirty
ninja explain: output .h of phony edge with no inputs doesn't exist
ninja explain: .h is dirty
ninja explain: obj/ABS_PATH/home/mpawlowski/dev/work/desktop/a/a.o is dirty
```

Note how `a[1].h` got split into three paths: `a`, `1` and `h`. The braces were treated as separators. Same thing happens with the `%` and `@` characters.

maciej-pawlowski:
Note that there are paths in Chromium repository that fail to be properly parsed due to this bug, mainly in test_data folders.

PR #1228, opened 2017-01-18
Fix parsing some special chars in depfiles

maciej-pawlowski:
This allows paths with "[", "]" and "%" to appear in depfiles.
Previously, only "[" would be handled properly.
Fixes issue #1227 .


maciej-pawlowski:
Can I have any feedback whether this is going to get integrated?

ppluciennik:
@nico would it be possible to integrate it? Without this fix, ninja detects that something has changed and rebuilds some targets (It's reproducible on Opera's code - Chromium's based code).

src/lexer.cc
```diff
@@ -1,4 +1,4 @@
-/* Generated by re2c 0.13.5 */
+/* Generated by re2c 0.16 */
```
jhasse: Please remove this file from your commit.

src/depfile_parser.cc
```diff
@@ -1,4 +1,4 @@
-/* Generated by re2c 0.13.5 */
+/* Generated by re2c 0.16 */
```
jhasse: Can you either generate the file with re2c 0.13.5 or add a separate commit which updates the file in Git from re2c 0.13.5 to 0.16?

jhasse:
Thanks for the PR and sorry we didn't look at this earlier!

I think the merge conflict is due to https://github.com/ninja-build/ninja/commit/265a6eaf399778c746c7d2c02b8742f3c0ff07e9.

maciej-pawlowski-opera:
I've since lost access to that account due to my stupidity and 2FA working as intended...
I've rebased the changes onto current master and split regenerating re2c outputs into separate commits here, in a new fork:
https://github.com/maciej-pawlowski-opera/ninja/tree/issue-1227-2

Should I make a new pull request? I'm not sure how github can handle such a situation more gracefully.

jhasse:
Yes, please just create a new PR :) Add "See #1228" in the description so that it references this one.

Issue #1214, opened 2016-12-09
Ninja in verbose mode fails smart terminal capabilities


dvdhrm:
Hey

Using `ninja -v` fails to use smart-terminal capabilities. The easiest way to reproduce is using gcc color output. It works fine in normal mode (`ninja`), but is stripped in verbose mode (`ninja -v`).

Looking at `build.cc`:

```cc
    // ninja sets stdout and stderr of subprocesses to a pipe, to be able to
    // check if the output is empty. Some compilers, e.g. clang, check
    // isatty(stderr) to decide if they should print colored output.
    // To make it possible to use colored output with ninja, subprocesses should
    // be run with a flag that forces them to always print color escape codes.
    // To make sure these escape codes don't show up in a file if ninja's output
    // is piped to a file, ninja strips ansi escape codes again if it's not
    // writing to a |smart_terminal_|.
    // (Launching subprocesses in pseudo ttys doesn't work because there are
    // only a few hundred available on some systems, and ninja can launch
    // thousands of parallel compile commands.)
    // TODO: There should be a flag to disable escape code stripping.
    string final_output;
    if (!printer_.is_smart_terminal())
      final_output = StripAnsiEscapeCodes(output);
    else
      final_output = output;
    printer_.PrintOnNewLine(final_output);
```

The comment is correct, and I understand the logic there. Only problem is, using `-v` causes `set_smart_terminal(false)`. I think merging the `verbose` logic into the `smart-terminal` logic is the culprit. If it was two flags, one `smart-terminal` flag which tells whether escape-codes should be stripped, and a `verbose` flag, telling whether **ninja** should make use of escape-codes, this would work. Could also be combined into a tri-state.

If verbose-mode is not supposed to do coloring, feel free to close this. It seems counter-intuitive to me, though.

Thanks
David

jhasse:
> It seems counter-intuitive to me, though.

I agree. "verbose" doesn't sound like something would be stripped from the output ;)

Thanks for the report, I've fixed this using two flags, as you suggested. I've also added a new Python script `misc/output_test.py` which checks that this works correctly and runs on Travis CI.

dvdhrm:
Thanks a lot!

Issue #916, opened 2015-02-25
Add environment variable and/or command line option to prevent ninja from stripping color escape codes


ngladitz:
Ninja strips out color escape codes from e.g. gcc/clang when it detects that it isn't connected to a terminal.

In a CMake/Ninja based project the QtCreator IDE invokes ninja and is apparently capable of interpreting color codes.

It would be nice if ninja would provide the means to disable stripping of color codes in this case as to allow them to be visualized by QtCreator.

martine:
Ninja doesn't strip the escapes.  gcc/clang turn off the color when they're run under ninja.

There's more discussion here, that I think summarizes all the pros and cons of all the workarounds.
https://github.com/martine/ninja/issues/174

martine:
I totally misremembered the conclusion of that issue!  Reopening.

ngladitz:
Yes, I did read up on that discussion as well.
To clarify I am forcing color output of gcc with -fdiagnostics-color=always.

nico:
See also #672, #746.

pseyfert:
See also #1268 and https://github.com/ninja-build/ninja/blame/master/src/build.cc#L146

jhasse:
Fixed by bf7107bb864d0383028202e3f4a4228c02302961.

PR #1230, opened 2017-01-27
Add -o flag for output options for verbosity and raw output

mgiuffrida:
Adds the general-purpose -o flag to address #746, with options for
verbosity (#480) and control sequence stripping (#581, #672, #916). Also
provides the ability to enable both verbose and raw output as a
workaround for #1214 without changing the existing verbose behavior.

-o verbose  show all command lines while building
-o quiet    hide command lines and outputs while building
-o raw      never strip control sequences from output
-o strip    always strip control sequences from output
-o color    strip most control sequences from output, but retain color codes

This patch does not affect Ninja's defaults of normal verbosity and
smart terminal detection for escape sequence stripping.

"-o color" is particularly useful for utilities like `head` and
`less -R` that interpret color codes from stdin and pass this colored
output to stdout. Any valid ANSI color code, of the form:

    `'ESC' '[' [ colors ] 'm'`

where `colors` is a semicolon-delimited list of optional integers:

   `[ n ] [ ';' colors ]`

is retained in its entirety when using "-o color" (any other CSI escape
sequences besides ANSI color codes are still stripped for non-smart
terminals).

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -79,7 +79,7 @@ BuildStatus::BuildStatus(const BuildConfig& config)
       overall_rate_(), current_rate_(config.parallelism) {
 
   // Don't do anything fancy in verbose mode.
-  if (config_.verbosity != BuildConfig::NORMAL)
+  if (config_.verbosity == BuildConfig::VERBOSE)
     printer_.set_smart_terminal(false);
 
   progress_status_format_ = getenv("NINJA_STATUS");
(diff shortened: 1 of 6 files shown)


johandc:
This is related to pull-request #1268

jonesmz:
@mgiuffrida Could you rebase this PR on the latest master branch?

jhasse:
I actually want to implement this in a different way (add a dedicated `--quiet` flag, leave the other options to external tools).

PR #1235, opened 2017-02-04
Suppress warning 4244 to bootstrap on vs2017

refack:


nico:
Thanks!

What's warning 4244? Does it make sense to fix the code? Can you paste an example of the warning?

refack:
It's a down cast
```
algorithm(946): warning C4244: '=': conversion from 'int' to 'char', possible loss of data
```

PR #1236, opened 2017-02-07
Allow commands with implicit outputs in the depslog

danw:
Compilers such as GCC and Clang can output auxiliary files when certain
arguments are passed. In order to add these to the build graph today, we
must turn off the depslog for these commands (deps = gcc -> deps = )

One example is '--coverage', which will output the standard .o/.d files,
but also writes a .gcno file. Then we'd like to distribute a bundle of
these files for the test systems to create the full coverage
information, so they need to be in the build graph.

This is no worse than using depfiles today, since the current depfile
code only cares about the first output. And the depslog code also only
looks at the first output. For auxiliary (implicit) outputs, this is
completely fine. For commands with multiple output that don't have a
"primary" output, the current depslog code that uses the first output
may not be correct, so I'm leaving the check in for that case.

### DIFF
--- a/src/manifest_parser.cc
+++ b/src/manifest_parser.cc
@@ -385,7 +385,7 @@ bool ManifestParser::ParseEdge(string* err) {
 
   // Multiple outputs aren't (yet?) supported with depslog.
   string deps_type = edge->GetBinding("deps");
-  if (!deps_type.empty() && edge->outputs_.size() > 1) {
+  if (!deps_type.empty() && edge->outputs_.size() - edge->implicit_outs_ > 1) {
     return lexer_.Error("multiple outputs aren't (yet?) supported by depslog; "
                         "bring this up on the mailing list if it affects you",
                         err);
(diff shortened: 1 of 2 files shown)


jgunthorpe:
I think you should revise the warning as well:

"When using deps only a single file can be in $out, use implicit outputs for other files"

I also tested this patch and it works for my use cases too which is a bit different, more like

```
rule foo
 command = foo -o a -o b -o c --deps a.d
 deps = gcc
 depfile = ${out}.d
build a | b c : foo
```
However, unlike your case, 'foo' emits something like this for the deps file:

```
b : inp1 inp2
```

Ie it does not emit $out in the makefile. It looks like the ninja dep parse ignores the output filename when parsing the makefile so this is OK.

I tried various situations with this patch and all worked as I would expect after modifying the generator script to use this syntax, including restat rules and edges that flow through the implicit outputs. Looks good to me, and certainly solves my problem.

melak47:
I'm using this so I can have `.pdb`s as implicit outputs, and `ninja -t clean` deletes them.

plinss:
I have a project using a custom transpiler that generates multiple outputs and needs a depfile. While multiple output support for depfiles would be ideal, this change allows me work around the lack of multiple output support by making one of the outputs implicit.

jhasse:
Superseded by #1364.

Issue #874, opened 2014-12-03
Restat rules that generate build.ninja can crash ninja


colincross:
I'm seeing a crash in ninja (master and v1.5.3):
(gdb) bt
#0  0x00007f3b268dbbb9 in __GI_raise (sig=sig@entry=6)

```
at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
```
#1  0x00007f3b268defc8 in __GI_abort () at abort.c:89
#2  0x00007f3b26918e14 in __libc_message (do_abort=do_abort@entry=1,

```
fmt=fmt@entry=0x7f3b26a27668 "*** Error in `%s': %s: 0x%s ***\n")
at ../sysdeps/posix/libc_fatal.c:175
```
#3  0x00007f3b269250ee in malloc_printerr (ptr=<optimized out>,

```
str=0x7f3b26a23759 "free(): invalid pointer", action=1) at malloc.c:4996
```
#4  _int_free (av=<optimized out>, p=<optimized out>, have_lock=0)

```
at malloc.c:3840
```
#5  0x0000000000409d65 in deallocate (this=<optimized out>,

```
__p=<optimized out>) at /usr/include/c++/4.8/ext/new_allocator.h:110
```
#6  _M_put_node (this=<optimized out>, __p=<optimized out>)

```
at /usr/include/c++/4.8/bits/stl_tree.h:374
```
#7  _M_destroy_node (this=<optimized out>, __p=<optimized out>)

```
at /usr/include/c++/4.8/bits/stl_tree.h:396
```
#8  _M_erase_aux (__position=..., this=0x2c14a40)

```
at /usr/include/c++/4.8/bits/stl_tree.h:1746
```
#9  erase (__position=..., this=0x2c14a40)

```
at /usr/include/c++/4.8/bits/stl_tree.h:826
```
#10 erase (__position=..., this=0x2c14a40)

```
at /usr/include/c++/4.8/bits/stl_map.h:711
```
#11 BuildStatus::BuildEdgeFinished (this=0x2c14a20, edge=edge@entry=0x2391220,

```
success=<optimized out>, output=..., 
start_time=start_time@entry=0x7fff9a8bfeb8, 
end_time=end_time@entry=0x7fff9a8bfebc) at src/build.cc:116
```
#12 0x000000000040a14c in Builder::FinishCommand (

```
this=this@entry=0x7fff9a8c00f0, result=result@entry=0x7fff9a8bffa0, 
[...]

PR #1237, opened 2017-02-08
Support restat when rebuilding manifest

### LINKED ISSUE #987: `restat` is broken when rebuilding build.ninja in 1.6.0

danw:
As a fix for #874, we started reloading the entire manifest even if the
manifest was never rebuilt due to a restat rule. But this can be slow,
so call State::Reset instead, which also fixes the original crash.

Fixes #987

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -247,10 +247,19 @@ bool NinjaMain::RebuildManifest(const char* input_file, string* err) {
   if (builder.AlreadyUpToDate())
     return false;  // Not an error, but we didn't rebuild.
 
-  // Even if the manifest was cleaned by a restat rule, claim that it was
-  // rebuilt.  Not doing so can lead to crashes, see
-  // https://github.com/ninja-build/ninja/issues/874
-  return builder.Build(err);
+  if (!builder.Build(err))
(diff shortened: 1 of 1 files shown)


danw:
And State::Reset should be safe -- it's used for this purpose extensively in the ninja tests.

nico:
Thanks, that was easier than expected. If you haven't, can you afl-fuzz with this merged for a bit? I think that found the other crash iirc (see readme)

Issue #1037, opened 2015-10-15
Ninja always recompiles sources with certain Intel Compiler remarks


bradking:
The Intel Compiler:

    >icl
    Intel(R) C++ Intel(R) 64 Compiler XE for applications running on Intel(R) 64, Version 15.0.0.108 Build 20140726

prints diagnostics to stderr and showIncludes to stdout.  Ninja listens with only one pipe for both, and it appears that `icl` scrambles the two together in this case:

    >type test_icl.cpp
    #include <QtGui/QWidget>
    int main(void) { return 0; }

    >icl -c test_icl.cpp -W3 -Qinline-max-size=1 -showIncludes -nologo -Ic:\Qt\4.8.6\include >out 2>&1 && type out
    ...
    Note: including file:      c:\Qt\4.8.6\include\QtCore/qsharedpointer.h
    Note: including file:       c:\Qt\4.8.6\include\QtCore/../../src/corremark #11074: Inlining inhibited by limit max-size
    remark #11075: To get full report use -Qopt-report:3 -Qopt-report-phase ipo
    elib/tools/qsharedpointer.h
    Note: including file:    c:\Qt\4.8.6\include\QtGui/qsizepolicy.h
    ...

    >del test_icl.obj

This leads Ninja to detect an incorrect dependency on a scrambled file path and to display scrambled output:

    >type build.ninja
    rule ICL
      deps = msvc
      command = icl -c $in -W3 -Qinline-max-size=1 -showIncludes -nologo -Ic:\Qt\4.8.6\include
    build test_icl.obj: ICL test_icl.cpp

    >ninja
    [1/1] icl -c test_icl.cpp -W3 -Qinline-max-size=1 -showIncludes -nologo -Ic:\Qt\4.8.6\include
    remark #11075: To get full report use -Qopt-report:3 -Qopt-report-phase ipo
    elib/tools/qsharedpointer.h

    >ninja -t deps
        ...
        ../../../../../../Qt/4.8.6/src/corremark #11074: Inlining inhibited by limit max-size
        ...

The incorrect dependency causes Ninja to always re-run the compiler:

    >ninja -d explain
    ninja explain: output ../../../../../../Qt/4.8.6/src/corremark #11074: Inlining inhibited by limit max-size of phony edge with no inputs doesn't exist
    ninja explain: ../../../../../../Qt/4.8.6/src/corremark #11074: Inlining inhibited by limit max-size is dirty
[...]

PR #1239, opened 2017-02-10
Separate stderr and stdout on windows

nafest:
Too avoid corrupt output of tools that output to stderr and
stdout on windows (as the Intel Compiler), use two pipes to catch the
output to stderr and stdout separately on windows.
In Subprocess::GetOutput simply concatenate both outputs.

This should fix the issues with the Intel Compiler reported in #1037 and #972

### DIFF
--- a/src/subprocess-win32.cc
+++ b/src/subprocess-win32.cc
@@ -21,14 +21,20 @@
 
 #include "util.h"
 
-Subprocess::Subprocess(bool use_console) : child_(NULL) , overlapped_(),
-                                           is_reading_(false),
+Subprocess::Subprocess(bool use_console) : child_(NULL),
                                            use_console_(use_console) {
+  memset(&overlapped_[PIPE_StdOut], 0, sizeof(overlapped_[PIPE_StdOut]));
+  memset(&overlapped_[PIPE_StdErr], 0, sizeof(overlapped_[PIPE_StdErr]));
+  is_reading_[PIPE_StdOut] = is_reading_[PIPE_StdErr] = false, false;
 }
 
 Subprocess::~Subprocess() {
-  if (pipe_) {
-    if (!CloseHandle(pipe_))
+  if (pipe_[PIPE_StdOut]) {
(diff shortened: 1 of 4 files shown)


nafest:
@nico can I do anything to get this change upstream?

jporkka:
Please!

src/subprocess-win32.cc
```diff
@@ -114,13 +134,18 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
...
-      CloseHandle(pipe_);
+      if (child_out_pipe)
+        CloseHandle(child_out_pipe);
+      if (child_err_pipe)
+        CloseHandle(child_err_pipe);
+      CloseHandle(pipe_[PIPE_StdOut]);
+      CloseHandle(pipe_[PIPE_StdErr]);
       CloseHandle(nul);
-      pipe_ = NULL;
+      printf("pipe to null\n");
```
nico: ?

nico:
Thanks for the patch. Some thoughts:
* The patch looks clean (except for what looks like a debug printf in there) and has a test - nice!
* It feels arguably like an icc bug to me that they disable line buffering on their stdout and stderr. So this is a workaround for a not super heavily used commercial compiler, which makes me not super excited about the goal of the patch.
* If I remember correctly, separating stdout and stderr is hairier on non-Windows, yet if this gets merged on Windows, chances are someone will want to do it on non-Windows too.
* Reliably putting all stderr after all stdout will change the output of programs that flush after writing to stdout and stderr and expect this to be interleaved; I could imagine some people or programs not being happy about that.
* GetOutput() now needs to copy the output, and the output can be large.

So I'm overall a bit ambivalent; that's why I haven't merged this yet. The patch itself is high-quality though :-)

jhasse:
In order to lower the number of open PRs I'm going to close this one. This doesn't mean it's out of the pictures, please address the questions raised by nico and jporkka and reopen it then :)

PR #1243, opened 2017-02-19
std:: corrections

kvart2006:
I made these changes because on NetBSD the compiler asks them.
I replaced "naked" string, map etc. by std::string, std::map etc.
./configure.py --bootstrap works!

### DIFF
--- a/src/edit_distance.cc
+++ b/src/edit_distance.cc
@@ -36,7 +36,7 @@ int EditDistance(const StringPiece& s1,
   int m = s1.len_;
   int n = s2.len_;
 
-  vector<int> row(n + 1);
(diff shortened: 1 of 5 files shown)


kvart2006:
I only changed in string_piece.h:
std::size_t .

nico:
Thanks for the patch! Can you show an example of the compiler diagnostic? There are `using namespace std` statements in a few places, so compilers should be happy.

ilyapopov:
Looks like this is because there are `using namespace std;` in headers, which is considered bad practice.

nico:
Closing since my question didn't get an answer.

PR #1244, opened 2017-02-22
WIP: build: add a trailing newline to rspfile contents

mathstuf:
Some compilers (at least the icecream parallel wrappers) drop the last
flag if there is not trailing whitespace in the file.

See: https://gitlab.kitware.com/cmake/cmake/issues/16664

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -745,6 +745,8 @@ bool Builder::StartEdge(Edge* edge, string* err) {
   string rspfile = edge->GetUnescapedRspfile();
   if (!rspfile.empty()) {
     string content = edge->GetBinding("rspfile_content");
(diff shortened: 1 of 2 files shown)


jhasse:
Do you still need this now that it seems to have been fixed in icecream?

mathstuf:
Do you have a link to the icecream fix? Mainly for posterity and for pointing users to a workable version of icecream.

jhasse:
Hm ... the CMake issues links to https://github.com/icecc/icecream/issues/188, but that looks like a different problem, doesn't it?

mathstuf:
The CMake issue submitter also authored the icecream issue. I'd assume that this is fixed unless someone runs into it again with icecream 1.2+.

jhasse:
Okay, please reopen this PR, if this is still an issue.

Issue #1248, opened 2017-02-23
UB found by UBSAN (-fsanitize=undefined)


matthiaskrgr:
part two of #1247

there was some UB reported in the ./ninja_test, if compiled as described in #1247 .
````
src/test.cc:101:3: runtime error: member call on null pointer of type 'testing::Test'
SUMMARY: AddressSanitizer: undefined-behavior src/test.cc:101:3 in 
src/test.cc:102:3: runtime error: member call on null pointer of type 'testing::Test'
SUMMARY: AddressSanitizer: undefined-behavior src/test.cc:102:3 in 
src/test.cc:139:3: runtime error: member call on null pointer of type 'testing::Test'
SUMMARY: AddressSanitizer: undefined-behavior src/test.cc:139:3 in 
[61/226] BuildTest.DepsGccWithEmptyDepfileErrorsOutsrc/graph.cc:464:36: runtime error: member call on null pointer of type 'DepsLog'
SUMMARY: AddressSanitizer: undefined-behavior src/graph.cc:464:36 in 
[214/226] SubprocessTest.SetWithLotsRaise [ulimit -n] above 1025 (currently 1024) to make this test go
[226/226] ElideMiddle.ElideInTheMiddle
passed
````

To aid debugging, you can obtain stacktraces by running
````
 UBSAN_OPTIONS=print_stacktrace=1 ./ninja_test
src/test.cc:101:3: runtime error: member call on null pointer of type 'testing::Test'
    #0 0x71d455 in AssertParse(State*, char const*) /home/matthias/vcs/github/ninja/src/test.cc:101:3
    #1 0x71d1a1 in StateTestWithBuiltinRules::AddCatRule(State*) /home/matthias/vcs/github/ninja/src/test.cc:88:3
    #2 0x71d163 in StateTestWithBuiltinRules::StateTestWithBuiltinRules() /home/matthias/vcs/github/ninja/src/test.cc:84:3
    #3 0x4f1d3b in (anonymous namespace)::BuildLogTest::BuildLogTest() /home/matthias/vcs/github/ninja/src/build_log_test.cc:33:8
    #4 0x4f1c3b in (anonymous namespace)::BuildLogTestWriteRead::BuildLogTestWriteRead() /home/matthias/vcs/github/ninja/src/build_log_test.cc:44:1
[...]

PR #1249, opened 2017-02-23
Fix part of UBSAN report in #1248

sgraham:
This fixes the ubsan reported errors in graph.cc. This happened to work before because the node passed in earlied out of DepsLog::GetDeps() before any members were accessed.

(The ones in test.cc are a bit trickier.)


jhasse:
Thanks! I've decided to fix it like this: 28a7d1491367de7b39c854d166114f76d272f04f

PR #1250, opened 2017-02-23
Fix UBSAN (out of order test setup) from #1248

sgraham:
Seems like abort()ing is reasonable, as this shouldn't really ever come up in practice.


jhasse:
Closing in favor of #1562.

PR #1255, opened 2017-03-09
browse: Bind to localhost by default

tchajed:
Previously would bind to all interfaces, which exposes an HTTP server over the network. This is insecure and almost certainly unnecessary for development purposes.

The hostname is now "localhost" by default and is configurable with `--hostname` or `-a` (for "address"), so the original behavior is achieved with `ninja -t browse -a ""`.

### DIFF
--- a/src/browse.py
+++ b/src/browse.py
@@ -193,6 +193,8 @@ class RequestHandler(httpserver.BaseHTTPRequestHandler):
 parser = argparse.ArgumentParser(prog='ninja -t browse')
 parser.add_argument('--port', '-p', default=8000, type=int,
     help='Port number to use (default %(default)d)')
+parser.add_argument('--hostname', '-a', default='localhost', type=str,
+    help='Hostname to bind to (default %(default)s)')
 parser.add_argument('--no-browser', action='store_true',
     help='Do not open a webbrowser on startup.')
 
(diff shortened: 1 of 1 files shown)


tchajed:
I made this configurable but I also think it's sensible to only allow `localhost`. It makes little sense to me to expose this server over the network. However, strictly speaking to follow standards the best default is `127.0.0.1` - `localhost` is generally mapped to this address but could in principle resolve to anywhere.

tchajed:
I also didn't update the [manual's description of `-t browse`](doc/manual.asciidoc#extra-tools), but would be happy to do so if you think this option should be listed in the documentation (though it's automatically available in `ninja -t browse --help` due to `argparse`).

src/browse.py
```diff
@@ -205,9 +207,11 @@ def log_message(self, format, *args):
 
 args = parser.parse_args()
 port = args.port
-httpd = httpserver.HTTPServer(('',port), RequestHandler)
+hostname = args.hostname
+httpd = httpserver.HTTPServer((hostname,port), RequestHandler)
```
nico: nit: space after ,

src/browse.py
```diff
@@ -205,9 +207,11 @@ def log_message(self, format, *args):
...
 try:
-    hostname = socket.gethostname()
+    if hostname == "":
```
nico: nit: `if not hostname:`

nico:
Makes sense, thanks.

PR #1264, opened 2017-04-11
DirName's separators and their length are known at compile time

gahr:


nico:
Thanks for the patch! What's the motivation here? Does it improve generated code?

gahr:
Well, I guess it's not going to make a huge difference but as I spotted it while I was having a look at the code, I thought I'd report it, just in case.

Actually, it improves the object size (from 116k to 110k) and reduces the number of instructions taken by the `MakeDirs` method from 260 to 160.

PR #1269, opened 2017-04-13
Visual Studio 2017 compatibility fix

TDannhauer:
Some minor changes to make ninja compilable with MS Visual Studio 2017

### DIFF
--- a/src/clparser.cc
+++ b/src/clparser.cc
@@ -53,7 +53,7 @@ string CLParser::FilterShowIncludes(const string& line,
 
 // static
 bool CLParser::IsSystemInclude(string path) {
-  transform(path.begin(), path.end(), path.begin(), ::tolower);
+  transform(path.begin(), path.end(), path.begin(), [](const char c) { return static_cast<char>(::tolower(c)); });
   // TODO: this is a heuristic, perhaps there's a better way?
   return (path.find("program files") != string::npos ||
           path.find("microsoft visual studio") != string::npos);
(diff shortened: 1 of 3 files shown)


src/clparser.cc
```diff
@@ -53,15 +53,15 @@ string CLParser::FilterShowIncludes(const string& line,
 
 // static
 bool CLParser::IsSystemInclude(string path) {
-  transform(path.begin(), path.end(), path.begin(), ::tolower);
+  transform(path.begin(), path.end(), path.begin(), [](const char c) { return static_cast<char>(::tolower(c)); });
```
nico: Ninja doesn't use any c++11 features and doesn't pass -std=c++11 to gcc/clang (where one has to explicitly opt in to that).

TDannhauer: MS VC compiler has afaik no switches to select specific c++ versions. Regardless which c++ version is used in ninja, it is not compilable with Visual Studio 2017 and needs to be fixed. 

Any suggestions?

paulsapps: MS VC actually does have switches now https://blogs.msdn.microsoft.com/vcblog/2016/06/07/standards-version-switches-in-the-compiler/

"We’re creating a series of new switches that correspond to versions of the C++ language, starting with /std:c++14 and /std:c++latest."

TDannhauer:
@paulsapps thanks. Does this also apply to VS 2017? They will provides only switches for C++14 and later.

Someone else if the ninja team hast to chip in to make it compatible with VS 2017. I'll quit now since my time is limited and ninja not my focus.

Thanks so far and good luck.

paulsapps:
@TDannhauer it applies to 2015+

jhasse:
AppVeyor uses VS2017 and it seems to work fine without this. Closing.

Issue #1135, opened 2016-04-26
Always rebuild output files on build errors


colincross:
There have been previous discussions about importing make's .DELETE_ON_ERROR behavior (https://www.gnu.org/software/make/manual/html_node/Errors.html) into ninja to handle the case where a rule partially writes the output file and then exits with an error (https://groups.google.com/d/msg/ninja-build/-IOv-2K6CkI/CJW9a0LKD1oJ).  ninja assumes that all tools write atomically, so the next build will see a new mtime on the output file and will not rerun the rule that failed.

.DELETE_ON_ERROR behavior is controversial, but https://groups.google.com/d/msg/ninja-build/YQuGNrECI-4/ti-lAs9SPv8J suggests an alternate solution - always check the output file mtime using .ninja_log, and always update it on success.  If a rule fails the output file mtime will either not be present in .ninja_log (for a first build), or will have an old mtime (for an incremental build), and the next build will rerun the failure.

PR #1281, opened 2017-05-19
Rebuild edges that update their output file on error

colincross:
https://groups.google.com/forum/#!msg/ninja-build/YQuGNrECI-4/ti-lAs9SPv8J discusses a case where an rule updates its output file and then fails.  The next run of ninja considers the ouptut file clean and doesn't rebuild it.  Always stat output files after they are built, and write the mtime into .ninja_log.  Consider output files dirty if the recorded mtime is older than the most recent input file.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -800,9 +800,10 @@ bool Builder::FinishCommand(CommandRunner::Result* result, string* err) {
     return true;
   }
 
-  // Restat the edge outputs, if necessary.
-  TimeStamp restat_mtime = 0;
-  if (edge->GetBindingBool("restat") && !config_.dry_run) {
+  // Restat the edge outputs
+  TimeStamp output_mtime = 0;
+  bool restat = edge->GetBindingBool("restat");
+  if (!config_.dry_run) {
     bool node_cleaned = false;
 
     for (vector<Node*>::iterator o = edge->outputs_.begin();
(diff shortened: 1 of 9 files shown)


colincross:
This pull request requires ninja to stat all output files.  A test on clean builds of the bionic/ directory in AOSP Android (~2 minutes total compilation time) shows a relatively small impact on ninja statistics.

Before:
```
metric               	count 	avg (us) 	total (ms)
.ninja parse         	1     	2397076.0	2397.1
canonicalize str     	866296	0.2     	131.5
canonicalize path    	1298655	0.1     	146.4
lookup node          	1311549	0.2     	324.4
.ninja_log load      	1     	11.0    	0.0
.ninja_deps load     	1     	15777.0 	15.8
.ninja_deps recompact	1     	38495.0 	38.5
node stat            	27070 	2.4     	64.5
depfile load         	1     	917.0   	0.9
StartEdge            	9442  	3775.4  	35647.7
FinishCommand        	9244  	170.6   	1576.7

path->node hash load 0.55 (108645 entries / 196613 buckets)

real	1m47.361s
user	57m7.524s
sys	2m41.768s
```
After:
```
metric           	count 	avg (us) 	total (ms)
.ninja parse     	1     	2945194.0	2945.2
canonicalize str 	866296	0.2     	135.1
canonicalize path	1298655	0.1     	155.0
lookup node      	1312231	0.2     	303.8
.ninja_log load  	1     	7.0     	0.0
.ninja_deps load 	1     	12422.0 	12.4
node stat        	36012 	3.0     	107.6
depfile load     	1     	948.0   	0.9
StartEdge        	9442  	3593.2  	33927.5
FinishCommand    	9244  	175.3   	1620.8

path->node hash load 0.56 (109121 entries / 196613 buckets)

real	1m42.683s
user	56m53.308s
sys	2m36.028s
```
stat is called 8942 extra times taking an extra 43 ms.

Fixes #1135

nico:
(This was supposed to be a comment on the first of your 3 reviews.)

colincross:
I'll just drop that patch from here, it was useful to generate the before and after statistics but it can be its own pull request if you want it.

nico:
If the motivation is to fix metrics, I think I'd rather move the macro call around.

colincross:
Moved the metrics to RealDiskInterface

nico:
Thanks! let's hope this doesn't break something we didn't think of :-)

bradking:
This introduced a regression with subninja.  See #1290.

PR #1284, opened 2017-06-02
Use DETACHED_PROCESS to mitigate Windows close-process bug

### LINKED ISSUE #1283: Ninja needs to work-around Windows 10 process-destruction bug
On long-running Windows 10 systems (starting in Anniversary Update?) there is a bug that can cause process destruction to be both expensive and serialized. Processes spend a significant amount of time on this call stack:

  |- ntoskrnl.exe!KiSystemServiceCopyEnd
  |    ntoskrnl.exe!NtTerminateProcess
  |    ntoskrnl.exe!PspExitThread
  |    |- ntoskrnl.exe!ExCallCallBack
  |    |    win32kbase.sys!W32CalloutDispatch
  |    |    win32kfull.sys!W32pProcessCallout
  |    |    |- win32kbase.sys!GdiProcessCallout
  |    |    |    win32kbase.sys!NtGdiCloseProcess

and they do it while holding a global system lock. This means that process shutdown is serialized, which on a 48-thread machine effectively means it runs 48x slower than it should. And, it seems to take 5-8 ms, thus restricting process destruction rates to 125-200 per second - low enough that it measurably affects Chromium build times when doing distributed builds (a full build of Chrome uses about 60,000 processes which implies a minimum build time of 5-8 minutes).

This global lock is also used to protect message queues so the Windows UI repeatedly locks up when this situation is hit.

The effects of this bug are greatly reduced if processes are created with the DETACHED_PROCESS flag, and this doesn't seem to have any problematic side-effects. I will create a PR.

randomascii:
Windows 10 has a close-process performance bug which can significantly
affect the performance of ninja builds. This bug causes process
destruction to be quite expensive (5-8 ms). Because process destruction
happens while holding a system-global lock it is serialized and it
causes micro-hangs in the UI (mouse pointer freezes for seconds at a
time, especially when many processes are terminating simultaneously).

This change specifies DETACHED_PROCESS when possible because this avoids
the expensive shutdown path, for some reason. It is effectively only
applied to compiler processes, so the problem is not entirely avoided,
but in some cases this fix is enough to improve build speeds by 20-30%.

Fixes #1283, at least partially

### DIFF
--- a/src/msvc_helper-win32.cc
+++ b/src/msvc_helper-win32.cc
@@ -65,8 +65,10 @@ int CLWrapper::Run(const string& command, string* output) {
   startup_info.hStdOutput = stdout_write;
   startup_info.dwFlags |= STARTF_USESTDHANDLES;
 
+  // Use DETACHED_PROCESS when possible to work around Windows bug:
+  // https://github.com/ninja-build/ninja/issues/1283
   if (!CreateProcessA(NULL, (char*)command.c_str(), NULL, NULL,
-                      /* inherit handles */ TRUE, 0,
+                      /* inherit handles */ TRUE, DETACHED_PROCESS,
                       env_block_, NULL,
                       &startup_info, &process_info)) {
     Win32Fatal("CreateProcess");
--- a/src/subprocess-win32.cc
+++ b/src/subprocess-win32.cc
@@ -102,7 +102,9 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
   memset(&process_info, 0, sizeof(process_info));
 
   // Ninja handles ctrl-c, except for subprocesses in console pools.
-  DWORD process_flags = use_console_ ? 0 : CREATE_NEW_PROCESS_GROUP;
+  // Use DETACHED_PROCESS when possible to work around Windows bug:
+  // https://github.com/ninja-build/ninja/issues/1283
+  DWORD process_flags = use_console_ ? DETACHED_PROCESS : CREATE_NEW_PROCESS_GROUP;
 
   // Do not prepend 'cmd /c' on Windows, this breaks command
   // lines greater than 8,191 chars.


randomascii:
Unfortunately this change doesn't help in the use_console_ path, which means that a lot of processes are still subject to the slow shutdown problem. I've seen ~two minute savings on goma builds of Chrome, and would presumably get slightly more savings if we could apply this to all process creates, but this is plenty enough benefit to be worth doing. PTAL.

atetubou:
Great work!
Can I see difference of '-d stats' with and without your patch on your normal chrome build?
`ninja -C out/Release -d stats -j 1000 chrome`

randomascii:
Unfortunately I logged off/on which clears the Windows state which triggers the worst instances of this bug, so I can't usefully measure this right now. I will try to get my machine into a bad state as quickly as possible.

BTW, the close-process performance bug was found because of UI hitches (mouse pointer freezing especially) during goma builds of Chrome. ETW trace analysis and following the chain of locks led to the analysis in the issue.

sgraham:
I think use_console_ is determined based on (ninja) console=pool, which maps to (GN) console=true in actions https://cs.chromium.org/chromium/src/tools/gn/variables.cc?l=930 which I don't think Chrome even uses. So, shouldn't be much of a problem.

(Generally it wouldn't make much sense to give control of the console when there's a whole lot of process running in parallel so seems like it should be fine.)

randomascii:
> which I don't think Chrome even uses.

Unfortunately a significant number of processes in Chrome's build use use_console_ = true and therefore cannot use DETACHED_PROCESS. Changing the logic to ignore use_console_ (or to use DETACHED_PROCESS when it is true) causes a bunch of visible console windows to open during builds. It looks like only the compiler processes can use DETACHED_PROCESS. This roughly halves the effect of this OS bug, but it means we can't completely mitigate it.

I'm still waiting for my machine to get back into the truly horrible state so that I can run more tests.

sgraham:
Oh, sorry. I didn't see anything in https://cs.chromium.org/search/?q=console.*%3D.*true+file:gn+package:%5Echromium$&type=cs . I guess there's some other way it gets turned on.

randomascii:
It turns out that my testing was poor and this change does not actually help. It might even make things worse.

I was testing a new ninja.exe that include my change with the current version of ninja used by Chromium. However the new ninja.exe contains much more than just my change and it was those other changes (probably from tikuta@) that were causing the speedup.

The DETACHED_PROCESS flag does help in artificial tests, which helped convince me that my science was good, but I know know that it was misleading.

My artificial tests also demonstrate that the DETACHED_PROCESS doesn't help if user32.dll is loaded into the process, so presumably cl.exe must be doing that (traces say otherwise) or something equivalent.

Anyway, I'm going to close this PR. Other action items include:
1) We should roll a new version of ninja to get the other build performance improvements
2) We should log-off and back on again occasionally (every few days?) in order to get build performance improvements
3) I'll continue working with Microsoft to get the underlying OS bug fixed

PR #1291, opened 2017-06-16
Fix segfault on edge with no inputs

### LINKED ISSUE #1290: regression: segmentation fault on rebuild with no inputs
Changes in #1281 regressed this case:

```console
$ cat example.sh
rm -rf sub build.ninja .ninja_deps .ninja_log

echo '
subninja sub/build.ninja
default sub/out' > build.ninja

mkdir sub
echo '
rule touch
  command = touch $out
build sub/out: touch
' > sub/build.ninja

ninja
echo ---
ninja

$ sh example.sh
[1/1] touch sub/out
---
Segmentation fault
```

colincross:
PR #1281 added a deference of most_recent_input without checking
for NULL, which can occur if a build rule has no inputs.
Check it for null before dereferencing, and add a test.

Fixes #1290.

### DIFF
--- a/src/build_test.cc
+++ b/src/build_test.cc
@@ -1305,6 +1305,37 @@ TEST_F(BuildWithLogTest, RebuildAfterFailure) {
   EXPECT_EQ("", err);
 }
 
+TEST_F(BuildWithLogTest, RebuildWithNoInputs) {
+  ASSERT_NO_FATAL_FAILURE(AssertParse(&state_,
+"rule touch\n"
+"  command = touch\n"
+"build out1: touch\n"
+"build out2: touch in\n"));
+
+  string err;
+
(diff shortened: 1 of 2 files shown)


bradking:
I can confirm that this fixes the case I reported.  However, with assertions enabled (e.g. `configure.py --debug`) the new test fails:

```
BuildWithLogTest.RebuildWithNoInputsninja_test: src/build.cc:661: bool Builder::Build(std::__cxx11::string*): Assertion `!AlreadyUpToDate()' failed.
Aborted
```

colincross:
I updated the test to have two outputs, one if which will rebuild on the second build so that AlreadyUpToDate() is false.

nico:
Thanks Brad for the report, and Colin for the quick fix!

PR #1292, opened 2017-06-18
Work around mtime being set to 0 sometimes

gtristan:
For the time being, I'm proposing Patrick Griffis' patch which has been in use for some months in the Flatpak freedesktop SDK as I need a solution to this also for using build sandboxes with intentionally zeroed timestamps.

Not sure this should be the best solution long term but it's a decent stop gap, the 1 timestamp is much more rare than a 0 timestamp.


nico:
Thanks!

PR #1293, opened 2017-06-18
Untangle mtime state

### LINKED ISSUE #1120: Files with zero mtime are treated as nonexistent
On a CentOS 7 system:

```
$ tar xf ome-cmake-superbuild-0.1.0.tar.xz
tar: ome-cmake-superbuild-0.1.0/cmake/GitVersion.cmake: implausibly old time stamp 1970-01-01 01:00:00
$ ls -l ome-cmake-superbuild-0.1.0/cmake/GitVersion.cmake
-rw-r--r-- 1 rleigh lsd 236 Jan  1  1970 ome-cmake-superbuild-0.1.0/cmake/GitVersion.cmake
```

This source archive was packed incorrectly, and a single file has an mtime of zero (i.e. start of Unix epoch).

```
$ mkdir build
$ cd build
$ cmake -G Ninja ../ome-cmake-superbuild-0.1.0
$ ninja
[1/1] Re-running CMake...
[cmake re-run repeated 100 times]
ninja: error: manifest 'build.ninja' still dirty after 100 tries
```

The cause:

```
$ ninja -d explain
ninja explain: output /tmp/ome-cmake-superbuild-0.1.0/cmake/GitVersion.cmake of phony edge with no inputs doesn't exist
```

Clearly, the file does exist.  If I `touch /tmp/ome-cmake-superbuild-0.1.0/cmake/GitVersion.cmake` to update the timestamp to the current time, then `ninja` runs perfectly.  It appears to be the case that Ninja is treating files with a zero mtime as being nonexistent, which seems incorrect.  Please could you consider changing that behaviour?  While this is an unusual corner case--files shouldn't have a timestamp this old--it can and does happen for various reasons, and it would be nice if Ninja could work correctly under these circumstances.

Kind regards,
Roger

gtristan:
This branch fixes #1120 more cleanly.

This changes the `Stat()` semantics so that it does not try to store additional state in the mtime, any mtime is valid, including 0 and negative mtimes.

The graph Node now should have a reduced memory signature, by storing various boolean states as single bits, storing the file existence state separately from the mtime of course.

Test cases are updated and still passing.

### DIFF
--- a/src/test.cc
+++ b/src/test.cc
@@ -146,13 +146,23 @@ void VirtualFileSystem::Create(const string& path,
   files_created_.insert(path);
 }
 
-TimeStamp VirtualFileSystem::Stat(const string& path, string* err) const {
+bool VirtualFileSystem::Stat(const string& path, StatResult* result, string* err) const {
   FileMap::const_iterator i = files_.find(path);
+
+  *err = "";
+
   if (i != files_.end()) {
-    *err = i->second.stat_error;
-    return i->second.mtime;
+    if (i->second.mtime < 0) {
+      *err = i->second.stat_error;
+      return false;
+    }
+    result->mtime = i->second.mtime;
+    result->exists = true;
+    return true;
   }
-  return 0;
+  result->mtime = 0;
+  result->exists = false;
+  return true;
 }
 
 bool VirtualFileSystem::WriteFile(const string& path, const string& contents) {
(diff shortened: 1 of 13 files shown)


src/test.cc
```diff
@@ -146,13 +146,23 @@ void VirtualFileSystem::Create(const string& path,
   files_created_.insert(path);
 }
 
-TimeStamp VirtualFileSystem::Stat(const string& path, string* err) const {
+bool VirtualFileSystem::Stat(const string& path, StatResult* result, string* err) const {
   FileMap::const_iterator i = files_.find(path);
+
+  *err = "";
```
jhasse: Why do we need to do this?

jhasse:
I'm closing due to merge conflicts. Definitely need to look at this some time in the future. Feel free to reopen.

PR #1294, opened 2017-06-28
Track in Plan whether wanted edges have been scheduled

bradking:
Refactor the `want_` map to track for wanted edges whether they have been scheduled or not.  This gives `ScheduleWork` a direct place to keep this information, making the logic more robust and easier to follow.  It also future-proofs `ScheduleWork` to avoid repeat scheduling if it is called after an edge has been removed from `ready_` by `FindWork`.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -318,18 +318,18 @@ bool Plan::AddSubTarget(Node* node, Node* dependent, string* err) {
     return false;  // Don't need to do anything.
 
   // If an entry in want_ does not already exist for edge, create an entry which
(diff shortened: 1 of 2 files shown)


bradking:
@nico this is the next step toward Fortran support.  It is also a nice cleanup on its own.

bradking:
Rebased on `master`.

janusw:
Ping! Any progress here? It would be really great to see an official release of ninja with Fortran support in the near future!

nico:
Thanks! Sorry for the delay, I was away for a bit.

PR #1296, opened 2017-06-30
Two concurrency related fixes

davezarzycki:
1) Change concurrent smart console builds to "blame" the right task that is taking too long.
2) Change the amount of concurrency to be conservative. Users can still use their own heuristics with -j if they want to.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -18,6 +18,7 @@
 #include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <limits.h>
 #include <functional>
 
 #ifdef _WIN32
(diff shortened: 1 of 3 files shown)


nico:
Please do one thing per pull request.

Do you have any data to support the new default -j flag? (The current default didn't have lots of measuring put into it either, but if we're changing things I'd like to have a good reason for why.)

davezarzycki:
I worked at Apple from 1996 to 2014 in the CoreOS department and later the developer tools department. We always found that "-j $CPU_COUNT" was/is the right default and all projects built that way.

Said differently, you’d need to have a *really* slow I/O system for ninja’s current “guess” logic to improve build times. Otherwise the current "guess" logic can only slow things down due to avoidable context switching and CPU cache thrashing.

PR #1306, opened 2017-08-03
Support building without "browse" command

gahr:
This commit adds a configure argument --without-browse to build without support for the browse command.
This might help packagers provide an option to users to build ninja in a way that it does not run-depend on python.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -218,6 +218,8 @@ parser.add_option('--with-python', metavar='EXE',
 parser.add_option('--force-pselect', action='store_true',
                   help='ppoll() is used by default where available, '
                        'but some platforms may need to use pselect instead',)
+parser.add_option('--without-browse', action='store_true',
+                  help='build without support for the "browse" command')
 (options, args) = parser.parse_args()
 if args:
     print('ERROR: extra unparsed command-line arguments:', args)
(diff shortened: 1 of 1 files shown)


jhasse:
Only the browse tool will run-depend on Python, doesn't it?

gahr:
That's correct. This is useful to produce ninja binaries that do not list `browse` in the list of tools.

jhasse:
IMHO that's not very useful. Someone might install Ninja and then install Python later. Distros should add Python as a recommended dependency instead of a required one, if that really is a problem.

gahr:
On FreeBSD, we register run-time dependencies between packages, so ninja now run-time depends on python because it might need it. If you uninstall python, it'll uninstall ninja too.
I just thought it sucked to be forced to have python installed to run ninja.. but fine, thanks.

jhasse:
> On FreeBSD, we register run-time dependencies between packages, so ninja now run-time depends on python because it might need it. If you uninstall python, it'll uninstall ninja too.

Aren't there optional run-time dependencies?

gahr:
Yes - they can be chosen at build time, and they normally correspond to configure parameters. We could provide an option for the python dependency, but what I don't like is having a ninja binary installed that exposes a non-working functionality because of turning off python at build time didn't turn the functionality off. But I agree this is not a major thing.

jhasse:
I see.

Even for non-supported platforms I'm thinking about exposing the tool, as it currently confuses people why it isn't there, see #1478.

gahr:
I see - fair enough. Can we have a configure option to have the tool just dump "Browse is not supported here" ? ;)

jhasse:
Done. With the new error message, I hope that FreeBSD is confident enough to turn Python into an optional dependency.

gahr:
Thanks!

PR #1307, opened 2017-08-16
Fix too small buffer in AbsPath

moroten:
Even if relative paths are not allowed to be longer than _MAX_PATH,
combined with the working directory may become larger. This commit
handles those problems.

This fixes the IncludesNormalize::LongInvalidPath test which broke in
commit a2d1cf2f26bdd7a326a3ef6186ee507fd87868d0,
"Fix crash in attempting to canonicalize paths longer than _MAX_PATH".

Signed-off-by: Fredrik Medley <fredrik.medley@autoliv.com>


src/includes_normalize-win32.cc
```diff
@@ -53,16 +53,41 @@ bool SameDriveFast(StringPiece a, StringPiece b) {
...
 
+void GetFullPathNameSafeLongPath(const string &filename, size_t absoluteSize, char *absolute) {
+  // NOTE: Avoiding malloc for speed
+  wchar_t wfilename[_MAX_PATH + 1];
+  {
+    size_t charsToConvert = filename.size() + 1;
+    size_t convertedChars = 0;
+    mbstowcs_s(&convertedChars, wfilename, charsToConvert, filename.c_str(), sizeof(wfilename) / sizeof(wfilename)[0]);
+    if (convertedChars != charsToConvert)
+      Fatal("path too long (A2W) \"%s\"", filename.c_str());
```
nico: what does A2W mean?

src/includes_normalize-win32.cc
```diff
@@ -53,16 +53,41 @@ bool SameDriveFast(StringPiece a, StringPiece b) {
   return IsPathSeparator(a[2]) && IsPathSeparator(b[2]);
 }
 
+void GetFullPathNameSafeLongPath(const string &filename, size_t absoluteSize, char *absolute) {
+  // NOTE: Avoiding malloc for speed
+  wchar_t wfilename[_MAX_PATH + 1];
+  {
+    size_t charsToConvert = filename.size() + 1;
+    size_t convertedChars = 0;
+    mbstowcs_s(&convertedChars, wfilename, charsToConvert, filename.c_str(), sizeof(wfilename) / sizeof(wfilename)[0]);
```
nico: We don't currently do any multibyte conversion anywhere -- this seems orthogonal to this patch (?)

(also, 80 cols)

nico:
@sgraham

moroten:
I totally missed the UTF-8 detail, thanks for pointing out. I'm not myself very comfortable in the string conversion APIs. I just wrote the patch as I was annoyed about one test always failing.

As sgraham said, should we do a small patch here or should Ninja go all in for the W Win32 API?

nico:
There's an open issue about multibyte file handling somewhere. I'm not an expert in the Windows multibyte APIs, but if someone wanted to do what Scott described and it's not overly messy, it'd probably be a good change. But it should probably be a dedicated patch :-)

nico:
If your motivation here is a test always failing, can you check if #1318 happens to make that test go for you?

moroten:
#1318 is a good test. I converted my patch into just checking return value from GetFullPathName to avoid just crashing.

src/includes_normalize_test.cc
```diff
@@ -108,36 +111,57 @@ TEST(IncludesNormalize, LongInvalidPath) {
...
+
   string forward_slashes(kExactlyMaxPath);
   replace(forward_slashes.begin(), forward_slashes.end(), '\\', '/');
   // Make sure a path that's exactly _MAX_PATH long is canonicalized.
-  EXPECT_EQ(forward_slashes,
+  EXPECT_EQ(forward_slashes.substr(cwd_len + 1),
             NormalizeAndCheckNoError(kExactlyMaxPath));
 }
+
+TEST(IncludesNormalize, ShortRelativeButTooLongAbsolutPath) {
```
nico: nit: typo AbsolutePath (missing e)

src/includes_normalize-win32.cc
```diff
@@ -26,6 +26,22 @@
 
 namespace {
 
+bool InternalGetFullPathName(LPCTSTR lpFileName, DWORD nBufferLength,
```
nico: nit: I get that this is supposed to match GetFullPathName's declaration, but no polish notation please :-) Also, every caller passes NULL for lpFilePart, so I'd omit this parameter here.

nico:
Looks good after rebasing, thanks.

This isn't a new regression, right? If so, I'd like to wait until after tagging 1.8.1 (which should happen within 24h) with merging, if that's cool

moroten:
This is a very old bug in the code base. I discovered it while investigating the failing long paths test.

src/includes_normalize_test.cc
```diff
@@ -58,9 +58,12 @@ TEST(IncludesNormalize, Simple) {
 }
 
 TEST(IncludesNormalize, WithRelative) {
+  string err;
   string currentdir = GetCurDir();
   EXPECT_EQ("c", NormalizeRelativeAndCheckNoError("a/b/c", "a/b"));
-  EXPECT_EQ("a", NormalizeAndCheckNoError(IncludesNormalize::AbsPath("a")));
+  EXPECT_EQ("a",
+            NormalizeAndCheckNoError(IncludesNormalize::AbsPath("a", &err)));
+  EXPECT_TRUE(err.empty());
```
jhasse: Can you use `ASSERT_EQ("", err);` instead? That way the error gets printed when the test fails.

src/includes_normalize_test.cc
```diff
@@ -138,3 +141,27 @@ TEST(IncludesNormalize, LongInvalidPath) {
   EXPECT_EQ(forward_slashes.substr(cwd_len + 1),
             NormalizeAndCheckNoError(kExactlyMaxPath));
 }
+
+TEST(IncludesNormalize, ShortRelativeButTooLongAbsolutePath) {
+  string result, err;
+  IncludesNormalize normalizer(".");
+  // A short path should work
+  EXPECT_TRUE(normalizer.Normalize("a", &result, &err));
+  EXPECT_TRUE(err.empty());
```
jhasse: Here too.

src/includes_normalize_test.cc
moroten: Done

src/includes_normalize_test.cc
moroten: Done

moroten:
Rebased and replaced `EXPECT_TRUE(err.empty())` with `EXPECT_EQ("", err)`.

PR #1318, opened 2017-09-04
fix normalizer test for _MAX_PATH

atetubou:
LongInvalidPath test sometimes fails due to GetFullPathName which does not allow to have more than _MAX_PATH length path.
The test sometimes successes depends on initial value of stack variable for GetFullPathName.

I fix this test to have cwd prefix in input path and confirmed the test passed in
* clang-cl with/without --debug in configure.py
* vs2017 cl.exe with/without --debug in configure.py

### DIFF
--- a/src/includes_normalize_test.cc
+++ b/src/includes_normalize_test.cc
@@ -108,36 +108,33 @@ TEST(IncludesNormalize, LongInvalidPath) {
       normalizer.Normalize(kLongInputString, &result, &err));
   EXPECT_EQ("path too long", err);
 
-  const char kExactlyMaxPath[] =
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "0123456789";
+
+  // Construct max size path having cwd prefix.
+  // kExactlyMaxPath = "$cwd\\a\\aaaa...aaaa\0";
+  char kExactlyMaxPath[_MAX_PATH + 1];
+  ASSERT_NE(_getcwd(kExactlyMaxPath, sizeof kExactlyMaxPath), NULL);
+
+  int cwd_len = strlen(kExactlyMaxPath);
+  ASSERT_LE(cwd_len + 3 + 1, _MAX_PATH)
+  kExactlyMaxPath[cwd_len] = '\\';
+  kExactlyMaxPath[cwd_len + 1] = 'a';
+  kExactlyMaxPath[cwd_len + 2] = '\\';
+
+  kExactlyMaxPath[cwd_len + 3] = 'a';
+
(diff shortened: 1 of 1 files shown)


src/includes_normalize_test.cc
```diff
@@ -108,36 +108,30 @@ TEST(IncludesNormalize, LongInvalidPath) {
...
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "012345678\\"
-      "0123456789";
+
+  // Construct max size path having cwd prefix.
+  // kExactlyMaxPath = "$cwd\\a\\aaaa...aaaa\0";
+  char kExactlyMaxPath[_MAX_PATH];
```
nico: Shouldn't this be _MAX_PATH + 1 (260 chars for the path, plus 1 for the terminating \0)?

src/includes_normalize_test.cc
```diff
@@ -108,36 +108,30 @@ TEST(IncludesNormalize, LongInvalidPath) {
...
+  kExactlyMaxPath[cwd_len + 1] = 'a';
+  kExactlyMaxPath[cwd_len + 2] = '\\';
+
+  kExactlyMaxPath[cwd_len + 3] = 'a';
+
+  for (int i = cwd_len + 4; i < _MAX_PATH - 1; ++i) {
+    kExactlyMaxPath[i] = 'a';
+  }
+
+  kExactlyMaxPath[_MAX_PATH - 1] = '\0';
```
nico: ...and then drop the -1 on this line and the next.

src/includes_normalize_test.cc
```diff
@@ -108,36 +108,30 @@ TEST(IncludesNormalize, LongInvalidPath) {
...
+  int cwd_len = strlen(kExactlyMaxPath);
+  ASSERT_LE(cwd_len + 3 + 1 + 1, _MAX_PATH)
+  kExactlyMaxPath[cwd_len] = '\\';
+  kExactlyMaxPath[cwd_len + 1] = 'a';
+  kExactlyMaxPath[cwd_len + 2] = '\\';
+
+  kExactlyMaxPath[cwd_len + 3] = 'a';
+
+  for (int i = cwd_len + 4; i < _MAX_PATH - 1; ++i) {
+    kExactlyMaxPath[i] = 'a';
```
nico: Also, could you do something like:

    if (i > cwd_len + 4 && i < _MAX_PATH-1 && i % 10 == 0)
      kExactlyMaxPath[i] = '\\';
    else
       kExactlyMaxPath[i] = 'a';

?

src/includes_normalize_test.cc
atetubou: Done

src/includes_normalize_test.cc
atetubou: Done

src/includes_normalize_test.cc
atetubou: Done

src/includes_normalize_test.cc
```diff
@@ -108,36 +108,33 @@ TEST(IncludesNormalize, LongInvalidPath) {
...
+  // kExactlyMaxPath = "$cwd\\a\\aaaa...aaaa\0";
+  char kExactlyMaxPath[_MAX_PATH + 1];
+  ASSERT_NE(_getcwd(kExactlyMaxPath, sizeof kExactlyMaxPath), NULL);
+
+  int cwd_len = strlen(kExactlyMaxPath);
+  ASSERT_LE(cwd_len + 3 + 1, _MAX_PATH)
+  kExactlyMaxPath[cwd_len] = '\\';
+  kExactlyMaxPath[cwd_len + 1] = 'a';
+  kExactlyMaxPath[cwd_len + 2] = '\\';
+
```
nico: Why this newline?

nico:
Thanks!

PR #1320, opened 2017-09-05
Fix confusing smart console output from concurrent builds

davezarzycki:
Developers tend to blame the last printed line when a build takes too
long. Unfortunately, when building concurrently, the last printed line
may have actually finished a long time ago. Under the current system,
ninja does not update the status line to reflect what jobs are still
running. This change makes ninja always print the longest running job
instead. In other words, the likely build bottlenecks.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -18,6 +18,7 @@
 #include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <limits.h>
 #include <functional>
 
 #ifdef _WIN32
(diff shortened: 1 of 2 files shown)


davezarzycki:
Hi @nico – Any chance you could please accept or reject this? Thanks!

nico:
First, sorry about the long delay for getting around to this (cf https://groups.google.com/forum/#!topic/ninja-build/yUOLRh19M-w).

Thanks for the patch, this seems like a really clean and clever solution to this problem. I merged it in #1405 (with attribution).

davezarzycki:
Neat-o! Thanks!

jhasse:
Hi! Due to #1418, I needed to revert this, sorry :/
nico also noticed a performance regression, see https://github.com/ninja-build/ninja/pull/1405#issuecomment-380080735.

Maybe we can think of a solution which avoids those two issues. For #1418 you can use `./misc/output_test.py` now to test it btw.

To keep the performance it might be a good idea to wait a second or two, before updating the status line due to a finished edge.

PR #1321, opened 2017-09-05
Make the default concurrency level conservative

davezarzycki:
The current "guess" logic assumes that I/O is slow and/or there isn't enough RAM to cache the build inputs. On machines with enough RAM, the current "guess" logic thrashes the CPU scheduler and CPU caches.

Finally, users can still override with -j if they need slightly more or less concurrency.

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -216,19 +216,6 @@ void Usage(const BuildConfig& config) {
           kNinjaVersion, config.parallelism);
 }
 
-/// Choose a default value for the -j (parallelism) flag.
-int GuessParallelism() {
-  switch (int processors = GetProcessorCount()) {
-  case 0:
-  case 1:
-    return 2;
-  case 2:
-    return 3;
-  default:
-    return processors + 2;
-  }
-}
-
 /// Rebuild the build manifest, if necessary.
 /// Returns true if the manifest was rebuilt.
 bool NinjaMain::RebuildManifest(const char* input_file, string* err) {
@@ -1025,7 +1012,7 @@ int ExceptionFilter(unsigned int code, struct _EXCEPTION_POINTERS *ep) {
 /// Returns an exit code, or -1 if Ninja should continue.
 int ReadFlags(int* argc, char*** argv,
               Options* options, BuildConfig* config) {
-  config->parallelism = GuessParallelism();
+  config->parallelism = GetProcessorCount();
 
   enum { OPT_VERSION = 1 };
   const option kLongOptions[] = {


davezarzycki:
Hi @nico – Any chance you could please accept or reject this? Thanks!

jhasse:
Why did you close this?

davezarzycki:
I’m closing my ninja pull requests in order to clean up my GitHub account.

Ninja, in my opinion, is an example of a reluctant open source project. If one looks at the ninja pull request queue, the project owner(s) are clearly uninterested in third-party contributions. They struggle to even acknowledge pull requests, let alone accept or reject them.

ocroquette:
This change might help too: https://github.com/ninja-build/ninja/pull/1398

PS: I agree that the feedback on pull requests could be better in this project...

nico:
Again, sorry about the slow reply, I went shortly after this pull request appeared. It's perfectly understandable that you're no longer interested in pursuing this.

Just to follow up: I vaguely remember there being some discussion on this pull request that I can't anymore. Maybe I'm misremembering. The discussion I remember was me saying that the default parallelism value was set fairly arbitrarily and that I buy that there's a better value for it – but that I'd like to see data to back up the new choice. Not because I don't believe you, but because I'd like to have an objective measure the next time someone wants to change the default.

nico:
@ocroquette This isn't setting the default -j to 1.

ocroquette:
My understanding is that the patch decreases the default value for -j to avoid to overload the CPU (I have the same problem on my Windows laptop, the default makes the whole OS very slow).

My point is that instead of making the default value for -j lower, one could also set the additional load limit (-l $CPUCOUNT). That would help in single-build, single-user case, but also for parallel builds, for instance in multi-users environments or on CI servers. I cannot see a case where it's a good idea to spawn more processes when all cores are busy.

davezarzycki:
Hi @nico – Sorry for getting impatient. I saw via another link that you were on paternity leave. Congrats!

Setting the default concurrency value doesn't need to be perfect or data driven. It just needs to be reasonable for most systems most of the time. The current behavior is awful for resource poor machines that cannot handle being overcommitted. This is one of a couple reasons why Apple sets `-j` to the number of CPUs available for their projects and projects created by their IDE (the other reason is because `make` defaults `-j` to 1).

I also wouldn't overthink `-j`. There will always be weird systems where the right value is beyond what any program can reasonably be expected to intuit. For example, I have a 4-way SMT box under my desk where the optimal `-j` value is half of the logical CPU count due to how the hardware SMT resources are subdivided. There is no way any program could be reasonably be expected to figure that out and I would never ask `ninja` to try.

Finally, in response to @ocroquette, I've think the `-l` feature needs some work before I'd recommend it to anybody. The problem is that the "load average" is not an instantaneous value. It is computed over a long time horizon relative to how quickly `ninja` can spawn/reap processes. For example, if I launch two `ninja -l $NCPU` jobs, then the real time load (not average!) will oscillate slowly between 0 and 2*$NCPU over the course of the two builds.

ocroquette:
I agree with Dave, the optimal value cannot be found automatically. For some systems, the current value $NCPU+2 is apparently OK, for my laptop, it's $NCPU/2 if SMT is enabled in the BIOS, $NCPU otherwise. On our CI server, $NCPU+2 is OK if a single build is running, but when several builds are running in parallel, $NCPU gives better overall results, while not impacting single builds significantly.

When the user has found the optimal value, it must be possible to store it easily so that it is not necessary to provide it on the command line on each call (it's not even possible when ninja is called from another tool). That's why something like https://github.com/ninja-build/ninja/pull/1399 is so important too.

davezarzycki:
Hi @ocroquette – How confident are you that your laptop has enough RAM? In general and unless one is low on RAM, SMT should be a small win for compile jobs due to how compiler algorithms and data structures work.

In any case, and to be totally honest, I don't think the current "guess" algorithm is defensible. When $NCPU equals one and RAM is in short supply, then running two jobs is the worst "guess" ever because both the CPU and the RAM end up being thrashed.

ocroquette:
My laptop has 16GB of RAM, which should be sufficient for 10 parallel builds. I just checked, the build requires up to 1GB for 6 parallel builds, and 2GB for 10, so the RAM is not the bottleneck in my case. I suspect it was the antivirus back then, but I cannot test this configuration anymore. This backs up the fact the optimal number for -j can only be found empirically, there are just too many factors: cores, RAM, I/O performance, antivirus... The current algorithm assumes that the CPU is always the bottleneck, but it's not.

PS: I am aware that an antivirus is not the best program to have running on a development machine, but there is no other choice for many employees.

src/ninja.cc
```diff
@@ -1025,7 +1012,7 @@ int ExceptionFilter(unsigned int code, struct _EXCEPTION_POINTERS *ep) {
 /// Returns an exit code, or -1 if Ninja should continue.
 int ReadFlags(int* argc, char*** argv,
               Options* options, BuildConfig* config) {
-  config->parallelism = GuessParallelism();
+  config->parallelism = GetProcessorCount();
```
sgraham: Does this need to handle 0 like the previous code?

davezarzycki: If `GetProcessorCount()` can return 0, then we should fix that, not the callers.

jhasse:
Also it's the least code lines.

ocroquette:
> While the anecdotal customization details are fun to read about

Sorry for hacking the thread ;) 

I actually understand the rational behind the -j NCORE + x, I used this heuristic myself with make on Linux systems to get better build performance than with -j NCORE.  So I can also understand the reluctance to use -j NCORE as a default, if NCORE + 2 worked fine so far for many people.

The customization would solve the problem, since anyone could set the default that works best, if the hard-coded heuristic picks unsatisfying values for the particular system.

Issue #1322, opened 2017-09-07
phony cycles not previously diagnosed now break since PR 1111


bradking:
CMake versions 2.8.12.x and 3.0.x incorrectly produce build statements of the form `build a: phony a` in certain cases.  This was a bug in CMake introduced by commit `v2.8.12~248^2` and fixed by commit `v3.1.0-rc1~353^2~2`.  See [CMake Issue 14972](https://gitlab.kitware.com/cmake/cmake/issues/14972).

Ninja didn't previously detect cycles involving only phony build statements, but does since PR #1111.  This broke `ninja`'s tolerance of the range of CMake versions with the above bug.

bradking:
Cc: @nico

The changes in #1111 do not break on correct build graphs and are very important in the work toward Ninja+Fortran support.  The detection of these cycles is a bug fix that has simply exposed a bug in old CMake versions that has also long been fixed.  IMO we should not revert the changes or introduce additional complexity in the graph/build logic just to support people using such old CMake versions.  If one is using a CMake that old one can use a Ninja that old too.

A middle-ground solution would be to tolerate the specific case produced by the buggy CMake versions without trying to tolerate phony cycles in general.  Prior to PR 1111 we tolerated both full cycles:

```
build a: phony b
build b: phony a
```

and self-referencing phony build statements:

```
build a: phony a
```

CMake 2.8.12 and 3.0 only produce the latter.  We can tolerate that with a hack in the parser to drop inputs of phony rules that correspond to their outputs.

nico:
Can we handle this similarly to #931? I.e.

1. Add `-w phonycycle=warn|err` flag, make that default to `warn` for now
2. If that's set to `warn`, do your #1323, but emit a warning
3. If that's set to err, keep current behavior (but maybe include "-w phonycycle" in the cycle error message if the cycle is from a phony edge)
4. At some point in the future, change the default to `err`
5. Eventually, remove the flag

?

bradking:
@nico yes a deprecation path will be good to help prevent similar generator bugs in the future.  I can work on that.  Do you intend to add such a warning in a 1.8 patch release, or start with the quick-fix from #1323 for 1.8.2 and add the warning in post-1.8 development?

bradking:
@nico I updated #1323 to add the `-w phonycycle=...` diagnostic.

PR #1323, opened 2017-09-07
Restore tolerance of self-referencing phony build statements

bradking:
Since PR #1111 we correctly reject self-referencing phony build statements like

    build a: phony a

as cycles.  Unfortunately this breaks support for CMake 2.8.12.x and
3.0.x because those versions incorrectly produce edges of this form
(that we used to tolerate).  In order to preserve compatibility with
those CMake versions we need to restore tolerance of these edges.

Add a special case to the manifest parser to filter out self-referencing
inputs of phony edges of the form produced by those CMake versions.
Warn by default, but add a `-w phonycycle={err,warn}` option to make it
an error.

Fixes: #1322

### DIFF
--- a/src/manifest_parser.h
+++ b/src/manifest_parser.h
@@ -31,10 +31,15 @@ enum DupeEdgeAction {
   kDupeEdgeActionError,
 };
 
+struct ManifestParserOptions {
+  ManifestParserOptions(): dupe_edge_action_(kDupeEdgeActionWarn) {}
+  DupeEdgeAction dupe_edge_action_;
+};
+
 /// Parses .ninja files.
 struct ManifestParser {
   ManifestParser(State* state, FileReader* file_reader,
-                 DupeEdgeAction dupe_edge_action);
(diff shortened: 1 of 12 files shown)


src/manifest_parser.h
```diff
@@ -31,10 +31,23 @@ enum DupeEdgeAction {
...
 };
 
+enum PhonyCycleAction {
+  kPhonyCycleActionWarn,
+  kPhonyCycleActionError,
+};
+
+struct ManifestParserOptions {
+  ManifestParserOptions():
+      dupe_edge_action_(kDupeEdgeActionWarn),
```
atetubou: Style like below is preferred.
```
  ManifestParserOptions()
      : dupe_edge_action_(kDupeEdgeActionWarn),
```
e.g. https://github.com/ninja-build/ninja/blob/master/src/graph.cc#L282

nico:
This looks great, thanks for the quick fix!

PR #1326, opened 2017-09-10
Two fixes for NetBSD

0-wiz-0:
Based on current patches in pkgsrc, please include them in the next release to reduce diffs.


jhasse:
Thanks for the PR!

(nitpick for next time: Please don't add a dot at the end of the first line of your commit message)

PR #1327, opened 2017-09-13
update RELEASING

atetubou:
This is update for RELEASING adding some steps what I did to release v1.8.2.

### DIFF
--- a/RELEASING
+++ b/RELEASING
@@ -1,19 +1,20 @@
 Notes to myself on all the steps to make for a Ninja release.
 
 Push new release branch:
-1. Consider sending a heads-up to the ninja-build mailing list first
-2. Make sure branches 'master' and 'release' are synced up locally
-3. update src/version.cc with new version (with ".git"), then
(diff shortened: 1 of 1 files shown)


RELEASING
```diff
@@ -1,20 +1,23 @@
 Notes to myself on all the steps to make for a Ninja release.
 
 Push new release branch:
-1. Consider sending a heads-up to the ninja-build mailing list first
-2. Make sure branches 'master' and 'release' are synced up locally
-3. update src/version.cc with new version (with ".git"), then
+1. File a tracking issue in bugs.chromium.org (e.g. crbug.com/762354)
```
nico: We shouldn't mention the chromium bug tracker in ninja's releasing doc. Ninja is an independent project and it logically sits below chromium (and many other projects). It's true that we use crbug to track ninja releases in chromium, but that shouldn't be in this doc here.

atetubou: I see, removed.

PR #1331, opened 2017-09-20
missingdeps tool, take 2

ilor:
This is a tool that detects some classes of random build failure causes around generated files (in Chrome, these are usually generated headers for C/C++). It takes several seconds to run, (more if the optional scan for files with no generator is used), and allows fixing a significant number of build system issues before they happen to randomly break someone's build.

The check relies on the correctness of depfile information, and on having all generator outputs listed. These things are usually easier to get right than proper deps. There's some overlap between this and gn check, but unlike gn check, this checks the actual ninja build graph and sees all issues.

This is a redo of https://github.com/ninja-build/ninja/pull/1031, it's been a while so I think it maybe better to start from scratch.


src/ninja.cc
```diff
@@ -505,6 +507,220 @@ int NinjaMain::ToolDeps(const Options* options, int argc, char** argv) {
...
+  }
+  vector<Edge*> missing_deps;
+  for (set<Edge*>::iterator de = deplog_edges.begin(); de != deplog_edges.end();
+       ++de) {
+    if (!state_->PathExistsBetween(*de, edge)) {
+      missing_deps.push_back(*de);
+    }
+  }
+
+  if (!deplog_nodes_without_generator.empty()) {
```
atetubou: Feel redundant.

src/ninja.cc
```diff
@@ -505,6 +507,220 @@ int NinjaMain::ToolDeps(const Options* options, int argc, char** argv) {
...
+              (*ne)->rule().name().c_str());
+          generator_rules_.insert(&(*ne)->rule());
+          missing_deps_rule_names.insert((*ne)->rule().name());
+        }
+      }
+    }
+    missing_dep_path_count_ += missing_deps_rule_names.size();
+    for (set<string>::iterator re = missing_deps_rule_names.begin();
+         re != missing_deps_rule_names.end(); ++re) {
+      //printf("Error: %s lacks a dep path to %s\n", node->path().c_str(),
```
atetubou: remove or uncomment?

ilor:
Thanks for the sugesstions so far. I'm also uploading example output of this tool, when run on a fairly recent chrome-linux build for the chrome target: [missingdeps-chrome-output.gz](https://github.com/ninja-build/ninja/files/1318295/missingdeps-chrome-output.gz)

The summary is:
```
Processed 39546 nodes.
Error: 161 file(s) under /src/chromium/src/out are deps log dependencies of other targets, but have no generator target listing them as outputs.
Error: There are 156 missing dependency paths.
75 targets had depfile dependencies on 87 distinct generated inputs (from 34 rules) without a non-depfile dep path to the generator.
There might be build flakiness if any of the targets listed above are built alone, or not late enough, in a clean output directory.
```

There's one case that's arguably a false-positive: `Missing dep: gen/components/resources/about_credits.html uses build.ninja (generated by gn)` which may require a special case to not treat it as an error.

I also would like to have some sort of test for this, but don't really know where such a test (that needs an after-a-build state, basically) would hae tobe.

ilor:
Ping?

ilor:
Ping? I still believe this would be useful to have in ninja.

ilor:
Ping? This could help fix a lot of rare clean-build build flakes in the chrome gn/ninja build, see https://bugs.chromium.org/p/chromium/issues/detail?id=655123 for a list of issues found.

jcrben:
I'm just a random web developer wanting to try building chromium on macOS 10.12.6 using the instructions at https://chromium.googlesource.com/chromium/src/+/67.0.3360.2/docs/mac_build_instructions.md, and running into issues.

Hoping that any efforts to reduce flakiness / missing deps can be merged, so +1 on the ping.

When I ran this tool, I got 
image

(note: this was not enough to help me fix my issue, which is apparently that mojom files are not generating their C++ files - chromium could perhaps use more instructions or more of a scripted setup)

bratell-at-opera:
I just ran into missing mojom.h files in a Chromium build and I wonder if this tool would have caught the problem when it was introduced. This seems like a good thing to have so I hope some maintainer can take a look.

src/state.cc
```diff
@@ -159,6 +159,23 @@ bool State::AddDefault(StringPiece path, string* err) {
...
 
+bool State::PathExistsBetween(Edge* from, Edge* to) {
+  EdgeAdjacencyMap::iterator it =
+      edge_adjacency_memoization_.find(make_pair(from, to));
+  if (it != edge_adjacency_memoization_.end())
+    return it->second;
+  bool found = false;
+  for (size_t i = 0; i < to->inputs_.size(); ++i) {
+    Edge* e = to->inputs_[i]->in_edge();
+    if (e && ((e == from) || PathExistsBetween(from, e))) {
```
atetubou:
```suggestion
    if (e && (e == from || PathExistsBetween(from, e))) {
```

src/ninja.cc
```diff
@@ -505,6 +507,213 @@ int NinjaMain::ToolDeps(const Options* options, int argc, char** argv) {
...
+      gen_dir = optarg;
+      break;
+    case 'h':
+    default:
+      printf("usage: ninja -t missingdeps [options] [targets]\n"
+"\n"
+"options:\n"
+"  -g DIR   assume files under DIR are generated (default: .)\n"
+"    DIR is relative to the build directory, empty string disables\n"
+"    checking for files with a missing generator\n"
```
atetubou: indent L690~L694

src/ninja.cc
```diff
@@ -505,6 +507,213 @@ int NinjaMain::ToolDeps(const Options* options, int argc, char** argv) {
...
+  if (!missing_deps.empty()) {
+    set<string> missing_deps_rule_names;
+    for (vector<Edge*>::iterator ne = missing_deps.begin();
+         ne != missing_deps.end(); ++ne) {
+      for (int i = 0; i < dep_nodes_count; ++i) {
+        if (dep_nodes[i]->in_edge() == *ne) {
+          generated_nodes_.insert(dep_nodes[i]);
+          printf("Missing dep: %s uses %s (generated by %s)\n",
+              node->path().c_str(), dep_nodes[i]->path().c_str(),
+              (*ne)->rule().name().c_str());
```
atetubou: indent L630~L631

src/ninja.cc
```diff
@@ -505,6 +507,213 @@ int NinjaMain::ToolDeps(const Options* options, int argc, char** argv) {
...
+  }
+}
+
+bool MissingDependencyScanner::IsNodeInGenDir(Node* node) {
+  if (abs_gen_dir_.empty())
+    return false;
+  string path = node->path();
+  char abs_path_buf[PATH_MAX];
+  char* rp = realpath(path.c_str(), abs_path_buf);
+  return (rp && (strlen(rp) > abs_gen_dir_.size())) &&
```
atetubou:
```suggestion
  return (rp && strlen(rp) > abs_gen_dir_.size()) &&
```

src/ninja.cc
```diff
@@ -505,6 +507,213 @@ int NinjaMain::ToolDeps(const Options* options, int argc, char** argv) {
...
+         (memcmp(rp, abs_gen_dir_.c_str(), abs_gen_dir_.size()) == 0);
```
atetubou:
```suggestion
         memcmp(rp, abs_gen_dir_.c_str(), abs_gen_dir_.size()) == 0;
```

ilor:
Thanks for taking a look. What would be the preferred form of these tests?

My initial idea was that to show the behavior, I'd need several build directories with manually written ninja files for some types of mistakes. Then from some runner script I'd build in each of these and then run the tool to make sure I get the expected diagnostic. Would that be okay? I don't recall anything quite like it in the repo, so I'm not quite sure where I'd put it.

Something like this, assuming there's a generator for foo.h but foo.o doesn't dep on it:

    cd missing-deps-test1
    ninja clean  # assuming this is clean enough
    ninja foo.h   # must be manually built first -- full builds are flaky
    ninja foo.o
    ninja -t missingdeps   # should report foo.o lacks a dep to foo.h's generator

then repeated for a few other cases.

src/state.cc
```diff
@@ -159,6 +159,23 @@ bool State::AddDefault(StringPiece path, string* err) {
   return true;
 }
 
+bool State::PathExistsBetween(Edge* from, Edge* to) {
```
atetubou: Better to detect cycle to prevent infinite recursion?

src/ninja.cc
```diff
@@ -505,6 +507,213 @@ int NinjaMain::ToolDeps(const Options* options, int argc, char** argv) {
...
+  if (!CollectTargetsFromArgs(argc, argv, &nodes, &err)) {
+    Error("%s", err.c_str());
+    return 1;
+  }
+  RealDiskInterface disk_interface;
+  MissingDependencyScanner scanner(&deps_log_, &state_, &disk_interface,
+                                   gen_dir);
+  for (vector<Node*>::iterator it = nodes.begin(); it != nodes.end(); ++it) {
+    scanner.ProcessNode(*it);
+  }
```
atetubou: Can we have gtest base test for MissingDependencyScanner?

ilor: OK, I will have to see what I can come up with. Are there any existing tests that set up a build graph and depfile?

src/state.cc
ilor: Hm, is it possible to have cycles by this point? I thought we'd would error out before getting there. Needs checking in a test I guess.

jhasse:
> My initial idea was that to show the behavior, I'd need several build directories with manually written ninja files for some types of mistakes. Then from some runner script I'd build in each of these and then run the tool to make sure I get the expected diagnostic. Would that be okay? I don't recall anything quite like it in the repo, so I'm not quite sure where I'd put it.

Not sure if this is what you're looking for, but there's https://github.com/ninja-build/ninja/blob/master/misc/output_test.py which runs the ninja binary on a temporary file and checks its output.

src/ninja.cc
```diff
@@ -514,6 +516,218 @@ int NinjaMain::ToolDeps(const Options* options, int argc, char** argv) {
...
+    : deps_log_(deps_log), state_(state), disk_interface_(disk_interface),
+      missing_dep_path_count_(0) {
+  if (!gen_dir.empty()) {
+    char buf[PATH_MAX];
+    if (getcwd(buf, PATH_MAX) == NULL)
+      Fatal("getcwd failed");
+    abs_gen_dir_ = buf;
+    abs_gen_dir_.append("/");
+    abs_gen_dir_.append(gen_dir);
+    if (!realpath(abs_gen_dir_.c_str(), buf))
```
ilor: msvc/mingw don't like the realpath calls (AppVeyor build failed). Any recommendations?

ilor:
Admittedly, the timing is not great with regards to me being able to spend as much time as I'd like on this. That said, I've rebased to master and fixed some minor issues, but will need a while for the tests.

The "main" obstacle is that the code uses classes like ImplicitDepLoader which want to load a real file and aren't triviallyfed test data; I guess I can use a temp file for now as I see some tests do that.

src/ninja.cc
jhasse: You can use `std::filesystem::canonical`: https://en.cppreference.com/w/cpp/filesystem/canonical

You can use `#if _MSC_VER > 1910` to detect if its available and otherwise disable your tool (not really a problem, as our Windows release is build with Visual Studio 2017).

ilor: Hm, not that I mind using C++17, but conditionally disabling the tool may be fairly ugly (no tools do that as far as I see?)

I noticed there's a Windows-specific AbsPath in src/includes_normalize.h, I can try moving it to src/util.h/cc so I can use it in the tool (may have to move the currently-local  InternalGetFullPathName as well to avoid changing too much other code), I think that'd be preferable over #including a weird header in a dependency analysis tool.

jhasse: The browse tool is disabled on Windows and on any platform when `NINJA_PYTHON` isn't available at runtime.

ilor: oh, thanks. I'll try it out and see how it goes, time permitting.

jonesmz:
@ilor Could you rebase this PR on the latest master branch?

ilor:
I'm working on this and should have something this week. It will be simplifies without the "files that lack a generator target" part (-g switch) because that adds a lot of complexity that's not needed to see the primary feature of the tool.

ilor:
Highlights from the update:
* Rebased onto master
* Separate cc/h pair for the tool, as it is large enough
* Unit test for the primary thing the tool should catch
* Remove -g switch and related code
* Avoid touching state.h/cc, keep the edge adjacency map

Removing -g and extracting to a separate source file simplified things a lot, and as a bonus now Windows should be able to build this without excessive ifdeffery or build config. I never found the -g tool to be very useful, it can be added later if there's a serious need.

Please take a look

src/missing_deps.h
```diff
@@ -0,0 +1,79 @@
...
+
+ private:
+  struct EdgePair {
+    Edge* from;
+    Edge* to;
+    bool operator<(const EdgePair& other) const {
+      return (from < other.from) || ((from == other.from) && (to < other.to));
+    }
+  };
+  // TODO: use a hash map here (~2x performance in a quick test)
```
jhasse: Try:

```cpp
#if __cplusplus >= 201103L
  using EdgeAdjacencyMap = std::unordered_map<EdgePair, bool>;
#else
  typedef std::map<EdgePair, bool> EdgeAdjacencyMap;
#endif
```

src/missing_deps_test.cc
```diff
@@ -0,0 +1,160 @@
...
+  AssertMissingDependencyBetween("generated_header", "compiled_object",
+                                 &compile_rule_);
+}
+
+TEST_F(MissingDependencyScannerTest, CycleInGraph) {
+  CreateInitialState();
+  CreateGraphDependencyBetween("compiled_object", "generated_header");
+  CreateGraphDependencyBetween("generated_header", "compiled_object");
+  // The missing-deps tool doesn't deal with cycles in the graph, beacuse
+  // there will be an error loading the graph befoer we get to the tool.
```
jhasse: typo: befoer

src/missing_deps.h
ilor: Will do. How do I configure the build to be in c++11 mode on Linux, by the way? I checked with a patched configure.py for now.

jhasse: You can use the `CXXFLAGS` environment variable. Newer compilers default to C++11 mode btw.

src/missing_deps.h
```diff
@@ -60,8 +60,6 @@ namespace std {
 template <>
 struct hash<EdgePair> {
   size_t operator()(const EdgePair& k) const {
-    // Compute individual hash values for first, second and third
-    // http://stackoverflow.com/a/1646913/126995
     uintptr_t uint_from = (uintptr_t)(k.from);
     uintptr_t uint_to = (uintptr_t)(k.to);
     return hash<uintptr_t>()(uint_from ^ (uint_to >> 3));
```
ilor: this is just a simplistic "hash" to get an unordered_map going, ended up plenty faster than map

src/missing_deps.h
```diff
@@ -0,0 +1,103 @@
...
+    return (from < other.from) || ((from == other.from) && (to < other.to));
+  }
+};
+
+#if __cplusplus >= 201103L
+namespace std {
+template <>
+struct hash<EdgePair> {
+  size_t operator()(const EdgePair& k) const {
+    uintptr_t uint_from = (uintptr_t)(k.from);
```
jhasse: Please use `static_cast` instead of C-style cast. Or maybe the following will also work:
```cpp
uintptr_t uint_from{k.from};
```

src/ninja.cc
ilor: this is now obsolete because the -g tool that required abspath is no longer part of this pull request to keep it more manageable

src/missing_deps.h
ilor: done

src/missing_deps.h
ilor: uinptr_t(k.from) works

src/missing_deps_test.cc
ilor: fixed in fixup

src/ninja.cc
ilor: gtests now included

src/state.cc
ilor: checked in a gtest that the code will error out before we get there

src/state.cc
ilor: done (code moved to src/missing_deps.cc)

src/ninja.cc
ilor: obsolete after code move to src/missing_deps.cc

src/ninja.cc
ilor: obsolete after -g tool removal

src/ninja.cc
ilor: obsolete after -g tool removal

src/ninja.cc
ilor: obsolete after code move to src/missing_deps.cc (with cleanups like this)

src/ninja.cc
ilor: I think this is also obsolete after code move to src/missing_deps.cc and -g tool removal

ilor:
Thanks for the comments so far, should be all addressed. Please let me know what I can do next to get this into a merge-able state.

ilor:
Friendly ping

src/graph.cc
```diff
@@ -554,24 +554,26 @@ bool ImplicitDepLoader::LoadDepFile(Edge* edge, const string& path,
...
-
-  // Add all its in-edges.
   for (vector<StringPiece>::iterator i = depfile.ins_.begin();
-       i != depfile.ins_.end(); ++i, ++implicit_dep) {
+       i != depfile.ins_.end(); ++i) {
     uint64_t slash_bits;
     if (!CanonicalizePath(const_cast<char*>(i->str_), &i->len_, &slash_bits,
                           err))
       return false;
-
```
jhasse: why did you remove this line?

ilor: no strong reason or strong opinion, I'll just put it back in

ilor:
Thanks for replying, I understand. While I think this can be very useful, it's not something I have to rush to any particular version.

I'll try to ping again in a week or a couple, maybe clean-up the branch history in the meantime (unless that's not welcome at this stage; I'm not sure how helpful/disruptive it'd be in the middle of a  pull request).

src/missing_deps_test.cc
```diff
@@ -0,0 +1,160 @@
+// Copyright 2011 Google Inc. All Rights Reserved.
```
atetubou: 2019?

ilor:
ping?

src/graph.cc
```diff
@@ -507,9 +507,9 @@ bool ImplicitDepLoader::LoadDeps(Edge* edge, string* err) {
   return true;
 }
 
-bool ImplicitDepLoader::LoadDepFile(Edge* edge, const string& path,
-                                    string* err) {
-  METRIC_RECORD("depfile load");
+bool ImplicitDepLoader::LoadDepfileDeps(Edge* edge, string path,
+                                        vector<Node*>* depfile_deps,
```
jonesmz: Consider using std:: for types from the standard library for newly added code.

src/graph.cc
```diff
@@ -554,24 +554,27 @@ bool ImplicitDepLoader::LoadDepFile(Edge* edge, const string& path,
...
     Node* node = state_->GetNode(*i, slash_bits);
-    *implicit_dep = node;
-    node->AddOutEdge(edge);
-    CreatePhonyInEdge(node);
+    depfile_deps->push_back(node);
   }
+  return true;
+}
 
+bool ImplicitDepLoader::LoadDepFile(Edge* edge, const string& path,
```
jonesmz: Is it practical for this string to instead be StringPiece?

src/graph.cc
```diff
@@ -591,15 +594,18 @@ bool ImplicitDepLoader::LoadDepsFromLog(Edge* edge, string* err) {
...
-  vector<Node*>::iterator implicit_dep =
-      PreallocateSpace(edge, deps->node_count);
-  for (int i = 0; i < deps->node_count; ++i, ++implicit_dep) {
-    Node* node = deps->nodes[i];
+  ApplyLoadedDeps(edge, deps->nodes, deps->node_count);
+  return true;
+}
+
+void ImplicitDepLoader::ApplyLoadedDeps(Edge* edge, Node** nodes, int node_count) {
+  vector<Node*>::iterator implicit_dep = PreallocateSpace(edge, node_count);
```
jonesmz: Only now being made aware of the PreallocateSpace function, I think it's an abomination.

Can you instead use std::vector::reserve()?

I'd also argue that you should eliminate that function entirely, but if you don't I certainly will.

src/graph.cc
```diff
@@ -554,24 +554,27 @@ bool ImplicitDepLoader::LoadDepFile(Edge* edge, const string& path,
...
+}
 
+bool ImplicitDepLoader::LoadDepFile(Edge* edge, const string& path,
+                                    string* err) {
+  METRIC_RECORD("depfile load");
+  std::vector<Node*> depfile_deps;
+  if (!LoadDepfileDeps(edge, path, &depfile_deps, err))
+    return false;
+  if (!depfile_deps.empty())
+    ApplyLoadedDeps(edge, &depfile_deps[0], depfile_deps.size());
```
jonesmz: std::vector<>::data()

But would you consider taking an iterator pair instead of a pointer-to-pointer, and size_t?

src/graph.h
```diff
@@ -247,6 +254,10 @@ struct ImplicitDepLoader {
   /// @return false on error (without filling \a err if info is just missing).
   bool LoadDepsFromLog(Edge* edge, string* err);
 
+  /// Insert the loaded implicit dependencies (\a nodes) of \a edge
+  /// into the graph.
+  void ApplyLoadedDeps(Edge* edge, Node** nodes, int node_count);
```
jonesmz: node_count should be size_t, that's the representation for "number of objects in memory"

src/missing_deps.cc
```diff
@@ -0,0 +1,160 @@
...
+#include "disk_interface.h"
+#include "graph.h"
+#include "state.h"
+#include "util.h"
+
+MissingDependencyScannerDelegate::~MissingDependencyScannerDelegate() {}
+
+void MissingDependencyPrinter::OnMissingDep(Node* node, const std::string& path,
+                                            const Rule& generator) {
+  printf("Missing dep: %s uses %s (generated by %s)\n", node->path().c_str(),
```
jonesmz: You might consider using iostreams, instead of printf. Printf is so.... ugh. For lack of better terminology to use.

src/missing_deps.cc
```diff
@@ -0,0 +1,160 @@
...
+    if (deps)
+      ProcessNodeDeps(node, deps->nodes, deps->node_count);
+  } else {
+    DepfileParserOptions parser_opts;
+    parser_opts.depfile_distinct_target_lines_action_ =
+      kDepfileDistinctTargetLinesActionWarn;
+    ImplicitDepLoader dep_loader(state_, deps_log_, disk_interface_,
+                                 &parser_opts);
+    vector<Node*> depfile_deps;
+    string depfile = edge->GetUnescapedDepfile();
```
jonesmz: consider `std::` for newly added code.

jonesmz:
Given the upcoming rule change (https://github.com/ninja-build/ninja/pull/1616/files):

> * Any code merged into the Ninja codebase which will be part of the main
>   executable must compile as C++03. You may use C++11 features in a test or an
>   unimportant tool if you guard your code with `#if __cplusplus >= 201103L`.

You might consider adopting C++11 entirely for this new tool. It would allow for things like ranged based for loops, which would benefit several places in your newly added code.

src/missing_deps.cc
ilor: Fully agree on the terminology -- I used printf solely based on the fact that there are 0 (zero) usages of iostreams across ninja right now. If iostreams are "okay" (especially in a tool) then sure,

src/missing_deps.cc
```diff
@@ -0,0 +1,160 @@
...
+  vector<Edge*> missing_deps;
+  for (set<Edge*>::iterator de = deplog_edges.begin(); de != deplog_edges.end();
+       ++de) {
+    if (!PathExistsBetween(*de, edge)) {
+      missing_deps.push_back(*de);
+    }
+  }
+
+  if (!missing_deps.empty()) {
+    set<string> missing_deps_rule_names;
```
jonesmz: Could you use StringPiece here, instead of string? Saves allocations.

src/graph.cc
ilor: I don't think it would be useful, there is a grand total of two call-sites for this and both have a std::string instance at the ready.

src/missing_deps.cc
```diff
@@ -0,0 +1,160 @@
...
+    printf(
+        "There might be build flakiness if any of the targets listed above are "
+        "built alone, or not late enough, in a clean output directory.\n");
+  } else {
+    printf("No missing dependencies on generated files found.\n");
+  }
+}
+
+bool MissingDependencyScanner::PathExistsBetween(Edge* from, Edge* to) {
+  EdgePair key;
```
jonesmz: Why not provide a constructor for EdgePair, so you can formulate this as

EdgePair key(from, to);

?

src/missing_deps.h
```diff
@@ -0,0 +1,103 @@
...
+  DiskInterface* disk_interface_;
+  std::set<Node*> seen_;
+  std::set<Node*> nodes_missing_deps_;
+  std::set<Node*> generated_nodes_;
+  std::set<const Rule*> generator_rules_;
+  int missing_dep_path_count_;
+
+ private:
+#if __cplusplus >= 201103L
+  using EdgeAdjacencyMap = std::unordered_map<EdgePair, bool>;
```
jonesmz: Wouldn't the hash_map implementation found in hash_map.h be sufficient for this?

src/ninja.cc
```diff
@@ -915,7 +940,7 @@ const Tool* ChooseTool(const string& tool_name) {
     printf("ninja subtools:\n");
     for (const Tool* tool = &kTools[0]; tool->name; ++tool) {
       if (tool->desc)
-        printf("%10s  %s\n", tool->name, tool->desc);
+        printf("%11s  %s\n", tool->name, tool->desc);
```
jonesmz: Could you elaborate on this change?

src/graph.cc
jonesmz: Alright.

src/graph.cc
ilor: As much as I don't fancy _either_ PreallocateSpace _or_ vector::reserve(), this looks a bit too close to the hot path to fiddle with as part of a larger change. I'd much rather see such changes done separately.

src/missing_deps.cc
jonesmz: While there's zero use of iostreams in terms of `std::cout`, there are several places where std::stringstream is used to construct a string (which, in a few situations, is then passed to fprintf...?)

While I am not an official project maintainer, and therefore one of the official maintainers can just overrule me, personally I think (for all it's warts...) iostreams is miles and miles ahead of fprintf, not the least reason of which is that it doesn't explicitly require nul-terminated strings.

src/ninja.cc
ilor: This is to align the list of tools properly -- the "missingdeps" line stands out otherwise with its short description. That's because "missingdeps" happens to be longer than the other tool names.
Roughly:
```
     browse  browse dependency graph in a web browser
      clean  clean built files
   commands  list all commands required to rebuild given targets
       deps  show dependencies stored in the deps log
missingdeps  check deps log dependencies on generated files
```
vs
```
    browse  browse dependency graph in a web browser
     clean  clean built files
  commands  list all commands required to rebuild given targets
      deps  show dependencies stored in the deps log
missingdeps  check deps log dependencies on generated files
```

src/missing_deps.h
ilor: As far as I can tell no, it seems only useful for maps keyed by a StringPiece.

src/graph.cc
jonesmz: Sure.

src/missing_deps.h
jonesmz: Ah, yes, you're correct. my mistake.

src/ninja.cc
jonesmz: Sounds good.

src/graph.cc
ilor: Thanks for pointing out the upcoming rule change!

src/graph.h
ilor: DepsLog::Deps node_count is an int though, around src/deps_log.h:84.

src/graph.cc
ilor: The other callsite uses deps->nodes, deps->node_count; DepsLog::Deps has a good ol' plain array of pointers.

ilor: Also, vector::data() is c++11 :(

src/missing_deps.cc
ilor: It's somewhat awkward. This StringPiece doesn't work in a set since it has no operator<, and there's no hash_set typedef either.

src/missing_deps.cc
ilor: Let's see how it goes with std::cout.

src/graph.h
jonesmz: Alright. That's horrible, but you're right that it should be an int.

src/graph.cc
jonesmz: It is? Did not know that. Thanks for the insights. 

All the more reason that Ninja should switch to C++11 fully, but *shrug*.

src/missing_deps.cc
jonesmz: Makes sense.

src/missing_deps.h
```diff
@@ -45,14 +45,15 @@ class MissingDependencyPrinter : public MissingDependencyScannerDelegate {
 };
 
 struct EdgePair {
-  Edge* from;
-  Edge* to;
+  EdgePair(Edge* from_, Edge* to_) : from(from_), to(to_) {}
```
jonesmz: I'm not trying to nitpick you, sorry :(

I think that the typical Ninja coding conventions have the underscore for member variables, and the constructor parameters are underscore-less.

src/missing_deps.cc
```diff
@@ -0,0 +1,158 @@
+// Copyright 2019 Google Inc. All Rights Reserved.
```
jonesmz: IANAL -- but I am very familiar with copyright issues, having worked with them plenty over my career.

Please consider carefully the license header, unless you are a Google employee.

While the Ninja project does have the explicit guideline that all files have "Copyright *year* Google Inc.", this is very much not the correct thing to do for non-Google employees.

If you are not a Google employee, then you are the owner of the copyright for newly written code. In such case, you can't simply say "Google owns this" without some explicit agreement with them to have the copyright transfer.

src/missing_deps.h
ilor: I don't mind at all

src/missing_deps.cc
jonesmz: Allow me to clarify my poor wording.

*obviously* you can *say* whatever you want.

I meant "Saying that Google holds the copyright on this code is not factually correct simply because the file says so. The copyright is retained by the actual author in non-"work for hire" situations unless there's an explicit agreement between the author and Google saying that the rights have been transferred".

ilor: I believe this would be more appropriate to discuss in https://github.com/ninja-build/ninja/pull/1616 that explicitly mentions the licensing, vs in a random change that happens to add a file.

And anyway, the change here is from a corporate email address with a google CLA in place already due to contributions elsewhere; and https://opensource.google.com/docs/releasing/contributions/#copyright doesn't indicate worrying about the copy header is required.

jonesmz: No need for you to do anything. I only wanted to make sure you were aware.

src/missing_deps.h
jhasse: btw: There's no need to use a different name,  `EdgePair(Edge* from, Edge* to) : from(from), to(to) {}` works.

src/graph.cc
jhasse: What's the advantage of std::vector<>::data()?

ilor: Only that it's arguably nicer as &v[0] is unusual-looking.

src/missing_deps.h
ilor: yeah, I'm aware, but I really don't like relying on this particular nook of c++ name scoping

src/graph.cc
jhasse: Actually I just found out that there's another important difference: It's always okay to call `data()` on an empty vector. See 95e0cb5156d295e0510b5296e441eecb1f8b1806

ilor:
Friendly/festive ping

jhasse:
Please have a look at the conflict in src/graph.cc.

ilor:
Thanks for pointing that out. Rebased on top of master to fix the conflict an cleaned up branch history while at it. Then noticed there's a cmake build now too, which the last fixup deals with.

jhasse:
It's really time for 1.10 to be released and I think we already included lots of useful stuff, which needs testing.

Therefore I'm assigning this PR to 1.11, hope you're okay with that. I just didn't have the time to test this thoroughly and it includes some non-tool-only changes ;)

ilor:
That's fine. I assume I should wait until the 1.10 release and then rebase the docs then so they say ... since 1.11 etc.?

jhasse:
Yes. No need to wait for the release for updating the docs though.

ilor:
Apologies for the delas. Rebased with no conflicts, marked doc to indicate 1.11 as the version.

ilor:
Friendly ping

WhiteHalmos:
This PR has been super useful whenever I change dependencies (I compiled ninja with this patch). Looking forward to this being merged into master.

ilor:
ping

ilor:
Ping

deepakcse87:
Nice tool. Looking forward to use it when it gets merged into master.

ilor:
@jhasse friendly ping

jhasse:
This PR is on the 1.11.0 milestone, I will definitely have a look at it before doing that release. Currently I'm working on 1.10.2 though, please be patient.

ilor:
That's good to hear, I'll be waitng.

jhasse:
Can you rebase on master? I think I found a crash in the unit test due to an assert.

ilor:
The test crash seems to be because the test code that writes a depfile needs to be adjusted after cc79afbc0510b7320d35fb481f0f200dec92cb43 changed some bits in deps_log.cc

src/missing_deps.h
```diff
@@ -94,12 +68,14 @@ struct MissingDependencyScanner {
   int missing_dep_path_count_;
 
  private:
-#if __cplusplus >= 201103L
-  using EdgeAdjacencyMap = std::unordered_map<EdgePair, bool>;
+#if (__cplusplus >= 201103L) || (_MSC_VER >= 1900)
```
jhasse, line 71: `_MSC_VER >= 1900` shouldn't be needed as MSVC sets `__cplusplus` correctly with `/Zc:__cplusplus`.

ilor: ack

ilor:
ping -- this should be better now, the macos failure looks like a infra issue

jhasse (changes requested):
Thanks! The code looks good :) There's one performance impact though:

While `implicit_dep` is still preallocated, the new temporary vector `depfile_deps` will grow on the fly. It's also now two loops because of the temporary vector.

Either

1. show that this has no significant performance impact with an appropriate benchmark,
2. or pass a callback to `LoadDepfileDeps` instead of a pointer to an vector. Call it and pass the node instead of doing `depfile_deps->push_back(node);`.

ilor:
@jhasse thanks, with all the quaint pre-C++11 support going on "use a callback" is a bit less enticing than it should otherwise be but I'll probably end up doing that. I may follow up with some perf experiments later if I find the time because I do feel like one extra vector and some pointer copies there should be immaterial.

jhasse:
I hope it's okay I've tried the "Allow changes from maintainers" feature of GitHub and pushed to your branch 🙈 

It isn't pretty, especially `PreallocateSpace` and `CreatePhonyInEdge` shouldn't be public. Maybe you have a better idea.

ilor:
I think the virtual function approach works out better now, especially when viewed vs the base branch. Should be little effectively no changes on the main code path now.

ilor:
friendly ping

src/graph.cc
```diff
@@ -525,9 +525,14 @@ struct matches {
   std::vector<StringPiece>::iterator i_;
 };
 
-bool ImplicitDepLoader::LoadDepFile(Edge* edge, const string& path,
-                                    string* err) {
+bool ImplicitDepLoader::LoadDepFile(Edge* edge, const std::string& path,
+                                    std::string* err) {
```
jhasse, line 529: please remove `std::` in both cases to show that this function's signature hasn't changed.

src/graph.cc
```diff
@@ -660,3 +670,26 @@ void ImplicitDepLoader::CreatePhonyInEdge(Node* node) {
   // this node, it will simply set outputs_ready_ to the correct value.
   phony_edge->outputs_ready_ = true;
 }
+
+bool ImplicitDepNodeLoader::LoadDepfileDepNodes(Edge* edge, std::string path,
+                                                std::vector<Node*>* dep_nodes,
+                                                std::string* err) {
+  dep_nodes_ = dep_nodes;
+  bool ret = LoadDepfileDeps(edge, path, err);
+  dep_nodes_ = NULL;
```
jhasse, line 679: What if dep_nodes_ wasn't `NULL` before?

src/graph.h
```diff
@@ -271,6 +282,30 @@ struct ImplicitDepLoader {
   DepfileParserOptions const* depfile_parser_options_;
 };
 
+/// ImplicitDepLoader variant that returns dep nodes without updating graph
+/// deps.
+struct ImplicitDepNodeLoader : public ImplicitDepLoader {
```
jhasse, line 287: Move the whole class to missing_deps.h/cc?

src/missing_deps_test.cc
```diff
@@ -0,0 +1,162 @@
...
+                                 &compile_rule_);
+}
+
+TEST_F(MissingDependencyScannerTest, CycleInGraph) {
+  CreateInitialState();
+  CreateGraphDependencyBetween("compiled_object", "generated_header");
+  CreateGraphDependencyBetween("generated_header", "compiled_object");
+  // The missing-deps tool doesn't deal with cycles in the graph, beacuse
+  // there will be an error loading the graph before we get to the tool.
+  // This test is to illstrate that.
```
jhasse, line 157: typo: illustrate

jhasse (commented):
That's a good solution :)

src/graph.cc
jhasse: Okay that can't happen. But why even set it to `NULL` here?

ilor: I did this to avoid holding on to a potentially-dangling pointer. This is admittedly a bit dodgy if someone were to call LoadDeps on this loader afterwards. I'm considering removing LoadDepfileDepNodes and adding a direct setter for dep_nodes_, so the calling code would be
```
dep_loader.SetDepNodesOutput(&depfile_deps);
dep_loader.LoadDeps(edge, depfile, &err);
```
instead of the current
```
dep_loader.LoadDepfileDepNodes(edge, depfile, &depfile_deps, &err);
```
Or maybe even just take the vector pointer in the constructor of the custom dep loader to keep things more straightforward since the pointer can then be easily shown as something necessary. That seems the way to go.

jhasse: > Or maybe even just take the vector pointer in the constructor of the custom dep loader to keep things more straightforward since the pointer can then be easily shown as something necessary.

+1

src/graph.cc
ilor: done, a lot less diff all over graph.h/cc now

src/graph.cc
ilor: PTAL now

jhasse (approved):

Issue #376, opened 2012-07-27
Use better logic for picking which node to build first


evmar:
See discussion in https://github.com/martine/ninja/pull/60

orenbenkiki:
Is this likely to be merged? I have a pretty involved old-style Makefile I would like to speed up. My main problem is critical path scheduling - I have some tests that take a long time (almost a minute) to run, and it is nearly impossible to trick make -j to schedule these ahead doing a lot of other smaller tasks that could happen in parallel. I can live with the ~10 seconds overall make tax, but waiting another minute for the build to finish because of bad scheduling is starting to get to me. I was hoping ninja could solve my problem, but if it doesn't do critical path analysis either, then it seems all I'll gain is the 10 seconds or so, not worth the effort of converting the Makefile...

nico:
I tried that on larger projects, and in practice it seemed to do way worse than the current scheduling approach, maybe because it's worse for the disk cache. In practice, most processes seem to be shortlived and there are many of them, so being nice to the disk cache (which the current behavior is) seems more important than being smart about the order. If there are long-running tasks like in your case, the tradeoff is different though.

Have you tried using the branch linked to above and measure how much it helps?

orenbenkiki:
What I did is log (out of the existing Makefile) when does each task start, end, and how much CPU % it consumes (this would be lower for an IO intensive task). I then visualized this on a diagram. I can see that my few long-living, high-CPU tasks come "too late", while I a ton of short-lived, low CPU % tasks come at the start (where Make doesn't even manage to spawn enough of them to saturate the CPUs). If I could manually rearrange the order they were executed, I could reduce my 3.5 minute build time by about 1 minute (the dependencies allow for that). Also, if I could say "run more of the low-CPU % tasks in parallel" I could probably get it down for 2m for a full build. I think I should be able to do that by using pools in Ninja... 

The Makefile is pretty involved (several hundred lines of rules, using GNU make macros and wildcard expansions and so on). Converting it to Ninja would take "some effort" - I'm not even certain what the best way would be (would I be able to control everything I want using CMAKE, such as pools etc.? Should I just write some shell/python/ruby script that directly spits out the Ninja rules? etc.).

[...]

PR #1333, opened 2017-09-21
support explicit build order

### LINKED ISSUE #232: explicit control of the build order
It seems to me that all ready to build Edges are contained in a std::set<Edge*> (Plan::ready_). That is, they are sorted by the pointer value comparison. In mostly cases, the order comes out in line with the order build commands are read and parsed from ninja files. And this is ok for us as we generate projects in order of their dependency orders. 

However, sometimes, it may happen that a cpp file from a lower level project is not schedules to compile until much later into the build process, and therefore blocking up all the linking actions, making the total build time unnecessarily long. This typically happens with a clean build, when at the start of ninja, thousands of cpp files are ready to be compiled.

I propose we add some explicit control of the build order for ready targets. Let's say every Edge object should have a member of buildorder, then Plan::ready_ can be changed to set with a compare function, which compare the buildoder member of edges.

There are a few options to determine the build order value for the edges. By default, ninja can assign a sequential number for every Edge object created. We may also introduce a build_order variable to build commands, therefore giveing the generator full control of the order.

A further benefit when above is acomplished, we can indroduce a ninja option that generates randomised build order for all edges. This would be useful to find subtle dependency problems in generated ninja files.

comicfans:
this work based on nico's critical path schedule code, plus suporting
user controlled explicit build order. when targets is given in command line ,
try to build earlier appeared target first, for example
ninja verylongjob all
user can use this feature to let ninja schedule long job first to
improve overall build time. after initial build ,following builds can
also pick history data to find shortest build plan.
 
depends on target build time , this implementation may have a limitation which user may not control more than 42 targets build in order (use uint64_t to determine critical time ,if last build job is 1 hour ,about 2^22 ms , one higher priority job may double that number), ninja still work when this limit exceeds, just can not pick the ordered target (they may have same uint64 max critical time)  I think this should be enough for most build.

my test result posted here #376 
it may also resolve #232


src/build.cc
```diff
@@ -474,6 +486,133 @@ bool Plan::CleanNode(DependencyScan* scan, Node* node, string* err) {
...
+              done.insert(in);
+          }
+      }
+    }
+    prev_top_crit = this_top_crit;
+  }
+
+  // 2. Build priority list in decreasing order of critical times.
+  sort(edges.begin(), edges.end(), EdgeCom);
+  if (true) {
```
atetubou: need this if?

src/build.cc
```diff
@@ -474,6 +486,133 @@ bool Plan::CleanNode(DependencyScan* scan, Node* node, string* err) {
...
+namespace {
+bool EdgeCom(const Edge* lhs, const Edge* rhs) {
+  // Note: > to sort in decreasing order.
+  return lhs->critical_time() > rhs->critical_time();
+}
+}  // namespace
+
+void Plan::ComputePriorityList(BuildLog* build_log) {
+
+  for (vector<Node*>::iterator it=targets_.begin();it!=targets_.end();++it){
```
atetubou: space after ';'?

src/build.cc
```diff
@@ -474,6 +486,133 @@ bool Plan::CleanNode(DependencyScan* scan, Node* node, string* err) {
...
+      //remove duplicate targets
+      targets_.resize(find(it+1,targets_.end(),*it)-targets_.begin());
```
atetubou: space after ','?

src/build.cc
```diff
@@ -474,6 +486,133 @@ bool Plan::CleanNode(DependencyScan* scan, Node* node, string* err) {
...
+    }
+
+    int duration = entry->end_time - entry->start_time;  // XXX: + 1?
+    edge->run_time_ms_ = duration;
+    // XXX: also count on rule, and globally
+  }
+
+
+  // Dump graph to stdout for debugging / prototyping.
+  if (false) {
```
atetubou: need this if?

src/build.cc
```diff
@@ -474,6 +486,133 @@ bool Plan::CleanNode(DependencyScan* scan, Node* node, string* err) {
...
+      continue;
+    Edge* e = it->first;
+    edges.push_back(e);
+  }
+  
+
+  // Critical path scheduling.
+  // 0. Assign costs to all edges, using:
+  // a) The time the edge needed last time, if available.
+  // b) The average time this edge type needed, if this edge hasn't run before.
```
atetubou: b) is implemented?

src/build.cc
```diff
@@ -474,6 +486,133 @@ bool Plan::CleanNode(DependencyScan* scan, Node* node, string* err) {
...
+    // source file cxx lines are added. Without this, the edge generating that
+    // binary would be added ot the queue 50 times.
+    set<Edge*> done;
+
+    Edge* first = (*tit)->in_edge();
+    if (!first){
+        continue;
+    } 
+    edgesQ.push(first);
+
```
atetubou: insert first to done here?

src/build.cc
```diff
@@ -474,6 +486,133 @@ bool Plan::CleanNode(DependencyScan* scan, Node* node, string* err) {
...
+      Node* output = edge->outputs_[0];
+      printf("%s %s %d\n", input->path().c_str(), output->path().c_str(),
+             edge->run_time_ms_);
+    }
+  }
+
+
+  uint64_t prev_top_crit=0;
+
+  for (vector<Node*>::reverse_iterator tit = targets_.rbegin(),
```
atetubou: can you improve order of this algorithm?
It looks like O(|targets| * |build edge|), it can be long.

src/build.h
```diff
@@ -73,6 +74,7 @@ struct Plan {
 
   /// Reset state.  Clears want and ready sets.
   void Reset();
+  void ComputePriorityList(BuildLog* build_log);
```
atetubou: No test for this function?

src/build.cc
comicfans: for my inital idea, earlier built target should uses previous pass max critical time as inital value, so it is O(|targets| * |build edge|)  but if you do not care to save bits of type of critical time(uint64_t) very much , we can use the longest path length (suppose  N ) of whole graph as lowest priority target max critical time,  and assign all targets initial critical time as 
pow(2,M) N......  4N  2N  N 0
so we can just walk whole graph once to get N, then just follow nico's previous implementation, mark all critical time for all edges at once.  it may waste more bits than O(|targets| * |build edge|) , but I think which is also desirable.

this is a quick implementation which improve my own build process immediately,  useful when the longest job build time much more than all other small jobs.  in my own test (3500 sources),  total saved build time (8 min )is much more than this algorithms spent( which I havn't measured this exactly, at least I havn't seen big latency before first job is starting build) . 

to my suprise, without explicit targets specified, this critical priority list didn't show big win over default ninja(just as nico mentioned),  objects on ciritical path do build much earlier than default, but the link step, didn't being picked earlier very much. so I think the priority logic may still have something  missed.

src/build.cc
```diff
@@ -474,6 +486,133 @@ bool Plan::CleanNode(DependencyScan* scan, Node* node, string* err) {
...
+    // binary would be added ot the queue 50 times.
+    set<Edge*> done;
+
+    Edge* first = (*tit)->in_edge();
+    if (!first){
+        continue;
+    } 
+    edgesQ.push(first);
+
+    first->set_critical_time(max(first->critical_time(),prev_top_crit));
```
comicfans: this should be corrected  as  :  first->set_critical_time(max(max(first->critical_time(),prev_top_crit),first->run_time_ms));

comicfans:
Hi @atetubou , thank you for your review, I've re-construct my code , now it use total build time of all edges as weight factor, much simpler and only added a little modification on nico's previous implementation.  for easy reviewing my modification, I didn't delete the test codes ( the if(true)  if (false) block and printf ),  just make them not crash when you test them.

comicfans:
Hi @domidimi , yes we should move debug output out of here, I'll address this in following commits. and for the vector usage, we need to preserve the order of targets appeared in command so I choose vector,   sort-uniq and set indeed gives better performance for deduplicates, but both require more code to recovery their original order. when user provides targets through command line , they may hit command line length limit eariler before hitting deduplicate bottlelock.  if this is really a problem (such as making ninja as a library and supplie targets by API), I'll recreate a better one (but with more codes).  any suggestion ?

domidimi:
Here a possible implementation for a tool. Not completely tested yet:

```c++
void NinjaMain::MarkDirty(Node * n) {
  Edge* edge = n->in_edge();
  if (!edge)
    return;

  n->MarkDirty();
  for (vector<Node*>::iterator in = edge->inputs_.begin();
       in != edge->inputs_.end(); ++in) {
      MarkDirty(*in);
  }
}

int NinjaMain::ToolSchedule(const Options* /*options*/, int argc, char* argv[]) {
  string err;
  vector<Node*> targets;
  Plan plan;

  if (!CollectTargetsFromArgs(argc, argv, &targets, &err)) {
    Error("%s", err.c_str());
    return 1;
  }

  for (size_t i = 0; i < targets.size(); ++i) {
    Edge* edge = targets[i]->in_edge();
    targets[i]->MarkDirty();

      (*o)->MarkDirty();
    }

    for (vector<Node*>::iterator in = edge->inputs_.begin();
         in != edge->inputs_.end(); ++in) {
      MarkDirty(*in);
    }

    if (!plan.AddTarget(targets[i], &err)) {
      if (!err.empty()) {
        Error("%s", err.c_str());
        return 1;
      }
    }
  }

  plan.ComputePriorityList(&build_log_);

  printf("priority list: \n");
  for (list<Edge*>::iterator it = plan.priority_list_.begin(), end = plan.priority_list_.end();
       it != end; ++it) {
    Edge* edge = *it;
    string input_str;
    if (edge->inputs_.empty()) {
        input_str = "empty";
    } else {
        input_str = edge->inputs_[0]->path();
    }
    Node* output = edge->outputs_[0];
    printf("%s %s crit %d edge %d\n", input_str.c_str(),
           output->path().c_str(), edge->critical_time(), edge->run_time_ms_);
  }

  return 0;
}

```

rkhlebnikov:
What's the status of this pull request? Can I help somehow in landing this?

In addition, I have a question/proposed change - it seems that having a separate file containing the previous runtimes of edges as opposed to always relying on log may be better as it prevents subsequent runs of ninja for different targets from removing the available information (e.g., assuming `target1` and `target2` are independent, `ninja target1` will record information for `target1` and its dependencies, running `ninja target2` not only will not have any timing information, but will also remove any info available for `target1`). This information could be stored in the same format as the `.ninja_log` for simplicity.

comicfans:
hi @rkhlebnikov I have no idea neither ... seems no one is reviewing this PR now . And for separate log topic, if I remember correctly , ninja won't purge log even if you build independent target , it will keep all command log until GC happened.  so this should not be a problem (please correct me if this is not the case)

src/build.h
```diff
@@ -92,6 +94,10 @@ struct Plan {
 
   set<Edge*> ready_;
 
+  /// user provided targets in build order, earlier one have higher priority
+  vector<Node*> targets_;
+  list<Edge*> priority_list_;
```
jhasse: Anything stopping you from using a `std::vector` for the `priority_list_`, too?

src/graph.h
```diff
@@ -59,17 +59,11 @@ struct Node {
   }
 
   /// Mark the Node as already-stat()ed and missing.
-  void MarkMissing() {
-    mtime_ = 0;
-  }
+  void MarkMissing() { mtime_ = 0; }
```
jhasse: Please don't include style changes in your PR.

src/graph.h
```diff
@@ -134,7 +128,7 @@ struct Edge {
     VisitDone
   };
 
-  Edge() : rule_(NULL), pool_(NULL), env_(NULL), mark_(VisitNone),
+  Edge() : rule_(NULL), pool_(NULL), env_(NULL), mark_(VisitNone),run_time_ms_(0),critical_time_(0),
```
jhasse: Add spaces after the `,`s.

src/build.cc
```diff
@@ -838,6 +1000,11 @@ bool Builder::FinishCommand(CommandRunner::Result* result, string* err) {
     disk_interface_->RemoveFile(rspfile);
 
   if (scan_.build_log()) {
+// XXX debug output: check if past run times predict future runtimes reasonably
```
jhasse: Please remove these commented lines.

src/build.cc
```diff
@@ -474,6 +486,154 @@ bool Plan::CleanNode(DependencyScan* scan, Node* node, string* err) {
...
+
+  METRIC_RECORD("ComputePriorityList");
+  set<Node*> dedup;
+  vector<Node*> deduped;
+  for (vector<Node*>::iterator it = targets_.begin(); it != targets_.end();
+       ++it) {
+    pair<set<Node*>::iterator, bool> insert_result = dedup.insert(*it);
+    if (!insert_result.second)
+      continue;
+    deduped.push_back(*it);
```
jhasse: Instead of calling `push_back` in a loop, it should be faster to call `targets_.assign(dedup.begin(), dedup.end()` afterwards.

src/build.h
comicfans: I thought at build.cc line 359

``` C++
  for (list<Edge*>::iterator it = priority_list_.begin(),
                             end = priority_list_.end();
       it != end; ++it) {
    i = ready_.find(*it);
    if (i != ready_.end()) {
      priority_list_.erase(it);
      break;
    }
  }

```
this may remove element from priority_list_ at middle pos , list is more friendly ,and erase don't need a extra resize . (maybe vector is faster for iterate?  I didn't benchmark this. I didn't even notice this changes because I just added minimal logic on original cpsched) , if vector is better choice ,I'll modify this.

src/build.cc
comicfans: but it will break user input target order?

jhasse: why?

comicfans: set is order by Node* , but that maybe not the order user typed in command line.

jhasse: Ah right! Sorry that I've missed that.

jonesmz:
@comicfans Could you rebase this PR on the latest master branch?

comicfans:
@jonesmz long time since my last try, I just make this 'compile' on top of master, but have no time/resource to verify these logic still works as expected (of course built binary accept explicit target and successfully build).  If you have build environment powerful enough, feel free to test/improve this.

Issue #828, opened 2014-10-09
ninja: error: remove(someDirectory) Permission denied. Windows


agalapovs:
After running "ninja -t clean" got "Permission denied" while deleting a folder.

MSDN describe stdio.h remove() as function for deleting files (not directories), on the other hand Linux man pages describe it as function for deleting files and folders. Could be a platform specific problem.

Calling _rmdir() (http://msdn.microsoft.com/en-us/library/wt8es881.aspx) do the trick.

crispyzeal:
+1

dkreuter:
On linux directories are only removed when they're empty. Can the behaviour be changed to recursive removal? Or maybe give the buildscript an opportunity to do clean up before -t clean tries to remove() the directory?

mbodmer:
Same issue here.
Directory was generated by rule created from CMake add_custom_command()

ninja clean 
[1/1] Cleaning all built files...
FAILED: /usr/bin/ninja -t clean
ninja: error: remove(mydirectory): Directory not empty

KUGA2:
Issue is still present in 3.7.2 on Windows.

jhasse:
> On linux directories are only removed when they're empty. Can the behaviour be changed to recursive removal?

Would be a pretty big change and people might lose their files. See #1334 btw.

Let's keep this issue about the Windows-only "Permission denied" error though.

jralls:
Microsoft's documentation for [`remove`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/remove-wremove?view=vs-2019) says that it "sets errno either to EACCES to indicate that the path specifies a read-only file, **specifies a directory,** or the file is open...", so that explains what's going on. It doesn't matter if the directory is empty, one simply can't call `remove` on a directory if using msvcrt.

PR #1334, opened 2017-09-25
Silently ignore non-empty directories during 'clean'

andron:
Instead of returning an error if 'remove()' fails to remove a non-empty
directory let this pass as a success and silently ignore the directory. (A
Warning() can be added if deemed necessary).

This will allow targets that generates a lot of files, e.g. a docs directory
generated by Doxygen, to be used without having to list each file as a
byproduct or an explicit output for a rule.  Depending on the wrapper (sh,
.bat, or any other script-language) it can be varingly difficult to ignore the
error which will then propagate further up the build chain.

Using CMake as generator the following provides a solution for Doxygen though:

  add_custom_command(OUTPUT docs/Doxygen
    COMMAND ${CMAKE_COMMAND} -E make_directory docs
    COMMAND ${CMAKE_COMMAND} -E chdir docs doxygen -g)

  add_custom_command(OUTPUT docs/.generate
    COMMAND ${CMAKE_COMMAND} -E chdir docs doxygen
    COMMAND ${CMAKE_COMMAND} -E touch docs/.generate
    MAIN_DEPENDENCY docs/Doxygen)


jhasse:
Not sure if we'd want to do this. This could result in someone losing a file he left in the directory and a warning wouldn't help, as it would be already too late, wouldn't it?

jhasse:
Ah right, your unit test makes that clear. Sorry should have read that.

Now the question remains if we want a warning or not.

andron:
I guess a warning should be issued since this partially changes the behavior. I'll add it and update the PR.

andron:
MSVC build fails with permission denied after the PR has been updated. Guess the PR has not been tested for a while :). This would be the error you got when trying to remove a directory on Linux as well, prior to this change.

jhasse:
Seems to be #828.

andron:
Ah interesting, that seem to originate from the same problem I had. I'll have a look at it and see if there is a common solution.

TorstenRobitzki:
I've stumbled over the same problem. I create documentation into a directory and I can simply not keep track of all created files. Removing just the whole directory would be a solution to me. But that's somehow contradictory to andron's solution.

andron:
Torsten I started out with the same problem. Generated a lot of doxygen files which I cannot track with CMake. Removing the whole directory is the only option. The purpose of this change is to avoid having ninja exit with non-zero when the directory is non-empty when the clean target is executed.

Of course having the directory be deleted recursively would be another option but requires a bigger patch I guess. It does not really change any behavior since removing the directory should mean the content is also unwanted.

doc/manual.asciidoc
```diff
@@ -271,6 +271,9 @@ the given rules.
 Files created but not referenced in the graph are not removed. This
 tool takes in account the +-v+ and the +-n+ options (note that +-n+
 implies +-v+).
++
+Non-empty directories created by the generator or any build targets are not
+recursively removed, these are silently ignored.
```
jhasse: "silently" is no longer true, is it?

Issue #1262, opened 2017-03-30
Backslashes not escaped when using response files on Windows with GCC


ChilliScott:
It is quite an unusual compile path but I'm compiling on Windows using a GCC cross compiler. The issue occurs when using rspfile and rspfile_content where the contents are written to the rsp_file with backslash path separators. When GCC reads the response file the backslashes are not present (as they are not escaped) and the compiler/linker cannot find any of the files.

I'll see if I can fix it and submit a pull request if that would help. In the meantime I'll need to handle the response file generation manually to get round the issue

EDIT:

I forgot to mention that I tried escaping the file paths before passing them into rspfile_content but ultimately had no effect.

ChilliScott:
Have realised that since I'm using GCC it works with Unix style paths (i.e. separated by forward slashes) so I've just changed the paths I'm feeding to rspfile_content to use '/' rather than '\'. Since it is unusual to be compiling with GCC on Windows and that GCC handles Unix paths this is probably not really an issue

Lekensteyn:
Can you check whether #1342 solves the problem for you?

jhasse:
Sorry for the delay. Can you rebase one more time? Thanks in advance!

PR #1342, opened 2017-10-11
Follow GCC/Clang behavior wrt depfiles

Lekensteyn:
Option is called "depfile = gcc" which should support depfiles generated
by GCC. GCC does not escape backslashes and GNU Make does not try to
unescape it, so neither should Ninja try to "unescape" it.

Fixes https://github.com/ninja-build/ninja/issues/1262
___
Performance is **better** than current master, but slightly worse than regenerating current master using re2c 1.0.2 (on an i7-6700HQ).

Elapsed time (n=10) for `./depfile_parser_perftest deps/*.cpp.d` on a desktop with a i7-3770:

| code                                | min    | avg    | max    | mdev    |
| --- | --- | --- | --- | --- |
| git master 887eccf-re2c-0.13.5      | 39.706 | 40.268 | 40.565 | 0.24238 |
| git master 887eccf-re2c-1.0.2       | 41.757 | 42.015 | 42.331 | 0.20303 |
| this patch 0e5cd97-re2c-1.0.2       | 39.662 | 40.099 | 40.385 | 0.26273 |

A bit surprising, but with exactly the same binaries, master+re2c 1.0.2 is faster on an i7-6700HQ. Following is a n=3 result, but I observed this in many repeated tests within an hour (to double-check if I did something wrong).

| code                                | min    | avg    | max    | mdev    |
| --- | --- | --- | --- | --- |
| git master 887eccf-re2c-0.13.5      | 45.094 | 45.453 | 46.006 | 0.48577 |
| git master 887eccf-re2c-1.0.2       | 38.546 | 38.882 | 39.192 | 0.32378 |
| this patch 0e5cd97-re2c-1.0.2       | 43.453 | 43.564 | 43.637 | 0.09771 |

The 170 cpp.d files are taken from the Wireshark build using clang (for convenience: [deps.tar.gz](https://github.com/ninja-build/ninja/files/1373861/deps.tar.gz)).
Above tests were done using Ninja, built using GCC 7.2.0-1 on Arch Linux with files located on a tmpfs.

Note for self:
```
mkdir res && for bin in ./depfile_parser_perftest-*; do what=${bin##*perftest-}; mkdir "res/$what" && date -R && for i in {0..9};do (time "$bin" deps/*.d) &>"res/$what/$i"; done; done
for f in res/*; do awk -F'[ms]' '/^real/{print $2}' "$f"/* | ~/scripts/calc-stats.py | awk '/min/{min=$2} /max/{max=$2} /Mean/{mean=$2} /sd/{sd=$2} END{ printf("| %-35s | %.3f | %.3f | %.3f | %.5f |\n","'"${f##**/}"'", min, mean, max, sd) }'; done
```

### DIFF
--- a/src/depfile_parser.in.cc
+++ b/src/depfile_parser.in.cc
@@ -29,9 +29,15 @@ DepfileParser::DepfileParser(DepfileParserOptions options)
 // How do you end a line with a backslash?  The netbsd Make docs suggest
 // reading the result of a shell command echoing a backslash!
 //
-// Rather than implement all of above, we do a simpler thing here:
-// Backslashes escape a set of characters (see "escapes" defined below),
-// otherwise they are passed through verbatim.
+// Rather than implement all of above, we follow what GCC/Clang produces:
+// Backslashes escape a space or hash sign.
+// When a space is preceded by 2N+1 backslashes, it is represents N backslashes
+// followed by space.
+// When a space is preceded by 2N backslashes, it represents 2N backslashes at
+// the end of a filename.
+// A hash sign is escaped by a single backslash. All other backslashes remain
+// unchanged.
+//
 // If anyone actually has depfiles that rely on the more complicated
 // behavior we can adjust this.
 bool DepfileParser::Parse(string* content, string* err) {
@@ -68,12 +74,33 @@ bool DepfileParser::Parse(string* content, string* err) {
       re2c:indent:string = "  ";
 
       nul = "\000";
-      escape = [ \\#*[|\]];
       newline = '\r'?'\n';
 
-      '\\' escape {
-        // De-escape backslashed character.
-        *out++ = yych;
+      '\\\\'* '\\ ' {
+        // 2N+1 backslashes plus space -> N backslashes plus space.
+        int len = (int)(in - start);
+        int n = len / 2 - 1;
+        if (out < start)
+          memset(out, '\\', n);
+        out += n;
+        *out++ = ' ';
+        continue;
+      }
+      '\\\\'+ ' ' {
+        // 2N backslashes plus space -> 2N backslashes, end of filename.
+        int len = (int)(in - start);
+        if (out < start)
+          memset(out, '\\', len - 1);
+        out += len - 1;
+        break;
+      }
+      '\\'+ '#' {
+        // De-escape hash sign, but preserve other leading backslashes.
+        int len = (int)(in - start);
+        if (len > 2 && out < start)
+          memset(out, '\\', len - 2);
+        out += len - 2;
+        *out++ = '#';
         continue;
       }
       '$$' {
@@ -81,13 +108,7 @@ bool DepfileParser::Parse(string* content, string* err) {
         *out++ = '$';
         continue;
       }
-      '\\' [^\000\r\n] {
-        // Let backslash before other characters through verbatim.
-        *out++ = '\\';
-        *out++ = yych;
-        continue;
-      }
-      [a-zA-Z0-9+,/_:.~()}{%@=!\x80-\xFF-]+ {
+      '\\'+ [^\000\r\n] | [a-zA-Z0-9+,/_:.~()}{%=@\x5B\x5D!\x80-\xFF-]+ {
         // Got a span of plain text.
         int len = (int)(in - start);
         // Need to shift it over if we're overwriting backslashes.
--- a/src/depfile_parser_test.cc
+++ b/src/depfile_parser_test.cc
@@ -101,15 +101,36 @@ TEST_F(DepfileParserTest, Spaces) {
             parser_.ins_[2].AsString());
 }
 
+TEST_F(DepfileParserTest, MultipleBackslashes) {
+  // Successive 2N+1 backslashes followed by space (' ') are replaced by N >= 0
+  // backslashes and the space. A single backslash before hash sign is removed.
+  // Other backslashes remain untouched (including 2N backslashes followed by
+  // space).
+  string err;
+  EXPECT_TRUE(Parse(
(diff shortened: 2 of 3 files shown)


nico:
Thanks for the patch! Can you paste a link to gcc's / clang's depfile writing logic, so I can cross-check what exactly they do? (Just for my own education; it's very likely this patch is good as-is.)

Lekensteyn:
There is a very helpful comment in the Clang source: https://github.com/llvm-mirror/clang/blob/44c160f916a1b080098b17b466b026aa07475ec2/lib/Frontend/DependencyFile.cpp#L316

For the GCC behavior, see: https://github.com/gcc-mirror/gcc/blob/22a8377023d59cc01ab0a84a1df56d0e1336efa3/libcpp/mkdeps.c#L47

src/depfile_parser.in.cc
```diff
@@ -55,25 +55,22 @@ bool DepfileParser::Parse(string* content, string* err) {
       re2c:indent:string = "  ";
 
       nul = "\000";
-      escape = [ \\#*[|];
+      escape = [ \t#];
```
nico: Why is "\t" included?

src/depfile_parser_test.cc
```diff
@@ -101,6 +101,18 @@ TEST_F(DepfileParserTest, Spaces) {
             parser_.ins_[2].AsString());
 }
 
+TEST_F(DepfileParserTest, SlashesSpaces) {
+  // Slashes are unescaped, except when they follow whitespace or a hash sign.
```
nico: s/Slashes/Backslashes/ (also in test name). A slash is '/'.

src/depfile_parser_test.cc
```diff
@@ -101,6 +101,18 @@ TEST_F(DepfileParserTest, Spaces) {
...
+  string err;
+  EXPECT_TRUE(Parse(
+"some\\\\ slash\\\\es:",
```
nico: Include a hash in here too to match the comment.

nico:
Sorry about the delay in getting back to this, I was away for a bit. This looks good to me, with two nits above.

src/depfile_parser.in.cc
Lekensteyn: To match the GCC behavior here as closely as possible:
https://github.com/gcc-mirror/gcc/blob/22a8377023d59cc01ab0a84a1df56d0e1336efa3/libcpp/mkdeps.c#L66

src/depfile_parser_test.cc
```diff
@@ -101,6 +101,18 @@ TEST_F(DepfileParserTest, Spaces) {
             parser_.ins_[2].AsString());
 }
 
+TEST_F(DepfileParserTest, BackslashesSpaces) {
+  // Backslashes are unescaped, except when they follow whitespace or a hash sign.
```
nico: Sorry for being super nitpicky, but could you s/whitespace/spaces, tabs,/? I didn't realize on first read that whitespace here includes tabs, hence my question above. Maybe add a tab to the test string too.

Lekensteyn: I appreciate your feedback, nitpicks are fine!

There is also an existing `Escapes` test that already checks for `\\#`. The new [back]`SlashesSpaces` test was intended to specifically check for one or more backslashes followed by slashes/spaces. For completeness, the comment also mentions hash sign (these are the only characters that can be escaped).

Should I, instead, modify the Escapes test or maintain this new test that demonstrates that multple backslashes are not consumed (unless followed by a space)?

Would you prefer it if I rebase the series on master and squash the fixup changes in the first patch?

Lekensteyn:
Pushed a new version that changes the behavior even more to match GCC/Clang closely. Previously (https://github.com/Lekensteyn/ninja/commits/depfile-fixes-old) it would only strip the last backslash before a space, tab or hash sign. Now it will also reduce the number of backslashes by a half. In practice nobody should be affected, but I did it anyway for full consistency.

I also removed support for tabs because it is an odd edge case (why would someone have tabs in a filename?) and Clang does not do it. There is also support for an even number of backslashes at the end of a filename (again, why would someone like to do that?), I can remove (resulting in `foo\\ bar` being treated as `foo\\ bar` instead of `foo\\` and `bar`) if that's better (but it would be inconsistent with GCC/Clang).

Re-ran the tests, these are the results (again n=10, i7-3770). Time has improved a little.

| code                                | min    | avg    | max    | mdev    |
| --- | --- | --- | --- | --- |
| git master ca041d8-re2c-0.13.5      | 40.537 | 40.764 | 41.115 | 0.17077 |
| git master ca041d8-re2c-1.0.3       | 41.103 | 41.392 | 41.620 | 0.17505 |
| this patch bead1eb-re2c-1.0.3       | 39.979 | 40.180 | 40.497 | 0.20417 |

Lekensteyn:
Rebased on master to retrigger CI (no further changes). Are you happy with this version?

jhasse:
Sorry for the delay. Can you rebase one more time? Thanks in advance!

Lekensteyn:
Rebased. Since GCC does not actually interpret `\[` as `[`, I had to change one test. @maciej-pawlowski-opera is that ok with you?

Tested using commands and deps.tar.gz from the initial post (n=10, i7-6700HQ). Performance remains unchanged.

| code                                | min    | avg    | max    | mdev    |
| --- | --- | --- | --- | --- |
| master 28a7d14-re2c-1.1.1           | 27.592 | 27.826 | 28.152 | 0.16997 |
| patched fba5ce0-re2c-1.1.1          | 27.152 | 27.919 | 28.578 | 0.50886 |

maciej-pawlowski-opera:
That's fine by me :)

PR #1344, opened 2017-10-17
Fix potential buffer overrun

loganchien:
This commit rearranges record size comparison and fread() to make sure
fread() only reads the data that can fit into the buffer.


nico:
Whoa, thanks!

Issue #1168, opened 2016-06-23
-d doesn't actually print debugging modes


kirbyfan64:
`ninja -h` says:

```
ryan@DevPC-LX /media/ryan/stuff/ninja master $ ./ninja -h
usage: ninja [options] [targets...]

if targets are unspecified, builds the 'default' target (see manual).

options:
  --version  print ninja version ("1.7.1.git")

  -C DIR   change to DIR before doing anything else
  -f FILE  specify input build file [default=build.ninja]

  -j N     run N jobs in parallel [default=6, derived from CPUs available]
  -k N     keep going until N jobs fail [default=1]
  -l N     do not start new jobs if the load average is greater than N
  -n       dry run (don't run commands but act like they succeeded)
  -v       show all command lines while building

  -d MODE  enable debugging (use -d list to list modes)
  -t TOOL  run a subtool (use -t list to list subtools)
    terminates toplevel options; further flags are passed to the tool
  -w FLAG  adjust warnings (use -w list to list warnings)
ryan@DevPC-LX /media/ryan/stuff/ninja master $ 
```

But it doesn't actually work:

```
ryan@DevPC-LX /media/ryan/stuff/ninja master $ ./ninja -d
./ninja: option requires an argument -- 'd'
usage: ninja [options] [targets...]

if targets are unspecified, builds the 'default' target (see manual).

options:
  --version  print ninja version ("1.7.1.git")

  -C DIR   change to DIR before doing anything else
  -f FILE  specify input build file [default=build.ninja]

  -j N     run N jobs in parallel [default=6, derived from CPUs available]
  -k N     keep going until N jobs fail [default=1]
  -l N     do not start new jobs if the load average is greater than N
  -n       dry run (don't run commands but act like they succeeded)
  -v       show all command lines while building

  -d MODE  enable debugging (use -d list to list modes)
  -t TOOL  run a subtool (use -t list to list subtools)
    terminates toplevel options; further flags are passed to the tool
  -w FLAG  adjust warnings (use -w list to list warnings)
ryan@DevPC-LX /media/ryan/stuff/ninja master $ 
```

kirbyfan64:
Wow, I'm an idiot and completely misread the help info. Sorry.

tambry:
@kirbyfan64 What did you misunderstand? I too am having the exact same problem/misunderstanding.

evmar:
"-d list". Perhaps we should quote it in the help text? But if you quote it in the command line it won't work

[...]

PR #1345, opened 2017-10-19
escape usage examples

refack:
Fixes: https://github.com/ninja-build/ninja/issues/1168

quick glance at the output of `ninja -h` can skip the `list` argument in the example
image
Maybe as part of the Gestalt Effect:
image


nico:
thanks!

Issue #1285, opened 2017-06-05
subprocess.py error: Ninja not building from source on Windows 8


saxenarohan97:
I tried building from source. I did:
```
$ git clone git://github.com/martine/ninja.git && cd ninja
$ git checkout release
```
Then in a Visual Studio command line I ran `configure.py` with Python:
`$ C:\Python27\python.exe configure.py --bootstrap`

But here I get the following error:

```
C:\ninja>C:\Python27\python.exe configure.py --bootstrap
bootstrapping ninja...
Traceback (most recent call last):
  File "configure.py", line 320, in <module>
    if platform.msvc_needs_fs():
  File "configure.py", line 84, in msvc_needs_fs
    stderr=subprocess.PIPE)
  File "C:\Python27\Lib\subprocess.py", line 390, in __init__
    errread, errwrite)
  File "C:\Python27\Lib\subprocess.py", line 640, in _execute_child
    startupinfo)
WindowsError: [Error 2] The system cannot find the file specified
```

Any ideas what to do?

nico:
What happens if you run `cl.exe`? Is this in a shell where you ran vcvars32 / vcvarsall?

If that doesn't mean anything to you, try Start->All Programs->Visual Studio->Visual Studio Tools->VS x64 Native Tools Command Prompt, and run the script in that shell. Does that help?

zhaoyongboy:
I had the same problem, and it works when I do the following:
" Start->All Programs->Visual Studio->Visual Studio Tools->VS x64 Native Tools Command Prompt, and run the script in that shell. Does that help?"
Many thanks!

nico:
Thanks for the comment.

I'm afraid that this is such a common issue that we probably want to add 2 lines to configure.py to detect this and give a better error message :-/

saxenarohan97:
Yes, that solves the problem. Thanks guys!

PR #1346, opened 2017-10-20
print nice error when running configure.py without rc.exe on PATH

nico:
Might help with issues like #1285. It's quite a bit of additional code though, so not sure if this is worth it.


nico:
Let's close this one out for now; if we get a bunch of additional support requests for this we can reconsider.

PR #1348, opened 2017-10-31
Fix building on Windows in UNICODE mode

ngg:
Some WinAPI calls were already suffixed with A to use the Ansi versions, but not all of those. With this commit it is possible to build ninja with -DUNICODE on Windows.


atetubou:
Better to set -DUNICODE by default in configure.py to prevent future usage of non-prefixed function?

PR #1354, opened 2017-11-07
Add "-m <N>" option to dynamically limit jobs

maxim-kuvyrkov:
... on memory threshold.  So far only Unix-style OS are supported.
No functional change for other configurations.


metti:
you can refer to a previous pull request (#660) that implemented this feature and its discussion back then.

src/util.cc
```diff
@@ -604,3 +687,5 @@ bool Truncate(const string& path, size_t size, string* err) {
   }
   return true;
 }
+
+// vim: et sts=2 st=2 sw=2:
```
nico: ?

src/ninja.cc
```diff
@@ -209,6 +209,10 @@ void Usage(const BuildConfig& config) {
 "  -j N     run N jobs in parallel [default=%d, derived from CPUs available]\n"
 "  -k N     keep going until N jobs fail [default=1]\n"
 "  -l N     do not start new jobs if the load average is greater than N\n"
+"  -m N     do not start new jobs if the memory usage exceeds N percent\n"
+#if !(defined(__APPLE__) || defined(linux) || defined(_WIN32))
+"           (not yet implemented on this platform)\n"
+#endif
 "  -n       dry run (don't run commands but act like they succeeded)\n"
 "  -v       show all command lines while building\n"
```
nico: This doesn't document the new -D and -M parameters. With those included, this increases the number of switches in ninja by 25%. Says HACKING.md: "I am very wary of changes that increase the complexity of Ninja (in
particular, new build file syntax or command-line flags)". Maybe there's a way to handle low-memory situations with less ninja flag surface?

maxim-kuvyrkov: @nico, I agree that this needs to be re-worked.  There needs to be a single top-level parameter for constraints, similar to debug output option.
I'm currently using https://github.com/maxim-kuvyrkov/ninja/commit/4fb7ab82334efcc5cf4fc45664a632a8d9ff0813 quite successfully, but this code needs to be refactored into a loop over different constraints.  If the constraints in above commit look useful to upstream ninja, I'll clean this up together with option handling, and re-submit.

jonesmz:
@maxim-kuvyrkov Could you rebase this PR on the latest master branch?

Bidski:
This would be a very useful feature to have. What will it take to get this merged?

jonesmz:
@Bidski 

At a minimum, the change needs to be rebased on master.

If you want to see this feature merged, that's the first thing that has to happen.

You might also look at #660, as mentioned by @metti to see if there are better ways to accomplish the goal.

skardach:
Please correct me if I'm wrong but this pull request currently does not contain any functional code. Instead it's located on the [limit-on-ram](https://github.com/maxim-kuvyrkov/ninja/tree/limit-on-ram) branch?

maxim-kuvyrkov:
> Please correct me if I'm wrong but this pull request currently does not contain any functional code. Instead it's located on the [limit-on-ram](https://github.com/maxim-kuvyrkov/ninja/tree/limit-on-ram) branch?

Hi @skardach ,
That's correct, but I don't consider code on that branch (or this pull request) useful anymore.

I've found that for builds that can, potentially, exhaust RAM, it's much more effective to
1. Put such builds into containers with memory cgroup limiting amount of available RAM.
2. Tweak parallelism based on ratio of on-CPU/off-CPU time that processes inside this container get.
3. When container hits the limit of available-to-container RAM, it starts swapping, thus causing a lot of off-CPU time for the processes.  This causes ninja to reduce parallelism due to (2).

The above approach is implemented here https://github.com/maxim-kuvyrkov/ninja/tree/limit-on-cpu , but it depends on CONFIG_HZ kernel setting, which I didn't find a public API for.  Therefore I don't see how to make this approach generic enough to be included in upstream ninja.

tmilev:
I'd like to note that our team has a use case where

1) Containerization is not an option.
2) We have a situation where we can run a very efficient RAM-wise compilation (some 200 parallel jobs) but we can't run efficient linkining. So, if it so happens that many link jobs happen to run in parallel at the same time, we run out of RAM. Furthermore this happens somewhat late in the build, as linking happens last.

maxim-kuvyrkov:
> I'd like to note that our team has a use case where
> 
> 1. Containerization is not an option.
> 2. We have a situation where we can run a very efficient RAM-wise compilation (some 200 parallel jobs) but we can't run efficient linkining. So, if it so happens that many link jobs happen to run in parallel at the same time, we run out of RAM. Furthermore this happens somewhat late in the build, as linking happens last.

@tmilev , your usecase can, potentially, be addressed with https://ninja-build.org/manual.html#ref_pool .

maxim-kuvyrkov:
Hi @puetzk , thanks for the suggestion!

I have implemented another approach (see [1] and [2]), which allows for several containers to gracefully compete for limited RAM.  Your suggestion applies equally well to that new approach.

The idea in [1] and [2] is that we increase parallelism until we run into CPU waiting.  The CPU waiting can come from direct CPU share limit (our container / cgroup used up its fair share of CPU cycles) or from RAM allowance getting exhausted, and we have started to use swap. While swapping we are "waiting" for CPU just like when hitting CPU share limit.

This new approach allows ninja to dance on the edge of swapping when there is high demand for RAM from other containers, while increasing parallelism to the maximum when other containers aren't using RAM.

[1] https://github.com/maxim-kuvyrkov/ninja/commit/70ef9be89327756e12cb6cbc31f8880bfa9a64bb .
[2] https://github.com/maxim-kuvyrkov/ninja/commits/limit-on-cpu

maxim-kuvyrkov:
Dropping this pull request in favour of https://github.com/maxim-kuvyrkov/ninja/commits/limit-on-cpu .

puetzk:
Yeah, my suggested related more to your limit-on-cpu branch, but this was the thread where it was getting talked about.
What the PSI feature offers you is a direct way to read the "wait ratio" as measured within the kernel, and distinguish whether it is being caused by competition for CPU, memory(swap) time, or other io. And the possibility to get a waitable fd for thresholds.

puetzk:
That seemed like the solution to your comment about

> it depends on CONFIG_HZ kernel setting, which I didn't find a public API for. Therefore I don't see how to make this approach generic enough to be included in upstream ninja.

maxim-kuvyrkov:
The reworked support using Linux PSI is posted here: https://github.com/ninja-build/ninja/pull/2300
@skardach , @tmilev , @puetzk

maxim-kuvyrkov:
> I'd like to note that our team has a use case where
> 
> 1. Containerization is not an option.
> 2. We have a situation where we can run a very efficient RAM-wise compilation (some 200 parallel jobs) but we can't run efficient linkining. So, if it so happens that many link jobs happen to run in parallel at the same time, we run out of RAM. Furthermore this happens somewhat late in the build, as linking happens last.

The implementation in https://github.com/ninja-build/ninja/pull/2300 should work for non-containerized environments as well.

PR #1356, opened 2017-11-09
Improve vim syntax definition

RedBeard0531:
Only highlights comments where the ninja lexer would treat them as such.
Also correctly scopes the rule- and pool-specific variables highlighting.

The comment highlighting was bugging me because we have some variable definitions that include the '#' character, and the old highlighting made it look like that would be ignored by ninja. The ambiguous phrasing in the manual didn't help: "Comments begin with # and extend to the end of the line."

### DIFF
--- a/misc/ninja.vim
+++ b/misc/ninja.vim
@@ -21,7 +21,10 @@ set cpo&vim
 
 syn case match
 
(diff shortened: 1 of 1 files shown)


nico:
Thanks!

PR #1358, opened 2017-11-15
Make -j and -j0 act like on GNU Make.

xim:


src/ninja.cc
```diff
@@ -1058,11 +1058,14 @@ int ReadFlags(int* argc, char*** argv,
...
+        if (!optarg)
+          break;
         char* end;
         int value = strtol(optarg, &end, 10);
-        if (*end != 0 || value <= 0)
+        if (*end != 0 || value < 0)
           Fatal("invalid -j parameter");
-        config->parallelism = value;
+        if (value)
+          config->parallelism = value;
```
stefanb2: That just means that `-j` or `-j0` will be cause ninja to use the default parallelism, which is set up in line 1039 above.

I definitely doesn't mean the same as GNU make `-j` or `-j0` i.e. run as many jobs in parallel as are ready to run.

stefanb2:
This PR has nothing to do with PR #1140 and is not required for it.

src/ninja.cc
xim: True, I could make it INT_MAX. I'll also drop the reference to PR #1140. The reason I mention it is that our build system passes a sanitized version of the `MAKEFLAGS` variable to the command line of the builders. As that by default contains a `-j` in an environment where Make <4.2 has a command called with a `+`-prefixed line, I feel they're related (although not needed to be useful individually).

xim:
Pull request updated with INT_MAX jobs and reference to other PR dropped.

nico:
Why?

nico:
I'm ok with supporting `-j 0` since that's consistent with `-k 0` (you should update --help output to say "(0 means infinity)" like the help now says for -k).

I don't see the point of supporting `-j` without an argument.

jhasse:
Thanks for the PR! As nico said, please add some documentation to `--help` and drop the support of `-j` without an argument. I'm closing this in order to reduce the number of open PRs, but with those changes we would merge this.

PR #1361, opened 2017-11-24
Flush changes into .ninja_log right away.

ppluciennik:
This change flushes new records to the disk (as soon as possible). The same approach is used in .ninja_deps log.
It's also helpful to read log file on the fly (without waiting possibly a few seconds to read new lines).


nico:
Thanks!

Issue #1302, opened 2017-07-17
Running install target with root privilages breaks subsequent invocations


Psirus:
A "common pattern" for a library is to run `make` and then `sudo make install`. This works nicely with CMake. However, when using Ninja as the underlying build system, `sudo ninja install` changes the owner of `.ninja_log` to root. Subsequent invocations of `ninja` than fail with

     ninja: error: opening build log: Permission denied

Just deleting the file afterwards seems fine, but this is not the way to go, right? When I raised this [issue on CMake](https://gitlab.kitware.com/cmake/cmake/issues/17073), one suggestion was for Ninja to try and write the log file with the original owner if it is run as root. What do you think?

evmar:
If you have a ninja target that builds anything, "sudo ninja ..." will build those intermediate outputs as root, which is also undesirable (the files will have the wrong owner and running the compiler as root).

I think better would be if the install commands themselves used sudo.  That is, the build file would be configured such that `ninja install` would run a command that looks like `sudo cp ...`.

mathstuf:
Putting it into the install rules is nasty: not all installs should happen as root and not all machines use sudo.

evmar:
Right, I meant it's up to the generator.

Psirus:
So the CMake dev thinks this should be solved by ninja, and you think it should be solved by CMake. Understandable, and I'm not knowledgable enough to give an opinion, but unlikely to resolve the issue.

evmar:
Can you link to the relevant cmake thread?

mathstuf:
It's linked to in the description: https://gitlab.kitware.com/cmake/cmake/issues/17073.

marcelhollerbach:
Hello,

How about checking if there is already a log file, if there is, ensure the permissions stay the same, if not, just rename it (and take the uid and gid of the running user)?

ajaybd:
I faced the same issue while building the sources.
The issue is some folders are built with root permission and some other folder with other username permission.
To get rid of this, change the complete permission of the source code to a single user.
sudo chown -R username:username <android_source_code>

jgrocha:
Same problem here. Always have to do a `chown $USER:$USER .ninja_log` after `sudo ninja install`. Thank you all for looking into this issue. You are great!

jerome-pouiller:
It seems we can consider that as a major issue and the problem seems well understood.

[...]

PR #1362, opened 2017-11-24
build_log: make sure to carry the old uid and gid

marcelhollerbach:
see #1302 as a normal workflow you do things like "ninja all && sudo
ninja install" for the case that the install target executes a build
target then ninja_log will be writte ntwo which will result in its
permissions to be root / root, which is a problem since you cannot
execute the workflow from above again.

### DIFF
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -25,6 +25,8 @@
 #include <errno.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/types.h>
+#include <sys/stat.h>
 
 #ifndef _WIN32
 #include <inttypes.h>
@@ -364,6 +366,8 @@ bool BuildLog::Recompact(const string& path, const BuildLogUser& user,
 
   Close();
   string temp_path = path + ".recompact";
+  struct stat old_file;
+  bool found_uid_gid = true;
   FILE* f = fopen(temp_path.c_str(), "wb");
   if (!f) {
     *err = strerror(errno);
@@ -393,6 +397,12 @@ bool BuildLog::Recompact(const string& path, const BuildLogUser& user,
   for (size_t i = 0; i < dead_outputs.size(); ++i)
     entries_.erase(dead_outputs[i]);
 
+  //get the stat
+  if (stat(path.c_str(), &old_file) < 0) {
+    //save if we found the stat
+    found_uid_gid = false;
+  }
+
   fclose(f);
   if (unlink(path.c_str()) < 0) {
     *err = strerror(errno);
@@ -402,6 +412,14 @@ bool BuildLog::Recompact(const string& path, const BuildLogUser& user,
   if (rename(temp_path.c_str(), path.c_str()) < 0) {
     *err = strerror(errno);
     return false;
+  }
+
+  //apply uid and gid again so we always stay as the uid and gid of the first caller that created a file
+  if (found_uid_gid) {
+    if (chown(path.c_str(), old_file.st_uid, old_file.st_gid)) {
+      *err = strerror(errno);
+      return false;
+    }
   }
 
   return true;
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -366,8 +366,6 @@ bool BuildLog::Recompact(const string& path, const BuildLogUser& user,
 
   Close();
   string temp_path = path + ".recompact";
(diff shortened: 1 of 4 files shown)


src/build_log.cc
```diff
@@ -393,6 +398,18 @@ bool BuildLog::Recompact(const string& path, const BuildLogUser& user,
   for (size_t i = 0; i < dead_outputs.size(); ++i)
     entries_.erase(dead_outputs[i]);
 
+  //secure owning uid and gid
+  if (access(path.c_str(), R_OK) < 0) {
```
mathstuf: Why is this call necessary? Why not just use `stat` and set `found_uid_gid` based on its return code?

marcelhollerbach: Hello!

This call is not directly nesessary, but i wanted it to keep it out of stat for one reason.
It is more code to check after a failed access(...) if it is due to a missing file / directory, or some internal error that we should report. So i just stayed with a seperated access, which would indicate that we cannot read the file in any case and nothing with internal stats went wrong.

Anyway, i dont have a _that_ strong opinion on that, we can also go the path of "if stat failed just stay with the uid / gid from now"

What would you like to see?

mathstuf: I'd rather we just call `stat` rather than trying to ask if we're OK. "Ask for forgiveness rather than permission" kind of thing.

marcelhollerbach: Done! :)

marcelhollerbach:
Okay its resolved, i saw that the problem is also in deps_log, so i added the same there!

ilyapopov:
How portable is this? (What about Windows?)

marcelhollerbach:
I have 2 patches that are making them portable, i just can compile them, but cannot run them (dont have a windows maschine arround ... ) I hope i can get tomorrow a running windows maschine and run it on there ...

marcelhollerbach:
I made it now work on windows (took a while to dive into theire api). Works now on windows and linux!

Any other comments? Or can it proceed ?

Stefan-Schmidt:
Is there anything blocking this PR to get accepted?
We have seen this issue several times in the process of getting our project moved from autotools+make to meson+ninja

src/util.cc
```diff
@@ -604,3 +606,51 @@ bool Truncate(const string& path, size_t size, string* err) {
...
+
+  //apply uid and gid again so we always stay as the uid and gid of the first caller that created a file
+  if (found_uid_gid) {
+    if (chown(path.c_str(), old_file.st_uid, old_file.st_gid)) {
+      *err = strerror(errno);
+      return false;
+    }
+  }
+#else
+  if (!CopyFile(old_path.c_str(), path.c_str(), false))
```
nico: This is the same behavior as the unlink/rename you're replacing, right? can we just keep using unlink/replace then?

And then that part could be shared still and it'd look like

```
#if !defined(_WIN32)
stat()
#endif
unlink()
rename()
#if !defined(_WIN32)
chown()
#endfi

Comment (marcelhollerbach) [reply src/util.cc]: Well, as far as I remember CopyFile keeps the owner of the new file as long as it exists. And only copies the content, i cannot find any source anymore saying that. However, when doing only unlink() rename() the same issue we have on linux hits on windows ...

Comment (nico) [reply src/util.cc]: Ah, thanks. Sounds subtle. I could imagine someone coming along and "simplifying" this code to what I just proposed some time later. I wonder if there's something we could do to stop them since using CopyFile() is important here. Do you think it's possible to write an automated test for that? Maybe create a file with non-standard permissions, call CopyFile() and check the new file has the non-standard permissions? Ideally in a way that the test fails if someone does the change I proposed here.

File: src/util.cc [hunk 3]
```diff
@@ -604,3 +606,51 @@ bool Truncate(const string& path, size_t size, string* err) {
...
+  struct stat old_file;
+  bool found_uid_gid = true;
+
+  //get the stat
+  if (stat(path.c_str(), &old_file) < 0) {
+    //save if we found the stat
+    found_uid_gid = false;
+  }
+
+  if (unlink(path.c_str()) < 0) {
```
mathstuf: Per #1572, this should probably ignore `ENOENT` errors.

marcelhollerbach:
This is replaced with #1588, i cannot update this one here anymore

PR #1363, opened 2017-12-14
Added support for environment variable NINJA_PARALLEL_JOBS

ocroquette:
Introduced `NINJA_PARALLEL_JOBS`, the maximum number of jobs to run in parallel.

This is the same as the -j option and can be used to override the default value guessed by Ninja. For instance, you might want to use only the number of physical cores even when virtual cores are available, especially when working with slow storage mediums. Or in a shared environment, using all cores could disturb other users. Instead of providing the -j option every time on the command line, set this environment variable for the system, the user or the current shell, as appropriate.

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -191,7 +191,7 @@ you don't need to pass `-j`.)
 Environment variables
 ~~~~~~~~~~~~~~~~~~~~~
 
-Ninja supports one environment variable to control its behavior:
+Ninja supports two environment variables to control its behavior:
 `NINJA_STATUS`, the progress status printed before the rule being run.
 
 Several placeholders are available:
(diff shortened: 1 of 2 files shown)


mathstuf:
There's also the `-l` option to stop spawning jobs when the system load is above the given limit.

ocroquette:
Thanks, I still need to try the -l option out, it seems more useful than the -j option. Still, Ninja lacks a way to specify this setting more permanently. Environment variables are an easy way to do, but it's not very elegant. Maybe Ninja should support preference file?

mathstuf:
I don't think it should. There's `GNUMAKEFLAGS` for `make`, but I doubt that would be wanted either. Actually, this function probably works just as well until ninja itself supports it (if ever):

```sh
ninja () {
  local jobs=
  [ -n "$NINJA_PARALLEL_JOBS" ] && jobs="-j$NINJA_PARALLEL_JOBS"
  "$(which ninja)" $jobs
}
```

ocroquette:
If you mean a shell function, that approach doesn't work on Windows, neither does it work when Ninja is called from another tool like CMake.

Why don't you think that user should be able to specify the default values the maximal number of jobs and the maximal load? It seems to me like a useful feature with minimal impact.

mathstuf:
Because some projects may have different requirements. One might be known to have many small targets which allows me to do `-j100`. Another might have gigabyte-chomping link process and I can only do `-j2`.

ocroquette:
It would still be possible to specify a value on the command line, like today, with -j, so you don't loose anything. But if you don't provide an explicit value, the idea is to use a value that's smaller than Ninja's default, for instance 4 on machine with 16 cores. That would already help in the 2 cases I mentioned (slow storage mediums or shared environment). Enabling the CPU load limit by default could also help, but I still need to try it out.

ocroquette:
Superseded by https://github.com/ninja-build/ninja/pull/1399

PR #1364, opened 2017-12-15
Multiple outputs with deps

moroten:
This PR is a rebase https://github.com/ninja-build/ninja/pull/1236 with more tests added. When adding the tests, I realised that deps for the implicit outputs also have to be recorded.

### DIFF
--- a/src/build_test.cc
+++ b/src/build_test.cc
@@ -1882,6 +1882,308 @@ TEST_F(BuildWithDepsLogTest, Straightforward) {
   }
 }
 
+/// Run a build. Then rerun with the implicit output as the only output.
+TEST_F(BuildWithDepsLogTest, MoveOutputFromImplicitToExplicit) {
+  string err;
+  // Note: in1 was created by the superclass SetUp().
+  const char* manifest_exp =
+      "rule touch\n"  // Will touch all outputs, explicit and implicit.
+      "  command = touch foobar\n"
+      "build outexp | outimp: touch in1\n"
+      "  deps = gcc\n"
+      "  depfile = in1.d\n";
+  const char* manifest_imp =
+      "rule touch\n"
+      "  command = touch foobar\n"
+      "build outimp: touch in1\n"
+      "  deps = gcc\n"
+      "  depfile = in1.d\n";
+  {
+    State state;
+    ASSERT_NO_FATAL_FAILURE(AddCatRule(&state));
+    ASSERT_NO_FATAL_FAILURE(AssertParse(&state, manifest_exp));
+
+    // Run the build once, everything should be ok.
+    DepsLog deps_log;
+    ASSERT_TRUE(deps_log.OpenForWrite("ninja_deps", &err));
+    ASSERT_EQ("", err);
+
+    Builder builder(&state, config_, NULL, &deps_log, &fs_);
+    builder.command_runner_.reset(&command_runner_);
+    EXPECT_TRUE(builder.AddTarget("outexp", &err));
+    ASSERT_EQ("", err);
+    fs_.Create("in1.d", "outexp: in2");
+    EXPECT_TRUE(builder.Build(&err));
+    EXPECT_EQ("", err);
+
+    // The deps file should have been removed.
+    EXPECT_EQ(0, fs_.Stat("in1.d", &err));
+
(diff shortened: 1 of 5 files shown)


src/build_test.cc
```diff
@@ -1882,6 +1882,308 @@ TEST_F(BuildWithDepsLogTest, Straightforward) {
...
+
+    // We should have rebuilt the output due to in2 being
+    // out of date.
+    EXPECT_EQ(1u, command_runner_.commands_ran_.size());
+
+    builder.command_runner_.release();
+  }
+}
+
+/// Verify that the deps are recorded for each output output.
```
nico: nit: output output

src/manifest_parser.cc
```diff
@@ -404,7 +404,7 @@ bool ManifestParser::ParseEdge(string* err) {
 
   // Multiple outputs aren't (yet?) supported with depslog.
```
nico: Update comment.

src/manifest_parser.cc
```diff
@@ -404,7 +404,7 @@ bool ManifestParser::ParseEdge(string* err) {
...
   string deps_type = edge->GetBinding("deps");
-  if (!deps_type.empty() && edge->outputs_.size() > 1) {
+  if (!deps_type.empty() && edge->outputs_.size() - edge->implicit_outs_ != 1) {
     return lexer_.Error("multiple outputs aren't (yet?) supported by depslog; "
```
nico: Maybe say "non-implicit outputs" here too.

nico:
Thanks much for working on this!

The approach here is to write a full deps entry with all inputs for each implicit output (meaning for an edge with N outputs and M inputs, you store N*M ints). This is simple, which is good.

Can you update the description to say "Multiple implicit outputs"? Multiple real outputs still aren't supported, right? Could this approach work for multiple non-implicit outputs as well?

What will happen if ninja is killed when it has written deps for some of the outputs but not all of them (both if old deps are in the log and if not)? I think this will effectively take the union of old and new deps for the edge, which should do the right thing (?).

It looks @danw didn't think of adding implicit outputs as necessary – maybe he has thoughts on the addition?

And @evmar has designed the deps log and always comes up with corner cases I can't think of.

moroten:
Regarding multiple real/non-implicit outputs: What is the logic of the .d file? My assumption is that the .d file should have one entry, which points to the single real output, and that all the implicit outputs just inherits that.

Supporting multiple entries in the .d file would require a lot more work and probably break a lot of setups as there is currently no check what the entry in the .d file is called.

Regarting writing partial deps log, I think, but am not sure, that `LoadDepsFromLog()` will check the mtime recorded in the deps log and conclude that the deps information is invalid and therefore rebuild the target.

I will update the pull request later when all your questions has got a response.

nico:
To be clear, I'm not asking you to add support for multiple outputs. I agree it'd require depfile parser changes and probably changes all over the place; I'm just curious if you think that the deps log part of such a change could work with this approach here.

I'll wait a bit if danw or evmar want to comment, but if they don't I'll take this patch (with the minor comments above addressed). Thanks again!

danw:
I don't think it's harmful, my approach would have just preferred to rebuild if the the structure of the outputs of the rule changed that much (we've run into occasional incremental build bugs due to inputs/outputs not being listed directly in the command line, so not being hashed)

Related to this discussion, I've discovered that the protobuf compiler (protoc) writes out a dependency file with multiple outputs, at least when generating C bindings. My current plan is just to rewrite the dep file using sed before giving it back to ninja.

mathstuf:
How does this interact with #1364 (sorry, I missed this when I implemented it)? It certainly conflicts, but the implicit output side might not be taken into account in that PR.

moroten:
> How does this interact with #1364 (sorry, I missed this when I implemented it)? It certainly conflicts, but the implicit output side might not be taken into account in that PR.

This is #1364, I don't understand your concerns.

mathstuf:
Sorry, I meant #1534.

moroten:
I am not very active any more in the Ninja development as I am moving to Bazel. Please feel free to pick the best from both pull requests (#1364 and #1534).

jhasse:
Closing for now until someone wants to pick it up.

PR #1365, opened 2017-12-15
Fix disk_interface_test.cc on Windows for 64-bit timestamp

moroten:
subdir/subsubdir/.. seems to get the time of subdir/subsubdir on NTFS
(Windows 7), not the time of subdir. Therefore, with the update to 64-bit time stamps, tests start to fail.

### DIFF
--- a/src/disk_interface_test.cc
+++ b/src/disk_interface_test.cc
@@ -87,6 +87,8 @@ TEST_F(DiskInterfaceTest, StatExistingDir) {
   string err;
   ASSERT_TRUE(disk_.MakeDir("subdir"));
   ASSERT_TRUE(disk_.MakeDir("subdir/subsubdir"));
+  EXPECT_GT(disk_.Stat("..", &err), 1);
+  EXPECT_EQ("", err);
   EXPECT_GT(disk_.Stat(".", &err), 1);
   EXPECT_EQ("", err);
   EXPECT_GT(disk_.Stat("subdir", &err), 1);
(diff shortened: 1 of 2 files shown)


src/disk_interface_test.cc
```diff
@@ -135,9 +135,11 @@ TEST_F(DiskInterfaceTest, StatCache) {
   EXPECT_EQ(disk_.Stat("subdir", &err),
             disk_.Stat("subdir/.", &err));
   EXPECT_EQ("", err);
+  /* This is not necessarily true, not even uncached.
   EXPECT_EQ(disk_.Stat("subdir", &err),
```
atetubou: Why L99 is OK, but here is not?

moroten: I didn't see that first. It seems to be different behaviour of FindFirstFileExA() in the cached behaviour and GetFileAttributesEx() in the non-cached flavour. Have no time to investigate it further today. Does anyone else know any details or have a suggestion how to fix it?

atetubou: Is it better to do path canonicalization around
https://github.com/ninja-build/ninja/blob/e234a7bdb6c42f4539c0ab09b624f191287c2c10/src/disk_interface.cc#L165 ?

moroten:
This updated change set fixes the code to pass the given test, and adds another test with its fix as well.

nico:
Thanks!

nico:
This test seems to not pass:

https://ci.appveyor.com/project/nico/ninja/build/1.0.737

```
[119/246] DiskInterfaceTest.StatCache
*** Failure in src\disk_interface_test.cc:164
disk_.Stat("subdir", &err) == disk_.Stat("subdir/.", &err)
*** Failure in src\disk_interface_test.cc:167
disk_.Stat("subdir", &err) == disk_.Stat("subdir/subsubdir/..", &err)
[120/246] DiskInterfaceTest.ReadFile
[121/246] DiskInterfaceTest.MakeDirs
```
Can you take a look?

moroten:
See the discussion at #1419. Basically, I also see flakiness on my computer which needs further investigation.

PR #1366, opened 2017-12-18
Canonicalize targets of clean command.

fjhenigman:
This is impeding our ability on the ANGLE project to use GN with Visual Studio:
https://bugs.chromium.org/p/chromium/issues/detail?id=791295
Thanks for any suggestions.

### DIFF
--- a/src/clean.cc
+++ b/src/clean.cc
@@ -180,15 +180,22 @@ int Cleaner::CleanTargets(int target_count, char* targets[]) {
   Reset();
   PrintHeader();
   for (int i = 0; i < target_count; ++i) {
-    const char* target_name = targets[i];
-    Node* target = state_->LookupNode(target_name);
-    if (target) {
-      if (IsVerbose())
-        printf("Target %s\n", target_name);
-      DoCleanTarget(target);
-    } else {
-      Error("unknown target '%s'", target_name);
+    string target_name = targets[i];
+    uint64_t slash_bits;
+    string err;
+    if (!CanonicalizePath(&target_name, &slash_bits, &err)) {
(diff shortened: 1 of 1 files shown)


null77:
@nico @sgraham PTAL

null77:
@atetubou are you an owner who would be able to review/land this?

src/clean.cc
```diff
@@ -180,15 +180,21 @@ int Cleaner::CleanTargets(int target_count, char* targets[]) {
...
-    if (target) {
-      if (IsVerbose())
-        printf("Target %s\n", target_name);
-      DoCleanTarget(target);
-    } else {
-      Error("unknown target '%s'", target_name);
+    string target_name = targets[i];
+    uint64_t slash_bits;
+    string err;
+    if (!CanonicalizePath(&target_name, &slash_bits, &err)) {
```
atetubou: Can I ask you to show err if it fails to canonicalize?
```
Error("Failed to canonicalize '%s': %s", target_name,c_str(), err.c_str());
```

atetubou:
Sorry, I'm developer interested in this project, but I'm not owner.

null77:
@evmar PTAL at this CL, it's been waiting for an owner to review it for a month.

sgraham:
Sorry, I went on vacation on Dec 20 so I just missed your ping, and am just getting back. I think this is probably fine, but as mentioned offline, it might be more effective for you to fix this in the GN generator as the turnaround time for a new released ninja version will be pretty long.

src/clean.cc
nico: This doesn't look addressed.

fjhenigman: done

nico:
Thanks!

PR #1371, opened 2018-01-01
Canonicalize parent paths

moroten:


nico:
This needs a more detailed change description. It's a fairly complicated change without any motivation.

* What is this needed for? We use `-I ...` with qualified paths and things seem to work.
* It changes CanonicalizePaths() which has a "this is perf sensitive" comment at the top but doesn't include any benchmarks on perf impact.
* It adds global state to util.cc which seems not good.

jhasse:
Thanks for the PR! I'm closing this in order to reduce the number of open PRs. Please feel free to reopen it after adressing nico's comments.

moroten:
Apparently, I missed @nico's comments, answers below.

@jhasse: Would you like to reopen this PR or should I create a new PR? The updated code is available at https://github.com/moroten/ninja/tree/canonicalize-parent-paths at the moment.

In summary, it solves the problem class of combining e.g. `gcc -I ..` with generated header files but adds a few percent of processing time for no-op builds.

> This needs a more detailed change description. It's a fairly complicated change without any motivation.

The commit message in "Teach CanonicalizePaths ..." has now been extended. I've also added a new test to the commits to show the problem.

> * What is this needed for? We use `-I ...` with qualified paths and things seem to work.

Do you mean absolute paths? I'm uncertain of the term "qualified paths".
Using a mixture of absolute and relative paths does not work for generated files. See a comment about it in the manual.

> * It changes CanonicalizePaths() which has a "this is perf sensitive" comment at the top but doesn't include any benchmarks on perf impact.

Running on my reasonably big project, aborting as soon as I can (the FinishCommand is close to zero). As you can see in the numbers, it adds a few percent on my Ubuntu virtual machine.
**Without this PR**
```
.ninja parse            4       1076230.8       4304.9
canonicalize str        1217113 0.2             208.4
canonicalize path       1257834 0.1             108.6
lookup node       	1260150	0.3     	388.8
node stat         	217664	3.2     	701.0
.ninja_log load   	1     	4387.0  	4.4
.ninja_deps load  	1     	1844.0  	1.8
StartEdge         	13    	415.8   	5.4
FinishCommand     	7     	170.9   	1.2
depfile load      	5393  	5.6     	30.4

path->node hash load 0.54 (242130 entries / 444487 buckets)
```
**After applying this PR**
```
.ninja parse            4       1081573.0       4326.3
canonicalize str        1217113 0.2             213.5
canonicalize path       1257834 0.1             114.8
lookup node       	1260150	0.3     	393.8
node stat         	217655	3.7     	796.7
.ninja_log load   	1     	4879.0  	4.9
.ninja_deps load  	1     	1888.0  	1.9
StartEdge         	10    	328.3   	3.3
FinishCommand     	4     	176.5   	0.7
depfile load      	5393  	5.6     	30.1

path->node hash load 0.54 (242130 entries / 444487 buckets)
```

> * It adds global state to util.cc which seems not good.

That is definitely not good, but I did not find an way around it. Improvement suggestions are welcome.

jhasse:
GitHub doesn't let me reopen the PR since you've updated your branch. Can you reopen it? If not, please create a new one and reference this one in it :)

moroten:
I've reset moroten/ninja:canonicalize-parent-paths to 
the old 2c26061. I give it a last change for you, @jhasse , to reopen it. If it does not work, I'll create a new PR.

src/build_test.cc
```diff
@@ -959,6 +960,46 @@ TEST_F(BuildTest, OrderOnlyDeps) {
...
+TEST_F(BuildTest, DepsReferencedThroughParentFolder) {
+  string foo_depfile_content = "foo.o: ../project/blah.h\n";
+
+  string err;
+  ASSERT_NO_FATAL_FAILURE(AssertParse(&state_,
+"rule cc\n  command = cc $in\n  depfile = $out.d\n"
+"rule touch\n  command = touch $out\n"
+"build foo.o: cc foo.c || blah.h\n"
+"build blah.h: touch blah.stamp\n"));
+  SetWorkingDirForCanonicalizePath("/root/project");
```
moroten: Removing this line with `SetWorkingDirForCanonicalizePath()` will fail the test.

moroten:
This PR is related to #1393 as both affects generated header files when building C/C++ code.

nico:
If I read the test in https://github.com/ninja-build/ninja/pull/1371/commits/f5d1da93904b8ceaa906ea493d5e20b90b3c9a13 , does this happen if you use the build dir as your cwd, but put your generated headers outside the build dir? I.e. your dir layout is

```
  src/
    build/
    generated_headers/
```

?

moroten:
The structure we have is more like
```
.../component_a/
  .git/
  build/
  external/  (prebuilt libraries downloaded outside git)
    component_b/
      build/
        bar.a
      bar.h
  foo.c
```
When developing `component_a`, my working directory is the root of the git repository, i.e. in `component_a/`, I use the prebuilt library of `component_b`. Include files are references using `#include "component_a/..."` and `#include "component_b/..."`. This results in `gcc -iquote .. -iquote external`.

For some legacy reasons, we also generate code inside our source repository. Generating inside `component_a/build/genfiles/component_a/` would result in `gcc -iquote build/genfiles -iquote external` and the problem would be solved.

If you think this PR is a too big change to be motivated to merge, no problem for me. It might though, as it did for us, result in obscure bugs. The path handling, and its pitfalls, could possibly be explained in more the documentation instead.

nico:
Does component_b include headers from component_a?

moroten:
> Does component_b include headers from component_a?

No, but component_b might include headers from component_c. All sub components are located in the external/ folder or in a sub directory of external/.

As I said, by updating our build system to generate files under `build/genfiles/component_a/` instead of in the source tree, we would get around the issue. The problem is that this behaviour was unexpected as the path resolution strategy is not well documented.

jhasse:
> With this change, given working dir '/a/b/proj', the path '../proj/foo'
will be translated into 'baz'.

Where does 'baz' come from?

moroten:
> > With this change, given working dir '/a/b/proj', the path '../proj/foo'
> > will be translated into 'baz'.
> 
> Where does 'baz' come from?

It must be by mistake and should be `foo`. I'll upload a rebased versioned with that fixed. Are there any other comments to sort out?

jhasse:
I didn't have a closer look yet. I would strongly prefer a way which avoids the global state though.

moroten:
I've rebased this PR onto `master` and fixed the commit message regarding `foo` and `baz`.

PR #1373, opened 2018-01-12
Update Usage to show "-k 0" behavior

BillyDonahue:
For "-k N", N==0 is interpreted as infinite. It's useful but not documented in the help, unfortunately.


nico:
Thanks!

PR #1386, opened 2018-02-19
Don't clean up after ourselves when exiting.

jmgao:
Destruction of NinjaMain can be an expensive operation when dealing
with stupidly large ninjafiles. exit directly instead of returning out
of real_main to avoid doing so.

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -1118,7 +1118,9 @@ int ReadFlags(int* argc, char*** argv,
   return -1;
 }
 
-int real_main(int argc, char** argv) {
+NORETURN void real_main(int argc, char** argv) {
+  // Use exit() instead of return in this function to avoid potentially
+  // expensive cleanup when destructing NinjaMain.
   BuildConfig config;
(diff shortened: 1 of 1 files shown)


src/ninja.cc
```diff
@@ -1118,7 +1118,9 @@ int ReadFlags(int* argc, char*** argv,
   return -1;
 }
 
-int real_main(int argc, char** argv) {
+NORETURN void real_main(int argc, char** argv) {
+  // Use exit() instead of return in this function to avoid potentially
```
nico: Couldn't real_main keep using return and main below then does `exit(real_main())`?

nico:
Generally looks like a great change; one comment on making the patch smaller above.

src/ninja.cc
nico: Nevermind; then we don't prevent destruction of NinjaMain.

nico:
I'd be curious how much this helps, and which object's destruction ends up being expensive.

PR #1393, opened 2018-03-07
Make path matching case insensitive on Windows

moroten:
This commit makes the path matching on Windows case insensitive because
the header paths reported by Microsoft's Visual C++ compiler cl.exe
sometimes are reported as lowercase.

The following files has been created in C:\ninja-case-test\DiR\subdir
a.c: #include "a.h"
a.h is empty

Then compile using
C:\ninja-case-test\DiR\subdir>cl -c /showIncludes /nologo a.c
a.c
Note: including file: c:\ninja-case-test\dir\subdir\a.h

Apparently, the "DiR" casing becomes lowercase "dir".

Now consider changing your build.ninja to generate DiR/subdir/a.h.
This will trigger a rebuild of DiR/subdir/a.h, but if case sensitive
matching is done, Ninja will not understand that DiR/subdir/a.c needs
dir/subdir/a.h generated before compiling.

To solve the problem, case insensitive matching has been introduced in
the node lookup. Another solution would be to match the dependencies
from cl.exe with all the order-only and other inputs for the
compilation to find the correct casing. That would require generated
headers to be specified as input and not just as e.g. a phony grouped
input. The benefit would to isolate the case insensitive matching
to the CLParser only.

Signed-off-by: Fredrik Medley <fredrik.medley@autoliv.com>

### DIFF
--- a/src/state.cc
+++ b/src/state.cc
@@ -14,6 +14,8 @@
 
 #include "state.h"
 
+#include <algorithm>
+
 #include <assert.h>
 #include <stdio.h>
 
@@ -106,6 +108,13 @@ Node* State::GetNode(StringPiece path, uint64_t slash_bits) {
     return node;
   node = new Node(path.AsString(), slash_bits);
   paths_[node->path()] = node;
+#ifdef _WIN32
+  // Also store as lowercase.
+  string lowerpath = node->path();
+  transform(lowerpath.begin(), lowerpath.end(), lowerpath.begin(), ::tolower);
+  path_storage_.push_back(lowerpath);
+  paths_[path_storage_.back()] = node;
+#endif  // _Win32
   return node;
 }
 
@@ -114,6 +123,17 @@ Node* State::LookupNode(StringPiece path) const {
   Paths::const_iterator i = paths_.find(path);
   if (i != paths_.end())
     return i->second;
+#ifdef _WIN32
+  string lowerpath = path.AsString();
+  transform(lowerpath.begin(), lowerpath.end(), lowerpath.begin(), ::tolower);
+  i = paths_.find(lowerpath);
+  if (i != paths_.end()) {
+    // Cache this casing of the path as well
+    path_storage_.push_back(path.AsString());
+    paths_[path_storage_.back()] = i->second;
+    return i->second;
+  }
+#endif  // _WIN32
   return NULL;
 }
 
--- a/src/build_test.cc
+++ b/src/build_test.cc
@@ -1956,6 +1956,94 @@ TEST_F(BuildWithDepsLogTest, ObsoleteDeps) {
   }
 }
 
+#ifdef _WIN32
+/// cl.exe sometimes reports deps with lowercase. Make sure a mix of
+/// correct casing and lowercase works.
+TEST_F(BuildWithDepsLogTest, CaseInsensitiveOnWindows) {
+  string err;
+  // Note: in1 and in2 were created by the superclass SetUp().
+  const char* manifest =
+      "build out: cat in1 || ExtraDep\n"
+      "  deps = gcc\n"
+      "  depfile = in1.d\n"
+      "build ExtraDep: cat in2\n";
+  {
+    State state;
+    ASSERT_NO_FATAL_FAILURE(AddCatRule(&state));
+    ASSERT_NO_FATAL_FAILURE(AssertParse(&state, manifest));
+
+    // Run the build once, everything should be ok.
+    DepsLog deps_log;
+    ASSERT_TRUE(deps_log.OpenForWrite("ninja_deps", &err));
+    ASSERT_EQ("", err);
+
+    Builder builder(&state, config_, NULL, &deps_log, &fs_);
+    builder.command_runner_.reset(&command_runner_);
+    EXPECT_TRUE(builder.AddTarget("out", &err));
+    ASSERT_EQ("", err);
+    fs_.Create("in1.d", "out: eXTRADep");
+    EXPECT_TRUE(builder.Build(&err));
+    EXPECT_EQ("", err);
+
+    // Both ExtraDep and out should have been produced.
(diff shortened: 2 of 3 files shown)


moroten:
I have found a problem when using generated header files with MSVC. If including "DiR/a.h", MSVC will report "dir/a.h". Apparently, the "DiR" casing becomes lowercase "dir".

`C:\ninja-case-test\DiR\subdir>cl -c /showIncludes /nologo a.c`
`a.c`
`Note: including file: c:\ninja-case-test\dir\subdir\a.h`

Now consider changing your build.ninja to generate DiR/a.h. This will trigger a rebuild of DiR/a.h, but if case sensitive matching is done, Ninja will not understand that DiR/a.c needs dir/a.h generated before compiling.

Run the commands below together with the attached 
[build.ninja](https://github.com/ninja-build/ninja/files/1815319/build.ninja.txt) to, hopefully, reproduce the behaviour.

`mkdir DiR`
`echo #include "a.h" > DiR\a.c`
`echo int main() { return A_RETURN_VALUE; } >> DiR\a.c`
`echo #define A_RETURN_VALUE 0 > foo.h`
`REM Build for the first time`
`ninja`
`echo #define NO_VALUE 0 > foo.h`
`ninja`
`REM Ninja should have failed compiling a.c but it did succeed.`
`ninja`
`REM Ninja now fails to compile a.c, without any code change.`
`ninja -t deps`
`REM You can find dir/a.h, but not DiR/a.h.`

I'm using Windows 10 and Visual Studio 2015.

nico:
Would it work to do this in src/clparser.cc when parsing the /showIncludes notes?

moroten:
Unfortunately, that does not work in the following scenario:
1. Build `A.c` which is including `A.h`, so `A.h` is an order only dependency to `A.c`.
2. `a.h` is shown by `cl.exe /showIncludes` and added to `.ninja_deps` for `A.c`.
3. Add a build target to generate `A.h` for one of the header files so that it is generated rather than read from disk.
4. Build `A.c` again and ninja concludes there is nothing to do.

Expected output would be to generate `A.h` and rebuild `A.c`, but as `A.c` only has `a.h` and not `A.h` recorded in `.ninja_deps`, `A.c` will be thought to be up to date.

jhasse:
Have you thought about reporting this as a bug in MSVC?

moroten:
> Have you thought about reporting this as a bug in MSVC?

No, I haven't thought about it before. The problem is also that I'm running Visual Studio 2015. If someone with MSVC 2017 or the associated build tools can verify the behaviour in the description at the top, please file a bug report and link it to this issue.

I assume no patch will be made for older versions of MSVC, so it is still a problem for probably a good amount of ninja users.

jonesmz:
My company is using VS2015, and plans to continue doing so for at least another year. I'm pretty positive there won't be any patches to VS2015 anymore.

src/state.cc
```diff
@@ -114,6 +123,17 @@ Node* State::LookupNode(StringPiece path) const {
   Paths::const_iterator i = paths_.find(path);
   if (i != paths_.end())
     return i->second;
+#ifdef _WIN32
+  string lowerpath = path.AsString();
+  transform(lowerpath.begin(), lowerpath.end(), lowerpath.begin(), ::tolower);
```
mathstuf: Might it not be better to instead get the real case of the path returned? Then case sensitivity can be preserved everywhere else?

moroten: Sorry, but I don't understand what you mean. The code first tries matching with the given casing and as a second choice also tries the lower case. The given casing can come from any input source, such as directory listing, .ninja file or .d file.

mathstuf: This just blindly does lowercasing of the path. This is not really well-defined on a character-by-character basis (e.g., the classic ẞ or I → ı in Turkish cases). Instead, for each path component, ask the Windows API "what is the actual name for this component?" and build up a properly cased path. That, to me, seems like a better thing to cache than a blind lowercased pathname since it is actually likely to show up in other locations.

Plus, Windows isn't the only case insensitive platform/filesystem out there (macOS, ext4 in Linux can now be case-insensitive).

moroten: I agree that lower case is not enough. Asking the filesystem might take time (I'm not sure) when doing for a lot of paths. What happens if a path does not exist on disk? We still want to be able to compare a generated output with an input with another casing, even if the file has not been generated yet.

Is there any other good normalizing API that can be used?

mathstuf: > What happens if a path does not exist on disk?

It's assumed to be correct. Since this is mainly an issue for `/showIncludes` reporting, I think we can assume that all paths will exist at that time?

> We still want to be able to compare a generated output with an input with another casing

I think we can assume/require that generators will be sane and say that this is invalid:

```ninja
build out: mk
build uses: mk OUT
```

> Is there any other good normalizing API that can be used?

Unfortunately, I don't think so :( .

mathstuf: > Since this is mainly an issue for /showIncludes reporting

Hmm. maybe that's where the "what's the real case of this path?" logic should go?

moroten: I agree. We should assume that generators are correct and keep this to `/showIncludes`. Please feel free to create such a PR as my interest in Ninja has dropped as I am moving to Bazel.

jonesmz:
Starting with Windows insider build 17093, Microsoft introduced a new way to handle case sensitive files in Windows: per-directory case sensitivity.

Does this pull request work properly for that? Or should this PR be closed?

moroten:
This patch will not work in case sensitive Windows directories because the mapping from lowercase path to node won't be unique any more.

mathstuf:
Note that even `ext4` on Linux has case insensitivity via per-directory options now as well. This has always been the case if a `vfat` drive had been mounted as well. I think the actual case is the only thing to actually use. I think one place that case insensitivity could be useful is in detecting "did you mean" suggestions for targets that aren't found from the command line.

PR #1398, opened 2018-03-17
Added support for environment variable NINJA_FLAGS

ocroquette:
The value of the variable is parsed like command line arguments. You can for
instance use less cores than available (-j), or limit the parallelism based
on the system load (-l). For instance, on a build server used by multiple
users, you may want to set `NINJA_FLAGS=-l 1` on the system level, so that
the server never gets overloaded by parallel ninja builds. Command line
parameters override the flags set with NINJA_FLAGS.


ocroquette:
Superseded by https://github.com/ninja-build/ninja/pull/1399

Issue #1370, opened 2017-12-28
Support setting verbosity through the environment


SebastianBoe:
Sometimes, regrettably, using the environment is the cleanest way to configure applications.

e.g. MAKEFLAGS, DISTCC_HOSTS, etc.

It would be useful if one could control ninja's verbosity through the environment.

e.g. have

VERBOSE=1 ninja

be (nearly) equivalent to

ninja -v

A nice use-case that this enables is having a single UI nob to turn on verbosity for both Ninja, 
and any ninja-invoked scripts.

evmar:
I think in cases like this you can make a shell script to obey your environment variables.  Otherwise I can think of equivalent environments where someone wants VERBOSE=1 to control the verbosity of a program other than Ninja.

Earnestly:
> I think in cases like this you can make a shell script to obey your environment variables.

These environment variables are for ninja, so how would they be obeyed if ninja doesn't recognise them?

Additionally it's not always feasable, or reasonable, to modify the command-line of tools in scripts as there may be far too numerous.  What you're suggesting is each script adopt a unique way of adjusting flags passed to the command, which is also unreasonable.

I've read the other PRs and issues associated with this situation and find none of the replies to be very well thought through.  In each case it is simply assumed that the user has direct control over where and how `ninja` is executed or is even in a position to realistically do anything about it, especially when configuration systems generate `ninja.build` scripts as part of the pipeline.

jhasse:
> These environment variables are for ninja, so how would they be obeyed if ninja doesn't recognise them?

The script would recognise them and pass e.g. `-v` to ninja.

Earnestly:
So as I predicted, you expect every script, or even script-like thing to encode its own set of logic to pass options to ninja.  That's not reasonable at all.

jhasse:
No, you only need one script to do that, which you name `ninja` and put it in your `PATH` before the real ninja exectuable.

Earnestly:
I thought this was covered in the other issue reports already.  That's also not reasonable as it now starts to effect reproducibility build pipelines where environments are feasible to track.  It also assumes people have the ability to do what you're suggesting.

jhasse:
> That's also not reasonable as it now starts to effect reproducibility build pipelines where environments are feasible to track.

You would have your script always there, then you can track `VERBOSE` just as you would if ninja itself handled it.

> think secure `PATH`

What do you mean by that?

Earnestly:
[...]

PR #1399, opened 2018-03-17
Added support for environment variable NINJA_FLAGS

ocroquette:
The value of the variable is parsed like command line arguments. You can for
instance use less cores than available (-j), or limit the parallelism based
on the system load (-l). For instance, on a build server used by multiple
users, you may want to set `NINJA_FLAGS=-l 1` on the system level, so that
the server never gets overloaded by parallel ninja builds. Command line
parameters override the flags set with NINJA_FLAGS.

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -1034,9 +1034,8 @@ int ExceptionFilter(unsigned int code, struct _EXCEPTION_POINTERS *ep) {
 
 /// Parse argv for command-line options.
 /// Returns an exit code, or -1 if Ninja should continue.
-int ReadFlags(int* argc, char*** argv,
+int ReadFlagsFromArray(int* argc, char*** argv,
               Options* options, BuildConfig* config) {
-  config->parallelism = GuessParallelism();
 
   enum { OPT_VERSION = 1 };
   const option kLongOptions[] = {
@@ -1045,6 +1044,7 @@ int ReadFlags(int* argc, char*** argv,
     { NULL, 0, NULL, 0 }
   };
 
+  optind = 1;
   int opt;
   while (!options->tool &&
          (opt = getopt_long(*argc, *argv, "d:f:j:k:l:nt:vw:C:h", kLongOptions,
@@ -1118,15 +1118,79 @@ int ReadFlags(int* argc, char*** argv,
   return -1;
 }
 
+int ReadFlagsFromCommandLine(int* argc, char*** argv,
+              Options* options, BuildConfig* config) {
+  int exit_code = ReadFlagsFromArray(argc, argv, options, config);
+  if ( exit_code == -1 ) {
+    *argv += optind;
+    *argc -= optind;
+  }
+  return exit_code;
+}
+
+int ReadFlagsFromEnv(Options* options, BuildConfig* config) {
+  char *str = getenv("NINJA_FLAGS");
+  if (str == NULL || strlen(str) == 0 )
+    return -1;
+
+  char *str_copy = (char *)malloc(strlen(str)+1);
+  char **argv = (char **)calloc(strlen(str), sizeof(char *));
+  int argc = 0, i = 0;
+  bool in_arg = false;
+
+  argv[argc++] = ""; // fake argv[0] (command) for getopt
+
+  while(str[i] != 0) {
+    switch(str[i]) {
+      case ' ':
+      case '\t':
+        str_copy[i] = 0;  // replace separators by 0 in the copy
+        in_arg = false;
+        break;
+      default:
+        str_copy[i] = str[i];
+        if (in_arg) {
+          break;
+        }
+        else {
+          argv[argc++] = &(str_copy[i]);
+          in_arg =  true;
+          break;
+        }
+    }
+    i++;
+  }
+  char **argv_copy = argv; // ReadFlagsFromArray modifies argv, so create a copy
+  int exit_code = ReadFlagsFromArray(&argc, &argv_copy, options, config);
+
+  free(argv);
+  free(str_copy);
+
+  return exit_code;
+}
+
+int ReadConfig(int* argc, char*** argv, Options* options, BuildConfig* config) {
+  // Default values:
+  config->parallelism = GuessParallelism();
+  options->input_file = "build.ninja";
+
+  // Read config from the environment:
+  int exit_code = ReadFlagsFromEnv(options, config);
+  if (exit_code >= 0)
+    return exit_code;
+
+  // Read config from the command line:
+  return ReadFlagsFromCommandLine(argc, argv, options, config);;
+}
+
 int real_main(int argc, char** argv) {
   BuildConfig config;
   Options options = {};
-  options.input_file = "build.ninja";
 
   setvbuf(stdout, NULL, _IOLBF, BUFSIZ);
   const char* ninja_command = argv[0];
 
-  int exit_code = ReadFlags(&argc, &argv, &options, &config);
+  int exit_code = ReadConfig(&argc, &argv, &options, &config);
   if (exit_code >= 0)
     return exit_code;
 
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -1138,7 +1138,7 @@ int ReadFlagsFromEnv(Options* options, BuildConfig* config) {
   int argc = 0, i = 0;
   bool in_arg = false;
 
-  argv[argc++] = ""; // fake argv[0] (command) for getopt
+  argv[argc++] = (char *)""; // fake argv[0] (command) for getopt
 
   while(str[i] != 0) {
     switch(str[i]) {
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -191,7 +191,7 @@ you don't need to pass `-j`.)
 Environment variables
 ~~~~~~~~~~~~~~~~~~~~~
 
-Ninja supports one environment variable to control its behavior:
+Ninja supports two environment variables to control its behavior:
 `NINJA_STATUS`, the progress status printed before the rule being run.
 
 Several placeholders are available:
@@ -211,6 +211,16 @@ specified by `-j` or its default)
 The default progress status is `"[%f/%t] "` (note the trailing space
 to separate from the build rule). Another example of possible progress status
 could be `"[%u/%r/%f] "`.
+
+`NINJA_FLAGS`, flags to be set before the command line arguments are parsed.
+
+The value of the variable is parsed like command line arguments. You can for
+instance use less cores than available (-j), or limit the parallelism based
+on the system load (-l). For instance, on a build server used by multiple
+users, you may want to set `NINJA_FLAGS=-l 1` on the system level, so that
+the server never gets overloaded by parallel ninja builds. Command line
+parameters override the flags set with NINJA_FLAGS.
+
 
 Extra tools
 ~~~~~~~~~~~


ocroquette:
Not sure why the CI build is failing. Can somebody help me out?

nico:
Previously #797.

src/ninja.cc
```diff
@@ -1118,15 +1118,79 @@ int ReadFlags(int* argc, char*** argv,
...
+  char *str = getenv("NINJA_FLAGS");
+  if (str == NULL || strlen(str) == 0 )
+    return -1;
+
+  char *str_copy = (char *)malloc(strlen(str)+1);
+  char **argv = (char **)calloc(strlen(str), sizeof(char *));
+  int argc = 0, i = 0;
+  bool in_arg = false;
+
+  argv[argc++] = (char *)""; // fake argv[0] (command) for getopt
```
nico: travis is failing because

```
src/ninja.cc:1141:16: warning: deprecated conversion from string constant to ‘char*’ [-Wwrite-strings]
   argv[argc++] = ""; // fake argv[0] (command) for getopt
                ^
```

(click the red x to get to that: https://travis-ci.org/ninja-build/ninja/builds/354768286?utm_source=github_status&utm_medium=notification)

nico:
I'm generally wary of making env vars affect your build. If you have the power to set env vars for everyone, you also have the power to run `alias ninja='ninja -l 1`, which should address your use case as well.

ocroquette:
"alias" is not an option on Windows, nor it will work if Ninja is called from another tool, for instance cmake. In those cases, how can you ensure global settings for -j and -l ?

Would it be acceptable to have dedicated environment variables like NINJA_MAX_JOBS and NINJA_MAX_LOAD? That would affect only build performance and not the build output.

nico:
I'm somewhat skeptical of that too :-/ I've debugged so many build issues where in the end the resolution was "oh I had this one env var set that I set months ago and forgot about".

ocroquette:
Like I said, the two specific environment settings can not impact the build result, so the risk of "build issues" is limited to the performance, which is very easy to troubleshoot with the documentation or FAQ.

Other build tools have something similar, for instance make has MAKEFLAGS, gradle has ~/.gradle/settings.gradle (for instance to enable/disable the daemon), Maven has ~/.m2/settings.xml, cmake has [environment variables](https://cmake.org/cmake/help/v3.10/manual/cmake-env-variables.7.html). All these tools live pretty well with the risk you mentioned, because it's more than compensated by the added value. 

I am obviously not the only one who needs that, there are several valid and common use cases, so what is the alternative to give to the user the possibility to set more sensible default values?

nico:
I agree the risk is limited to performance, but that seems like a pretty big risk in a project whose purpose is performance.

The alternative depends on your exact use case. For the one mentioned in comment 0, on linux I'd probably set the default nice level in limits.conf for each user since it sounds like you don't want this just for ninja builds. For just ninja at an admin level, I'd set the alias. For personal use, I'd manually use the flags. I'm less familiar with Windows, but similar things are probably possible there (if ninja is system-provided, put `ninja -l 1` in a bat file and put that in %path%, say).

ocroquette:
Using an alias or a BAT script has the same issues, with additional problems and limitations mentioned. So for the users who really needs this, it's better to provide a clean way.

ocroquette:
Hi @nico. If the main concern is the troubleshooting, how about displaying a message on the console when the environment variables are used? Something like:

`
Setting flags from environment variable NINJA_FLAGS: -j 4 -v
`

Or if we want to limit this to the performance options:

`
Setting maximum parallel jobs from environment variable NINJA_MAX_JOBS: 4
`

People not using this feature would not be impacted.
People using this feature by mistake or troubleshooting someone's build would be clearly notified.
People using this feature on purpose could find the message annoying, but it's better than having nothing at all.

Hello71:
I mean, it is extremely common to install software to /usr/local then forget about it. will you also disable installation to /usr/local?

aureooms:
I build software from the AUR (Arch User Repository) using ninja. I am searching for a solution to always use `-j1` because I still want to be able to use my system while the build runs, and the packager of the software used ninja's default which I guess is the number of cores of the cpu. Is there another way to assign this parameter by default aside the suggestion made in this pull request?

jhasse:
Instead of an `alias`, you can use this simple Python script:

```python
#!/usr/bin/env python3

import sys
import subprocess

try:
	subprocess.check_call(['/usr/bin/ninja', '-j1'] + sys.argv[1:])
except subprocess.CalledProcessError as e:
	exit(e.returncode)
```

Name it `ninja`, make it executable with `chmod +x` and put it somewhere in your `PATH` (e.g. at `/usr/local/bin`).

Not sure if this will work on Windows, too?

@ocroquette Can you name some limitations with this script approach? It should work with CMake for example.

jonesmz:
While that should be functional, starting a python interpreter just for the sake of parsing a single environment variable seems like huge overkill.

jhasse:
It takes 50 ms on my system. Is that really noticeable in your builds? If I understand it correctly the use-case for this seems to be full rebuilds anyway.

ocroquette:
Thanks for the idea, but I can only repeat my comment above:

> Using an alias or a BAT [or python or whatever] script has the same issues, with additional problems and limitations mentioned. So for the users who really needs this, it's better to provide a clean way.

It looks like the design decision here is not to support any user settings. While I can understand the concern anything that could modify the build result, it doesn't apply for the parallelism factor. It just forces the user to resort to cumbersome solutions which doesn't need to be.

ocroquette:
Python doesn't solve anything. If I wanted to solve this with an additional script, I would use a language supported natively by the platform anyway (for instance CMD on Windows and SH on Unix-likes), so that I don't have to deploy Python on every developer PC and build server of the company just to pass an integer to Ninja.

Relying on a wrapper script is in general not a good option, our whole build system is complex enough by nature that we don't need a hack that relies on external files and the order of directories in PATH.

The issues I was referring to are the one mentioned by nico, the main one being that it is possible to forget about the custom settings values. Refusing to do this in Ninja just moves the problem elsewhere.

jhasse:
> If I wanted to solve this with an additional script, I would use a language supported natively by the platform anyway (for instance CMD on Windows and SH on Unix-likes)

Of course that's also possible, I've just used Python, because I know it better.

> we don't need a hack that relies on external files

It's only one external file.

> and the order of directories in PATH.

Because you don't like reliance on env variables? That's exactly the reason why we are against `NINJA_FLAGS`.

> The issues I was referring to are the one mentioned by nico, the main one being that it is possible to forget about the custom settings values. Refusing to do this in Ninja just moves the problem elsewhere.

Yes, but I think this way it's a bit more explicit. Also moving it out of Ninja is exactly the point for a feature that won't be used by most of Ninja's users.

aureooms:
A working solution then is to create a directory `/etc/mock/ninja/j1`, put the `ninja` python or sh or whatever script in there, then when one needs to set the option `-j1` by default while running another executable `some_executable`, use `env PATH=/opt/mock/ninja/j1:$PATH some_executable ...`.

Am I missing anything?

ocroquette:
I will try to explain one more time.

First, let's recall the user needs mentioned so far. There are users who:
- by default want to build in the background, duration is not important, system load must be low (-j 1)
- build a multiuser system, and need to share the resources (-j N/2)
- build on a system on which Ninja's default value (N+2) overloads the system, typically due to I/O or memory bottlenecks. Setting -j to N or N-x improves the duration of the builds AND the responsiveness of system

So the user wants Ninja to use a custom value instead of the default for all builds. This kind of value is commonly stored by applications in a configuration file or similar (registry on Windows). Look at your home directory, you will find plenty of examples. Imagine that the Git developers force you to use "git --author" on the command line, and tells you that you just have to create a shell alias or a wrapper script if you want to have your own default value. And VIM for your favorite editor settings. And... How attractive does that sound?

That's about the user friendliness. Now the second aspect is the technical difficulties of a wrapper script.

In my company, we have a configuration management tool (let's call it CMT) which manages the tools in the right version for the different project and even for the different project versions. When a developer checks out a project, CMT configures automatically the workspace with all required tools (Ninja, CMake, compilers, Doxygen, Innosetup, IDE...), with the proper versions. The developers don't install anything manually. Not only this saves effort, it makes sure that we use a consistent development environment on all developer's PCs and build servers. This is a common pattern in bigger teams or organisations.

We never call Ninja directly, not even CMake. The typical workflow is:

CMT configures the IDE in the workspace to use the proper tools
The user starts the IDE
The user builds from the IDE. The IDE calls CMake which in turn calls Ninja.

CMT also allows to run CMake directly from the CLI:

CMT -> CMake -> Ninja

In this environment, introducing a user specific wrapper script is a nightmare, as you can imagine, since CMT has full control over the PATH. That's its job. It would be straightforward though if Ninja supported a configuration file where the user can specify his own settings. Then it wouldn't matter which version it is, if it's manually installed or started through CMT, or CMake, or where it is in the PATH list. A single line in a text file like ~/.ninja_conf and it would just work in all the cases, independently of the operating system, just like .gitconfig for git. Before you propose any solution, please ask yourself if you would like it for all the settings you have in your ~/.* files.


PS: using an environment variable was just a try to provide the feature with the simplest possible implementation, and also to reuse a common concept in the landscape of building tools. autoconf, "./configure", gmake and many other support environment variables like CC, CXX, MAKEFLAGS... A user configuration file is maybe a more usual solution for the use case at hand and is fine too.

jhasse:
> Then it wouldn't matter which version it is, if it's manually installed or started through CMT, or CMake, or where it is in the PATH list.

It would matter, as the version needs to be one, which supports `NINJA_FLAGS` or the configuration file.

Also

> manually installed

but earlier you said

> The developers don't install anything manually.

?

> In this environment, introducing a user specific wrapper script is a nightmare,

Why?

> as you can imagine,

I can't sorry.

> since CMT has full control over the PATH.

Doesn't that ... help?

For who do you want to introduce an easy way to override Ninja's defaults. CMT or the user?

jonesmz:
To add to what ocroquette said, lets not forget that a wrapper script involves another program execution. If it's written in python, cmd.exe batch, or bash, that's still *significantly* more work than doing an environment variable access in c++.

A wrapper script as described here would be executed many many hundreds of times per build with the build system that I work on at work (it's a beast of a legacy build system. I'm working to improve this, but legacy systems are legacy systems). That additional processing overhead might add up to 10-20 minutes to a 4 hour build.

But checking an environment variable won't do that. It'll add mere seconds.

Since I keep being told that Ninja is all about speed, lets all at least agree that launching python/bash/cmd.exe is much more expensive than checking an environment variable.

jhasse:
> Because the tools and the versions used (e.g. the PATH) are managed centrally and not controllable by the user: in a CMT supported project, the developer doesn't get choose which Ninja is used., otherwise we loose the benefit of the configuration management for the project.

So the user can set new environment variables like `NINJA_FLAGS`, but not change the `PATH`? Couldn't CMT install the wrapper script for the user?

> Please see the beginning of my last message for more details on this aspect.

> by default want to build in the background, duration is not important, system load must be low (-j 1)

`NINJA_FLAGS` won't help, as Ninja will still run in the foreground. If he configures the Ninja invocation himself, he sure can add `-j 1` there.

> build a multiuser system, and need to share the resources (-j N/2)

That's just a workaround (e.g. what when a process that ninja spawns is multi-threaded?), you should rather have a fair OS-level process scheduler.

> build on a system on which Ninja's default value (N+2) overloads the system, typically due to I/O or memory bottlenecks. Setting -j to N or N-x improves the duration of the builds AND the responsiveness of system

That's definitely an issue, but I think `NINJA_FLAGS` isn't the right solution here (again just a workaround).  Let's discuss this use-case and possible solutions in #1441 instead, please.

> A wrapper script as described here would be executed many many hundreds of times per build with the build system that I work on at work (it's a beast of a legacy build system. I'm working to improve this, but legacy systems are legacy systems). That additional processing overhead might add up to 10-20 minutes to a 4 hour build.

10-20 minutes, let's say 15 minutes. 

15 min = 900 s = 900,000 ms

As I've measured the wrapper script adds about 50 ms, so that's

900,000 ms / 50 ms = 18,000

executions and that's definitely more than "many many hundreds of times per build".

jonesmz:
On the build system I spoke about, removing an invocation of cmd.exe by replacing a function call that "shelled out" the running of a program with directly executing the program, cut the build times by 15 minutes on average. Went from 4hrs 15 -> 4hrs, roughly speaking, on my build farm.

So I'm not just pulling numbers out of my butt :-)

Windows process creation is *slow*. Linux process creation can be much faster than windows.

But if it's taking Ninja 50ms to grab and parse an environment variable, I think we have much larger problems. I can't imagine it would involve more than a handful of pointer dereferences and a string comparison.

What language did you write your wrapper script in?

jhasse:
> On the build system I spoke about, removing an invocation of cmd.exe by replacing a function call that "shelled out" the running of a program with directly executing the program, cut the build times by 15 minutes on average. Went from 4hrs 15 -> 4hrs, roughly speaking, on my build farm.

Sounds to me like you have control over the way `ninja` is called and can easily add any needed flags to it yourself.

> What language did you write your wrapper script in?

Please see above.

ocroquette:
> So the user can set new environment variables like `NINJA_FLAGS` but not change the `PATH`?

Technically yes, CMT doesn't care about variables it doesn't take control of (PATH, CMAKE_MODULE_PATH and a few other).

>  Couldn't CMT install the wrapper script for the user?

No. First, adding wrapper scripts is a very smelly design. Additionally it would require CMT to order the PATH components so that the wrapper script comes first, which is not possible. It would also be necessary to update the wrapper script so that it finds itself in the PATH, and looks further to find the real ninja. It would also require a wrapper script for every OS we support (or worse, deploy Python or similar). Are you seriously advocating such a "solution" just to specify a user setting? Maintaining our custom ninja would be much less hassle.

> `NINJA_FLAGS` won't help, as Ninja will still run in the foreground. If he configures the Ninja invocation himself, he sure can add `-j 1` there.

Background was meant as "it should not have any noticeable impact on the system".
That's the use case of @aureooms above. Did you read it? He doesn't want to specify -j1 on every invocation. We are talking about a user setting. Why should he configure with a wrapper script when every reasonable tool out there offers a more standard and convenient way?

> > build a multiuser system, and need to share the resources (-j N/2)
> 
> That's just a workaround (e.g. what when a process that ninja spawns is multi-threaded?), you should rather have a fair OS-level process scheduler.

It's not only about the CPU, it's also about memory and I/O.
And are you seriously telling us to migrate to another OS to circumvent a Ninja limitation?

> That's definitely an issue, but I think `NINJA_FLAGS` isn't the right solution here (again just a workaround). Let's discuss this use-case and possible solutions in #1441 instead, please.

#1441 is about memory, which is not the problem in my case. The problem on my laptop is I/O. Ninja will never be able to recognize and adapt to all possible bottlenecks. Let the user decide easily what's the best default is on his system in a convenient way.

At this point, I would like to remind of the similar requests so far:

@urkle in #797 
@SebastianBoe in  #1370 
@aureooms in #1399 
@Alexander-- in #1441 
@kobalicek in #1482 
@jonesmz and @ocroquette  here

That's just the tip of the iceberg, e.g. the users who took the time to provide feedback (or even patches). The first request is from 2014. Don't you see yet that the added value for your users is real and immediate?

aureooms:
Command line arguments, configuration files and environment variables are standard. People complain.

kobalicek:
I gave up because anything I wrote was answered with the same arguments again and again. I seriously thought about forking ninja and adding support for something like `NINJA_DEFAULT_JOBS` (not even `NINJA_FLAGS` as that's much more complicated), but simplicity won and I'm now using make on machines where this was problematic. Pity for meson users that cannot pick their 'make' anymore.

jonesmz:
@kobalicek 

I've already decided to fork Ninja for my own use, and use at my job. I would be willing to collaborate with people there, and eventually distribute binaries in some fashion.

https://github.com/jonesmz/splinter

Currently I'm maintaining things as a series of patches on Ninja so I can keep things updated with improvements to Ninja. I'd prefer if contributors tried to do the same thing, but I'm willing to re-work patches as needed, up to the limits of my time availability anyway.

kobalicek:
Agreed!

Drop the damn status message to make ninja environment free! If we should write wrapper scripts then you should write too haha :)

jhasse:
> Additionally it would require CMT to order the PATH components so that the wrapper script comes first, which is not possible. It would also be necessary to update the wrapper script so that it finds itself in the PATH, and looks further to find the real ninja.

You could rename the real ninja binary.

> Background was meant as "it should not have any noticeable impact on the system".
That's the use case of @aureooms above. Did you read it? He doesn't want to specify -j1 on every invocation.

Yes and that use-case can be solved by the wrapper script.

> I seriously thought about forking ninja and adding support for something like NINJA_DEFAULT_JOBS (not even NINJA_FLAGS as that's much more complicated), but simplicity won and I'm now using make on machines where this was problematic. Pity for meson users that cannot pick their 'make' anymore.

So let me get this straight: Because you didn't want to type `ninja -j1` instead of `ninja` you're now typing `make`?

> Ninja already parses environment variables for the status line.

I think the idea is that environment variables shouldn't influence the build outcome or speed. That's why `NINJA_STATUS` is okay. (`TERM` can also influence the output btw)

> What platform? Linux?

Yes.

> Do you have JIT enabled?

No.

> And please address the speed problem.

The wrapper can be written in C or C++ which should bring down the execution time to a few ms.

kobalicek:
@jhasse just convinced me that any further discussion about this doesn't make any sense.

jonesmz:
@jhasse 

> So let me get this straight: Because you didn't want to type `ninja -j1` instead of `ninja` you're now typing `make`?

You know it's more complicated than that. Please consider being less antagonistic to your user base. The people commenting in this thread are by all appearances sophisticated users who are trying to help make ninja better for everyone, including themselves.

I'm not saying that you should agree to every request anyone makes, and I understand the reasoning you're using with the other things you and i have been discussing even though I disagree with you, but on this one I think you're seriously just disagreeing for the sake of disagreeing, and it's making at least one person other than myself extremely frustrated, if not others.

I'm personally very frustrated, and while you're doing a lot of legwork recently, you're winning about as many fans as nico did before you. The Ninja community, as far as the behavior on github and the mailing list, is widely considered a joke by my professional acquaintances. Collectively the maintainers of Ninja are seen as disagreeable, rather irrational, and unwilling to consider common use cases. Maybe you don't care one way or another about that, and my one anecdote is certainly not data. But I believe you're pushing away a lot of potential high quality contributors.

In the situation described, where the "CMT" tool is deploying a full set of tools at configuration time, the options for ninja are hard-coded into whatever the CMT tool is deploying, a the time of deployment, so while a wrapper script would provide the functionality desired, if ninja doesn't offer any kind of feature to handle the use case and there's no wrapper script, there's no "just type -j1", because the CMT tool has already run, and running it again may be very expensive.

Typing "-j1" isn't desirable by the CMT tool maintainer because it's the choice of the specific user. Most likely the CMT tool is defaulting to not setting the -j option at all so that ninja uses numcpus to determine how many jobs to run, and the user could provide an environment variable which would indicate a new default value.

The whole point of this feature discussion / pull request, is to allow the users of whatever higher level build tool to interact with the low level ninja behavior without requiring :

- The higher level build tool to needlessly put in various options to handle this
- The user to need to learn how to interact with those options (Lets see, cmake does this xyz way, meson does it this other way. The stupid thing at work does it abc way, and autotools does it 123 way. Gahahhhhh).
- Wrapper scripts which add unnecessary overhead.
- 

> > Ninja already parses environment variables for the status line.
> 
> I think the idea is that environment variables shouldn't influence the build outcome or speed. That's why `NINJA_STATUS` is okay. (`TERM` can also influence the output btw)
 
The speed of the build is not part of the build outcome. It's something that users want to influence for a variety of reasons that build system maintainers aren't able to predict or control, but does not affect the resulting artifacts beyond, potentially, timestamps embedded in the result by the underlying buildrules (why someone would do this, I don't know, but they would know to expect it to vary wildly).

The status is part of the outcome. Tools may be parsing the console output, and may prematurely stop the build based on that output.

Other tools, such as gnu make, have supported the concept of MAKEFLAGS for... forever?

I reject your excuse here. Please try a different rationale.

> > And please address the speed problem.
> 
> The wrapper can be written in C or C++ which should bring down the execution time to a few ms.

Yea, dude, no. Just no. You are being silly.

No one's interested in writing a C or C++ wrapper script, and dealing with the deployment headache of needing to precompile it for every possible platform they have to deal with, and then deploy it selectively based on the platform. That adds a significant amount of complexity, especially the part involving needing to compile a component of your tool chain by hand.

At this point it's clear that you're disagreeing with us just for the sake of being disagreeable.

Even if someone did write a wrapper script in C/C++, that's *still not* faster than having Ninja parse an environment variable directly. So really, pick your criteria. You are being inconsistent. Do you want speed or not? 

It's not fair to say "Well, Ninja has to be fast, but it's ok if you have to run a slow thing before ninja just so Ninja doesn't have to check the value of a single environment variable." that's just arguing from bad faith.

I honestly do not believe that there is a single user of Ninja who has an objection to allowing an environment variable to set the defaults for the various commandline flags, as long as the commandline overrides the environment variable, except for you, and I'm really struggling to put myself in your shoes here.

We're literally talking about far less than 1 ms worth of environment variable parsing for every time ninja is executed, which is an order of magnitude less time than is required to start a new process. And lets not forget, it's not going to have any effect on the time it takes to calculate and execute the build graph. It's just part of the program initialization, a one time minimal cost.

Further I honestly believe that no one who wants to use those environment variable settings is going to be willing to write any kind of wrapper script (or modify their build system) to support some kind of environment variable setting, unless somehow it's a hard requirement for them. Anyone for which this is already a requirement is willing to pay the price of a wrapper program, but is not going to be happy that they had to do it, or is going to switch away from Ninja to a different tool.

You already, apparently, lost one user over the lack of this feature.

Please reconsider.

But if you're not going to reconsider, then lock this PR, so the whole world can see how stubborn you are being.

kobalicek:
I have updated the issue I have opened:

  https://github.com/ninja-build/ninja/issues/1482

It provides a code as well and it's not as generic as NINJA_FLAGS. Maybe it has a greater chance of being accepted so guys if you are watching this issue you might want to support the one I have opened as well.

jonesmz:
> It provides a code as well and it's not as generic as NINJA_FLAGS. 

Honestly I would rather see a fully generic environment variable.

For me, the perfect solution is to have an environment variable that accepts any flags that the commandline can handle. It gets parsed with the same commandline parser currently used, and then after that the commandline flags get parsed and replaces anything defined by the environment variable.


As I mentioned before, I think the appropriate way to handle this kind of thing is some set (or subset) of

- environment variable
- system config File
- user config file
- ninjafile itself
- commandline

And in fact one could easily imagine a commandline flag (or environment variable) that allows the user to somehow indicate the order of preference.

ocroquette:
I can only second @jonesmz’s message.

@jhasse You clearly don't realize the complexity of the build environments we are dealing with. The people requesting a better solution here are obviously knowledgeable and reasonable. Every addition you make to your workaround makes it even worse.

We can close this discussion quickly, just write clearly that you know better than all of us combined what is good for us, our applications and our users. Then we will know what we are dealing with.

@nico @evmar  is that really the future you want for this project?

jhasse:
@jonesmz I'm working on Ninja in my free time and I'm not a Google employee or anything. Also English isn't my first language and I'm sometimes writing my responses in a rush (I think it's better than not to answer at all or take a very long time). So if anything I said was antagonistic (had to look that one up), it was never my intention and I'm very sorry.

@ocroquette No, I don't know better. I just wanted to give some context as to why this PR was closed. And yes - I do not understand your build environment. All of my questions where legitimate as I wanted to understand your use-case and why the wrapper isn't working for you. Sorry that that came of like I'm some kind of know-it-all - also not my intention at all.

aureooms:
Here is a brilliant idea: Why not make this jobs limit parameter a compile time constant? That would achieve some speed up I am sure.

fbbdev:
Hello everyone, I'm the last one to drop in here and I hope not to offend anyone, also I understand a lot of people got frustrated, but people please do not reply heatedly to a message which is clearly trying to apologize.

Thank you @jhasse also for your latest reply on #1482; I initially supported that issue and not this PR but after reading the whole discussion I'm starting to change my mind. Please everyone excuse me if I'm weighing in unsolicited, but I have a hope of bringing the threads together and maybe restart a reasonable discussion.

Initially I stood behind @nico's argument that a global `NINJA_FLAGS` variable would break assumptions and possibly introduce weird behavior; however, summing up all the arguments that have been brought in favor:
  - there is a big precedent in MAKEFLAGS
  - it would help with a lot of different environments and use cases, which is good since each of us came here with different problems
  - it would be future-proof as it's the most general possible solution

This reasoning is turning me to the NINJA_FLAGS side. However the big problem, as I see it, is that something like `NINJA_FLAGS=-t compdb` or `NINJA_FLAGS=-C somedir` or even `NINJA_FLAGS=target` would _definitely_ influence build output, and this ought not happen IMO. Per my understanding, output predictability is one of the core concerns of ninja, and I'm not sure whether a NINJA_FLAGS warning message as suggested by @ocroquette would be enough, or whether a filter should be applied to `NINJA_FLAGS`, which would make things uncomfortably (uselessly?) more complicated, or which other solution could be adopted.

So I'd definitely like to get support for `NINJA_FLAGS`, but if we all stand behind a more limited `NINJA_NUM_JOBS` (or maybe `NINJA_DEFAULT_JOBS`) as proposed in #1482 I still believe it would be the safest way to go.

It'd be great if we could finally get to discuss this peacefully with _you_ mantainers, and not against you as it's been unfortunately happening until now...
In this regard, could it be beneficial to the discussion if I or someone else were to open a new issue bringing all needs, proposals and arguments together on a clean sheet, so that we could also join our efforts?

Thanks everyone for reading and I hope I don't get caught in the crossfire :sweat_smile:

ocroquette:
Thank you @jhasse for your understanding and the time you spend on this project. If we agree that the need to change the default for -j is legitimate and the current options are 
 not satisfactory in general, let's think further what we can do about it.
  
The proposed generic solution based on NINJA_FLAGS in reference to MAKEFLAGS is easy to implement and use, however I can accept that the it is too dangerous since it can influence the build output.

I tried to classify the current CLI options and environment variables:

Clearly not user/system settings, it is unwanted to store them permanently:
*  CLI --version  print ninja version ("1.8.2")
*  CLI -C DIR   change to DIR before doing anything else
*  CLI -f FILE  specify input build file [default=build.ninja]
*  CLI -k N     keep going until N jobs fail [default=1]
*  CLI -n       dry run (don't run commands but act like they succeeded)
*  CLI -t TOOL  run a subtool


Doesn't influence the build result, could be seen as user/system setting:
*  CLI -j N     run N jobs in parallel
*  CLI -l N     do not start new jobs if the load average is greater than N
*  CLI -v       show all command lines while building
*  CLI -d MODE  enable debugging (use -d list to list modes)
*  CLI -w FLAG  adjust warnings (use -w list to list warnings)
*  ENV NINJA_STATUS

So I see following possible implementation:
* support for all options in NINJA_FLAGS: as mentioned, probably too dangerous, even if "make" is doing this.
* support for legitimate options only in NINJA_FLAGS
* support for legitimate options as specific environment variables, like proposed for the number of jobs in #1482

I could live with any option, since my concrete need is only to set the number of jobs, and all of them cover this.

I have however a preference for the restricted NINJA_FLAGS, because:
* it is easy for users to understand and use
* it is easy for developers to maintain (code and documentation can be shared with CLI parsing)
* it is nicer if at some point the user needs to specify multiple options

While configuration files are still possible (~/.ninjaconf), I think it is too much effort and impact compared to the requirement at hand. It could be introduced later though, if need arises. In this case, I think the redundant combination "restricted NINJA_FLAGS + conf files" would be nicer than "multiple env vars + conf files". This is an additional (rather weak, I admit) argument for restricted NINJA_FLAGS.

@jhasse you are apparently the only maintainer involved in this discussion, so I am afraid the decision in on you :) I would be ready to update my PR to add the restrictions we can agree on, if you decide this is the best way to support our need.

kobalicek:
If you can update the PR to parse `NINJA_FLAGS` where certain flags will be restricted and would print warning when parsed (but will be ignored) then I'm definitely in favor of `NINJA_FLAGS` instead of `NINJA_NUM_JOBS` as I suggested in #1482. It would be a logical approach to me as I only created the issue because of the refusal of `NINJA_FLAGS`.

metti:
I think “CLI -k N keep going ...” can be considered safe as well. It does not change the output of successfully run rules.
In fact I often use “-k 0” in my alias when refactoring to suppress the early abort upon the first failure. I also came across many developers preferring “-k 0” as default.

ocroquette:
> If you can update the PR to parse `NINJA_FLAGS` where certain flags will be restricted and would print warning when parsed (but will be ignored) then I'm definitely in favor of `NINJA_FLAGS` instead of `NINJA_NUM_JOBS` as I suggested in #1482.

OK, good!

> I think “CLI -k N keep going ...” can be considered safe as well. 

I wasn't sure about this one, because one could argue that it changes the build result. I can completely understand the use case you describe though, and I think the ratio "benefits / risk" is in favor of allowing it in NINJA_FLAGS.

jonesmz:
> @jonesmz I'm working on Ninja in my free time and I'm not a Google employee or anything. Also English isn't my first language and I'm sometimes writing my responses in a rush (I think it's better than not to answer at all or take a very long time). So if anything I said was antagonistic (had to look that one up), it was never my intention and I'm very sorry.

I don't want you to think that it was a language issue. I had no suspicion that you were anything but a native English speaker.

I agree that a rushed answer is overwhelmingly better than no answer.

My issue is less about how you said things than what you said.

While I'm completely on board with requiring potential contributors / pull requests to strongly demonstrate the benefit of a contribution, as I think that's in keeping with typical best practices for software engineering, what I didn't like was what appeared to be dismissal of a variety of independently stated reasons for including the patch.

To illustrate, consider again the conversation revolving around the use of a wrapper script. All of the people responding to that topic, except for yourself, outright dismissed a wrapper script as a viable option. We would have been more willing to accept your refusal if the thing we were asking to include was very expensive at run time or ninja had no existing feature to support the concept.

None of us could understand your refusal to include the feature because "It can be done another way" when we were pointing out that it could not be done another way. Part of our criteria for wanting the NINJA_FLAGS feature was, specifically, that all of the other ways that the problem could be solved were not viable for the reasons we explained.

Personally, I don't want to see Ninja include every PR opened, or even most of them, but if something adds "on-topic" functionality to the tool, and has no cost, or very very minimal cost, it's probably wise to seriously consider accepting it.

jonesmz:
@ocroquette

> Thank you @jhasse for your understanding and the time you spend on this project. If we agree that the need to change the default for -j is legitimate and the current options are not satisfactory in general, let's think further what we can do about it.
> 
> The proposed generic solution based on NINJA_FLAGS in reference to MAKEFLAGS is easy to implement and use, however I can accept that it is too dangerous since it can influence the build output.

I don't want to be the stick in the mud... so forgive me for being contrary, but is this really a problem?

Having NINJA_FLAGS simply parse the environment variable as if it were the command line is much simpler to implement under the hood than parsing it, and then only evaluating certain options.

If a user set the environment variable to a bad setting, the only two possibilities are
- They understood what they were doing
- They did not understand what they were doing

In the first case, by filtering the allowed options, Ninja incurs more runtime cost (even if it is so small as to be nearly unmeasurable), and prevents the user that knows what they are doing from doing what they want to do.

If they do not understand what they are doing, all bets are off, as no amount of hand-holding is going to truly work-around a PEBKAC error.

A third case is that the environment variable has been set without the user knowing about it. In which case I suspect the user is in far more trouble than whether or not ninja is working properly...

fbbdev:
@jonesmz I agree with you that filtering NINJA_FLAGS would be inconvenient (and counterintuitive too). If it gets implemented, it should come with full power.

But I'm really concerned with a particular and IMO concretely possible instance of your third case, i.e. a sysadmin with arcane intentions, or some idiot build tool, setting NINJA_FLAGS to a perverse default, which an unknowing user will then be forced to debug. And in my debugging experience a lot of time will often be wasted before one remembers to check for things that are completely out of context.

So, this is not an instance of not knowing what you are doing, actually your very knowledge becomes an hindrance. Also the problem is more about collaboration than about single users.

On the other hand it's true that solving such kind of problems is not in any way ninja's responsibility, yet I strongly believe that some protection, like a warning, should be included, as @ocroquette himself suggested. Which is not convenient either, but still better than nothing.

E.g.:
```
ninja: warning: additional flags parsed from NINJA_FLAGS environment variable, resulting invocation is: ninja -j4 <etc>
```
(with all overloaded flags omitted).

However this should come in a form that avoids hurting build log parsers.

jonesmz:
I'd be elated to see ninja print out it's full command line, after evaluating all of the various places it can get run time options. Command line, env variable, other places).

It doesn't even need to be a warning, just a normal status line.

Perhaps something like

"Ninja build options : /path/to/ninja -j -n -k -l so-on-so-forth"
"    Configuration sources - command line, NINJA_FLAGS env dir, whatever else, but only things that were actually used"
"    Other status lines here"

That gives a nice reminder to the user where configuration information comes from.

jhasse:
> @jhasse you are apparently the only maintainer involved in this discussion, so I am afraid the decision is on you :)

I'm not the only one and didn't even close this PR. Also I can't make final decisions about Ninja.

jonesmz:
@jhasse 

> I'm not the only one and didn't even close this PR. Also I can't make final decisions about Ninja.

You're the only person who has commit access who's done anything for many months. You're the only maintainer that anyone who isn't an official project member cares about.

ocroquette:
@jhasse

> I'm not the only one and didn't even close this PR. Also I can't make final decisions about Ninja.

I tried to catch the attention of other maintainers, with no luck so far. How do you suggest we should proceed then?

Independently of the decision, your opinion as a maintainer matters, so it would definitely help. Do you personally think that the NINJA_FLAGS (restricted or not) is an acceptable way to solve our problem, since we didn't find any other good option in the lengthy discussion?

Just for the record, I am thankful for the free time you spend on Ninja, including on this topic. But keep in mind you are not the only one. I have already spent several hours here, time that I cannot expect my employee to account for. I am surely not the only user in this case.

jhasse:
> I tried to catch the attention of other maintainers, with no luck so far. How do you suggest we should proceed then?

Fastest way would be getting it merged in a fork, e.g. https://github.com/jonesmz/splinter
If you want it in upstream Ninja, I would start a thread on the mailing list, which the other maintainers pay more attention to than GitHub.

> Independently of the decision, your opinion as a maintainer matters, so it would definitely help to hear it.

I would rather stay out of it for a while, sorry.

kobalicek:
No meson and no ninja is the solution for me in the end. Thanks to everybody involved for trying to convince the maintainers the feature was worth it.

jonesmz:
I think the maintainers for Ninja need to stop recommending that people discuss things on the mailing list.

There's been 3 topics, and 12 emails total for the last 30 days.

Clearly the community for Ninja is here, on Github.

@ocroquette I don't think we should give up. 

Lack response from maintainer doesn't imply that the feature won't be merged, eventually.

And of course, the majority of people who are opposed to a feature like this say nothing about why.

Quoting from the mailing list

> Just so it's not "maintainers vs the world" here, I'd like to voice my opposition to using environment variables for ninja configuration. Just as not all users take the time to express their needs, even fewer users take the time to object to PRs they disagree with.

This person simply says they are opposed to it, without providing an explanation.

Since it's not possible for an environment variable that is not set to cause any harm, and all of the proposed alternatives have exactly the same downsides (someone might set the wrapper script up wrong, just like the environment variable being set up wrong) there's been no objection to really argue against.

ocroquette:
> Lack response from maintainer doesn't imply that the feature won't be merged, eventually.

We made our point and actually got responses from the 2 active maintainers, both negative.

> And of course, the majority of people who are opposed to a feature like this say nothing about why. [...]
> This person simply says they are opposed to it, without providing an explanation.
> 
> Since it's not possible for an environment variable that is not set to cause any harm, and all of the proposed alternatives have exactly the same downsides (someone might set the wrapper script up wrong, just like the environment variable being set up wrong) there's been no objection to really argue against.

I agree. But unfortunately, it is very easy for anyone (including maintainers) to be against any solution for a problem you are not impacted by.
I would expect more understanding from the maintainers, but if it is not there, I don't see what we can do.

jonesmz:
> I would expect more understanding from the maintainers, but if it is not there, I don't see what we can do.

Well, I'll continue improving my fork of Ninja, and do my own thing. At the moment, I'm waiting for a release before I update my code against the latest Ninja master, but I may start diverging soon enough *shrug*.

I guess other folks in this thread have spoken their piece as well about moving onto greener pastures.

eli-schwartz:
The independent Ninja build language reimplementation https://github.com/michaelforney/samurai supports this via the environment variable SAMUFLAGS. Simply use samu anywhere you'd otherwise use ninja, or install samu as your system /usr/bin/ninja (some linux distros have options to install this competing implementation as the default, or only, ninja).

lilydjwg:
@eli-schwartz unfortunately samu doesn't support `-l` which is what a multi-user build machine needs. I'll going to create a wrapper and hack the ninja command in a post-install hook instead. It's not a good and nice solution but I think we've run out of other options.

ddevault:
Consider asking samu to implement `-l`, or sending them a patch, I'm sure they'd be open to it.

Salamandar:
@lilydjwg Adding support for -l is **really** not difficult. Please do an issue and mention my name if they say they don't have time for it.

eli-schwartz:
@lilydjwg there's an open feature request to add it to samurai and the developer has even said he will accept a patch for it.

https://github.com/michaelforney/samurai/issues/41

lilydjwg:
@ddevault @Salamandar @eli-schwartz thank you but I've taken another suggestion to make those processes compete fairly with cgroups because I don't need to modify the build environment this way so it's much simpler to implement. Hopefully make & ninja will play fair and give enough CPU resources to system services. (make -l & ninja -l doesn't play very well: the process numbers are unbalanced and CPU cores are under-utilized.)

Salamandar:
I think I agree. They just "dont start jobs if load is too high", that's not a good idea for a multi-user machine. Cgroups are the way to go.

RJVB:
>I think I agree. They just "dont start jobs if load is too high", that's not a good idea for a multi-user machine. Cgroups are the way to go.

Are those available on all modern Unix variants?

jhasse:
Depends on your definition of "modern Unix".

But I guess that's a rhetorical question anyway and you already know that cgroups are a Linux kernel feature.

RJVB:
>But I guess that's a rhetorical question anyway and you already know that cgroups are a Linux kernel feature. 

Actually I didn't, just had a hunch. Here's another not-really rhetorical question: how many such a platform-specific feature can you use in what's supposed to be a lean and mean application, before it isn't anymore? ;)

I agree though that the current load-limited option is (AFAICT) mostly interesting/useful if you want to allow ninja to take over the entire host without shooting itself in the foot (too much).

eli-schwartz:
What does this have to do with anything?

Someone who previously thought samurai wasn't a good replacement for ninja because it doesn't implement `-l` has now decided it's perfectly fine because they don't need `-l`, because they'd much rather use linux cgroups *before* running ninja/samurai.

No one has suggested implementing or reimplementing `-l` to simply set up cgroups itself, so I don't know what "lean and mean application" you think you are referring to.

touraill-adobe:
Having the ability to configure a tool with environment variables also makes it much easier to control it when it is buried in a stack of subprocesses, which is often the case when dealing with build systems. For instance you can have `cmake` calling `vcpkg` calling `cmake` calling `ninja`: passing the appropriate options to `ninja` through this stack is at best a nightmare, at worst impossible, whereas with environment variables you can do that easily:
```NINJA_FLAGS="-d explain" cmake ...```

A wrapper script is much harder to maintain and to "activate" temporarily (you would need to push/pop from the `PATH`, reconfigure projects that store absolute paths to `ninja`, etc.).

jhasse:
> A wrapper script is much harder to maintain and to "activate" temporarily (you would need to push/pop from the PATH, reconfigure projects that store absolute paths to ninja, etc.).

Your wrapper script could pass $NINJA_FLAGS which would invalidate the points (no need for push/pop, no need to reconfigure).

RJVB:
On Thursday January 15 2026 13:06:32 Jan Niklas Hasse wrote:

>Your wrapper script could pass $NINJA_FLAGS which would invalidate the points (no need for push/pop, no need to reconfigure).


What's the point in doing that in a wrapper script if you can just do it in the parent environment?

A global wrapper script *could* provide support for a NINJA_FLAGS env. var, indeed. It'd really fit in with the philosophy of a lightweight, fast build tool to have to implement sought-after flexibility via a wrapper script (whereas the contents of that variable could be processed by the existing arguments parser with far less overhead).

lilydjwg:
> you would need to push/pop from the `PATH`, reconfigure projects that store absolute paths to `ninja`, etc.

You would need to properly package the wrapper script and then install it in the build environment when using a chroot packaging tool.

jhasse:
If we go forward with this I would like to keep it at one env var, so my roadmap would be:

1. Replace getopt with custom/better arg parsing (I have a WIP prototype).
2. Support `--status` in addition to `NINJA_STATUS`.
3. Add `NINJA_FLAGS` support.
4. Deprecate `NINJA_STATUS`.
5. Remove `NINJA_STATUS`.

@RJVB If I understand you correctly: There would be no point in it, but currently it isn't possible (as ninja doesn't support NINJA_FLAGS).

RJVB:
FWIW and just for reference: `samu` already supports `SAMUFLAGS` (but only for -v, -j and -l). As you'd expect these arguments are injected before (and so can be overriden by) the command line arguments.

To be expected but debatable for things like colour output that I've already seen being hardcoded by build systems.

eli-schwartz:
> To be expected but debatable for things like colour output that I've already seen being hardcoded by build systems.

Color is hardcoded by build systems because ninja insists to strip colors anyway based on tty checks, so the normal behavior of letting the tool check for a try and emit colors via --color=auto is:
- crippled (tty isn't passed through)
- redundant (it would be stripped if it did work)

There's no practical point in *ever* allowing any tool to utilize --color=never given the current state of ninja, which "handles it for you" (whether you like it or not) :D

I'm interested in ideas for solving this in ninja -- and without solving it in ninja then as the maintainer of a build system I'd prefer not to be bothered by "bright ideas" about what the build system should do :D

RJVB:
On Friday January 16 2026 09:11:27 Eli Schwartz wrote:

>Color is hardcoded by build systems because ninja insists to strip colors anyway based on tty checks, so the normal behavior of letting the tool 

Not to get side-tracked by what was just an example, but colours used by CLI tools are typically a PITA for me because 1) they weren't chosen with *my* terminal background colour in mind and/or 2) the output gets sent to file and opened in traditional utilities where the colour "opcodes" only get in the way.

eli-schwartz:
> Not to get side-tracked by what was just an example, but colours used by CLI tools are typically a PITA for me because 1) they weren't chosen with *my* terminal background colour in mind

Yes, normally one assumes that users who specifically don't want color for this reason, pass --color=never to the tools that don't support color schemes. --color=auto is the default for most tools for a very good reason. It is unfortunate that ninja as of today breaks this in such a way that the only way to fix it is to add --color=always unconditionally to all programs... just to enable the *effect* of --color=auto

>  and/or 2) the output gets sent to file and opened in traditional utilities where the colour "opcodes" only get in the way.

That being said if the output gets sent to a file I tell `less` to process colors. Other people just pipe through `ansifilter` to remove all color codes before (saving it to a file and/or) displaying it in their non-color-compatible utilities, since not all programs even support disabling color via an option.

PR #1405, opened 2018-04-05
Fix confusing smart console output from concurrent builds

nico:
Developers tend to blame the last printed line when a build takes too
long. Unfortunately, when building concurrently, the last printed line
may have actually finished a long time ago. Under the current system,
ninja does not update the status line to reflect what jobs are still
running. This change makes ninja always print the oldest still running job
instead. In other words, the likely build bottlenecks.

Patch from David Zarzycki, originally uploaded at #1320.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -16,6 +16,7 @@
 
 #include <assert.h>
 #include <errno.h>
+#include <limits.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <functional>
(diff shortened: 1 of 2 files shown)


src/build.cc
```diff
@@ -130,6 +131,18 @@ void BuildStatus::BuildEdgeFinished(Edge* edge,
...
+    int oldest_start = INT_MAX;
+    Edge* oldest = NULL;
+    for (i = running_edges_.begin(); i != running_edges_.end(); i++) {
+      if (i->second < oldest_start) {
+        oldest_start = i->second;
+        oldest = i->first;
+      }
+    }
+    if (oldest)
+      PrintStatus(oldest, kEdgeRunning);
```
colincross: This seems like the wrong place, if the finished edge has any output or fails this will overwrite the status line for the finished edge with the status line for a still running edge and then follow it with the output of the finished edge.  I think this whole block has to move to the end of the function.

nico:
Thanks for the report.

Another thing I noticed while using this is that Terminal.app feels less responsive. I think this might be because we now print 50% more output (1 line per started and 2 per completed edge, as opposed to 1+1 before). So that needs looking into too.

I'll try to take a look at both issues.

sorbits:
> I think this might be because we now print 50% more output (1 line per started and 2 per completed edge, as opposed to 1+1 before)

@nico Have a look at the patch I did for this, it only prints status when we are down to one task running, so for a regular build, you normally would not get any extra status printed, only when one task is holding up the build: https://github.com/sorbits/ninja/commit/192fed9c10025f721e63e931ad6aba001d8d59df

PR #1408, opened 2018-04-05
[wip/demo/do not merge] dramatically improve error messages in many c…

nico:
…ases

two changes (should be landed as two pull requests i suppose):
- Lexer::ReadIdent() now sets last_token_ before returning false, like
  Lexer::ReadEvalString() does.  So all "expected identifiers" and
  things that call ReadIdent (pool parser, rule parser, let parser,
  code parsing the rule name after a : in a build line) now point
  the "^ near here" at what was there instead of what we expected
  instead of the previous token. (landed in #1415)
- If last_token_ is a newline, the snippet would usually show an empty
  line. In that case, show the previous line instead, which is the line
  where the error would have been.

See the manifest_parser_test diff for what this changes.  Almost
everything is much better. Two things get slightly worse (the "expected
'command =' line for rules is now at the end of the rule line instead
of after it where it goes, and the same for the 'depth =' line for
pools). That's not ideal, but also not horrible.

### DIFF
--- a/src/lexer.cc
+++ b/src/lexer.cc
@@ -23,12 +23,17 @@
 bool Lexer::Error(const string& message, string* err) {
   // Compute line/column.
   int line = 1;
-  const char* line_start = input_.str_;
+  const char* line_start = input_.str_, * prev_line_start = NULL;
   for (const char* p = input_.str_; p < last_token_; ++p) {
     if (*p == '\n') {
       ++line;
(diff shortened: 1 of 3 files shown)


nico:
Landing the first part, which is 100% better, at #1415.

jhasse:
To reduce the number of open PRs I'm closing this one. Feel free to reopen (or open a new one for the second part).

PR #1409, opened 2018-04-06
Add appveyor.yml for build & test on windows

atetubou:
I'd like to ninja repository has build & test for windows.

This configuration build & test like below.
https://ci.appveyor.com/project/atetubou/ninja/build/1.0.15

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,14 @@
+version: 1.0.{build}
+image: Visual Studio 2017
+build_script:
+- cmd: >-
+    call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvars64.bat"
+
+    python configure.py --bootstrap
+
+    ninja.bootstrap.exe all
+
+    ninja_test
+
+    python misc/ninja_syntax_test.py
+test: off


atetubou:
> I'm pretty sure we never had an appveyor.yml – is that a new way to configure appveyor? Can I delete the webhook now? Do you know how my instance happens to know what to do? Do you know why my instance doesn't show up on pull requests?

I think that you need to do followings to enable appveyor in this project.
1. merge this PR to master branch
1. make ninja-build/ninja project on appveyor (done already?)
1. delete webhook after confirming that a new pull request kicks appveyor like https://github.com/atetubou/ninja/pull/1

(Maybe you also need to tweak settings on https://ci.appveyor.com/project/ninja-build/ninja/settings)

appveyor.yml
```diff
@@ -0,0 +1,14 @@
+version: 1.0.{build}
+image: Visual Studio 2017
+build_script:
+- cmd: >-
+    call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvars64.bat"
```
nico: Is there a way to test with multiple MSVS versions? The old bot used 2012 for example, and I think the cmake folks used to have a 2010 bot even.

nico:
Ok, 1 is done :-) I think 2 maybe isn't, there's only nico/ninja, not ninja-build/ninja as far as I know.

appveyor.yml
atetubou: Thank you for merging!

appveyor supports vs2013, vs2015, vs2017, but not vs2012.
https://www.appveyor.com/docs/build-environment/#build-worker-images

If you want, I will make another PR.

nico:
I figured out where I had configured the build -- at https://ci.appveyor.com/project/nico/ninja/settings/build there's a "script" thingy where one can paste a script. Having this checked in to the ninja repo seems nicer though.

sgraham:
[I don't remember whether I set up Appveyor for ninja before (maybe? I can't find the account that I might have used if I did), but when I did it for gyp I put all the config into the AppVeyor UI to avoid adding an appveyor file in the repo.]

sgraham:
Oops, what you said. :)

nico:
I had the bright idea to ask for help, at https://github.com/appveyor/ci/issues/2237 Within minutes, the good appveyor people told me what button to push to make appveyor builds show up again.

It looks like appveyor now uses appveyor.yml instead of the script I had configured in their web ui, which is probably a good thing :-)

Issue #673, opened 2013-10-19
configure.py/bootstrap.py uses CXX with CFLAGS 


geoff-nixon:
...rather than CXXFLAGS.
Thus,
`CC=clang CXX=clang++ CFLAGS=-std=c11 CXXFLAGS=-std=c++11 configure.py; ninja`
or the like raises
`error: invalid argument '-std=c11' not allowed with 'C++/ObjC++'`.

PR #1410, opened 2018-04-06
Honor CXXFLAGS in configure.py

ryandesign:
Prior request for this: https://github.com/ninja-build/ninja/issues/673

Prior attempted implementation: https://github.com/ninja-build/ninja/pull/676

As with the prior implementation, this tries to ensure existing workflows that depend on using `CFLAGS` are not affected. Supplied `CFLAGS` and `CXXFLAGS` are also now added to `LDFLAGS`, since some C++ flags like `-stdlib` are needed by the linker too.


nico:
Thanks!

PR #1412, opened 2018-04-06
Factor CreatePhonyInEdge into a new base class of ImplicitDepLoader

bradking:
Create a DepLoader base class to hold the method.  This will allow it to
be re-used by other kinds of dependency loaders.


jhasse:
Similar to #1413, please just add this commit in a PR which adds a different kind of dependency loader.

PR #1413, opened 2018-04-06
Factor out a base class of ManifestParser

bradking:
Create a Parser base class that holds parser functionality not specific
to the build manifest file format.  This will allow it to be re-used for
other parsers later.


src/parser.cc
```diff
@@ -0,0 +1,51 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
```
atetubou: Copyright 2018 ?

src/parser.h
```diff
@@ -0,0 +1,49 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
```
atetubou: Copyright 2018?

src/parser.cc
```diff
@@ -0,0 +1,51 @@
...
+
+bool Parser::Load(const string& filename, string* err, Lexer* parent) {
+  METRIC_RECORD(".ninja parse");
+  string contents;
+  string read_err;
+  if (file_reader_->ReadFile(filename, &contents, &read_err) !=
+      FileReader::Okay) {
+    *err = "loading '" + filename + "': " + read_err;
+    if (parent)
+      parent->Error(string(*err), err);
```
atetubou: parent->Error(*err, err);?

src/parser.cc
bradking: Updated.  I've been rebasing this change since it was first made in 2015.

src/parser.h
bradking: Updated.

src/parser.cc
bradking: The `string(*err)` came from the original `ManifestParser::Load`.  It is only moved by this PR.

The explicit temporary is necessary because `Lexer::Error` writes to the string passed as the second argument and reads from the string passed as the first argument in an unspecified order.  We cannot pass a reference and a pointer to the same string.

atetubou: Oh, I see. Sorry for wrong comment.

jhasse:
This looks find, but there's no need to merge this without any alternative parsers. Can you include this commit in the PR you're working towards?

bradking:
Sure, but the work is pretty extensive and this prepares it with incremental PRs.

jhasse:
That's not a problem, as we can look at the individual commits in a PR.

PR #1414, opened 2018-04-06
Allow EdgeFinished and NodeFinished to fail with errors

bradking:
Add an 'err' string argument and return a boolean for success.  Update
call sites to pass an 'err' string argument and check the return value.
This will be useful later for adding logic to these methods that may
fail.


jhasse:
Similar to #1413, please just add this commit in a PR which uses the err argument.

PR #1417, opened 2018-04-07
Silence GCC -Wimplicit-fallthrough warnings

stefanb2:
Add a comment after each case to indicate our intention to the compiler.

GCC 7.x -Wextra implies -Wimplicit-fallthrough=3.

### DIFF
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -76,11 +76,17 @@ uint64_t MurmurHash64A(const void* key, size_t len) {
   switch (len & 7)
   {
   case 7: h ^= uint64_t(data[6]) << 48;
+          NINJA_FALLTHROUGH;
   case 6: h ^= uint64_t(data[5]) << 40;
+          NINJA_FALLTHROUGH;
   case 5: h ^= uint64_t(data[4]) << 32;
+          NINJA_FALLTHROUGH;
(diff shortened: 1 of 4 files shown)


stefanb2:
Failures on AppVeyor are not caused by my changes.

nico:
Thanks for the patch!

If we want to fix this, I think we want a device like https://cs.chromium.org/chromium/src/third_party/boringssl/src/crypto/internal.h?type=cs&q=OPENSSL_FALLTHROUGH&sq=package:chromium&l=175 to fix the warning with clang too. But there are so few switch statements in ninja that we might just want to pass -Wno-implicit-fallthrough and not touch the code. (I'm a big fan of the warning in general – e.g. https://crbug.com/177475 – but I feel it might not be worth it for ninja.)

stefanb2:
Before making this change I checked the ninja source code for precedence. There is one in `src/util.cc`. So I assumed that simply copy & pasting that comment to the other places was the preferred solution by the project.

stefanb2:
I guess that is the solution you prefer.

stefanb2:
Ping? Is there an issue with getting this small change merged?

nico:
Thanks!

PR #1423, opened 2018-04-11
Remove broken disk_interface test for Stat

atetubou:
We don't want to pass non-canonicalized path to DiskInterface::Stat.

This makes test pass on appveyor.

### DIFF
--- a/src/disk_interface.h
+++ b/src/disk_interface.h
@@ -46,6 +46,7 @@ struct FileReader {
 struct DiskInterface: public FileReader {
   /// stat() a file, returning the mtime, or 0 if missing and -1 on
   /// other errors.
+  /// |path| should be canonicalized.
   virtual TimeStamp Stat(const string& path, string* err) const = 0;
 
   /// Create a directory, returning false on failure.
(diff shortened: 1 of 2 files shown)


jhasse:
Thanks for the PR, we definitely need to fix that AppVeyor build. Can you elaborate what's the problem with non-canonicalized paths?

ikifof:
Ninja relies on the path being canonicalized: so it does not seem to make sense to test the behavior of utility functions in ninja core on non-canonicalized path, mainly if some platforms do not properly support some non-canonicalize paths (see #1420 ). 
If we make the assumption that all path that reach this point are canonicalized, should we assert if we get one that is not? (although that will probably make a lot of the tests to fail).

Another possibility would be to simply handle that case in the code for windows: 

```
  string dir = DirName(path);
  string base(path.substr(dir.size() ? dir.size() + 1 : 0));
  if (base == "..") {
    // StatAllFilesInDir does not report any information for base = "..".
    base = ".";
    dir = path;
  } else if (!dir.size()) {
    dir = ".";
  }

```
With this change the test passes and can be kept as is without making the assumption that only canonical paths can be used.

That being said: to maintain quality, it seems it would make sense to insure that all paths that reach certain functions have been per-canonicalized.

atetubou:
Sorry, I cannot make a time to invest for this near future. So please feel free to take this if someone interested in this.

jhasse:
No problem. Thanks for the PR and the explanation, I will take a closer look (deactivated the tests for now).

PR #1424, opened 2018-04-13
Teach FakeCommandRunner to support multiple active commands

bradking:
Replace our single active edge pointer with a vector and add a
parameter that tests can set to limit the number of concurrent
edges.  Set the default to 1 to preserve the current behavior.
Specific tests will be able to override it later to simulate
concurrent builds.


jhasse:
Similar to #1413, please just add this commit in a PR which uses the functionality.

PR #1425, opened 2018-04-17
ignore duplicate outputs caused by phony edges

fbenkstein:
CMake generates phony edges for files that are inputs to the
'configure_file' command.  When configure_file is called from different
directories with the same file it will be listed as a phony output
twice.  Teach ninja to ignore this special case.

### DIFF
--- a/src/manifest_parser_test.cc
+++ b/src/manifest_parser_test.cc
@@ -351,6 +351,43 @@ TEST_F(ParserTest, NoDeadPointerFromDuplicateEdge) {
   // That's all the checking that this test needs.
 }
 
+TEST_F(ParserTest, DuplicatePhonyOutputIgnored) {
+  const char kInput[] =
+"build out out: phony\n";
+  ManifestParserOptions parser_opts;
+  parser_opts.dupe_edge_action_ = kDupeEdgeActionError;
+  ManifestParser parser(&state, &fs_, parser_opts);
+  string err;
+  EXPECT_TRUE(parser.ParseTest(kInput, &err));
+  EXPECT_EQ("", err);
+}
+
+TEST_F(ParserTest, DuplicatePhonyEdgeIgnored) {
+  const char kInput[] =
+"build out: phony\n"
+"build out: phony\n";
+  ManifestParserOptions parser_opts;
+  parser_opts.dupe_edge_action_ = kDupeEdgeActionError;
+  ManifestParser parser(&state, &fs_, parser_opts);
+  string err;
+  EXPECT_TRUE(parser.ParseTest(kInput, &err));
+  EXPECT_EQ("", err);
+}
+
+TEST_F(ParserTest, DuplicateNonPhonyEdgeNotIgnored) {
+  const char kInput[] =
+"rule cat\n"
+"  command = cat $in > $out\n"
+"build out: cat in\n"
+"build out: phony\n";
+  ManifestParserOptions parser_opts;
+  parser_opts.dupe_edge_action_ = kDupeEdgeActionError;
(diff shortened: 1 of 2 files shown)


jhasse:
Why does CMake do this exactly?

fbenkstein:
> Why does CMake do this exactly?

When such a file is touched the manifest needs to be regenerated so it needs to be a dependency of `build.ninja`.  But if such a file is deleted you also need to be able to regenerate without ninja bailing out with the `'<filename>', needed by 'build.ninja', missing and no known rule to make it` error.  Therefore the `phony` edge is added.  Lastly, the duplication likely comes from there being no way to track these rules in CMake (at least not currently).

@bradking maybe this should be a CMake bug?

Even if it were fixed in CMake however I see no harm to ignoring these edges in ninja.

bradking:
@fbenkstein yes, you're welcome to fix CMake's Ninja generator to not duplicate these in the first place.

fbenkstein:
I've proposed a fix to CMake here:  https://gitlab.kitware.com/cmake/cmake/merge_requests/2586.  It would be nice if this could be addressed in Ninja as well. Especially since 42eefc854b32b3c1efc27548baaacdebad199853 (which would make it into the next release, I guess) makes duplicate rules an error by default.

jhasse:
Hm ... does that mean that step 2 of #931 wasn't completed?

fbenkstein:
> Hm ... does that mean that step 2 of #931 wasn't completed?

This could be considered a CMake bug also.  It's now tracked here:  https://gitlab.kitware.com/cmake/cmake/issues/18584

For reference the generated edges in question look like this:

```
build ../CMakeLists.txt ../a ../s/../a ../s/CMakeLists.txt <system cmake modules>: phony
```

The problem comes `../s/../a` not being canonicalized by CMake but canonicalized to `../a` in ninja.  I think Ninja really complains about the wrong thing here, especially since it's a `phony` edge.  Hence I still think this PR should be merged to help people running Ninja from the master branch (or the next released version) without having the fix in CMake.

nico:
We can revert https://github.com/ninja-build/ninja/commit/42eefc854b32b3c1efc27548baaacdebad199853 for a while longer if it's useful, but since there's still an explicit opt-out and cmake bugs apparently didn't get reported and fixed without the default flip, maybe it's not needed. But I weakly feel we should probably do that and wait at least until there's a cmake release containing https://gitlab.kitware.com/cmake/cmake/commit/6199637e9540627b03d9018ff53a14f005274607 (thanks for the fix!).

This does sound more like a bug in the cmake generator to me.

I haven't seen this in the cmake-using projects I build (mostly llvm and a few small personal one-off projects). I suppose those don't use `configure_file` twice with the same input. Do you have a sense for how common that is?

So my suggested way forward here is to possibly revert https://github.com/ninja-build/ninja/commit/42eefc854b32b3c1efc27548baaacdebad199853 (depending on how many people use `configure_file` with the same input multiple times; https://gitlab.kitware.com/cmake/cmake/issues/18584 sounds like there's also a cmake-side workaround, and affected people can still explicitly `-w dupbuild=warm`), but not merge this PR. Does that sound fine?

bradking:
IMO Ninja should not revert 42eefc854b32b3c1efc27548baaacdebad199853 for 1.9.  The case on the CMake side that produces this, while not hard, is fairly rare and easy to work around.  End users can always use `-w dupbuild=warn`.

bradking:
FYI I backported the fix on the CMake side to be included in 3.13.0.

nico:
Cool, thanks. Sounds like we can close this then.

PR #1426, opened 2018-04-22
Add a new NINJA_STATUS format: %E for ETA time.

irishrover:
New '%E' format prints a time remaining to complete the build.
Now '%e' and '%E' both print not just seconds, but hh:mm:ss.


irishrover:
gcc compiler error.

PR #1428, opened 2018-04-22
Add a new NINJA_STATUS format: %E for ETA time.

irishrover:
'%E' prints a time remaining to complete the build.
Now '%e' and '%E' both print not just seconds, but h:mm:ss.

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -205,7 +205,8 @@ Several placeholders are available:
 `%o`:: Overall rate of finished edges per second
 `%c`:: Current rate of finished edges per second (average over builds
 specified by `-j` or its default)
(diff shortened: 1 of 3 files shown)


irishrover:
@nico, ping :-)

nico:
This assumes that all edges run at the same rate, right? I think previous discussions about this suggested using historical times from .ninja_log if present, which someone said makes the ETA much more precise.

irishrover:
> This assumes that all edges run at the same rate, right? I think previous discussions about this suggested using historical times from .ninja_log if present, which someone said makes the ETA much more precise.

OK, so I close this PR. Thanks.

LebedevRI:
Superseded by #1963.

PR #1432, opened 2018-04-28
Adding a tool to remove dead outputs.

ikifof:
The build directory should have the same contents whether it was an incremental build, a clean build, or a build done from scratch.

Currently if the manifest is updated and does not produce an output anymore, the given output is left being in the build directory. This could impact the behavior of the product being built.

In theory, if the manifest is generated, the generator could try to figure out the changes that forced a re-generation and take care of the eventual cleanup... But that seems a bit artificial, mainly since ninja holds the information needed and there should be no need to duplicate the information or the logic.

Initially I thought that cleaning these dead outputs should be done as part of the re-compact and could be done after the manifest has been updated by the generator, and before we start looking for what is dirty. But I found that:
  - when a tools is selected, it is executed and ninja exits after it is done (currently no possibility to perform the build after running a tool);
  - the re-compact uses a isPathDead() method to cleanup dead entries in the log, which I was expecting to return true when the path has no node, or the node is not inEdge()...  but the logic was different (and I think partially broken -- thus the modification I made part of this fix) , and justified in the commit comment that points to [this discussion](https://github.com/ninja-build/ninja/pull/697#issuecomment-37140762) to justify not removing the entries for the file that still exist.

Since I do not fully understand the requirement, and I cannot find test for NinjaMain::isPathDead() that would clarify the intent (what is the test methodology for NinjaMain?), and I do not have enough of the big picture yet, I decided to insulate the feature by creating a new tool (could not add just a new option to the existing clean tool since it needs the logs to be loaded):
 
       ninja -t recompact    # leaves the entries for the dead paths that exist in the build directory
       ninja -t cleandead     # removes the dead paths that exist in the build directory
       ninja -t recompact    # now this can remove all the dead paths 

I hope this helps.

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -123,6 +123,7 @@ struct NinjaMain : public BuildLogUser {
   int ToolTargets(const Options* options, int argc, char* argv[]);
   int ToolCommands(const Options* options, int argc, char* argv[]);
   int ToolClean(const Options* options, int argc, char* argv[]);
+  int ToolCleanDead(const Options* options, int argc, char* argv[]);
   int ToolCompilationDatabase(const Options* options, int argc, char* argv[]);
   int ToolRecompact(const Options* options, int argc, char* argv[]);
   int ToolUrtle(const Options* options, int argc, char** argv);
@@ -153,9 +154,6 @@ struct NinjaMain : public BuildLogUser {
   void DumpMetrics();
 
   virtual bool IsPathDead(StringPiece s) const {
-    Node* n = state_.LookupNode(s);
-    if (!n || !n->in_edge())
-      return false;
     // Just checking n isn't enough: If an old output is both in the build log
     // and in the deps log, it will have a Node object in state_.  (It will also
     // have an in edge if one of its inputs is another output that's in the deps
@@ -719,6 +717,11 @@ int NinjaMain::ToolClean(const Options* options, int argc, char* argv[]) {
   }
 }
 
+int NinjaMain::ToolCleanDead(const Options* options, int argc, char* argv[]) {
+  Cleaner cleaner(&state_, config_, &disk_interface_);
+  return cleaner.CleanDead(build_log_.entries());
+}
+
 void EncodeJSONString(const char *str) {
   while (*str) {
     if (*str == '"' || *str == '\\')
@@ -893,6 +896,8 @@ const Tool* ChooseTool(const string& tool_name) {
       Tool::RUN_AFTER_LOAD, &NinjaMain::ToolRecompact },
     { "rules",  "list all rules",
       Tool::RUN_AFTER_LOAD, &NinjaMain::ToolRules },
+    { "cleandead",  "clean built files that are no longer produced by the manifest",
+      Tool::RUN_AFTER_LOGS, &NinjaMain::ToolCleanDead },
     { "urtle", NULL,
       Tool::RUN_AFTER_FLAGS, &NinjaMain::ToolUrtle },
     { NULL, NULL, Tool::RUN_AFTER_FLAGS, NULL }
--- a/src/clean.cc
+++ b/src/clean.cc
@@ -124,6 +124,19 @@ int Cleaner::CleanAll(bool generator) {
   return status_;
 }
 
+int Cleaner::CleanDead(const BuildLog::Entries& entries) {
+  Reset();
+  PrintHeader();
+  for (BuildLog::Entries::const_iterator i = entries.begin(); i != entries.end(); ++i) {
+    Node* n = state_->LookupNode(i->first);
+    if (!n || !n->in_edge()) {
+      Remove(i->first.AsString());
+    }
+  }
+  PrintFooter();
+  return status_;
+}
+
 void Cleaner::DoCleanTarget(Node* target) {
   if (Edge* e = target->in_edge()) {
     // Do not try to remove phony targets
--- a/src/clean.h
+++ b/src/clean.h
@@ -20,6 +20,7 @@
 
 #include "build.h"
 #include "dyndep.h"
+#include "build_log.h"
 
 using namespace std;
 
@@ -58,6 +59,10 @@ struct Cleaner {
   /// Clean the file produced by the given @a rules.
   /// @return non-zero if an error occurs.
   int CleanRules(int rule_count, char* rules[]);
+  /// Clean the files produced by previous builds that are no longer in the
+  /// manifest.
+  /// @return non-zero if an error occurs.
+  int CleanDead(const BuildLog::Entries& entries);
 
   /// @return the number of file cleaned.
   int cleaned_files_count() const {
--- a/src/clean_test.cc
+++ b/src/clean_test.cc
@@ -15,7 +15,16 @@
 #include "clean.h"
 #include "build.h"
 
+#include "util.h"
 #include "test.h"
+
+#ifndef _WIN32
+#include <unistd.h>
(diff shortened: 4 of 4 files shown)


src/util.cc
```diff
@@ -56,7 +56,7 @@
 
 void Fatal(const char* msg, ...) {
   va_list ap;
-  fprintf(stderr, "ninja: fatal: ");
+  fprintf(stderr, ": ninja: fatal: ");
```
ikifof: When a test fails, the ninja was merged with the test name making it more difficult than necessary to copy/paste it in the exclusion list. At the minimum there should be a space or a \n before "ninja:".

src/ninja.cc
```diff
@@ -148,9 +149,6 @@ struct NinjaMain : public BuildLogUser {
   void DumpMetrics();
 
   virtual bool IsPathDead(StringPiece s) const {
-    Node* n = state_.LookupNode(s);
-    if (!n || !n->in_edge())
-      return false;
```
ikifof: I think this is a bug, right?

src/ninja.cc
```diff
@@ -827,6 +830,8 @@ const Tool* ChooseTool(const string& tool_name) {
       Tool::RUN_AFTER_LOAD, &NinjaMain::ToolCompilationDatabase },
     { "recompact",  "recompacts ninja-internal data structures",
       Tool::RUN_AFTER_LOAD, &NinjaMain::ToolRecompact },
+    { "cleandead",  "clean built files that are no more produced by the manifest",
```
jhasse: I'm not a native English speaker, but shouldn't this be "are not produced anymore" instead?

src/util.cc
jhasse: `Fatal` is also used by non-test code though.

ikifof: But leaving a space after « : » is standard, no?

src/ninja.cc
ikifof: Not a native speaker either... Cambridge says the « no more » is more formal than « not any more », and that « any more » is more English than « anymore »(one word) is more American English... so ... you decide what you prefer ;-)

jhasse: Okay, it's fine then.

src/util.cc
jhasse: Yes, but previously there wasn't even a `:`.

ikifof: This is not part of the main feature here, so if you do not want it, I can remove it.

That being said, the issue here is that this blindly get appended to the error message (for instance when the ninja test fails, “ninja” ends up mangled at the end of the failing test name making it impossible to efficiently select (double click on the failing test name also copies ninja). Ideally this message would make sure it start with a separator(new line or :, ...) or we need to enforce that any error will end with a separator... it seems simpler to do it centrally and a new line seems overkill, so I opted for the compact “: “. If the message ends up with a new line, the extra “: “ at the beginning of the line seems less of an issue than the problem it fixes.

jhasse: I see. Might be something fixed by using GoogleTest (see #1562).

> This is not part of the main feature here, so if you do not want it, I can remove it from this pull request.

Yes, please do.

mathstuf:
> If you have a better name for the tool

`reap`? Something zombie-related? `prune`?

src/ninja.cc
mathstuf: How about "that are no longer produced by the manifest"?

ikifof:
> > If you have a better name for the tool
> 
> `reap`? Something zombie-related? `prune`?

Updated to use reap. Thanks for the suggestion.

ikifof:
@jhasse I am done with the integration of the review comments. I messed up trying to rebase and following git recommendation when my command failed... so it created 77 noisy changes which I don't seem to be able to squash :(
Let me know if I need to do something specific before you can accept this pull request.

jhasse:
Thanks!

I really liked `cleandead` more, sorry :/ Since there's already a clean tool and it's established that it does delete files, having clean in the name is very important IMHO. Also as a non-native speaker I didn't know what "reap" meant before looking it up.

What do you two think?

ikifof:
> Thanks!
> 
> I really liked `cleandead` more, sorry :/ Since there's already a clean tool and it's established that it does delete files, having clean in the name is very important IMHO. Also as a non-native speaker I didn't know what "reap" meant before looking it up.
> 
> What do you two think?

It make sense to keep the clean prefix. So be it: reverted back to cleandead...

src/ninja.cc
jhasse: Can you explain a little bit? If it was a bug, it should be caught by a test.

ikifof: Well, if it had been covered, removing it would have made some test fail... Which is why I am asking the question: if there was a test, I could at least have understood the purpose of this.

If we check if s is a dead path, and we don't find it, or we find it, but it in not in edge, then why would it not be considered as dead? what is the logic? What is the corner case to which it applies? 
I could not find any reason for this code... and it is exactly the definition of "dead" I used in the Cleaner::CleanDead() implementation.

jhasse: Yeah makes sense.

Re-adding those lines won't make any of your test fail though, that's what I meant by "caught by a test".

ikifof: Actually, the right answer is that I should add a test for removing these 3 lines.
I don't have the time right now... I'll do it as soon as I can (unless you really disagree with removing these 3 lines).

Basically: if the code was right, if a file used to be built, but has both been removed from the manifest and from the build dir, but still has an entry in the .ninja_log, the command:
`    ninja -t recompact
should remove the entry from the log... With these 3 lines, it does not... Which means it leaves useless stuff in the .ninja_log which goes against the idea of speed.

Also: removing these 3 lines is kind of part of the feature here: the "recompact" tool considers "dead path" as candidate to be removed from the .ninja_log (even so, with these 3 lines, the answer IsPathDead answer is always false...)... but the "clean" tool does not remove the artifacts that are no more in the manifest... so the tool "cleandead" actually remove those, thus allowing the "recompact" tool (without those 3 lines) to do its job (remove entries that are now useless because duplicated or referencing dead paths).

Ideally, if I find some time I'd really like to have the cleandead tool also perform the recompact.

In any case: let me know how you would like to proceed.

jhasse: Thanks for the explanation. I still don't understand it 100%, that's why a test would be good.

I would also merge this PR as it is without the removal of those 3 lines and handle the change in a new PR.

jhasse: > These lines were added in the following commit, but the change seems unrelated... That's why I think it was added by mistake.

Actually they weren't added but moved from:

```cpp
    return (!n || !n->in_edge()) && disk_interface_.Stat(s.AsString()) == 0;
```

But I think it's a logic error. The equivalent code would be:

```cpp
    if (n && n->in_edge())
      return false;
```

Do you agree? If so, I would fix this in a commit on master.

@nico Can you also have a look? This was the change: https://github.com/ninja-build/ninja/commit/3beebde51a2089ecb01820f1428efe0263deaeea#diff-78294872cbf9d32f4f972288561fa718R146

jhasse: I've fixed the conditional in https://github.com/ninja-build/ninja/commit/1f02bc192a31ec2daac813c11d98dd939848d40f. When you're rebasing, just keep the three lines.

jhasse:
@ikifof If you rebase against master, I'll merge :)

jhasse:
I've fixed the conflict using GitHub's web editor :)

Thx again for the PR!

PR #1433, opened 2018-04-28
Remove 'using namespace std', properly namespace all std symbols

jonesmz:
I'm creating a fork of Ninja that focuses on keeping the source as close as possible with the Ninja code, but incorporates c++17-isms where ever I think it's practical.

So really it's a patch series, but github is an awful medium for doing that without making a fork and committing on top of it. It is what it is, I guess.

The first change is to properly namespace std::symbols.

I'm sure my motivation for doing that isn't really something that the Ninja maintainers care about, since it's one of those tabs vs spaces kinds of discussion that ultimately boils down to be mostly about personal preference, but suffice to say it's the first and by far the largest patch, and I'm not going to drop it from my patch series. *shrug*.

I hope that you'll be willing to merge it, as it'll drastically reduce the amount of work I have to do to continue incorporating changes from upstream into my patch series, but if you don't want to merge this pull request please don't feel that you should leave this pull request open. I'd rather rip the band-aid off quickly.


jonesmz:
Wrong branch. Sorry.

Issue #1411, opened 2018-04-06
MSVC 12 bot broken after high-res timestamp change


nico:
Appveyor builds don't show up on pull requests for some reason, so I haven't noticed this for a long time. First bad build: https://ci.appveyor.com/project/nico/ninja/build/1.0.652

```
deps_log.cc
.\src\deps_log.cc(138) : error C2065: 'uint32_t' : undeclared identifier
.\src\deps_log.cc(138) : error C2146: syntax error : missing ';' before identifier 'mtime_part'
.\src\deps_log.cc(138) : error C2065: 'mtime_part' : undeclared identifier
.\src\deps_log.cc(138) : error C2061: syntax error : identifier 'uint32_t'
.\src\deps_log.cc(139) : error C2065: 'mtime_part' : undeclared identifier
.\src\deps_log.cc(141) : error C2065: 'mtime_part' : undeclared identifier
.\src\deps_log.cc(141) : error C2061: syntax error : identifier 'uint32_t'
.\src\deps_log.cc(142) : error C2065: 'mtime_part' : undeclared identifier
```

Just using unsigned instead of uint32_t is probably an ok fix? @QuLogic

QuLogic:
`typedef unsigned __int32 uint32_t;` on old MSVC might also be a possibility.

PR #1435, opened 2018-05-01
Fix older VS compatibility issues and PDB files generation issue.

ikifof:
This fixes several windows builds related issue:
- the issue #1411 (with visual studio 2013 and earlier);
- another compilation issue with VS2012 (strtoll not defined)
- the PDB files related warnings when building ninja/ninja_test by providing a unique name per library/binary for the PDB file.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -414,7 +414,7 @@ n.newline()
 
 if platform.is_msvc():
     n.rule('cxx',
-        command='$cxx $cflags -c $in /Fo$out',
+        command='$cxx $cflags -c $in /Fo$out /Fd' + built('$pdb'),
         description='CXX $out',
         deps='msvc'  # /showIncludes is included in $cflags.
     )
@@ -485,6 +485,9 @@ else:
 n.newline()
 
 n.comment('Core source files all build into ninja library.')
+cxxvariables = []
+if platform.is_windows():
+    cxxvariables = [('pdb', 'ninja.pdb')]
 for name in ['build',
              'build_log',
              'clean',
@@ -505,15 +508,15 @@ for name in ['build',
              'string_piece_util',
              'util',
              'version']:
-    objs += cxx(name)
+    objs += cxx(name, variables=cxxvariables) 
 if platform.is_windows():
     for name in ['subprocess-win32',
                  'includes_normalize-win32',
                  'msvc_helper-win32',
                  'msvc_helper_main-win32']:
-        objs += cxx(name)
+        objs += cxx(name, variables=cxxvariables)
     if platform.is_msvc():
-        objs += cxx('minidump-win32')
+        objs += cxx('minidump-win32', variables=cxxvariables)
     objs += cc('getopt')
 else:
     objs += cxx('subprocess-posix')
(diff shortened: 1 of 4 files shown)


src/deps_log.cc
```diff
@@ -20,6 +20,9 @@
 #include <string.h>
 #ifndef _WIN32
 #include <unistd.h>
+#elif (_MSC_VER < 1900)
```
jhasse: You'd need the `defined(_MSC_VER)` here, too, as MinGW defines `_WIN32`, but not `_MSC_VER`.

src/build_log.cc
```diff
@@ -35,6 +35,9 @@
 #include "graph.h"
 #include "metrics.h"
 #include "util.h"
+#if defined(_MSC_VER) && (_MSC_VER < 1800)
+#define strtoll _strtoi64
```
jhasse: Why not a `typedef`?

jhasse: Ah because it's a function, missed that, sorry.

src/deps_log.cc
ikifof: My understanding is that MinGW needs these defines (but it does not need the strtoll) . 
Could somebody confirm?

configure.py
```diff
@@ -480,6 +480,9 @@ def has_re2c():
 n.newline()
 
 n.comment('Core source files all build into ninja library.')
+cxxvariables = []
+if platform.is_windows():
```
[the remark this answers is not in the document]
ikifof: Based on the rest of the code: correct. Sorry about that. It should be:
```suggestion
if platform.is_msvc():
```

configure.py
```diff
@@ -583,7 +588,9 @@ def has_re2c():
              'hash_collision_bench',
              'manifest_parser_perftest',
              'clparser_perftest']:
-  objs = cxx(name)
+  if platform.is_windows():
```
ikifof: This should also be:
```suggestion
  if platform.is_msvc():
```

src/deps_log.cc
jhasse: MinGW doesn't need them, otherwise Ninja wouldn't have compiled with it all the time ;)

I've added a MinGW build on AppVeyor, can you rebase your PR on the master branch?

ikifof: Make sense. Done.

jhasse:
Looks like you merge the changes? Can you do

```
git rebase origin/master
git push ikifof issue1411_MSVC12_broken --force
```

?

(I'm assuming that origin is https://github.com/ninja-build/ninja.git and ikifof is your fork)

ikifof:
Did I do something wrong? I hope not... Git has some cool features... but way too uselessly complex... :(

I just ran the 2 commands you gave. I think I answered all your review comments but somehow the CR summary still shows an open request... Let me know if I need to do something.

Note that the MinGW build has some bad warnings and fails the disk_interface_test that was disabled for msvc only. It probably should be disabled on windows.

jhasse:
What's your output of `git remote -v`?

> the disk_interface_test that was disabled for msvc only. It probably should be disabled on windows.

Will do that tomorrow.

ikifof:
origin  https://github.com/ikifof/ninja.git (fetch)
origin  https://github.com/ikifof/ninja.git (push)
upstream        https://github.com/ninja-build/ninja.git (fetch)
upstream        https://github.com/ninja-build/ninja.git (push)

jhasse:
Ah! Run these then please:

```
git rebase upstream/master
git push origin issue1411_MSVC12_broken --force
```

ikifof:
Let me know if I need to do anything else. I can create a new clean pull request if I made things complicated by messing up this one.

jhasse:
No it's okay, I squashed the commits using GitHub. Thank you!

ikifof:
Thanks!

PR #1438, opened 2018-05-12
Alternative implementation of a persistent build server mode.

jdarpinian:
I like the idea of having a persistent build server mode from #1389. It completely eliminates parsing time at startup and doesn't require a new on-disk file format. But #1389 requires the user to change how they invoke ninja, and requires manually restarting the build server if .ninja files change or if you want different build arguments.

This PR implements a persistent build server mode that works the same way as ninja's current non-persistent mode from a user perspective. All you have to do is add -p to your build command line and builds get faster. The server is started automatically and restarted if necessary so it's never out of date.

Using -p, a no-op build of Chromium for Android drops from 4.4 seconds to 1.0 seconds on my Linux workstation.


src/ipc-win32.cc
```diff
@@ -0,0 +1,251 @@
...
+    GetWin32EscapedString(argv[i], &escaped);
+    args += escaped;
+    args += ' ';
+  }
+  // Start the process and wait for it to create the pipe before continuing.
+  STARTUPINFO si = {0};
+  PROCESS_INFORMATION pi = {0};
+  HANDLE pipe_created_event =
+      CreateEvent(NULL, TRUE, FALSE, GetEventName().c_str());
+  if (!CreateProcess(NULL, (char *)args.c_str(), NULL, NULL, FALSE, 0, NULL,
```
jhasse: Although we are still using `CreateProcessA`, `CreateProcessW` might actually modify the second argument. Microsoft might also change the behavior. I would copy the args into a buffer instead of the `const_cast` here. (or at least use `const_cast` instead)

src/manifest_parser.cc
```diff
@@ -445,3 +453,10 @@ bool ManifestParser::ExpectToken(Lexer::Token expected, string* err) {
   }
   return true;
 }
+
+bool ManifestParser::OutOfDate() {
+  for (size_t i = 0; i < paths_.size(); i++)
```
jhasse: Please add some {} for readability here.

src/manifest_parser.h
```diff
@@ -58,6 +60,8 @@ struct ManifestParser {
     return Parse("input", input, err);
   }
 
+  bool OutOfDate();
```
jhasse: Please add a short doc for this method.

src/ninja.cc
```diff
@@ -1244,23 +1259,41 @@ NORETURN void real_main(int argc, char** argv) {
...
+      int result = ninja.RunBuild(argc, argv);
+      if (g_metrics)
+        ninja.DumpMetrics();
+      if (options.persistent) {
+        SendBuildResult(result);
+        // Loop around and wait for the next build request without reparsing.
+        cycle = 1;
+        ninja.state_.Reset();
+        continue;
+      } else {
```
jhasse: No need for `else` after `continue`.

src/disk_interface.cc
```diff
@@ -284,3 +284,9 @@ void RealDiskInterface::AllowStatCache(bool allow) {
     cache_.clear();
 #endif
 }
+
+void RealDiskInterface::ClearStatCache() {
+#ifdef _WIN32
```
jhasse: Why?

src/ipc-win32.cc
jdarpinian: Done. Hopefully CreateProcessW will never be required now that Microsoft is moving toward supporting UTF-8 as the system locale code page.

src/disk_interface.cc
jdarpinian: It seems like the cached timestamps would become stale if other processes touched the files after the server was started. I added this to invalidate the cache it at the start of each build.

jdarpinian:
Thanks for looking! I've reformatted the files with clang-format, but it didn't move the `&` and `*` by default so I changed the .clang-format file to do that.

jhasse:
That's weird, it does for me (without those two lines in .clang-format). What version are you using? I'm at 7.0.1.

jdarpinian:
I'm using nightly builds, but I think these options are needed for any version as far back as 3.4. It detects the most common pointer position from the file, so if more than 50% of the pointers are the wrong way (as in the new files I wrote) it will put them all the wrong way. If you test it by just changing one pointer in a file that already has many that are correct, it will appear to work.

src/disk_interface.cc
jhasse: Why is that the case only on Windows?

jhasse:
Ah I see! Thanks for the explanation, I didn't know clang-format looks at the way existing code is formatted.

src/disk_interface.cc
jdarpinian: The cache_ member variable only exists on Windows. All other functions dealing with the cache have #ifdef _WIN32 as well, such as the previous function in this file.

jhasse: I see. Can you wrap ClearStatCache in `#ifdef _WIN32` including the call instead of the function body?

jdarpinian: Done.

src/disk_interface.h
```diff
@@ -84,7 +84,7 @@ struct RealDiskInterface : public DiskInterface {
   /// Whether stat information can be cached.  Only has an effect on Windows.
   void AllowStatCache(bool allow);
 
-  /// Removes any cached stat information.  Only has an effect on Windows.
+  /// Removes any cached stat information. Only defined on Windows.
```
jhasse: Also wrap this declaration please - having a linker error when something tries to use it, would be weird. You can drop the note from the comment then.

jdarpinian: Done.

tmandry:
Hello, it looks like this is not being worked on anymore.

@jhasse do you see any obstacles to merging this if conflicts were resolved?

jhasse:
@tmandry There are lots of other features on the 1.11.0 and 1.12.0 milestones (see https://github.com/ninja-build/ninja/milestones) which I consider more important or am personally more interested in. Therefore I wouldn't have the time to review this any time soon, sorry.

jhasse:
closing in favor of #2030.

PR #1439, opened 2018-05-18
browse.py: fix delay with multiple connections

Lekensteyn:
When Firefox opens two TCP connections and sends a request over the
second connection, it is not processed until the first one is closed.
Use a threaded server to solve this issue. Tested with Python 2.7.15 and
3.6.5. (Also tested with 2.6, but that failed due to lack of argparse.)

Link: https://bugs.python.org/issue31639

### DIFF
--- a/src/browse.py
+++ b/src/browse.py
@@ -24,8 +24,10 @@ from __future__ import print_function
 
 try:
     import http.server as httpserver
+    import socketserver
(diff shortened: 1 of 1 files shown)


Lekensteyn:
Rebased on master (no further changes).

nico:
Thanks!

PR #1449, opened 2018-07-10
Do not add two to the default job count

Hello71:
Benchmarks show that increasing the number of make jobs past the number of CPUs decreases performance: https://blogs.gentoo.org/ago/2013/01/14/makeopts-jcore-1-is-not-the-best-optimization/. This was conducted 5 years ago. Since then, although context switching itself has become faster, both CPU cache size and compiler memory usage have increased (the latter particularly so for C++). Moreover, compilers spend even less time on I/O now than they did then; any reasonable software package's header files will all fit into main memory. Therefore, I believe the results should be even more significant today.

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -225,10 +225,8 @@ int GuessParallelism() {
   case 0:
   case 1:
     return 2;
-  case 2:
-    return 3;
   default:
-    return processors + 2;
+    return processors;
   }
 }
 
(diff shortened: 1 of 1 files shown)


Hello71:
Upon reconsideration, I think the value should be 1 for single CPU machines as well. Such machines are likely to also have low RAM, and so running more jobs than necessary is not only slower on the CPU, it may cause swapping.

jhasse:
In order to reduce the number of open PRs I'm closing this one. If you still think this should be changed, please open an issue or start a thread on the mailing list.

ocroquette:
See also: https://groups.google.com/forum/#!topic/ninja-build/E3s9gE7RiLQ
That would at least mitigate the issue when the default value is not optimal.

PR #1454, opened 2018-07-30
Support ANSI escape color codes in NINJA_STATUS

### LINKED ISSUE #713: Support color in NINJA_STATUS

mqudsi:
Adds support for ANSI escape sequences used for colorization to the job
status indicator, and updates `ElideMiddle()` to account for the
non-obvious change in text width.

As an example, the following will now emit the job status in green:

   `env NINJA_STATUS="\\033[32m[%f/%t] \\033[0m" ninja;`

Invalid color escape codes are treated as plain text (as before). ANSI
escape sequences apart from `\033...m` are not supported.

Closes #713. Closes #912.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -203,7 +203,21 @@ string BuildStatus::FormatProgressStatus(
   char buf[32];
   int percent;
   for (const char* s = progress_status_format; *s != '\0'; ++s) {
-    if (*s == '%') {
+    // Support ANSI color escape codes in NINJA_STATUS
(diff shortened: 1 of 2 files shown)


jhasse:
btw: ANSI escape color codes in `NINJA_STATUS` should already work using the following:

```sh
env $'NINJA_STATUS=\x1b[32m[%f/%t] \x1b[0m' ninja
```

It's true that this will break `ElideMiddle` though.

jhasse:
Superseded by  #1494.

PR #1455, opened 2018-07-31
Emit color job progress and job failure messages by default

mqudsi:
Set the default `NINJA_STATUS` and the "FAILED" text on job failure to
use ANSI color code escapes if running under a smart terminal.

Requires #1454 to be merged first.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -203,7 +203,21 @@ string BuildStatus::FormatProgressStatus(
   char buf[32];
   int percent;
   for (const char* s = progress_status_format; *s != '\0'; ++s) {
-    if (*s == '%') {
+    // Support ANSI color escape codes in NINJA_STATUS
+    if (strncmp(s, "\\033[", 5) == 0) {
+        const char *end = strchr(s + 5, 'm');
+        if (end == NULL) {
+            // Not a valid ANSI color sequence, treat as regular text
+        } else {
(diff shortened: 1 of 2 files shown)


nico:
This is inherently subjective. We think the current default is a good default; it looks like you'll be able to set the env var to include color codes to get this locally soon.

jhasse:
Eliding with ANSI escapes is slower (if done correctly, see WIP about this in #1494), that's why we're reluctant to change the default. You also can't use `smart_terminal` as that is still true for Windows <10, which doesn't support ANSI color escapes.

Coloring the "FAILED:" message does sound like a good idea to me though.

mqudsi:
Thanks, @jhasse

mqudsi:
@jhasse Thank you for providing legitimate reasons for not taking this. Do you want me to open a separate PR for the red failure messages or rebase this one for you guys to reopen? Also, what do you recommend in lieu of `smart_terminal`?

jhasse:
> Do you want me to open a separate PR for the red failure messages or rebase this one for you guys to reopen?

Open a new one.

> Also, what do you recommend in lieu of smart_terminal?

There's `supports_color_` now, which needs to be set to `false` on Windows if `SetConsoleMode(console_, mode | ENABLE_VIRTUAL_TERMINAL_PROCESSING);` didn't work.

PR #1458, opened 2018-08-12
Added const to methods which 'ought to be const' based on Const Checker

eyolfson:
I double checked that everything compiled without issue. These methods were found to by our Const Checker research tool. Let me know if these methods should not be const. Thanks!


Hi-Angel:
A bit offtopic: is that Const Checker tool available online? I'd be interested to run it on a few FOSS projects I care about.

jhasse:
Thanks for the PR! It fails on Travis because `lexer.cc` is generated from `lexer.in.cc`. Please change it in both files and reopen.

PR #1268, opened 2017-04-12
Add -c flag to force color output.

altimin:

PR #1462, opened 2018-08-27
Enable ANSI escape sequences on Windows 10 terminals

### LINKED ISSUE #1451: Support ANSI colors on Windows
Clang's `cl.exe` has an option `-fansi-escape-codes` which makes it output ANSI colors instead of using the usual Windows console API. The console of Windows 10 in one of the updates gained the ability to show those colors correctly. This means that theoretically I could set that option and make Ninja correctly forward the colors (which can never work with the usual colors in Windows, because they are not part of the output stream).

However, it doesn't work for some reason, which I was not yet able to find out. Ninja will forward the colors with `pool = console`, but otherwise it won't. Furthermore, the the mode for automatic parsing of include dependencies via `deps = msvc` will probably also need to learn to ignore, but forward the ANSI escape codes.

GoaLitiuM:
Fixes #1451

### DIFF
--- a/src/line_printer.cc
+++ b/src/line_printer.cc
@@ -18,6 +18,9 @@
 #include <stdlib.h>
 #ifdef _WIN32
 #include <windows.h>
+#ifndef ENABLE_VIRTUAL_TERMINAL_PROCESSING
+#define ENABLE_VIRTUAL_TERMINAL_PROCESSING 0x4
+#endif
 #else
 #include <unistd.h>
 #include <sys/ioctl.h>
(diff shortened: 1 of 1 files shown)


src/line_printer.cc
```diff
@@ -40,6 +43,12 @@ LinePrinter::LinePrinter() : have_blank_line_(true), console_locked_(false) {
   console_ = GetStdHandle(STD_OUTPUT_HANDLE);
   CONSOLE_SCREEN_BUFFER_INFO csbi;
   smart_terminal_ = GetConsoleScreenBufferInfo(console_, &csbi);
+  
+  // Try enabling ANSI escape sequence support on Windows 10 terminals.
+  DWORD mode;
+  if (GetConsoleMode(console_, &mode)) {
```
jhasse: Do we even want to call this even when `smart_terminal_` is `false`?

GoaLitiuM: Ah, I missed that completely. Fixed and rebased.

Boddlnagg:
This is required for `CLICOLOR_FORCE` (https://github.com/ninja-build/ninja/pull/1268#issuecomment-438289348) to work well on Windows.

jhasse:
How so? Disabling escape code stripping should only have an effect when not running in a terminal anyway.

TheOneRing:
@jhasse KDE Craft calls cmake and ninja and logs its output, in order to have nice colors CLICOLOR_FORCE is needed. So ninja is running in a terminal but wrapped by another process.

nico:
Thanks, this is great. One thing to think about is that the ansi mode is I believe inherited by subprocesses, and this probably sets it for all subprocesses. Only subprocesses in the console pool have access to the console, and letting them inherit this is probably a good thing anyhow. I'm just mentioning this to make sure that this is a conscious decision :-)

jhasse:
Thanks again, for the two corrections.

PR #1472, opened 2018-09-28
Fix wrong description of script installation

Hi-Angel:
The described way of installation makes zsh fail with `_arguments:comparguments:325: can only be called from completion function`. Per [zsh documentation](https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org#telling-zsh-which-function-to-use-for-completing-a-command) the correct way is to use `$fpath`.


jhasse:
Thanks!

PR #1473, opened 2018-09-29
Query terminal width from STDOUT_FILENO

zdave:
Rather than 0 (stdin). So it will work even if ninja's stdin isn't connected to
the terminal for whatever reason.

### DIFF
--- a/src/line_printer.cc
+++ b/src/line_printer.cc
@@ -82,7 +82,7 @@ void LinePrinter::Print(string to_print, LineType type) {
     // Limit output to width of the terminal if provided so we don't cause
     // line-wrapping.
     winsize size;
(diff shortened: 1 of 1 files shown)


zdave:
I've just rebased this. Any objection? Using the stdout handle makes more sense as that's where the output actually goes to. I've had reason before to redirect Ninja's stdin to /dev/null, which currently doesn't work properly because of this.

jhasse:
`STDOUT_FILENO` is 1, correct?

zdave:
Yeah.

jhasse:
As we are using `isatty(1)`, too, this makes perfect sense to me. Thanks for the PR!

PR #1479, opened 2018-10-12
Expand @rspfile to rspfile_contents in reported commands

jon-turney:
Move EvaluateCommandWithRspfile() to class Edge
Use it in verbose logging of command
Use it in logging of failed command
Tweak documentation

### DIFF
--- a/src/graph.h
+++ b/src/graph.h
@@ -146,6 +146,15 @@ struct Edge {
   /// full contents of a response file (if applicable)
   string EvaluateCommand(bool incl_rsp_file = false);
 
+  /// Expand all variables in a command and return it as a string.
+  /// if ECM_EXPAND_RSPFILE is specified, the @rspfile will be
+  /// substituted with the contents of the response file
+  enum EvaluateCommandMode {
+  ECM_NORMAL,
+  ECM_EXPAND_RSPFILE
+  };
+  string EvaluateCommandWithRspfile(EvaluateCommandMode mode = ECM_EXPAND_RSPFILE);
+
   /// Returns the shell-escaped value of |key|.
   string GetBinding(const string& key);
   bool GetBindingBool(const string& key);
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -812,8 +812,9 @@ keys.
 
 `description`:: a short description of the command, used to pretty-print
   the command as it's running.  The `-v` flag controls whether to print
-  the full command or its description; if a command fails, the full command
-  line will always be printed before the command's output.
+  the full command (with rspfile_content expanded) or its description;
+  if a command fails, the full command line will always be printed before
+  the command's output.
 
 `generator`:: if present, specifies that this rule is used to
   re-invoke the generator program.  Files built using `generator`
--- a/src/build.cc
+++ b/src/build.cc
@@ -152,7 +152,7 @@ void BuildStatus::BuildEdgeFinished(Edge* edge,
       outputs += (*o)->path() + " ";
 
     printer_.PrintOnNewLine("FAILED: " + outputs + "\n");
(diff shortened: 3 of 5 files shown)


jhasse:
Thanks for the PR!

Why did you move `EvaluateCommandWithRspfile` into `Edge`?

jon-turney:
> Why did you move `EvaluateCommandWithRspfile` into `Edge`?

I guess my thinking was something like:

* Keeping it in ninja.cc, I would either have had to create a ninja.h, since one doesn't exist, or chosen some other header to put the prototype in.
* It's unclear it really belongs in ninja.cc, since that seems to mainly contain code related to main()
* Since it's just a wrapper around some other methods of class Edge, it seems to belong there.

jhasse:
Makes sense.

I'm not sure whether we want this by default or if it should only be activated by something like `-vv` or `-v -x`.

jon-turney:
> I'm not sure whether we want this by default or if it should only be activated by something like `-vv` or `-v -x`.

I added an `-x` option (by analogy with `ninja -t compdb -x`), to enable expansion of `@rspfile` to `rspfile_contents`.

There doesn't seem to be any documentation of options, so there is no documentation change.

Note that `ninja -x -t compdb`  doesn't work as expected (the same as `ninja -t compdb -x`), because tools currently only have access to `struct Options` and not to `struct BuildConfig`.  Not sure if or how to fix that.

jhasse:
> Note that ninja -x -t compdb doesn't work as expected (the same as ninja -t compdb -x), because tools currently only have access to struct Options and not to struct BuildConfig. Not sure if or how to fix that.

Oh, that's not good. In #1223 @asankah said that he tested it manually (and I'm assuming that it worked).

jon-turney:
> Oh, that's not good. In #1223 @asankah said that he tested it manually (and I'm assuming that it worked).

I think what I have written is unclear. `ninja -t compdb -x` works.

After this PR, it's now possible to write `ninja -x -t compdb`, but the -x does nothing.

jhasse:
btw: Do you have an idea why there's the need for `EvaluateCommandMode` and not simply a `bool expand_rsp`?

jon-turney:
> Do you have an idea why there's the need for `EvaluateCommandMode` and not simply a `bool expand_rsp`?

No idea, sorry.

nico:
IIRC we intentionally don't do this. If you need an rsp file, you won't be able to run the command if it's expanded like this. And if you don't need and rsp file, don't use one :-)

nico:
Aha, see also discussion in #1223 (already linked to above).

src/graph.h
```diff
@@ -146,6 +146,15 @@ struct Edge {
   /// full contents of a response file (if applicable)
   string EvaluateCommand(bool incl_rsp_file = false);
 
+  /// Expand all variables in a command and return it as a string.
+  /// if ECM_EXPAND_RSPFILE is specified, the @rspfile will be
+  /// substituted with the contents of the response file
+  enum EvaluateCommandMode {
+  ECM_NORMAL,
```
jhasse: This needs indention.

jon-turney:
> IIRC we intentionally don't do this. If you need an rsp file, you won't be able to run the command if it's expanded like this. And if you don't need and rsp file, don't use one :-)

Title: Expand @rspfile to rspfile_contents in **reported** commands 
Commentary: Expand @rspfile to rspfile_contents with -x

Move EvaluateCommandWithRspfile() to class Edge
Add -x option to control expanding rspfile content **in output**
Use it in **verbose logging** of command
Use it in **logging** of failed command

src/graph.h
jon-turney: Thanks. Fixed.

jonesmz:
@jon-turney Could you update this PR based on the latest master branch?

PR #1484, opened 2018-10-19
Change gyp to gn

smeenai:
Chrome has switched from gyp to gn, and Fuschia also uses it now.

Note that the "Chromium Ninja documentation" link is dead, but I'm not sure what to replace it with. The closest I've found is https://chromium.googlesource.com/experimental/chromium/src/+/refs/wip/bajones/webvr/docs/ninja_build.md, but I'm not sure that's the original intended target.

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -154,10 +154,10 @@ design is quite clever.
 Ninja's benefit comes from using it in conjunction with a smarter
 meta-build system.
 
-http://code.google.com/p/gyp/[gyp]:: The meta-build system used to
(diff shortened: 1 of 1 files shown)


mathstuf:
I think `gn` should be added. `gyp` still seems to exist [here](https://gyp.gsrc.io/). Rebasing would fix the CI problem too.

jhasse:
IMHO removing GYP is fine, as there's also https://github.com/ninja-build/ninja/wiki/List-of-generators-producing-ninja-build-files.

nico:
The "Chromium Ninja" link target no longer really exists, now that ninja is the only way to build Chromium. Maybe that part can just go away.

Issue #1477, opened 2018-10-09
Needs release


Twey:
There are a variety of bugs in ninja that have been fixed on master but still exist elsewhere because they have not been incorporated into a release.  For example, we're trying to use ninja internally at my company, but are suffering through https://github.com/ninja-build/ninja/pull/1219 ; we are using an internally-maintained patchset that includes the fix, but this is obviously suboptimal.

Is there a roadmap for a ninja release that distros can package?  The last (even patch) release was over a year ago.

nico:
Yes, let's try to do a release in the next month or two. Sorry about the radio silence.

Boddlnagg:
But https://github.com/ninja-build/ninja/issues/1418 should be investigated first ...

jhasse:
> Is there a roadmap for a ninja release that distros can package?

I've created https://github.com/ninja-build/ninja/milestone/1 as some kind of roadmap what needs to get fixed before the next release.

Twey:
Thank you so much, folks! It's a great relief to hear this — I was a little worried that ninja wasn't active any more. ☺

Twey:
It looks like the roadmap is all done! ^_^

Pwera:
+1

jonesmz:
@jhasse 

Could https://github.com/ninja-build/ninja/milestone/1  be updated, so contributors know which PRs need attention first?

jhasse:
As far as I'm concerned there's nothing left to be done for 1.9.0.

jonesmz:
@nico @evmar Any advice?

ignatenkobrain:
There's RELEASING file in repository. Please release new version. Thanks!

nico:
Yes, see RELEASING. Step 1 is to send something like the "Call for 1.8 testing: vfork spawn, larger path component, faster clparser, vs2017, uclibc, dragonfly" thread for 1.9 and wait a bit, and not merge any commits in the meantime.

jhasse:
Okay, I've posted to the mailing list :) https://groups.google.com/forum/#!topic/ninja-build/F5CWcXiv4TU

If I forgot any important change that needs more testing, please add a post there.

ignatenkobrain:
@jhasse seems that no one has any objections, could you make new release?

Thanks!

jhasse:
I don't have access to a Mac, so I'll fix #1509 first.

jonesmz:
It's been over 30 days.

[...]

PR #1490, opened 2018-11-09
Make -j 0 run unlimited parallel builds

### LINKED ISSUE #1309: Support unlimited number of parallel jobs #1379
Hey @nico,

I am looking to add a new mode, possibly a value passed to `-j` (or reading an environment variable to decide to enable this special mode or not) that runs as many parallel tasks as possible.
The easy way would be to just pass `-j 1000` or something silly but I would like to actually get the number of concurrent tasks possible, which depends on the project, it could be limited by one part of a build depending on something previous being built and linked by the same project

Looking through the code, `wanted_edges_` and `command_edges_ ` pop out in `src/build.cc`
Could i get your input on how I would achieve this mode?

I assume this part is not possible because ninja just does what the it is told by the script but is there a way to differentiate between compile and link jobs for this. 

Best,
Martell

martell:
Just ended up doing it myself in 5mins over a random lunch time in the end.
closes #1309

@nico @jhasse please review.

Would it be possible to get a release once this is merged also?
There hasn't been one in a long time and it is nearing 100 commits now.

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -206,7 +206,7 @@ void Usage(const BuildConfig& config) {
 "  -C DIR   change to DIR before doing anything else\n"
 "  -f FILE  specify input build file [default=build.ninja]\n"
 "\n"
-"  -j N     run N jobs in parallel [default=%d, derived from CPUs available]\n"
+"  -j N     run N jobs in parallel (0 means infinity) [default=%d, derived from CPUs available]\n"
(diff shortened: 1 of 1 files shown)


mathstuf:
The release is tracked in #1477. I think this is likely to miss the tag however.

jhasse:
Thanks for the PR!

Can you add `, fix #1309` at the end of your commit message so that the issue is references and automatically closed?

jhasse:
Nevermind ;)

PR #1491, opened 2018-11-09
util: don't add ellipses width when deciding if they're necessary

mathstuf:
If the string fits, just use it. If we need the ellipses, *then* we need
to compute the width based on that.

---
Picked out from #912.


jhasse:
Thanks! I think this went unnoticed since the unit tests for `ElideMiddle` weren't comprehensive enough. I've extended them in another commit so that they only pass with your fix.

PR #1494, opened 2018-11-13
Elision with escape seq

### LINKED ISSUE #713: Support color in NINJA_STATUS

mathstuf:
Fixes #713.

---
Cc: @jhasse

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -442,26 +442,57 @@ bool islatinalpha(int c) {
   return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z');
 }
 
+/// Return the end of a span in `in` starting at `from` and whether it is an
+/// ANSI escapse sequence or not.
+///
+/// The index returned creates an half-open interval with the input index over
+/// the span.
+static pair<size_t, bool> AnsiSpan(const string& in, size_t from) {
+  size_t pos = from;
+  bool in_seq = in[pos] == '\33';
+
+  if (in_seq) {
+    if (pos + 1 >= in.size()) {
+      // A lone escape character at the end of the string.
+      pos = in.size();
+    } else if (in[pos + 1] != '[') {
+      // A non-CSI sequence. Just ignore the escape character.
+      ++pos;
+    } else {
+      pos += 2;
+
+      // Skip everything up to and including the next [a-zA-Z].
+      while (pos < in.size() && !islatinalpha(in[pos])) {
+        ++pos;
+      }
+      // Also take the last character of the escape sequence.
+      ++pos;
+    }
+  } else {
+    // Find the next sequence.
+    pos = in.find('\33', pos);
+    if (pos == string::npos) {
+      pos = in.size();
+    }
+  }
+
+  return make_pair(pos, in_seq);
+}
+
 string StripAnsiEscapeCodes(const string& in) {
   string stripped;
   stripped.reserve(in.size());
 
-  for (size_t i = 0; i < in.size(); ++i) {
-    if (in[i] != '\33') {
-      // Not an escape code.
-      stripped.push_back(in[i]);
-      continue;
+  size_t pos = 0;
+  while (pos < in.size()) {
+    pair<size_t, bool> span = AnsiSpan(in, pos);
+    if (!span.second) {
+      stripped.append(in.begin() + pos, in.begin() + span.first);
     }
 
-    // Only strip CSIs for now.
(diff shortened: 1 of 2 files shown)


src/util.cc
```diff
@@ -442,26 +442,55 @@ bool islatinalpha(int c) {
...
   string stripped;
   stripped.reserve(in.size());
 
-  for (size_t i = 0; i < in.size(); ++i) {
-    if (in[i] != '\33') {
-      // Not an escape code.
-      stripped.push_back(in[i]);
-      continue;
+  size_t pos = 0;
+  for (;;) {
```
jhasse: `while (pos < in.size())`?

jhasse:
Fixes #713.

src/util.cc
mathstuf: Indeed. I wasn't sure how many `break` statements there would be when starting.

src/util.cc
```diff
@@ -442,26 +442,52 @@ bool islatinalpha(int c) {
   return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z');
 }
 
+static pair<size_t, bool> AnsiSpan(const string& in, size_t from) {
```
jhasse: Can you add a `///` comment to this function, so that it's clear what it does?

src/util_test.cc
```diff
@@ -428,3 +428,13 @@ TEST(ElideMiddle, ElideInTheMiddle) {
...
 }
+
+TEST(ElideMiddle, NothingToElideMiddleWithEscape) {
+  string input = "Nothing \33[m string.";
+  EXPECT_EQ("Nothing \33[m string.", ElideMiddle(input, 80));
+}
+
+TEST(ElideMiddle, ElideMiddleWithEscape) {
+  string input = "0\33[m1234567890123456789";
+  EXPECT_EQ("0\33[m123...\33[m789", ElideMiddle(input, 10));
```
jhasse: What happens when the ANSI escape is inside the elided part?

src/util_test.cc
```diff
@@ -428,3 +428,13 @@ TEST(ElideMiddle, ElideInTheMiddle) {
   EXPECT_EQ("012...789", elided);
   EXPECT_EQ("01234567...23456789", ElideMiddle(input, 19));
 }
+
+TEST(ElideMiddle, NothingToElideMiddleWithEscape) {
+  string input = "Nothing \33[m string.";
+  EXPECT_EQ("Nothing \33[m string.", ElideMiddle(input, 80));
```
jhasse: 80 is enough for an non-ANSI-aware `ElideMiddle`. Please check that it correctly starts to elide at the right time ans move this nothing-to-elide check into the `ElideMiddleWidthEscape` test.

src/util_test.cc
mathstuf: All ANSI escapes are always output.

mathstuf: Hence the duplicate `\33[m` part on the second half.

mathstuf:
I'm seeing some issues with this when using it on projects locally. Will debug and figure out what's missing and add a testcase for it.

src/util_test.cc
jhasse: Hm ... why duplicate it?

mathstuf: It grabs the escape sequences from the elided part. I haven't dug into the issue I'm seeing here yet, but fixing it will probably make doing it better easier.

jhasse:
What do you think about #1600?

mathstuf:
Sure, that offers another frontend to be able to do what I want, but the thing is that today elision logic in ninja itself is broken with color (though this is also related to #1584). If #1600 is accepted, should color just be outright rejected from `NINJA_STATUS`? Or is it "basic" enough that having the support is worthwhile to to require some other frontend for color?

jhasse:
Good questions. We probably should also fix this in Ninja itself.

Btw: Did you have a look at my take on ANSI-aware elide_middle (in Python though)? https://github.com/jhasse/ja/blob/da264576a85d4d8b1de4517487bfe501b6337915/ja/native.py#L258

mathstuf:
Yes, I had looked at it. I started the code from my old `%0`-based algorithm though. I should probably just try and redo this code from scratch (using that as a template) rather than trying to adapt the older logic which wasn't span-based.

jhasse:
Or have consider using my frontend, comes with other awesome features, too ;)

jhasse:
I had another go at it in #2428.

PR #1495, opened 2018-11-14
Explicit python2

mathstuf:
---
Cc: @jhasse

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -30,14 +30,14 @@ Visual Studio.
 
 ##### Using Visual Studio
 Assuming that you now have python installed, then the steps for building under
- Windows using Visual Studio are:
- 
+Windows using Visual Studio are:
+
 Clone and checkout the latest release (or whatever branch you want). You
 can do this in either a command prompt or by opening a git bash prompt:
 
 ```
-    $ git clone git://github.com/ninja-build/ninja.git && cd ninja 
-    $ git checkout release 
+    $ git clone git://github.com/ninja-build/ninja.git && cd ninja
+    $ git checkout release
 ```
 
 Then:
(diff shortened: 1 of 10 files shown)


jhasse:
The scripts should rather be made compatible with Python 3 (I think configure.py and browse.py are). Did you run into any problems?

mathstuf:
> The scripts should rather be made compatible with Python 3

I agree, but this is what I have time for right now :) .

> Did you run into any problems?

A build was trying to run `python configure.py` after updating the repo which fails because I don't have a `python` executable anymore (Fedora 29).

> No need to CC me as I'm watching this repository.

Noted. Not everyone does it for their projects (personally, I tend not to due to email overload and instead rely on notifications).

jhasse:
> A build was trying to run python configure.py after updating the repo which fails because I don't have a python executable anymore (Fedora 29).

Running `./configure.py` is a better fix for that.

> Noted. Not everyone does it for their projects (personally, I tend not to due to email overload and instead rely on notifications).

Well, if I wouldn't like the email overload, I would sure start to disable mentions, too, after people start mentioning me in their PRs ;)

jhasse:
> Running ./configure.py is a better fix for that.

Nvm, this doesn't help either. I wonder why Fedora doesn't symlink `python` to `python3`?

mathstuf:
> I wonder why Fedora doesn't symlink python to python3?

Upstream recommends against it. Python2 scripts have been doing `/usr/bin/env python` forever and it's easier to detect their breakage by a bad interpreter than a wrong one. However, it looks like it is [slated for Fedora 32](https://fedoraproject.org/wiki/FinalizingFedoraSwitchtoPython3#Phase_3:_Switch_python_to_refer_to_python3) after the removal forces scripts to choose a version number.

Following links, it looks like [python-unversioned](https://fedoraproject.org/wiki/Changes/Move_usr_bin_python_into_separate_package) provides the symlink, but is not installed by default.

jhasse:
Okay, I would still recommend against hard-coding python2 (if anything we should use python3 to stay future-proof). Until Fedora 32 `sudo dnf install python-unversioned-command` is a good workaround.

The first two commits of your PR look useful though, can you create a new PR with those two?

PR #1498, opened 2018-11-15
export to CodeBlocks

dimkaip:
Ninja is an excellent tool and congratulations on the constructions.
We are all committed to private interest programs such as Microsoft's, which are closed-source.
But the codeblocks, which is open source and supports many types of projects, has happily come out on the market.
I personally have a serious problem with an application I'm making because I do not use visual studio but CodeBlocks.
So I want to ask if you can make an upgrade so that Ninja makes a project not only for Visual Studio but also for CodeBlocks.
We must respect and adopt the philosophy of open source software and tools.
I think we should not be a helper in a company that is constantly changing politics and its relationship with all of us is profit.
Thank you very much and I wish to listen.


jhasse:
Ninja is not a generator and can't create Visual Studio projects either.

PR #1501, opened 2018-11-16
NINJA_STATUS_SLEEP: Avoid too frequent status updates

moroten:
A build log on a dumb terminal can be big, even for a fast build. Reduce
the output on dumb terminal by printing less frequent if
NINJA_STATUS_SLEEP is set.

This is especially useful for build automation servers where the
terminal usually is dumb. You will be able to see progress, but do not
have to overflow the console output with status lines.

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -191,8 +191,10 @@ you don't need to pass `-j`.)
 Environment variables
 ~~~~~~~~~~~~~~~~~~~~~
 
-Ninja supports one environment variable to control its behavior:
-`NINJA_STATUS`, the progress status printed before the rule being run.
+Ninja supports two environment variables to control its behavior:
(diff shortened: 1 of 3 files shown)


nico:
I thought about doing this a few times. The way I imagined this to work:
- Don't have an option for this (http://neugierig.org/software/blog/2018/07/options.html). Do this by default in smart terminal mode, there's no need to update the status more than a few times per second. Don't do this for verbose or non-smart-terminal mode.
- To make this not cause stale output for when something happens during the "no update" window, also call seititimer or something to make sure we can update the status at the end of the dead window if necessary. (This might be a v2 feature depending on how much code it is, though.)

kbrowder:
Any chance of this getting merged? Seems nifty

jhasse:
-> should be done as part of (or after) #2516

PR #1508, opened 2018-12-05
Emit "FAILED: " in red if terminal supports ANSI color output

mqudsi:
Per [@jhasse's request](https://github.com/ninja-build/ninja/pull/1455#issuecomment-444128323)

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -138,7 +138,10 @@ void BuildStatus::BuildEdgeFinished(Edge* edge,
          o != edge->outputs_.end(); ++o)
       outputs += (*o)->path() + " ";
 
-    printer_.PrintOnNewLine("FAILED: " + outputs + "\n");
+    if (printer_.supports_color())
+        printer_.PrintOnNewLine("\x1B[31m" "FAILED: " "\x1B[0m" + outputs + "\n");
+    else
+        printer_.PrintOnNewLine("FAILED: " + outputs + "\n");
     printer_.PrintOnNewLine(edge->EvaluateCommand() + "\n");
   }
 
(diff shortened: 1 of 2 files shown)


mqudsi:
Thanks, done.

jhasse:
Thanks! Have you tested that this works on Windows? Should display a colored message on Windows 10 and the normal message on 7 and 8.

mqudsi:
I tested it under Windows 10 and confirmed it worked there. I don't have a Windows 7/8 machine set up with a development toolchain, but I can probably just copy the binary and try it there directly.

mqudsi:
@jhasse I've tested it on Windows 7 and can confirm that the latest build correctly avoids emitting the control codes under conhost without color support.

image

_Shown: without and with the latest commit to avoid color output if not supported_

src/build.cc
```diff
@@ -138,7 +138,10 @@ void BuildStatus::BuildEdgeFinished(Edge* edge,
          o != edge->outputs_.end(); ++o)
       outputs += (*o)->path() + " ";
 
-    printer_.PrintOnNewLine("FAILED: " + outputs + "\n");
+    if (printer_.supports_color())
+        printer_.PrintOnNewLine("\x1B[31m" "FAILED: " "\x1B[0m" + outputs + "\n");
```
shahzadlone: Perhaps a braced ```if``` might be better please.

src/build.cc
```diff
@@ -138,7 +138,10 @@ void BuildStatus::BuildEdgeFinished(Edge* edge,
...
+    else
+        printer_.PrintOnNewLine("FAILED: " + outputs + "\n");
```
shahzadlone: Perhaps a braced ```else``` might be better please.

src/build.cc
mqudsi: @shahzadlone My preference is for braced `if`/`else`, but I am sticking to the style of nearby code.

src/build.cc
mqudsi: @shahzadlone My preference is for braced `if`/`else`, but I am sticking to the style of nearby code.

Issue #1514, opened 2019-01-09
Windows: fopen cannot create file if its full path is longer than 254 chars


volth:
Chromium build started from a directory which is ~30 chars long fails with
```
[3572/20749] STAMP obj/services/network/public/mojom/mojom_ip_address_shared_cpp_sources__jumbo_merge.stamp
[3573/20749] STAMP obj/services/metrics/public/mojom/mojom_blink__jumbo_merge.stamp
ninja: error: WriteFile(__services_network_public_mojom_mutable_network_traffic_annotation_interface_shared_cpp_sources__jumbo_merge___build_toolchain_win_win_clang_x64__rule.rsp): Unable to create file. No such file or directory
```

The filename "__services_network_public_mojom_mutable_network_traffic_annotation_interface_shared_cpp_sources__jumbo_merge___build_toolchain_win_win_clang_x64__rule.rsp" is 155 chars long + paths inside Chromium source tree + prefix caused by superviser package manager (nixpkgs) result in full path being about 260 chars long

Alhough ```fopen()```'s argument has length of only 155 chars, it hits the limit set for the full path.

I propose to use raw WinAPI instead of ```fopen()``` and add ```\\?\``` prefix to the argument to allow paths up to 32767 chars long.

atetubou:
I think it is better to file a bug to crbug too so that jumbo build generates shorter file name.

volth:
> It's possible to lift the MAX_PATH limitation via a manifest setting, see https://docs.microsoft.com/en-us/windows/desktop/FileIO/naming-a-file#paths. I'm not sure if this will apply to fopen though

That seems to require Windows 10 and affects only Unicode API (CreateFileW but not CreateFileA)

jhasse:
We're using neither `CreateFileW` nor `CreateFileA` but `fopen`. I wonder if the setting will affect it. If not, it might affect `_wfopen` which is needed for #1195 anyway.

volth:
If we will fix it by adding the manifest, what about the next person who step into this bug?
Will he be advised to upgrade Windows to Windows 10?

jhasse:
I doubt that this would be an option for him, otherwise he would have done it already.

volth:
Well, even if ```ninja``` managed to create the file with long path (I fixed it with https://github.com/ninja-build/ninja/compare/master...volth:patch-1), `it does not fix Chromium jumbo build: ```python``` fails to open the file :(

```
Traceback (most recent call last):
[...]

PR #1515, opened 2019-01-10
Native Windows version of RealDiskInterface::WriteFile

volth:
The idea is to support creation of files longer than 254 characters (and also to allow creation files with names like "nul.txt", "com1" which also require ```\\?\```-prefix for not confusing with DOS devices)

Fixes https://github.com/ninja-build/ninja/issues/1514


src/disk_interface.cc
```diff
@@ -221,7 +221,72 @@ TimeStamp RealDiskInterface::Stat(const string& path, string* err) const {
 #endif
 }
 
+#ifdef _WIN32
+bool UTF8ToWide(wstring& wide, const string& utf8) {
```
atetubou: Can I ask you to add test for this function?

src/disk_interface.cc
```diff
@@ -221,7 +221,72 @@ TimeStamp RealDiskInterface::Stat(const string& path, string* err) const {
...
+    int dw = MultiByteToWideChar(CP_UTF8, 0, utf8.c_str(), utf8.size(), wdata.data(), wdata.size());
+    if (dw == 0) {
+      return false;
+    }
+    wide = wstring(wdata.data(), dw);
+    return true;
+  }
+}
+
+static bool fullPathName(wstring& result, const wstring& wpath) {
```
atetubou: Can I ask you to add test for this function?

src/disk_interface.cc
```diff
@@ -221,7 +221,72 @@ TimeStamp RealDiskInterface::Stat(const string& path, string* err) const {
...
+  vector<wchar_t> buf(100);
+  DWORD dw = GetFullPathNameW(wpath.c_str(), buf.size(), buf.data(), NULL);
```
atetubou: As far as I know GetFullPathName is slow API, can I ask you to show this change does affect performance of ninja execution?

volth: I hope, as for an addition to slow ```CreateFileW```+```WriteFile``` the overhead should be neglible.
But if we prefix all the filesystem API (including ```StatFile``` and others), then there might be some space for optimization: caching results, checking for the cases when ```GetFullPathName``` can be safely skipped, etc.

src/disk_interface_test.cc
```diff
@@ -188,6 +188,19 @@ TEST_F(DiskInterfaceTest, ReadFile) {
   EXPECT_EQ("", err);
 }
 
+TEST_F(DiskInterfaceTest, WriteFile) {
```
volth: This test is not passed on Windows (but passed on Linux) without the fix.
Long file name + some characters for the path overflow MAX_PATH

src/disk_interface.cc
volth: ```GetFullPathName``` is to be replaced with a hand-written function anyway: it puts names like ```"nul"``` to "\\.\"-namespace instead of "\\?\"

volth:
```CanonicalizePathW``` is mostly copy-paste from ```CanonicalizePath``` (only nested ```#ifdef _WIN32``` are removed and delimiter is set to '\' instead of '/').
Perhaps, C++ ```template<>``` over ```char``` and ```wchar_t``` should be used there to reduce number of code lines, but I am not sure about portability (especiall about using ```std::basic_string<T>```)

File: src/util.cc [hunk 5]
```diff
@@ -313,12 +411,83 @@ void GetWin32EscapedString(const string& input, string* result) {
...
+  return true;
+}
+
+static bool IsPathAbsoluteW(const wstring& wpath) {
+  return wpath.length() >= 2 && ( (wpath[0] == '\\' && wpath[1] == '\\') ||
+    (('a' <= (wpath[0]|0x20) && (wpath[0]|0x20) <= 'z' && wpath[1] == ':')) );
+}
+
+bool FullPathName(wstring* fullpath, const string& path, string* err) {
+  if (!UTF8ToWide(fullpath, path)) {
```
atetubou: can we do this just before L472 and use non-wide function for path manipulation?

volth: It is possible, but GetCurrentDirectory must be wide.
We cannot use result of GetCurrentDirectoryA, because it is not UTF8.
So using non-wide function for path manipulation means we have to
convert wide argument and also convert the result back ro wide.
This way we anyway get the wide function for path manipulation, only by wrapping non-wide one with the converters

src/disk_interface.cc
```diff
@@ -222,6 +222,32 @@ TimeStamp RealDiskInterface::Stat(const string& path, string* err) const {
...
+#ifdef _WIN32
+  wstring fullpath;
+  string err;
+
+  if (!FullPathName(&fullpath, path, &err)) {
+    Error("WriteFile(%s): %s", path.c_str(), err.c_str());
+    return false;
+  }
+
+  HANDLE handle = CreateFileW(fullpath.c_str(), GENERIC_WRITE, 0, NULL, CREATE_ALWAYS, FILE_ATTRIBUTE_NORMAL, NULL);
```
atetubou: Can I ask you to use newline so that this line (and some other places) does not exceed 80 chars?

src/util.cc
```diff
@@ -313,12 +314,112 @@ void GetWin32EscapedString(const string& input, string* result) {
...
+}
+
+// can it be memoized?
+static bool GetCurrentDirectoryUTF8(string* result) {
+  wstring wresult;
+  return GetCurrentDirectoryWide(&wresult) && WideToUTF8(result, wresult);
+}
+
+static bool IsPathAbsolute(const string& path) {
+  return path.length() >= 2 && ( (path[0] == '\\' && path[1] == '\\') ||
```
atetubou:
```suggestion
  return path.length() >= 2 && ((path[0] == '\\' && path[1] == '\\') ||
```

src/util.cc
```diff
@@ -313,12 +314,112 @@ void GetWin32EscapedString(const string& input, string* result) {
...
+    (('a' <= (path[0]|0x20) && (path[0]|0x20) <= 'z' && path[1] == ':')) );
```
atetubou:
```suggestion
    (('a' <= (path[0]|0x20) && (path[0]|0x20) <= 'z' && path[1] == ':')));
```

volth:
I think it is good enough for the first step.
The second should be replacing remaining A-functions with W-functions (in ```MakeDir```, ```StatFile```, ...), which should be trivial. There is no explicit problem as there was with Chromium build, just some inconsistency with codepages (A-functions use system encoding, which is not UTF8).

tambry:
> A-functions use system encoding, which is not UTF8

[Unless you set the default system encoding to UTF-8](https://vgy.me/DlrEAM.png).

jonesmz:
https://github.com/ninja-build/ninja/pull/1671

Issue #1265, opened 2017-04-11
Fortran support


albertziegenhagel:
Currently there is a fork maintained by @bradking from Kitware that adds Fortran support to ninja.

Support in CMake to generate ninja files including Fortran has been present for quiet a while now and we are using the fork successfully for many projects in my company. It would be great if that work could be added to upstream ninja, so that we no longer need to rely on a fork.

There are two merge requests (#1039 and #1111) related to this topic that for some reason do not seem to get a lot of attention...

janusw:
I have recently tried the fork at https://github.com/Kitware/ninja on a large Fortran project ( >300 kloc in over 800 files, mostly F90 with some F77) and it works absolutely great, without any flaws.

Also it finally makes cmake usable with Fortran by providing great performance improvements: The project mentioned above takes about 30s for a trivial (no-op) rebuild with cmake+make, while cmake+ninja does it in a split second.

I'ld really love to see the Fortran supported being merged into the official ninja repository.

petsche:
My team has also been using the Kitware ninja fork for a large project that contains both C++ and Fortran without any flaws. It does seem like it is time for these merge requests to get attention. Thanks @bradking for the good work.

janusw:
> There are two merge requests (#1039 and #1111) related to this topic

Apparently the latter is merged by now, so we're halfway there and only #1039 is still missing ...

bradking:
The next step is #1294.

mathomp4:
@bradking I haven't seen any new updates to [Kitware/ninja](https://github.com/Kitware/ninja) in a while (2017 Jun 21). Does ninja now support Fortran in such a way that I can use the canonical ninja? (I thought about asking on the Kitware/ninja issues...but there are none.)

mathstuf:
@mathomp4 PR #1294 is waiting for canonical Ninja to support Fortran (I don't know how many steps are left after it).

mathomp4:
@mathstuf Ah. Okay. I just hadn't seen a new Kitware ninja, so I thought it might have gone moribund because it wasn't needed. :)

mathstuf:
Nope, Fortran just doesn't change much these days ;) .

janusw:
> I just hadn't seen a new Kitware ninja, so I thought it might have gone moribund because it wasn't needed. :)

[...]

Issue #1303, opened 2017-07-17
Warn/deprecate `depfile` rules that generate their own edges


taktoa:
Note: the text below is slightly modified from a series of messages I wrote on the `ninja-build` IRC channel. Since no one replied to them, I figure that it'd be better to put them in a more permanent place.

I am currently developing a tool that generates a Nix package (derivation) from a Ninja file and a source path, thus allowing reproducible/incremental/distributed builds of anything that can emit a `build.ninja`. Unfortunately, I've run into what seems to be a dire problem, that is a direct result of what I see as somewhat of a misfeature in Ninja.

Essentially, the fact that it is possible to use a depfile rule to generate dependencies of that self-same rule is a big problem. This wouldn't be so bad if it weren't for the fact that the depfile documentation basically recommends this style (assuming I am not misunderstanding the semantics of GCC):

```
rule cc
  depfile = $out.d
  command = gcc -MMD -MF $out.d [other gcc flags here]
```

CMake generates rules like this, and I suspect many of the other Ninja generators I'd like to accept input from do as well, since it's the "obvious" way to get perfect dependencies in a C/C++ project. Admittedly there is a benefit in performance by having one rule instead of two.

The reason this basically makes my project impossible is because you can't make a clean phase separation between rules-that-generate-build-edges and rules-that-are-expensive. You end up having to build all the depfiles twice: once with all the files available, and once sandboxed with the correct dependencies, which destroys incrementalism. Note that, AFAICT, adopting the style I am suggesting does have correctness benefits: as it stands, it is possible for builds to be nondeterministic if you forget to add enough order-only dependencies.

tl;dr: I'm suggesting that instead of doing `gcc -MMD -MF $out.d [other flags]`, you split it into two rules: one that _only_ outputs the dependency data, and one that actually compiles the file. ninja could also disallow (or discourage) people from writing these self-dependent nondeterministic rules.

If I'm wrong about this being an issue, please let me know; I'd be ecstatic.

zlolik:
@taktoa , I do not understand clearly why you need separate depfile phase. But I want to add that some compilers can not generate depfile while compiling. So I had to do rule with separate calling `$CXX -ppd=$out.d` (like `gcc -MMD -MF $out.d`):
```
rule compile_c6x
  command = cmd.exe /c @"$rspfile"
  depfile = $out.d
  rspfile = $out.bat
[...]

PR #1521, opened 2019-01-23
dyndep: dynamically discovered dependencies for Fortran and C++20 modules

bradking:
Fortran modules (and likely C++20 modules) require sources that provide modules to be compiled before sources that consume modules.  Therefore the content of the source files defines dependencies.  Unlike preprocessor dependencies (handled by `depfile`), module dependencies must be discovered *before* compiling sources even on the first build.  This PR introduces a `dyndep` feature to provide buildsystem generators with a way to handle such dynamically discovered dependencies.

For some history, see the ninja-build list thread [Adding Ninja features needed for Fortran](https://groups.google.com/d/topic/ninja-build/tPOcu5EWXio/discussion) and issue #1265.

This PR includes the full implementation, documentation, and thorough tests.  The changes have been used in practice by CMake users for years (via a Kitware-maintained branch of ninja).  They have been used to build large Fortran projects with thousands of sources and hundreds of modules.

The feature works by providing a new `dyndep = ` binding for build statements.  It specifies one of the (order-only) inputs that will provide the dynamically discovered dependencies that must be loaded before the build statement's command is executed.  Once the "dyndep" file is ready, ninja loads it to inject additional dependencies in its build graph, re-evaluate the affected portion of the graph, and then continue the build normally.

For example:

```
rule f95
  command = f95 -o $out -c $in
rule fscan
  command = fscan -o $out $in

build foobar.dd: fscan foo.f90 bar.f90

build foo.o: f95 foo.f90 || foobar.dd
  dyndep = foobar.dd
build bar.o: f95 bar.f90 || foobar.dd
  dyndep = foobar.dd
```

In this example the order-only dependencies ensure that `foobar.dd` is generated before either source compiles.  The hypothetical `fscan` tool (provided by the buildsystem generator in practice) scans the source files and writes `foobar.dd` with content such as:

```
ninja_dyndep_version = 1
build foo.o | foo.mod: dyndep
build bar.o: dyndep |  foo.mod
```

The "dyndep" file specifies additional inputs and outputs for the relevant build statements beyond those specified in the build manifest.

See documentation added by this PR for more details and examples.

Fixes: #1265


mqudsi:
Please forgive me if I'm overstepping here, it seems like a lot of good work has gone into this - but is it necessary to required `foobar.dd` to be listed both as the `dyndep` source and explicitly as an order-only dependency? Can't ninja deduce the latter?

bradking:
@mqudsi I did think about that while designing this but decided against having dependencies specified in `build.ninja` by means other than the normal syntax.  Also, the dyndep input does not have to be an order-only input and the binding would not have a clear way to specify which type.  Currently it is an error for the `dyndep = ` binding to name a path that is not an input so we would have an option in the future to replace the error case with an implied input.

bradking:
Rebased on `master` for post-1.9 development.

dublet:
@bradking Is there any CMake branch where this feature is used?

bradking:
@dublet CMake 3.7 and higher already have support.  It is activated by recognizing a special `.dyndep-1` suffix on Ninja's version number that Kitware added to our branch of Ninja we maintain pending upstream integration.  One can activate it on top of this PR's branch by locally hacking `src/version.cc`:

```diff
-const char* kNinjaVersion = "1.9.0.git";
+const char* kNinjaVersion = "1.9.0.git.dyndep-1";
```

Of course once Ninja integrates this PR in a release then we can update CMake to check for that instead.

mathstuf:
This is probably future work, but outputting a build graph with dyndep information can help to find bottlenecks in the build. Is this already available with this PR or should I open an issue for it?

bradking:
@mathstuf I updated the branch here to load dyndep files for the clean, graph, and query tools.

bradking:
@jhasse thanks for the review.  I've updated the branch to address all review comments, either with fixes or additional source comments with answers to your questions.

src/build_test.cc
```diff
@@ -602,19 +648,33 @@ bool FakeCommandRunner::WaitForCommand(Result* result) {
...
-  last_command_ = NULL;
+
+  // Provide a way for test cases to verify when an edge finishes that
+  // some other edge is still active.  This is useful for test cases
+  // covering behavior involving multiple active edges.
+  string const& verify_active_edge = edge->GetBinding("verify_active_edge");
+  if (!verify_active_edge.empty()) {
+    bool verify_active_edge_found = false;
+    for (vector<Edge*>::iterator i = active_edges_.begin(),
+         e = active_edges_.end(); !verify_active_edge_found && i != e; ++i) {
```
jhasse: I would remove `!verify_active_edge_found && ` and use break below (maybe even skip that as this is only used in one test and not performance critical). While creating `e` is faster than calling `end()` in the condition, I wouldn't use that trick here as it makes the loop more complicated as it needs to be.

src/build_test.cc
```diff
@@ -602,19 +648,33 @@ bool FakeCommandRunner::WaitForCommand(Result* result) {
     result->status = ExitFailure;
   else
     result->status = ExitSuccess;
-  last_command_ = NULL;
+
+  // Provide a way for test cases to verify when an edge finishes that
+  // some other edge is still active.  This is useful for test cases
+  // covering behavior involving multiple active edges.
+  string const& verify_active_edge = edge->GetBinding("verify_active_edge");
```
jhasse: Please use west-const.

bradking:
@jhasse thanks for the review. I've updated the branch to address all review comments and rebase on `master`.

jhasse:
Awesome stuff! Again: Thank you so much for the hard work :)

bradking:
@jhasse great, thanks for the review and integration!

Ericson2314:
https://github.com/ninja-build/ninja/issues/1303 Per this long thread, I think it would be great to have an *optional* warning for `depfile` saying to use `dyndep` instead. I'm really stoked for `dyndep`; I would always like to use it instead of `depfile` so initial builds and rebuilds do the same thing.

melak47:
In general, how is the scanner supposed to know the name of the *output* file?
Say different source files produce their output in different directories, like
```
build bob/foo.o: f95 foo.f90
build alice/bar.o: f95 bar.f90
```
Now, generating a single `foobar.dd` with a single invocation of `fscan` would be difficult, the generator would have to pass the output file names for each input file into the scanner.
Then the scanner writes that information back out to the dyndep file, then ninja parses it back in.

Is that how it's meant to work?

Ericson2314:
@melak47 I think it's basically left to the user to ensure the dyndeps file contains the needed information. You can use less or more dyndeps as is convenient; more would help with not wanted to gather up everything for one big scan.

bradking:
> the generator would have to pass the output file names for each input file into the scanner....Is that how it's meant to work?

Yes.  For example, CMake's Fortran build rules pass `--obj=...` to the scanner to tell it what object file will be produced by compiling the source it is scanning.

Issue #1403, opened 2018-04-04
C++ standard in use?


jonesmz:
I've been reviewing the source for ninja, and see a lot of areas where c++11 or newer features would be nice to have, but aren't currently used, like the "= default;" keyword for constructors and destructors, ranged based for loops, nullptr, lambdas, constexpr, so on.

What's the maximum c++ version that ninja's codebase abides by?

evmar:
Ninja mostly just works as-is and doesn't see a lot of code churn, so I don't think it's worth rewriting bits of it in a newer dialect of C++.  Maybe Nico feels differently though.

jonesmz:
Fear of change for the sake of unspecified boogy-bugs doesn't really help much.

I'm considering contributing improvements to ninja, but I'm not going to do that if I'm required to use c++98, and can't take advantage of any language improvements from the last 10 years.

I'm willing to do the work to modernize the codebase, and understand that slow-and-steady wins the race, but I wanted to ask before I jumped in.

nico:
Ninja is intentionally written in C++98. C++11 is nice, but it means your program can't run on older systems, and build servers for some are often old systems.

Re old PRs: Like all projects, we don't merge all PRs. We keep the ones we don't merge open.

jonesmz:
When will Ninja adopt C++11?

I'm not asking to be rude, I seriously want to know. I've seen so many projects that use C89 in 2017-2018 for the same justification, but the oldest system the project maintainers know about is CentOS 6.

Are you planning to keep C++98 until 2028?

jonesmz:
Why are you keeping the PRs you don't merge open?

If the code isn't ever going to be merged, close the PR so people don't see the list and assume "The maintainers of this project ignore PRs", like I just did.

nico:
There are no plans on adopting C++11. I agree it makes writing code a bit nicer, but ninja's small enough that the benefits of not adopting it (ninja runs on older systems; ninja itself compiles a bit faster) outweigh that.

Re pull requests: Closing a pull request feels unfriendly to me, and there have been instances where I didn't want to merge something at first but changed my mind after enough people explained their use cases over time. Is closing pull request considered best practice? I don't know much about github etiquette.

jonesmz:
Thank you for the explanation. I disagree with your priorities, but I agree that you'be come to the appropriate conclusion based on your stated goals, and I respect that.

As for pull requests, I think it's subjective.

[...]

PR #1524, opened 2019-01-28
Minor improvement [Reserve vector]

shahzadlone:
Save on reallocation cost, by reserving the vector we know the size of before hand.

### DIFF
--- a/src/browse.cc
+++ b/src/browse.cc
@@ -47,6 +47,7 @@ void RunBrowsePython(State* state, const char* ninja_command,
       }
 
       std::vector<const char *> command;
+      command.reserve(argc + 7);
       command.push_back(NINJA_PYTHON);
       command.push_back("-");
       command.push_back("--ninja-command");


jhasse:
Thanks for the PR! While you're right, that this improves performance, the code path isn't hot and performance isn't really an issue there. Also calling Python takes the majority of the time anyway.

jonesmz:
Why did you close this PR without merging it?

You said it yourself, it improves the performance. What difference does it make if python is more expensive?

jhasse:
> performance isn't really an issue there.

shahzadlone:
@jhasse I agree this might not be a hot path, that being said this is better code than before, even if it might be on a small magnitude. This is either good (probably is) or does no negative effect at all.

I assumed the end goal usually is to achieve for better code. Even if the change is as small as changing a post-inc (i++) to pre-inc(++i), one can make an argument that compilers can optimize that but still you shouldn't compromise on code quality.

Anyways, even if this isn't merge I can understand. But could you please update the contribution guide lines to say you guys don't accept minor improvements.

jhasse:
It hard-codes the number of arguments, so it's not necessarely "better" code.

jonesmz:
You're moving the goal posts.

The PR, as of the first commit, is already better or equal to what's in the ninja codebase right now. (but its better for the vast majority of potential std::vector implementations).

There is no justification for not including this change.

Improvements don't need to be perfect to still be better. 

@shahzadlone : if you really want to make the PR slightly better, the suggestion I made in my previous comment would be how I would do it.

You're also welcome to submit a similar PR to my fork of ninja http://github.com/jonesmz/splinter

jhasse:
> Number of arguments, are already hard coded.

I meant the number `7`.

jonesmz:
> Your fork has ++14!? I thought this repository supported ++11, also I have only used string_view's for C++17 projects, does ++14 support it aswell?

Ah, sorry. Now that I refresh my memory by looking at the Splinter codebase, I see that I did indeed make the decision to go full C++17. std::string_view and std::array are fair game.

Please note that currently Splinter is several commits behind Ninja. I am waiting for Ninja to make a new release before I go through the work of rebasing. I'll probably manage to do so within a week or two of a Ninja release, but currently things are out of date.

The Ninja project is still C++98, and currently does not accept pull requests to enable modern C++. You can see part of the discussion about that here: https://github.com/ninja-build/ninja/issues/1403

jonesmz:
>> Number of arguments, are already hard coded.

> I meant the number 7.

The number of arguments is hard-coded at 7 + argc. Why does it matter?

jhasse:
> The number of arguments is hard-coded at 7 + argc. Why does it matter?

When adding another parameter, one has to remember changing 7 to 8.

jonesmz:
> When adding another parameter, one has to remember changing 7 to 8.

Wouldn't that be caught by the unit tests that such a PR would add to ensure the new parameter worked correctly?

jonesmz:
No actually it wouldn't. The only consequence would be an additional allocation.

Would my proposed solution be acceptable then?

jhasse:
> No actually it wouldn't. The only consequence would be an additional allocation.

Exactly ;)

There are times when performance is critical and times where developing comfort is more important. For the latter people often choose Python. This shows that this part of Ninja, is not meant to be performance-critical. So "better" code in this case is the one that is simpler, shorter, prettier, etc. IMHO using `reserve` in this case is overkill and unneeded.

I think if you would benchmark the change, there wouldn't be a measurable performance impact, making this a case of premature optimization.

jonesmz:
So again, you should have said all of that to the person who wrote the pull request initially.

And it still doesn't mean we should not improve this.

Is the suggestion that I made, such that the number of arguments is automatically handled if they change, acceptable or not? You can see a version of that suggestion here: https://github.com/jonesmz/splinter/pull/1

jonesmz:
> ??? This isn't a DM or private chat, I'm replying on the PR.

Yes, but you didn't provide these thoughts in your initial message that closed the PR.

> Sure, but I think that "performance" should be weighted very low for deciding what is an improvement to this code.

Ninja is all about speed. Just because it's not a hot path doesn't excuse the abuse of the memory allocator.

--------------------------------------------------------

What about this won't work with C++98?

The only difference between this and the code from the linked PR is I dropped the constexpr keyword.

```
static const char * staticArgs[] = {NINJA_PYTHON, "-", "--ninja-command", ninja_command, "-f", input_file};
static size_t argSize = sizeof(staticArgs)/sizeof(*staticArgs);
std::vector<const char *> command;
command.reserve(argSize + argc + 1);
for(size_t i = 0; i < argSize; ++i)
{
    command.push_back(staticArgs[i]);
}
for (int i = 0; i < argc; ++i)
{
    command.push_back(argv[i]);
}
command.push_back(nullptr);
```

jhasse:
> What about this won't work with C++98?

Nothing.

btw: If you add `cpp` after the first three backticks, the code snippet gets highlighted on GitHub.

shahzadlone:
https://github.com/shahzadlone/ninja/commit/7ffd251e288d5307cbf774cae3312ad968a9d0ce

This won't be updated here because this is closed, should I make a new PR ?

jonesmz:
Pull Requests can be re-opened by the project maintainer, and closed pull requests can be updated, and merged. You can make your changes here.

shahzadlone:
Perfect then, because this commit was on the same branch this pull request was made on.

jonesmz:
@jhasse 

What's wrong with

`
static const char * staticArgs[] = {NINJA_PYTHON, "-", "--ninja-command", ninja_command, "-f", input_file};
static size_t argSize = sizeof(staticArgs)/sizeof(*staticArgs);
std::vector<const char *> command;
command.reserve(argSize + argc + 1);
for(size_t i = 0; i < argSize; ++i)
{
    command.push_back(staticArgs[i]);
}
for (int i = 0; i < argc; ++i)
{
    command.push_back(argv[i]);
}
command.push_back(nullptr);
`

??

That code will work without C++11.

Are there plans for C++11, btw?

jhasse:
It's longer and harder to understand IMHO.

> Are there plans for C++11, btw?

No concrete ones. Support for Debian 8 ends on June 30, 2020, that might be a date to keep in mind.

shahzadlone:
@jhasse It is better code though, and what is so hard to understand about that code? It also fixes the problem of hard coding things which you yourself pointed out "When adding another parameter, one has to remember changing 7 to 8."

jhasse:
> It is better code though

That's subjective. I think we can agree that good code should be

1. short,
2. easy to understand,
3. and performant.

This is only better on 3. The weighting of those three factors is subjective, I gave some reasoning why I weight performance very low here. 

> what is so hard to understand about that code?

I didn't say it's hard to understand. It's hard**er**.

PR #1525, opened 2019-01-30
ninja_syntax.py: remove unused has_path argument

avallee:
The usage of the parameter was removed a long time ago and was never cleaned up. The argument is not provided in the test.


avallee:
Thanks Jan.

PR #1526, opened 2019-01-30
CI with Azure Pipelines (Linux, macOS and Windows at the same time)

### LINKED ISSUE #1509: Make artifact in travis/appveyor for released ninja binary
I think it is better to take ninja binary from CI instead of maintainer's personal machine.

Following docs seems relevant.
https://www.appveyor.com/docs/packaging-artifacts/
https://docs.travis-ci.com/user/deployment/releases/

jhasse:
Unlike Travis, Azure Pipelines has support for easy artifact uploading. And unlike AppVeyor it supports macOS.

For this to work, a project owner of ninja-build/ninja needs to enable it, and then it would look like this: https://dev.azure.com/jhasse/jhasse/_build/results?buildId=34

Fixes #1509.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,5 +1,6 @@
 trigger:
 - master
+- release
 
 jobs:
   - job: Linux
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -16,6 +16,7 @@ jobs:
         ./ninja_test --gtest_filter=-SubprocessTest.SetWithLots
         ./misc/ninja_syntax_test.py
         ./misc/output_test.py
+        strip ninja
         mkdir package
         mv ninja package
       displayName: Test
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -0,0 +1,73 @@
+trigger:
+- master
+
+jobs:
+  - job: Linux
+    pool:
+      vmImage: 'Ubuntu-16.04'
```
shahzadlone: just a question. Could it support multi-plat build. Example: 18.04 ?

jhasse: Only Ubuntu 16.04 is available, see https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#use-a-microsoft-hosted-agent

But other distributions could be supported via Docker.

jhasse:
superseded by #1681

Issue #1376, opened 2018-01-28
Ninja unnecessarily runs CMake just after it was run manually


Ede123:
Cross-posting [an issue from CMake bugtracker](https://gitlab.kitware.com/cmake/cmake/issues/17531) as I suspect this might be a ninja issue.

Just after running CMake manually, ninja decides to re-run CMake:
```
ninja explain: recorded mtime of build.ninja older than most recent input CMakeCache.txt (538856124 vs 538856322)
```
while the file is newer on disk

```
$ stat -c %Y build.ninja CMakeCache.txt
1517153123
1517153122
```

Ede123:
@QuLogic could this have been caused by 5fcdcf95cb62ab3d593c36ef90df27cef63874a1?

topilski:
Any updates?

nico:
Is this a regression? Does it happen with the current stable ninja? If so, it's not due to high-res timestamps, those aren't in a release yet.

Might be a dupe of #906.

topilski:
Hi @nico, i used 1.8.2 version.

tchernobog:
I was also heavily affected by this bug (a CMake run for us runs in about one minute and a half on Linux, about 6 minutes on Windows), so I tried compiling ninja from the master branch.

It appears I cannot reproduce this bug anymore. @Ede123 could you please try ninja master and see if the issue still persists for you?

ojura:
I have reproduced it. Ninja uses the timespamp in`.ninja_log`, it seems, and that one can fall out of date.

rnk:
I don't think this is #906, because my build.ninja is newer than my CMakeCache.txt:
```
$ stat build.ninja CMakeCache.txt
  File: build.ninja
...
Modify: 2018-09-05 13:38:28.056863700 -0700
  File: CMakeCache.txt
...
Modify: 2018-09-05 13:27:03.224376800 -0700
```
But if I compile some random target, ninja says the "recorded mtime" of build.ninja is older:
```
$ ninja -n -d explain not
ninja explain: recorded mtime of build.ninja older than most recent input LLVMBuild.cmake (557882386 vs 557883085)
[1 processes, 0/1 @ ?/s : 0.000s ] Re-running CMake...
```
As a workaround, if I run this sed command after manually running cmake, I get the behavior I want:
```
$ sed  '/\tbuild\.ninja\t/d' .ninja_log -i
```

maikelvdh:
[...]

PR #1527, opened 2019-02-04
graph: always use the on-disk mtime of generator rules

mathstuf:
Generator rules are likely to have been run outside of ninja, so the
build log should not be trusted for detecting whether or not to rerun
the rule.

---
Observed with CMake projects which exhibit behavior that `ninja` reruns `cmake` right after running `cmake` manually. Maybe it is better to just not add generator rules to the build log at all? I don't know of the implications of that.

See https://gitlab.kitware.com/cmake/cmake/issues/15830 for some context.

### DIFF
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -252,14 +252,21 @@ bool DependencyScan::RecomputeOutputDirty(Edge* edge,
         EXPLAIN("command line changed for %s", output->path().c_str());
         return true;
       }
-      if (most_recent_input && entry->mtime < most_recent_input->mtime()) {
+      // Compute the effective mtime for the rule.
+      TimeStamp effective_mtime = entry->mtime;
+      if (generator && output->status_known()) {
+        // For generator rules, always use the on-disk mtime since it is likely
+        // to have been regenerated outside of ninja.
+        effective_mtime = output->mtime();
+      }
+      if (most_recent_input && effective_mtime < most_recent_input->mtime()) {
         // May also be dirty due to the mtime in the log being older than the
         // mtime of the most recent input.  This can occur even when the mtime
         // on disk is newer if a previous run wrote to the output file but
         // exited with an error or was interrupted.
         EXPLAIN("recorded mtime of %s older than most recent input %s (%" PRId64 " vs %" PRId64 ")",
                 output->path().c_str(), most_recent_input->path().c_str(),
-                entry->mtime, most_recent_input->mtime());
+                effective_mtime, most_recent_input->mtime());
         return true;
       }
     }


fkaelberer:
So this should Fix #1376, I guess.

mathstuf:
Yes, it should. I suspect there are likely other duplicates of the issue around.

Siron777:
I have tested the patch and I do not observe anymore the issue #1376.

src/graph.cc
```diff
@@ -252,14 +252,21 @@ bool DependencyScan::RecomputeOutputDirty(Edge* edge,
         EXPLAIN("command line changed for %s", output->path().c_str());
         return true;
       }
-      if (most_recent_input && entry->mtime < most_recent_input->mtime()) {
+      // Compute the effective mtime for the rule.
+      TimeStamp effective_mtime = entry->mtime;
+      if (generator) {
+        // For generator rules, always use the on-disk mtime since it is likely
+        // to have been regenerated outside of ninja.
+        effective_mtime = output->mtime();
```
jhasse: How can you be sure that `mtime()` doesn't return `-1`?

mathstuf: Hmm. It is used above (line 220). I'll test out adding `&& output->status_known()` to the conditional guarding this.

mathstuf: Still works with that conditional. I don't know exactly how the status wouldn't be known at this point, but it doesn't happen in typical CMake usage patterns at least.

rubenvb:
This seems to fix the issue for me.

jhuels:
This fixes the issue for `generator` rule for me, but I don't think CMake provides a way to mark custom commands as `generator`.  I'm having an identical issue with a code generator where the disc time of the output is newer because it was run outside of ninja, but ninja_log is stale.

Do you know of a way around this, or should I request a feature for CMake to set custom commands as `generator`?

mathstuf:
Please open an issue over there (rather than muddying the discussion here), but include your example custom command. Ninja's rules on the `generator` rule variable doesn't really make sense with custom commands.

jhasse:
> Do you know of a way around this

Update the ninja_log in your custom command.

mathstuf:
> Update the ninja_log in your custom command.

I don't know that this is going to be easy. Ninja itself uses the tmpfile-and-rename style, so any command doing this would need to coordinate with ninja on what's going on. This just sounds 100% flaky to me.

jhasse:
> Ninja itself uses the tmpfile-and-rename style, so any command doing this would need to coordinate with ninja on what's going on.

But this shouldn't matter because we're talking about the case where the generator is run outside of ninja, right?

mathstuf:
How is the tool supposed to know if it is inside or outside ninja? Plus, the question asked was for a way to claim `generator = 1` in order to access the behavior introduced here. Maybe some other flag should be added that `generator = 1` should imply?

jhasse:
> How is the tool supposed to know if it is inside or outside ninja?

Pass itself a flag in its build rule for example.

mathstuf:
Maybe if the deplog format were documented and guaranteed to be stable, that route could be taken. Until then, is this a sufficient solution to the problem being seen?

jhasse:
I would rather add a tool to Ninja (e.g. `ninja -t updatecache`?) or add the functionality to an existing one. What about `ninja -t recompact`? Or the [new `cleandead` tool](https://github.com/ninja-build/ninja/pull/1432)? Running the latter on reconfigures is a good idea anyway, as it might catch bugs by deleting files hanging around from previous builds.

mathstuf:
Hmm. That could work. A pipeline like `ninja -t cleandead && ninja -t recompact && ninja -t restat` maybe?

jhasse:
Exactly! `restat` also sounds like a good name.

mathstuf:
Strawman MR filed for CMake here: https://gitlab.kitware.com/cmake/cmake/merge_requests/3316

jhasse:
Closing in favor of #1685.

PR #1534, opened 2019-03-01
manifest_parser: remove multi-output depslog restriction

mathstuf:
It seems to work as intended.

---
Cc: @RedBeard0531

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -1033,7 +1033,7 @@ bool Builder::FinishCommand(CommandRunner::Result* result, string* err) {
   }
 
   if (!deps_type.empty() && !config_.dry_run) {
-    assert(edge->outputs_.size() == 1 && "should have been rejected by parser");
+    assert(edge->outputs_.size() >= 1 && "should have been rejected by parser");
     for (std::vector<Node*>::const_iterator o = edge->outputs_.begin();
          o != edge->outputs_.end(); ++o) {
       TimeStamp deps_mtime = disk_interface_->Stat((*o)->path(), err);
--- a/src/build.cc
+++ b/src/build.cc
@@ -1034,13 +1034,15 @@ bool Builder::FinishCommand(CommandRunner::Result* result, string* err) {
 
   if (!deps_type.empty() && !config_.dry_run) {
     assert(edge->outputs_.size() == 1 && "should have been rejected by parser");
-    Node* out = edge->outputs_[0];
-    TimeStamp deps_mtime = disk_interface_->Stat(out->path(), err);
-    if (deps_mtime == -1)
-      return false;
-    if (!scan_.deps_log()->RecordDeps(out, deps_mtime, deps_nodes)) {
-      *err = string("Error writing to deps log: ") + strerror(errno);
-      return false;
+    for (std::vector<Node*>::const_iterator o = edge->outputs_.begin();
+         o != edge->outputs_.end(); ++o) {
+      TimeStamp deps_mtime = disk_interface_->Stat((*o)->path(), err);
+      if (deps_mtime == -1)
+        return false;
+      if (!scan_.deps_log()->RecordDeps(*o, deps_mtime, deps_nodes)) {
+        *err = std::string("Error writing to deps log: ") + strerror(errno);
+        return false;
+      }
     }
   }
   return true;
--- a/src/build_test.cc
+++ b/src/build_test.cc
@@ -488,6 +488,11 @@ struct BuildTest : public StateTestWithBuiltinRules, public BuildLogUser {
                 status_(config_) {
   }
 
+  BuildTest(DepsLog* log) : config_(MakeConfig()), command_runner_(&fs_),
+                            builder_(&state_, config_, NULL, log, &fs_),
+                            status_(config_) {
+  }
+
   virtual void SetUp() {
     StateTestWithBuiltinRules::SetUp();
 
@@ -582,6 +587,8 @@ bool FakeCommandRunner::StartCommand(Edge* edge) {
       edge->rule().name() == "cat_rsp" ||
       edge->rule().name() == "cat_rsp_out" ||
       edge->rule().name() == "cc" ||
+      edge->rule().name() == "cp_multi_msvc" ||
+      edge->rule().name() == "cp_multi_gcc" ||
       edge->rule().name() == "touch" ||
       edge->rule().name() == "touch-interrupt" ||
       edge->rule().name() == "touch-fail-tick2") {
@@ -641,6 +648,14 @@ bool FakeCommandRunner::WaitForCommand(Result* result) {
       result->status = ExitFailure;
     active_edges_.erase(edge_iter);
     return true;
+  }
+
+  if (edge->rule().name() == "cp_multi_msvc") {
+    const std::string prefix = edge->GetBinding("msvc_deps_prefix");
+    for (std::vector<Node*>::iterator in = edge->inputs_.begin();
+         in != edge->inputs_.end(); ++in) {
+      result->output += prefix + (*in)->path() + '\n';
+    }
   }
 
   if (edge->rule().name() == "fail" ||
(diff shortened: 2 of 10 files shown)


jhasse:
A test that it works as intended would be good (not only that it parses without error).

mathstuf:
Here's a test file (courtesy of @RedBeard0531)

```ninja
rule TOUCH
    command = touch $out

rule COPY
    command = echo 'using $in' && for file in $out; do cp $in $$file; done
    deps = msvc
    msvc_deps_prefix = using 

rule CHECK
    command = for file in $in; do if ! test -e $$file; then return 99; fi; done

build source: TOUCH
build dest1 dest2: COPY source
build test1: CHECK dest1
build test2: CHECK dest2
build test3: CHECK dest1 dest2
build test: phony test1 test2 test3
default test
```

`ninja -t deps dest1` and `ninja -t deps dest2` both output the expected depslog information. What's the best way to add this to the test suite?

mathstuf:
Tests should also cover gcc parsing for the following two cases:

```make
a b: in
```

and

```make
a: in
b: in
```

RedBeard0531:
It should also work with either of:

```make
a: in
```
```make
b: in
```

Especially when one of the outputs is an implicit output. This is important for cases like `-gsplit-dwarf -MMD` because the compiler seems to only list the "primary" output in its deps files.

That said, if it is harder to do this for make-style deps, getting it just for msvc-style would be an improvement for a few of my use cases, so I'd be happy to see msvc-style support land independently if that would get it in an earlier release.

mathstuf:
Also found out the place to implement the tests for this (`build_test.cc`). I'll work on that this weekend.

mathstuf:
@RedBeard0531 Noted, I'll add a test for that as well. Though we should get GCC to mention the `.dwo` file in the deps log if it doesn't already.

mathstuf:
Tests added. @RedBeard0531 he `b: in` (or even `b a: in`) is currently treated the same as if you had a depfile not agreeing on the "primary" output name and is basically ignored. It's up to the generator to get the "primary" output correct.

mathstuf:
Odd. Those failures aren't happening locally. Debugging.

mathstuf:
Ugh. I wish GTest gave more information :/ . Can anyone else reproduce the test failures?

mathstuf:
Passing on my local macOS build too. Will try with Windows later.

jonesmz:
> Ugh. I wish GTest gave more information :/ . Can anyone else reproduce the test failures?

Don't blame GTest for that. The Ninja code base has it's own special set of code that pretends to be GTest and does a horrible job of it.

Quoted from test.h:
```
// A tiny testing framework inspired by googletest, but much simpler and
// faster to compile. It supports most things commonly used from googltest. The
// most noticeable things missing: EXPECT_* and ASSERT_* don't support
// streaming notes to them with operator<<, and for failing tests the lhs and
// rhs are not printed. That's so that this header does not have to include
// sstream, which slows down building ninja_test almost 20%.
```

\<sarcasm> Obviously 20% is a huge concern when dealing with a codebase that compiles in under a minute \</sarcasm>

I encourage you to consider writing a pull request that makes the configure.py script's "--with-gtest" actually do something (currently it does nothing at all, as the letters "gtest" are found exactly once in configure.py).

But probably don't start working on that until you've asked on the Ninja mailing list https://groups.google.com/forum/#!forum/ninja-build , as there's a decent chance (in my opinion) that such a PR would be either ignored or rejected.

jhasse:
> I encourage you to consider writing a pull request that makes the configure.py script's "--with-gtest" actually do something (currently it does nothing at all, as the letters "gtest" are found exactly once in configure.py).

I think it would be better to remove that switch and always use googletest.

> But probably don't start working on that until you've asked on the Ninja mailing list 

Or opened an issue where you suggest the change.

> as there's a decent chance (in my opinion) that such a PR would be either ignored or rejected.

PRs should no longer be ignored, I'm trying to change that. If you think there's an open one which should be looked at, just ping me :)

jhasse:
PR for GoogleTest: #1562

You can cherry-pick the commit to get more information about the failed test.

mathstuf:
Thanks, will look at this in the next week or so.

mathstuf:
Ah ha. Installing re2c gets me to agree with the CI now. Tracking it down.

mathstuf:
@jhasse Now that the cc.in file has right code too, CI is passing.

src/depfile_parser_test.cc
```diff
@@ -266,25 +288,40 @@ TEST_F(DepfileParserTest, MultipleRulesTolerateMP) {
...
+  ASSERT_EQ("bar", parser_.outs_[1].AsString());
+  ASSERT_EQ(3u, parser_.ins_.size());
+  EXPECT_EQ("x", parser_.ins_[0].AsString());
+  EXPECT_EQ("y", parser_.ins_[1].AsString());
+  EXPECT_EQ("z", parser_.ins_[2].AsString());
+}
+
+TEST_F(DepfileParserTest, BuggyMP) {
+  string err;
+  EXPECT_TRUE(Parse("foo: x y z\n"
```
mathstuf: @bradking and I discussed this case here. This test passes as-is, but some logic will need to change if we want to continue rejecting this case.

RedBeard0531:
This is a rather severe limitation that prevents listing correct dependencies for a c++20 module scanner that outputs both a dyndeps file and a compiler flags file, when the compile task depends on both but order only for dyndeps and implicit for flags file. Additionally, the overall restriction has been lifted with the dyndeps patch because dyndeps can add implicit outputs to a `deps=gcc` build.

What is remaining to getting this merged?

mathstuf:
There's the conflict to resolve, but more importantly, there is the test case I commented on above. What was labeled a "buggy" depfile is no accepted. Should it continue to be rejected? If so, further work is necessary.

mathstuf:
Rebased and now rejects the case again.

petrhosek:
We've the limitations around multiple outputs with deps when building Rust targets with Ninja. I've tested this change locally and it seems to be working as expected. Can we get this merged?

mathstuf:
I've rebased on top of `master`.

mathstuf:
Any chance of this for 1.10 too? :) @jhasse

jhasse:
Let's see. It probably needs someone more experienced than me to review.

mathstuf:
Hmm. @bradking? @nico? Thoughts on this PR?

src/build.cc
```diff
@@ -1033,14 +1033,15 @@ bool Builder::FinishCommand(CommandRunner::Result* result, string* err) {
   }
 
   if (!deps_type.empty() && !config_.dry_run) {
-    assert(edge->outputs_.size() == 1 && "should have been rejected by parser");
```
bradking: Should we still assert that there is at least one output?

mathstuf: I've added back the assert. I can't test it because `build: foo` is invalid because at least one output is required in the grammar.

mathstuf: Rather, I can't test that the appropriate check in `manifest_parser` would be triggered; the error happens before that point).

src/build.cc
```diff
@@ -1033,14 +1033,16 @@ bool Builder::FinishCommand(CommandRunner::Result* result, string* err) {
...
-    Node* out = edge->outputs_[0];
-    TimeStamp deps_mtime = disk_interface_->Stat(out->path(), err);
-    if (deps_mtime == -1)
-      return false;
-    if (!scan_.deps_log()->RecordDeps(out, deps_mtime, deps_nodes)) {
-      *err = string("Error writing to deps log: ") + strerror(errno);
-      return false;
+    assert(edge->outputs_.size() >= 1 && "should have been rejected by parser");
+    for (vector<Node*>::const_iterator o = edge->outputs_.begin();
```
jonesmz: std::

src/build_test.cc
```diff
@@ -643,6 +650,14 @@ bool FakeCommandRunner::WaitForCommand(Result* result) {
     return true;
   }
 
+  if (edge->rule().name() == "cp_multi_msvc") {
+    const string prefix = edge->GetBinding("msvc_deps_prefix");
```
jonesmz: std::

src/depfile_parser.cc
```diff
@@ -294,41 +294,32 @@ bool DepfileParser::Parse(string* content, string* err) {
...
-            *err = "depfile has multiple output paths";
-            return false;
-          } else if (options_.depfile_distinct_target_lines_action_ ==
-                     kDepfileDistinctTargetLinesActionError) {
-            *err =
-                "depfile has multiple output paths (on separate lines)"
-                " [-w depfilemulti=err]";
+      StringPiece piece = StringPiece(filename, len);
+      // If we've seen this as an input before, skip it.
+      vector<StringPiece>::iterator pos = find(ins_.begin(), ins_.end(), piece);
```
jonesmz: would a std::set make more sense here? All of the items in the vector need to be unique, right?

bradking: `ins_` is the list of inputs and must preserve order, but one could consider using a separate local `std::set`.

src/graph.cc
```diff
@@ -541,9 +553,15 @@ bool ImplicitDepLoader::LoadDepFile(Edge* edge, const string& path,
...
+  if (depfile.outs_.empty()) {
+    *err = path + ": no outputs declared";
+    return false;
+  }
+
   uint64_t unused;
-  if (!CanonicalizePath(const_cast<char*>(depfile.out_.str_),
-                        &depfile.out_.len_, &unused, err)) {
+  vector<StringPiece>::iterator primary_out = depfile.outs_.begin();
+  if (!CanonicalizePath(const_cast<char*>(primary_out->str_),
```
jonesmz: This doesn't seem like a good idea.

Modifying the contents of a StringPiece in place violates the assumptions of the class.

src/build.cc
jonesmz: See new guidelines for the use of std:: here https://github.com/ninja-build/ninja/pull/1616/files

src/graph.cc
bradking: The existing code was doing this.  Fixing that is beyond the scope of this PR.

mathstuf:
Updated to use `std::` on all lines added/changed in this PR. I left `std::strerror` alone since `<cstring>` is not included anyways.

src/depfile_parser.cc
bradking: Actually since depfile-specified inputs do not affect `$in` the order does not matter.  One could convert the depfile parser's `ins_` member from `vector` to `std::set`.

mathstuf: I think converting it over is a larger change out-of-scope here. Lots of test suite checks would need to cope with the out-of-order-ness of a `std::set`.

bradking: Okay, that optimization can be reserved for a later PR.

PR #1536, opened 2019-03-07
Hpux

millert:
HP-UX is missing posix_spawn() so I've included the version by Ed Schouten which is used by FreeBSD and OpenBSD.  Since posix_spawn.c doesn't build with g++ I had to make some tweaks to configure.py to compile .c files with gcc instead of g++.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -62,12 +62,14 @@ class Platform(object):
             self._platform = 'aix'
         elif self._platform.startswith('dragonfly'):
             self._platform = 'dragonfly'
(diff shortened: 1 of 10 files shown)


jhasse:
This adds quite a bit of code to the repository which I can't test myself. Is there a repository for HP-UX where patches like this could be kept? I'm not sure if this should be part of upstream ninja. After a while we are probably going to break HP-UX again, because there's no CI or anything.

jhasse:
Closing for now.

PR #1543, opened 2019-03-22
Pass command-line options to child processes in an environment variable

bonzini:
This pull request adds support for a NINJAFLAGS environment variable that passes a subset of the ninja command line option down to child processes. When ninja starts a subprocess, they can use `NINJAFLAGS` to control their own verbosity and parallelism. This is particularly useful for test suites.

~~This pull request adds support for a NINJAFLAGS environment variable that accepts a subset of the ninja command line option and passes them down to child processes.~~

~~The possible uses are many:~~

* ~~when ninja is invoked by the user or an automated build system, this allows a build system to set up a global load average cap, working around the lack of support for a global jobserver (#1139). Build systems would probably want to add a global `-v` option to ensure that build logs are more complete and failures are more easily understandable.~~

* ~~when ninja starts a subprocess, they can use `NINJAFLAGS` to control their own verbosity and parallelism. This is particularly useful for test suites.~~

* ~~in the rare case when ninja is invoked recursively, it will obey the verbosity and maximum load average of the parent too.  An example is Meson's `ninja clean` shortcut, it makes `ninja -t clean` obey the verbosity that was passed to `ninja clean'~~

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -192,7 +192,6 @@ Environment variables
 
 Ninja supports one environment variable to control its behavior:
 `NINJA_STATUS`, the progress status printed before the rule being run.
-
 Several placeholders are available:
 
 `%s`:: The number of started edges.
@@ -210,6 +209,15 @@ specified by `-j` or its default)
 The default progress status is `"[%f/%t] "` (note the trailing space
 to separate from the build rule). Another example of possible progress status
 could be `"[%u/%r/%f] "`.
+
+In addition, Ninja passes one environment variable to the commands it
+starts, `NINJAFLAGS`. `NINJAFLAGS` contains the `-k`, `-j`, `-l` and
+`-v` options that are in use for the build, except that `-j` is
+only present if the command is started in the <<ref_console_pool,
+"console" pool>>.  External programs can inspect the contents of the
+`NINJAFLAGS` variable, for example to control their own parallelism.
+Note that the exact set of options that it contains may vary in the
+future, so programs should ignore unknown options.
 
 Extra tools
 ~~~~~~~~~~~
@@ -663,6 +671,7 @@ build heavy_object2.obj: cc heavy_obj2.cc
 
 ----------------
 
+[[ref_console_pool]]
 The `console` pool
 ^^^^^^^^^^^^^^^^^^
 
--- a/src/build.cc
+++ b/src/build.cc
@@ -491,7 +491,8 @@ void Plan::Dump() {
 }
 
 struct RealCommandRunner : public CommandRunner {
-  explicit RealCommandRunner(const BuildConfig& config) : config_(config) {}
+  explicit RealCommandRunner(const BuildConfig& config) : config_(config),
+    subprocs_(config_) {}
   virtual ~RealCommandRunner() {}
   virtual bool CanRunMore();
   virtual bool StartCommand(Edge* edge);
--- a/src/build.cc
+++ b/src/build.cc
@@ -16,9 +16,11 @@
 
 #include <assert.h>
 #include <errno.h>
+#include <limits.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <functional>
+#include <sstream>
 
 #ifdef _WIN32
 #include <fcntl.h>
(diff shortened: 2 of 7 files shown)


jonesmz:
Good luck...

https://github.com/ninja-build/ninja/pull/1399

bonzini:
Yeah I know. That's why I explicitly limited it to a few options and not all. Another possibility is to add a NINJA_PPID variable and only obey NINJAFLAGS if the parent pid matches, but for now I would like a general review of the code.

The use case is real. I am seriously thinking of using a build.ninja to Makefile translator for my project, no matter how stupid that might sound and probably is, if the developers cannot listen to their users' requests as is the case for this feature or the jobserver...

bonzini:
Trying to be more constructive:

* I can split the functionality between passing and receiving NINJAFLAGS if desired. The most important part is passing flags to testsuite runners, and passing NINJAFLAGS to child processes would be enough for that to work right. This part is what makes this pull request different from #1399.

* The NINJAFLAGS + NINJA_PPID pair I suggested previously would have the effect of allowing the third case (`ninja -v clean` -> `ninja -v -t clean` for meson-generated `build.ninja`), but that case is the least important and it is probably not worth the complication of adding NINJA_PPID. So I can do it, and it would work, but I don't really think it's a good idea. I'd rather remove entirely the functionality of reading NINJAFLAGS.

Regarding the last point:

* The main point made in #1399 against reading NINJAFLAGS was the user can forget about that random script that has set NINJAFLAGS to -j80. That is certainly a good point; if somebody is for example using distcc, he would place a script in /usr/local/bin or $HOME/bin that adds the flag, and `which ninja` would flag that easily. Therefore, it is probably a good idea to mention NINJAFLAGS in `ninja -v` output.

* On top of this, the patch to correct the getopt reset is an independent bugfix.

Therefore, I will limit this pull request to passing NINJAFLAGS to the child precesses, and create a separate pull request to correct the gteopt reset.

jhasse:
> This is particularly useful for test suites.

Can you elaborate?

bonzini:
Test suite drivers want to run multiple test jobs in parallel, and they can also vary their verbosity (e.g. print the command they run vs just the test name). This patch allows them to vary their behavior based on Ninja's command line options.

jhasse:
I see. If I understand it correctly you want to use ninja as the command line front end for the test suite. While that might be useful in some cases, I (and I think I'm not alone with this) would expect the underlying commands NOT to change their behavior or output, when I'm passing flags like `-v` or `-j` to Ninja.

Imagine a user who wants to limit the load to one CPU core, therefore passes `-j1` to Ninja. Now the test suite thinks "Oh I'm the only process running! Let's use all cores :)". Not what the user expected.

Furthermore this becomes problematic, when there's a bug in the test suite, which depends on the environment variable. Imagine a command errors, then you run with `ninja -v` to see how it was called, but then it succeeds (because it only fails when in verbose mode itself).

btw: Some build tools (e.g. Tup) warn you when any environment variable has changed, exactly for this reason.

Creating your own script as the front end for the test suit is a better approach. Sorry, that I'm closing this for now, I think it's out-of-scope for Ninja. It's definitely something we have to discuss in an issue or on the mailing list first, before making the change.

bonzini:
> Imagine a user who wants to limit the load to one CPU core, therefore passes -j1 to Ninja. Now the test suite thinks "Oh I'm the only process running! Let's use all cores :)". Not what the user expected.

Well, the idea behind the patch was the opposite, i.e. to *not* use all the cores when the user passes -j1, because there are already tools that are using all the cores by default just like ninja. Of course the best solution would be to support the jobserver not just as client (#1139) but also as server, so that the testsuite runner would not bother guessing the right number of jobs, and instead would just use the jobserver.

jhasse:
> Well, the idea behind the patch was the opposite, i.e. to not use all the cores when the user passes -j1, because there are already tools that are using all the cores by default just like ninja.

Wouldn't this mean that when the user passes `-j8` the tool would use 8 threads also, resulting in 64 threads?

bonzini:
Yes, that's exactly the case now if the tool uses #processors as the default. Ninja could run 8 copies of it, each of them using 8 threads. It is not a big deal because usually things such as "ninja test" will depend on all build products, but it is indeed a problem.

In fact this is why the patch only passes "-j" if the task is in the console pool, enabling the following:
* if NINJAFLAGS is found and lacks "-j", the tool should default to "-j1" instead of "-j#processors", because it means the tool was not placed in the console pool;
* if NINJAFLAGS is found and does have "-j", it will be possible to have up to 2*#ninjajobs-1 "things" running, which is not great but not worse than the current situation.

eli-schwartz:
The independent Ninja build language reimplementation https://github.com/michaelforney/samurai supports this via the environment variable SAMUFLAGS. Simply use samu anywhere you'd otherwise use ninja, or install samu as your system /usr/bin/ninja (some linux distros have options to install this competing implementation as the default, or only, ninja).

bonzini:
Samurai as far as I know can take flags from the environment, but not pass them to child processes if they were specified or overridden on the command line.

PR #1545, opened 2019-03-24
Improve portability of getopt reset

bonzini:
getopt reset is done sometimes with "optind = 0", sometimes with "optind = 1".  Unfortunately, neither is portable.  For example, if the bundled getopt see "optind = 1" on the very first call it will implicitly add a dash in front of the first argument; and while some glibc versions even had reports of crashes if reset with "optind = 1", Apple's getopt did the same for "optind = 0".
    
Since we know that there will always be two getopt calls only, of which the first will be in ReadFlags, place the ugly code there so that it is not replicated all over ninja.cc.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -97,6 +97,10 @@ class Platform(object):
     def is_aix(self):
         return self._platform == 'aix'
 
+    def uses_bundled_getopt(self):
+        # AIX supplies getopt but not getopt_long.
+        return self.is_windows() or self.is_aix()
+
     def uses_usr_local(self):
         return self._platform in ('freebsd', 'openbsd', 'bitrig', 'dragonfly', 'netbsd')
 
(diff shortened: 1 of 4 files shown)


jhasse:
Why not use the bundled getopt all the time?

bonzini:
That would be a questioni for you. :-) There could be prototype compatibility issues I guess, similar to AIX.

jhasse:
> That would be a questioni for you. :-)

I didn't write any of this, so I have no idea.

> There could be prototype compatibility issues I guess

You mean that the definition of getopt might conflict with the system provided one?

bonzini:
Yes, for example const differences could be a problem.

evmar:
I don't quite remember, but I think I first wrote the code assuming that getopt was always available, then once we were on a system that didn't have it we added the builtin one.  I agree that maybe always using the builtin one is a good idea.

jhasse:
Always using the builtin one is the way to go IMHO. Eventually even move on to a different solution with a better API.

PR #1546, opened 2019-03-24
Allow 'ninja -t compdb' accept one target

### LINKED ISSUE #1544: [Feature Request] Allow 'ninja -t compdb' accept one target

e.g.

New help message will be:

options:

lk-chen:
Fix #1544 

Test:
./ninja -t compdb -a wrong_target (fail)
./ninja -t compdb -a test_target (empty due to lack of rules)
./ninja -t compdb -a test_target cc cxx (ok)
./ninja -t compdb cxx (ok, regression test)
./ninja ninja_test && ./ninja_test (passed)

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -17,6 +17,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <unordered_set>
 
 #ifdef _WIN32
 #include "getopt.h"
@@ -699,6 +700,38 @@ string EvaluateCommandWithRspfile(Edge* edge, EvaluateCommandMode mode) {
   return command;
 }
 
+bool GetAllDependentEdges(Node* node, std::vector<Edge*>* depend_edges,
+                          std::unordered_set<Node*>* visited_nodes=NULL,
+                          std::unordered_set<Edge*>* visited_edges=NULL) {
+  if (visited_nodes == NULL || visited_edges == NULL) {
+    std::unordered_set<Node*> visited_nodes_;
+    std::unordered_set<Edge*> visited_edges_;
+    return GetAllDependentEdges(node, depend_edges, &visited_nodes_, &visited_edges_);
+  }
+  if (node == NULL || depend_edges == NULL) {
+    Error("Internal error");
+    return false;
+  }
+  if (visited_nodes->count(node)) {
+    return true;
+  } else {
+    visited_nodes->insert(node);
+  }
+  Edge* edge = node->in_edge();
+  // Leaf node
+  if (!edge || visited_edges->count(edge)) {
+    return true;
+  } else {
+    visited_edges->insert(edge);
+    depend_edges->push_back(edge);
+  }
+  for (Node* input_node : edge->inputs_) {
+    if (!GetAllDependentEdges(input_node, depend_edges, visited_nodes, visited_edges))
+      return false;
+  }
+  return true;
+}
+
 int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
                                        char* argv[]) {
   // The compdb tool uses getopt, and expects argv[0] to contain the name of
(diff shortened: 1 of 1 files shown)


src/ninja.cc
```diff
@@ -710,26 +743,51 @@ int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
 
   optind = 1;
   int opt;
-  while ((opt = getopt(argc, argv, const_cast<char*>("hx"))) != -1) {
+  string err;
+  Node* user_given_target = NULL;
```
jhasse: Please create a scope in for `case 'a'` and move those two variable declarations there.

See http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-introduce

src/ninja.cc
```diff
@@ -710,26 +743,51 @@ int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
...
+          Error("%s: %s", argv[optind], err.c_str());
+          return 1;
+        }
+        break;
+
       case 'h':
       default:
         printf(
-            "usage: ninja -t compdb [options] [rules]\n"
+            "usage: ninja -t compdb [options] [target] [rules]\n"
```
jhasse: Isn't `[target]` only useful when passing `-a`?

src/ninja.cc
```diff
@@ -710,26 +743,51 @@ int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
...
             "  -x     expand @rspfile style response file invocations\n"
+            "  -a     generate compilation database for given target\n"
             );
         return 1;
     }
   }
   argv += optind;
   argc -= optind;
 
+  std::vector<Edge*> user_interested_edges;
```
jhasse: Also move this variable into the `if (user_given_target) {`.

src/ninja.cc
```diff
@@ -699,6 +700,38 @@ string EvaluateCommandWithRspfile(Edge* edge, EvaluateCommandMode mode) {
...
+    std::unordered_set<Edge*> visited_edges_;
+    return GetAllDependentEdges(node, depend_edges, &visited_nodes_, &visited_edges_);
+  }
+  if (node == NULL || depend_edges == NULL) {
+    Error("Internal error");
+    return false;
+  }
+  if (visited_nodes->count(node)) {
+    return true;
+  } else {
```
jhasse: No need for `else` after `return`.

src/ninja.cc
```diff
@@ -699,6 +700,38 @@ string EvaluateCommandWithRspfile(Edge* edge, EvaluateCommandMode mode) {
...
+    visited_nodes->insert(node);
+  }
+  Edge* edge = node->in_edge();
+  // Leaf node
+  if (!edge || visited_edges->count(edge)) {
+    return true;
+  } else {
```
jhasse: Same here.

src/ninja.cc
```diff
@@ -699,6 +700,38 @@ string EvaluateCommandWithRspfile(Edge* edge, EvaluateCommandMode mode) {
...
+                          std::unordered_set<Node*>* visited_nodes=NULL,
+                          std::unordered_set<Edge*>* visited_edges=NULL) {
+  if (visited_nodes == NULL || visited_edges == NULL) {
+    std::unordered_set<Node*> visited_nodes_;
+    std::unordered_set<Edge*> visited_edges_;
+    return GetAllDependentEdges(node, depend_edges, &visited_nodes_, &visited_edges_);
+  }
+  if (node == NULL || depend_edges == NULL) {
+    Error("Internal error");
+    return false;
```
jhasse: I would use `assert` for this instead.

jhasse:
Before you continue working on this, I think we need to discuss if

1) only one target should be accepted,
2) `-a` is a good name for the flag,
3) this feature should be added at all.

lk-chen:
1. Technically, I don't see a way to pass multiple targets
2. I am surely open to suggestions. I chose 'a' from 'tArget' since 't' is used for 'tools'.
3. I am working on chromecast, and using cquery+compilation database. The compilation database is too big to work well on my laptop (~17G memory). Per my test, the size of compilatoin database JSON file decreased by ~50% if given '-a cast_shell'. I can even specify a smaller sub-target to improve performance. Thus I believe it is useful.

lk-chen:
ping?

jhasse:
> ping?

Please don't do this, I'm getting enough emails as it is.

> Technically, I don't see a way to pass multiple targets

Maybe separating them with `,`s?

> I am surely open to suggestions. I chose 'a' from 'tArget' since 't' is used for 'tools'.

`--targets` would be my suggestion.

> I am working on chromecast, and using cquery+compilation database. The compilation database is too big to work well on my laptop (~17G memory).

I see. Why does cquery load the whole file into memory though?

mg94c18:
I faced the same problem recently, and looking at the code I think it actually makes more sense to add a feature to 'targets' tool which would output all the rules that are used by given target.  Then, you could use 'compdb' tool easily by giving it all those rules as arguments.  In my case I could deduce the rules used by my targets from 'rules' tool with -d flag, so I didn't need the changes but here are my two cents: "-t targets rule X" gives you targets that use the rule X (existing feature, though not documented) and "-t targets rules X" could give you rules that are used by target X (possible new feature).

Alternatively, to solve the problem with parameters ([options] [target] [rules] mess), and having in mind that most people would try to put target name instead of rule name in "-t compdb [rules]", I think it would be elegant to just specify a new flag like -t with semantics "treat rules parameter as targets", leading to "-t compdb -t [rules]" which would then interpret the rules arguments differently.

Again, even without all these changes you may be able to discover rules used by your target from "-t rules -d" and then you could use existing compdb feature.

jhasse:
closing in favor of #2319

PR #1549, opened 2019-04-05
Add reading of TERM variable for win32 dumb terminals

jhuels:
This change is definitely a nice-to-have, let me know if controlling output via environment variables is not something you want.  It will help integration with some non-standard windows terminals.  I also prefer the multiple line output.

It is kind of weird using "TERM" on windows though... any other idea?

Would address:
https://stackoverflow.com/questions/46970462/how-to-enable-multiline-logs-instead-of-single-line-progress-logs
https://github.com/vector-of-bool/vscode-cmake-tools/issues/382


jhuels:
1 sec, will reopen when failure fixed.

PR #1557, opened 2019-04-13
Add '--purge' to configure.py to remove files after build

Leedehai:
This option will remove those generated by --bootstrap: 'ninja', 'build.ninja', and directory 'build/'.
This is nice to have because sometimes we want to remove all generated files.


jhasse:
Thanks for the PR! We would like to keep the configure.py as short as possible though and that task can already be handled by other means, e.g. `git clean -xfd`.

Leedehai:
Sure, but not all users check out a Git repo (me). They may just download a zip since they worry a Git directory is huge (they don't know the size before fully downloading).

PR #1558, opened 2019-04-16
make inline.sh more portable

goshhhy:
some implementations of the 'od' utility, in hex mode, will not print a leading 0 if the byte is less than 0x10. this causes the existing sed script in inline.sh to create a file which does not compile.

this patch fixes this issue in a manner that should also be portable to sed implementations which do 
not support gnu extensions.

i have tested it with the following:
* gnu sed and od
* netbsd sed and od
* suckless sed and od

ninja built successfully in each case.

### DIFF
--- a/src/inline.sh
+++ b/src/inline.sh
@@ -20,6 +20,5 @@
 
 varname="$1"
 echo "const char $varname[] ="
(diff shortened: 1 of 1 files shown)


goshhhy:
pushed the wrong file, ignore

PR #1562, opened 2019-04-18
Use GoogleTest instead of our own framework

jhasse:
This replaces the re-implementation of GoogleTest with the real version.

Advantages:

* Shows values when checks fail.
* Less code in Ninja repository.
* Colored output! :D
* IDE integration, e.g. https://marketplace.visualstudio.com/items?itemName=DavidSchuldenfrei.gtest-adapter

Disadvantages, therefore I don't want to merge this:

* ~~Adds GoogleTest as a Git Submodule, which makes it harder for people not using Git to build the tests.~~
* ~~Compiling GoogleTest in `configure.py` is pretty hacky:~~
  * ~~Include paths are added to everything~~
  * ~~`gtest-all.cc` is added with `../` resulting in an object file being created outside of the build directory.~~
* Compiling Unit Tests with configure.py instead of CMake has been removed.
* ~~MSVC no longer compiles because `#define unlink _unlink` doesn't seem to work. No idea why.~~
* ~~Warning when compiling with Clang, see https://github.com/google/googletest/issues/1752~~
* Dependency on GoogleTest in general (longer compile times)
* ~~#1583 is now an issue since it happens when just running `./build-cmake/ninja_test`~~
* ~~GoogleTest now requires C++11, see https://github.com/google/googletest/issues/1519~~


mathstuf:
The MSVC test seems to be due to `build_log_test` having all of its symbols optimized out? Or not generating any symbols at all?

Hi-Angel:
> Adds GoogleTest as a Git Submodule, which makes it harder for people not using Git to build the tests.

Well, I don't think using the submodule is necessary. Most systems has Google Test framework in their package repository.

Hi-Angel:
> Well, I don't think using the submodule is necessary

I mean, it's optional: if one builds from a tar archive, they could just install framework in the system.

Hi-Angel:
Fun fact: b2fe56caaf0bed497ee480003f10486c18d8de9a

jhasse:
> Well, I don't think using the submodule is mandatory. Most systems has Google Test framework in their package repository. So if one builds from a tar archive, they could just install framework in the system.

configure.py won't detect a that.

Hi-Angel:
> configure.py won't detect a that.

But if you want to wipe that point from your list of cons, it can be made to?

jhasse:
> But if you want to wipe that point from your list of cons, it can be made to?

Well people need to install googletest, they didn't have to before. So it makes it harder ;)

Hi-Angel:
> > But if you want to wipe that point from your list of cons, it can be made to?
> 
> Well people need to install googletest, they didn't have to before. So it makes it harder ;)

Compared to initializing a git submodule it's even less text on most systems, e.g.: 
```haskell
git submodule init && git submodule update
```

vs

```haskell
sudo pacman -S gtest
```

Myself I'd sure prefer the latter variant.

jhasse:
1. `git submodule update --init` is enough.
2. Some people might not have root access.
3. GoogleTest can also be out-of-date in the package manager.
4. Package name differs.
5. Windows doesn't come with a package manager by default.

jonesmz:
There are plenty of OS's beyond windows that wouldn't have Google Test at all. I don't know any off of the top of my head, but I can imagine various enterprise-y OSes might not. Perhaps QNX could be an example? It's been a long time since I've used it though.

A configure.py that detects a system version, and uses that instead of a submodule. Or provides CLI switches for the various ways things might be set up, could be attractive though.

jhasse:
Given #1563 I don't want invest too much in configure.py. Please ignore the "added submodule / dependency" disadvantage for now - the other disadvantages block merging this anyway.

jonesmz:
BTW: I would be very happy to see this change.

When debugging some changes I was experimenting with, the custom "google-test-like" test setup that Ninja had made debugging the problem very difficult, especially since it doesn't provide value outputs.

jhasse:
Unfortunately still no activity here: https://github.com/google/googletest/issues/1752

jonesmz:
I don't understand why that GoogleTest bug is a showstopper.

Could you explain in more detail?

jhasse:
We would need to disable the warning if we want to build with `-Werror` on CI (which is good practice).

Also the response (or the lack thereof) in the GitHub issue shows, how much support we will get from Google if we run into other problems with GoogleTest.

jonesmz:
Good point about using -werror.

Can the warning be worked around by explicitly casting, or if comparing against numeric constants, using the 'u' suffix?

Or am I misunderstanding the problem?

jhasse:
Yeah, but that would need to happen GoogleTest itself IIRC.

jonesmz:
Any alternatives that you're fond of?

I use Boost Test, but I imagine that's far too heavyweight for Ninja.

jhasse:
We could try with a PR for GoogleTest first.

I use Boost.Test myself, too. I heard that [doctest](https://github.com/onqtam/doctest) is really nice. Changing the framework would require quite a lot of code changes though.

Hi-Angel:
> We would need to disable the warning if we want to build with -Werror on CI (which is good practice).

A project I work on at work ran into the same problem. We worked that around by disabling warnings for "tests" target, specifically:

```
-Wno-sign-conversion -Wno-switch-default  -Wno-switch-enum
```

it's not so bad of a workaround because the rest of the project still builds with
these 3 warnings enabled.

mathstuf:
> how much support we will get from Google if we run into other problems with GoogleTest.

Google folks have a history with ninja. Maybe we could call on them to help prod that PR along?

jhasse:
GoogleTest seems to have fixed the compiler warnings with 1.8.1 :)

mathstuf:
1.9 is out now as well.

jonesmz:
@mathstuf 

Typo? Latest release is 1.10.

https://github.com/google/googletest/releases/tag/v1.10.x

mathstuf:
Oh, thought I saw a release announcement saying 1.9 somewhere. Guess they're doing the even/odd scheme. Yep, 1.10 :) .

nico:
FWIW we used to use Google Test and moved off of it (see git history). It made the build faster and simpler – several people had build problems with the dep on gtest, given that ninja doesn't have any other dependencies. Unless there's some huge reason for undoing the old change, I don't think we want to do this.

jhasse:
Yeah, see Disadvantages ;)

mathstuf:
If *ninja* is the block in your build pipeline, I'd be envious of you :) . But I highly doubt it is. The problem is that CI can fail and all you get is "something failed", no indication of what the sides of the comparison was. It makes it really frustrating to track down bugs the CI is finding but you can't reproduce locally.

jhasse:
Updated, looking for feedback :)

mathstuf (approved):

CMakeLists.txt
```diff
@@ -249,7 +260,8 @@ if(BUILD_TESTING)
     target_sources(ninja_test PRIVATE src/includes_normalize_test.cc src/msvc_helper_test.cc
       windows/ninja.manifest)
   endif()
-  target_link_libraries(ninja_test PRIVATE libninja libninja-re2c)
+  find_package(Threads REQUIRED)
+  target_link_libraries(ninja_test PRIVATE libninja libninja-re2c gtest Threads::Threads)
```
digit-google, line 264: Thanks for this change.

I found out that this CMakeList.txt file does not work when using the GTEST_ROOT environment variable to point to a local installation (the compiler complains on every `#include <gtest/gtest.h>` line). However, this can be fixed by changing the `gtest` above here into `GTest::gtest` instead.

digit-google: And it seems that using `GTest::gtest` breaks the configure step on GitHub actions (see [1]). Even though it works fine with a locally built 1.10.0 release here. I am so confused :-/ CMake differences maybe?

[1] https://github.com/digit-google/ninja/actions/runs/6786109288

jhasse: Have you tried `GTest::GTest`?

digit-google: It is possible to make all cases work correctly by defining an alias conditionally, i.e.:

```
find_package(Threads REQUIRED)
if (NOT TARGET GTest::gtest)
  add_library(GTest::gtest ALIAS gtest)
endif()
target_link_libraries(ninja_test PRIVATE libninja libninja-re2c GTest::gtest Threads::Threads)
```

Note that `GTest::gtest` is what the official GoogleTest docs are using these days.
I just found this out by trial and error, I don't really understand why CMake's behavior changes here. Any advice welcome :-/

digit-google: Good idea. I just tried `GTest::GTest` locally and it fails in all cases :-(

What is surprising is that even when not using `GTEST_ROOT` (i.e. downloading googletest-1.10.0 and expanding it under build_dir/_deps/googletest-src/), files like `build_dir/_deps/googletest/CMakeFiles/Export/0c08b8e77dd885bfe55a19a9659d9fc1/GTestTargets.cmake` contain definitions for the GTest::gtest target.

However, these are generated when building googletest, but apparently are not available when configuring the build before that. I suspect these are generated by CMake based on input from the googletest build rules.

I am using cmake-3.27.1 by the way.

digit-google: So in other words, when doing a local install of googletest-1.10.0, these CMake files defining GTest::gtest will be copied to the installation directory, then later picked up by find_package() when directed to look there with GTEST_ROOT.

On the other hand, when fetching the package directly from the internet, these files do not exist, and I assume the FetchContent_MakeAvailable() magic can only rely on the top-level library declarations in googletest's own CMakeList.txt files, but does not scope them with `GTest::` (or create an alias like `GTest::gtest` for them).

jhasse: Sounds like a bug in GoogleTest to me. Maybe we could workaround it using `if (TARGET GTest::GTest)`, but I'm not sure if we should.

Out of curiosity: Why are you using GTEST_ROOT in the first place though?

digit-google: Yes, it definitely looks like a GTest bug. I have reproduce the example project from the official GTest documentation at [1], where they use FetchContent_Declare() and FetchContent_MakeAvailable() in the same way as this CMakeLists.txt file, but also links against GTest::gtest.

This works perfectly fine when downloading the archive from the default URL on this page (i.e. https://github.com/google/googletest/archive/03597a01ee50ed33e9dfd640b249b4be3799d395.zip).

However, when I switch it to https://github.com/google/googletest/archive/refs/tags/release-1.10.0.zip, the configure step fails, complaining that there is no GTest::gtest target defined.

Apparently, this has been fixed in 1.11.0 though.

I suggest to conditionally add the ALIAS declaration in the code path that uses FetchContent_Declare() only, and stick to using GTest::gtest as the linked target. This way, updating GoogleTest to a future release will not break the CMake configure step if needed.

[1] https://google.github.io/googletest/quickstart-cmake.html

digit-google: Apart from that, my project (Fuchsia) requires that all prebuilt binaries that we use come from certified and reproducible local CI builds, which must only use sources taken from our own git or archive mirrors, or other "clean" prebuilts generated in the same way. This is both important for legal and technical reasons.

Which means random downloads from the network are not allowed when building these. Hence the need to point to a local installation for dependencies.

digit-google: I have pushed the fix at [1] on my personal github branch, feel free to cherry-pick it onto your branch if you feel so. (My googletest branch is just a rebase of your own on top of current master + my fix, I have no intention to make a PR out of it).

[1] https://github.com/digit-google/ninja/commit/590dd80b2e1c329584ef5789755bdbc0aa2083f3

mathstuf: I would guess that the namespace is only applied at install time. My solution for this has been to make a build-time `ALIAS` target of the same name that will be used when installed and then use that everywhere (that [you can](https://gitlab.kitware.com/cmake/cmake/-/issues/19445)).

```cmake
add_library(foo …)
add_library(Foo::foo ALIAS foo)
```

Issue #1560, opened 2019-04-17
Build on android fails due to missing getloadavg()


goshhhy:
Attempting to build ninja on an android device with clang, python, and the ndk headers and libraries fails due the bionic libc not including an implementation of the getloadavg() function.

Other linux based platforms exist which do not have this function, such as those based on uclibc. This is currently handled in that case.

Suggestion to improve portability is, rather than hardcoding features based on detected / specified platform, detect the existence and usability of needed functions within configure.py. 

This issue was found while attempting to test #1513, and then I attempted a build against master and found the same issue. 

Log from the build is attached.

[ninja-android-build.log](https://github.com/ninja-build/ninja/files/3091034/ninja-android-build.log)

PR #1563, opened 2019-04-18
Add CMake build file as an alternative to configure.py

jhasse:
This has come up [on the mailing list](https://groups.google.com/d/topic/ninja-build/q6FNVe8ziLc/discussion) multiple times. @evmar summed up the pros and cons already:

Pros:
* lets people who have cmake installed (in particular the cmake buildbots) build ninja without python
* more familiar to people who know how to use cmake

Cons:
* build system files listed in two places; any change that adjusts build rules needs to update both
* built executables unlikely to be identical (see e.g. the random cflags/ldflags within configure.py); debugging/diagnosing user issues may require asking where the executables came from

I'd also like to add some other pros:
* Everything is build out-of-tree:
  * `src/lexer.cc` and `src/depfile_parser.cc` don't get overwritten all the time.
  * Build multiple configurations in separate build directories at the same time.
* Faster Bootstrapping with e.g. `make -j$(nproc)`
* Visual Studio project generation, see https://groups.google.com/d/topic/ninja-build/wxjMv8jBfYU/discussion
* More portable / easier to add feature tests as suggested in #1560
* Easier to add dependencies also using CMake, e.g. GoogleTest

I've added [WIP] to the title, as there's a lot of stuff missing and I'm also not a CMake expert. This should also be tested in CI and maybe have a check which compares the binaries created by `configure.py` and CMake to see if they are identical (if that's even possible to achieve).

Then there's the question if this should be added at all (IMHO yes). Discuss ;)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@
 *.pdb
 *.ilk
 TAGS
-/build
+/build*/
 /build.ninja
 /ninja
 /ninja.bootstrap
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,84 @@
+cmake_minimum_required(VERSION 3.1)
+project(ninja)
+
+if(MSVC)
+	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /W4 /GR-")
+else()
+	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-deprecated -fdiagnostics-color")
+endif()
+
+# the depfile parser and ninja lexers are generated using re2c.
+function(re2c IN OUT)
+	add_custom_command(DEPENDS ${IN} OUTPUT ${OUT}
+		COMMAND re2c -b -i --no-generation-date -o ${OUT} ${IN}
+	)
+endfunction()
+re2c(${CMAKE_SOURCE_DIR}/src/depfile_parser.in.cc ${CMAKE_BINARY_DIR}/depfile_parser.cc)
+re2c(${CMAKE_SOURCE_DIR}/src/lexer.in.cc ${CMAKE_BINARY_DIR}/lexer.cc)
+add_library(libninja-re2c OBJECT ${CMAKE_BINARY_DIR}/depfile_parser.cc ${CMAKE_BINARY_DIR}/lexer.cc)
+target_include_directories(libninja-re2c PRIVATE src)
+
+# Core source files all build into ninja library.
+add_library(libninja OBJECT
+	src/build_log.cc
+	src/build.cc
+	src/clean.cc
+	src/clparser.cc
+	src/dyndep.cc
+	src/dyndep_parser.cc
+	src/debug_flags.cc
+	src/deps_log.cc
+	src/disk_interface.cc
+	src/edit_distance.cc
+	src/eval_env.cc
+	src/graph.cc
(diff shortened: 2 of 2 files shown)


mqudsi:
> build ninja without python

I think this makes it worth adding an alternative build system, no questions asked. 

Looking at your list of "other pros" however, I think it's unclear whether the intent here is to simply let people bootstrap ninja itself or ultimately replace the python build system entirely. I have no opinion either way (and am not in a position where my opinion on that deserves any weight) but want to point out merely that it makes a big difference.

If the goal is just to bootstrap ninja, given the extremely few dependencies, a simple `Makefile` would probably be possible (one that is even both GNU and BSD `make` compatible). That would lower the barrier to bootstrapping/building ninja even further, introduce fewer complications and dependencies, reduce the duplication w/ the existing build system, etc.

But if the intent is to open the door to migrating to CMake as the official build system (and why not, `build.ninja` is after all intended to be machine-generated and CMake undeniably offers 1st tier ninja support), then of course that opens the door for all kinds of other rationalizations and makes for a more compelling case over the simple `Makefile` approach.

jhasse:
> [...] simple Makefile would probably be possible (one that is even both GNU and BSD make compatible). That would [...] introduce fewer complications [...] reduce the duplication w/ the existing build system, etc.

Why?

evmar:
Re "Everything is build out-of-tree" -- if you wanted to, you could make the existing system do this too.  I think I checked in the generated lexer.cc to make it easier to build ninja when you don't have re2c installed.

I think otherwise you have captured all the pros/cons well.

One last thing to consider is whether you're "blessing" cmake over other systems.  I kinda liked that ninja demonstrated that it was system agnostic, and also demonstrating how to generate a nice build file.  I worry that cmake generates some suboptimal output for cmake reasons and then users learn from that.  (For example I've seen some tools, not sure if cmake is one, that just create a separate `rule` per build step.)

evmar:
The reason we didn't go the simple Makefile route is that it's pretty troublesome on Windows -- we'd require people to install make just to build ninja.  If you look at ninja's configure.py you'll see there's a unfortunately surprising number of config tweaks.

jhasse:
Another pro:
* Fixes some unconventional behavior of configure.py, e.g. #673

mqudsi:
@jhasse but it really doesn't address any existing issues unless the answer to my earlier question is yes:

> But if the intent is to open the door to migrating to CMake as the official build system (and why not, build.ninja is after all intended to be machine-generated and CMake undeniably offers 1st tier ninja support), then of course that opens the door for all kinds of other rationalizations and makes for a more compelling case [...]

Is the intention to ultimately drop `./configure.py`?

jhasse:
The idea would be to support both, so that if someone wants a conventional way to build Ninja he can use CMake without everyone else having to adjust to what they are used to.

atetubou:
I think this is good as a prepration to use google test because it seems much more useful for ninja's test. ninja's test library has too less verbosity when something failed.
But better to update document about how to build and have coverage of cmake build in CI too?

jonesmz:
See https://github.com/ninja-build/ninja/pull/1562 for using Google Test instead of Ninja

I don't believe there is currently anything for replacing configure.py with a call to CMake

I'm not sure if @jhasse wants to completely remove configure.py, or have both configure.py and cmake as equal choices.

PR #1568, opened 2019-05-02
add c++ 98 build config

atetubou:
This is to check we don't use c++11 feature.


jhasse:
No need, GCC 4.8 already runs in this mode by default. See https://travis-ci.org/ninja-build/ninja/jobs/510697426 for example.

PR #1572, opened 2019-05-13
build_log, deps_log: ignore ENOENT for unlink()

mathstuf:
The file is already gone, so it's not an error.

### DIFF
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -406,7 +406,7 @@ bool BuildLog::Recompact(const string& path, const BuildLogUser& user,
     entries_.erase(dead_outputs[i]);
 
   fclose(f);
-  if (unlink(path.c_str()) < 0) {
+  if (unlink(path.c_str()) < 0 && errno != ENOENT) {
     *err = strerror(errno);
     return false;
   }
--- a/src/deps_log.cc
+++ b/src/deps_log.cc
@@ -353,7 +353,7 @@ bool DepsLog::Recompact(const string& path, string* err) {
   deps_.swap(new_log.deps_);
   nodes_.swap(new_log.nodes_);
 
-  if (unlink(path.c_str()) < 0) {
+  if (unlink(path.c_str()) < 0 && errno != ENOENT) {
     *err = strerror(errno);
     return false;
   }


jhasse:
This would happen if one calls the tool in a directory without a `.ninja_deps` / `.ninja_log` file, right?

mathstuf:
Yes. In making CMake call the tools post-generation, ninja errors that its metadata isn't available.

jhasse:
The question is: Should these tools really succeed if they are called in a folder without a `.ninja_*` file? Or should CMake check if those files exist before calling them?

mathstuf:
Why shouldn't they? They did their job, right? The most salient point is that these errors are for "failed to remove non-existent file". The desired end state has already been achieved.

As for CMake (and other generators) knowing this behavior:

  - Where is it documented that these are the only files ninja cares about?
  - Where is the tool/needed file association documented?
  - Where is `subninja` behavior for these tools documented in relation to build and deps logs?

jhasse:
> They did their job, right?

If I understand it correctly they did nothing: E.g. there's no `.ninja_deps` so nothing to recompact.

> * Where is it documented that these are the only files ninja cares about?
> * Where is the tool/needed file association documented?
> * Where is subninja behavior for these tools documented in relation to build and deps logs?

Not sure, but I think there's no documentation at all for these tools.

mathstuf:
> E.g. there's no .ninja_deps so nothing to recompact.

Yeah, but that's a valid recompaction, no? Note that this error is on *removal*. If the non-existence of the file was a problem, why wouldn't it show up on an `open` call?

jhasse:
> Yeah, but that's a valid recompaction, no?

That's debatable. You could also argue that calling `ninja` without a build manifest is a valid build, too ;)

> Note that this error is on removal. If the non-existence of the file was a problem, why wouldn't it show up on an `open` call?

Because it isn't a problem for a normal `ninja` call, therefore `BuildLog::Load` doesn't propagate the error, see https://github.com/ninja-build/ninja/blob/20b30dac6698d119e7797b34d6ed2c4ed8f48417/src/build_log.cc#L249

mathstuf:
As stated in #1685, an alternative is to document what files must exist for each tool to be run successfully. Is that more palatable?

jhasse:
Yes, that sounds good.

btw: Does CMake ignore an non-zero exit code when calling these tools?

mathstuf:
It looks at the exit code (which, IIRC, is why this was found).

mathstuf:
Pending the merge #1685, this should use the new enum introduced there.

jhasse:
Is this still needed after the second commit of #1685?

mathstuf:
Possibly not. I'll double check when I get the chance.

mathstuf:
Seems that this is still necessary.

mathstuf:
Oops, helps if I test the right binary. Seems to be working without this.

PR #1577, opened 2019-05-20
WIP: Build recently modified files first

### LINKED ISSUE #1570: [RFE] compile modified files first
**Problem**: if a project has 300 cpp files and a single header that's included everywhere, and then someone changed the header and a single cpp file, most likely one wants to know whether changes in these 2 files caused any build errors. However, after executing `ninja -C build`, in the worst case, the changed cpp file may get built after the 299 others, and fail the build.

Ultimately, when rebuilding a project, one is interested *(well, besides of course having the binaries)* in "did my changes cause any problem?" question.

**Solution**: before building, sort files to build by "last changed" timestamp, and compile changed ones first.

Hi-Angel:
Currently targets are built in arbitrary order, which is not at all useful. Building them in "modified first" order would allow to spot errors in files just modified earlier.

Fix #1570


src/build.cc
```diff
@@ -320,7 +320,7 @@ bool Plan::AddTarget(Node* node, string* err) {
 }
 
 bool Plan::AddSubTarget(Node* node, Node* dependent, string* err,
-                        set<Edge*>* dyndep_walk) {
+                        set<Edge*,EdgeComparator>* dyndep_walk) {
```
jhasse: please add a space after ,

src/graph.h
```diff
@@ -217,6 +217,18 @@ struct Edge {
   bool is_phony() const;
   bool use_console() const;
   bool maybe_phonycycle_diagnostic() const;
+
+  // May be actual mtime as well as just 0
```
jhasse: Please use `///` here. IDEs will use the comment as a documentation for the variable then.

src/graph.cc
```diff
@@ -135,6 +135,7 @@ bool DependencyScan::RecomputeDirty(Node* node, vector<Node*>* stack,
       } else {
         if (!most_recent_input || (*i)->mtime() > most_recent_input->mtime()) {
           most_recent_input = *i;
+          edge->most_recent_input_mtime = (*i)->mtime();
```
jhasse: What happens when Edge is already part of `ready_` at this point?

Hi-Angel: Nice question!

Per local experiments, it stays in the same order as it was. I.e. if I iterate over the map after I changed the first element to be last one, it stays the first. Worth noting though that upon asking for the element directly, it will not be found. Out of interest, I'll look if it's possible to re-sort a container — it looks like nobody have ever asked that question.

It doesn't matter though, because in ninja all of the work with mtime happens before adding the edges, so it's not affected.

But of course alternatives are welcome.

src/build.cc
Hi-Angel: Thanks, done.

src/graph.h
Hi-Angel: Thanks, done. I also added another comment, but with double `//` slashes because it documents why the `EdgeComparator` sorts like this. See if you want me to change it to `///`, I've used the `//` because I see such kind of comment in other places uses the double slashes.

src/graph.h
```diff
@@ -217,6 +217,19 @@ struct Edge {
...
   bool maybe_phonycycle_diagnostic() const;
+
+  /// May be actual mtime as well as just 0
+  TimeStamp most_recent_input_mtime;
+};
+
+
+struct EdgeComparator {
+  bool operator()(const Edge* lhs, const Edge* rhs) const {
+    // sort by recently modified files first, and then arbitrarily
```
jhasse: Do you mean this one? Yeah, I don't think `///` would have any effect here.

src/graph.cc
jhasse: My idea:

```cpp
struct EdgeHandle {
  Edge* const p;
  const TimeStamp mtime;
};
```

And use this for `ready_`. This way the compiler will make sure, that an element can't be changed after it has been added.

Still one could add an Edge that's already part of the set a second time ... hm ... also it would make accessing the set pretty cumbersome.

Okay forget the idea :D

Maybe we should just use `std::multiset` instead, so that it becomes apparent that uniqueness isn't guaranteed?

Hi-Angel: > Maybe we should just use `std::multiset` instead, so that it becomes apparent that uniqueness isn't guaranteed?

Problem isn't lack of uniqueness, but that the container relies on the order to search for an element. So after mtime in an Edge was modified, the container needs to be reshuffled, otherwise `find()` method won't find an element. This holds for multiset as well.

Honestly, I think the only way to make this future-proof, is to rewrite a lot of ninja code to provide more immutability guaranteees. In particular to make it explicit that once `Plan` started gathering Nodes/Edges, they won't be modified.

I made a quick stab at constifying the Plan — I'm not sure yet if it's easy, but for sure it would be a lot of changes.

jhasse: Ah I see. Maybe just use a plain old std::vector and do the sorting on insert ourself?

Hi-Angel: > Ah I see. Maybe just use a plain old std::vector and do the sorting on insert by ourselves?

Well, yeah, I can do it. I'd probably need to write a wrapper around `std::vector`, to make sure sorting is always done on `push_back`.

> I made a quick stab at constifying the Plan — I'm not sure yet if it's easy, but for sure it would be a lot of changes.

Yeah, definitely not easy. One would need to decouple various initialization code to make things immutable, just inserting `const` here and there unfortunately doesn't work.

Hi-Angel: I asked on IRC #C++, turns out [keys to an associative container must be immutable](https://stackoverflow.com/a/15701171/2388257#why-not-modify-key-of-associative-container). Mutating them is undefined behavior, so they should ideally be const everywhere.

One of sad consequences of C++ not enforcing immutability by default, having it as an opt-in feature.

Hi-Angel: @jhasse phew, that should be ready now. First commit adds the wrapper and migrates `ready_` there, the second adds mtime to edges and acc. comparisons.

src/sorted_vec.h
```diff
@@ -0,0 +1,93 @@
...
+#include <cassert>
+#include <vector>
+
+/// Modifying a key in associative containers leads to UB. This is an std::set
+/// reimplementation with worse characteristics (such as re-sorting the
+/// container on every push), but one that allows for elements to be
+/// out-of-order (which can happen after a key was modified)
+template<class Elem, class Compare = std::less<Elem> >
+class SortedVec {
+  vector<Elem> data;
```
jhasse: Please move private at the end of the class. https://google.github.io/styleguide/cppguide.html#Declaration_Order

src/sorted_vec.h
```diff
@@ -0,0 +1,93 @@
...
+    return data.size();
+  }
+
+  void insert(const Elem& elem) {
+    data.push_back(elem);
+    sort(data.begin(), data.end(), comp);
+  }
+
+  bool empty() const {
+    return data.size() == 0;
```
jhasse: `return data.empty();`

src/sorted_vec.h
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  void clear() {
+    data.clear();
+  }
+
+  class Iterator {
+    SortedVec<Elem,Compare>& parent_;
```
jhasse: Space after ,

src/sorted_vec.h
```diff
@@ -0,0 +1,93 @@
...
+    assert(!empty());
+    return Iterator{*this, data.begin()};
+  }
+
+  Iterator end() {
+    return Iterator{*this, data.end()};
+  }
+};
+
+#endif // SORTED_SET_H
```
jhasse: wrong comment.

src/build.cc
```diff
@@ -376,12 +376,7 @@ void Plan::EdgeWanted(Edge* edge) {
 }
 
 Edge* Plan::FindWork() {
-  if (ready_.empty())
-    return NULL;
-  set<Edge*>::iterator e = ready_.begin();
-  Edge* edge = *e;
-  ready_.erase(e);
-  return edge;
+  return (ready_.empty())? NULL : ready_.begin().drop();
```
jhasse: Unnecessary ()

src/sorted_vec.h
```diff
@@ -0,0 +1,93 @@
...
+  const Compare comp = Compare();
+
+public:
+  size_t size() const {
+    return data.size();
+  }
+
+  void insert(const Elem& elem) {
+    data.push_back(elem);
+    sort(data.begin(), data.end(), comp);
```
jhasse: This doesn't check for uniquness.

Hi-Angel: Okay. Initially I did the check *(the wrong comment you found above leaked from that time)*, but then I wondered, what wrong could happen if I have a duplicate… But if you think something can happen, sure, I'll add the check.

jhasse: > [...] but then I wondered, what wrong could happen if I have a duplicate… But if you think something can happen, [...]

I also don't know^^ Since you named it SortedVec, you're right: It shouldn't check for uniqueness.

Hi-Angel:
> Might be easier not to introduce a new class, but to use `std::vector` and provide a helper function for insertion?

This is not future-proof, because later in time someone can make use of `push_back` instead of our custom function. And if we don't care about being future-proof then there's no need to do any changes at all, because right now nobody inserts elements to the set after they were modified. I.e. the first patch version *(ofc with comments addressed)* is fine.

>  Or keep `std::set` and sort by mtime when retrieving an Edge to build?

This falls into the same category. I.e. it's not future proof, and if we don't care, then it's already fine. Also, associative containers can not be resorted, because modifying a key after it was added is UB.

Hi-Angel:
> nobody inserts elements to the set after they were modified

err, I meant "nobody inserts elements before modifying them".

jhasse:
> This is not future-proof, because later in time someone can make use of push_back instead of our custom function. And if we don't care about being future-proof then there's no need to do any changes at all, because right now nobody inserts elements to the set after they were modified. I.e. the first patch version (ofc with comments addressed) is fine.

The difference is that having an Edge in `ready_` twice or checking if it's part of `ready_` returning false, although it's there (but with different mtime) can lead to a severe bug. But having someone push_back to the vector will only result in the ordering being wrong which will just break a nice-to-have feature.

> Also, associative containers can not be resorted, because modifying a key after it was added is UB.

We wouldn't resort it, but just remove an element which wouldn't necessarily be the first one.

Hi-Angel:
> > This is not future-proof, because later in time someone can make use of push_back instead of our custom function. And if we don't care about being future-proof then there's no need to do any changes at all, because right now nobody inserts elements to the set after they were modified. I.e. the first patch version (ofc with comments addressed) is fine.
> 
> The difference is that having an Edge in `ready_` twice or checking if it's part of `ready_` returning false, although it's there (but with different mtime) can lead to a severe bug. But having someone push_back to the vector will only result in the ordering being wrong which will just break a nice-to-have feature.

Bugs are bugs. Maybe I'd prefer your option about writing tests for the class. Do I need to just test insertion/erasing and iteration?

src/sorted_vec.h
Hi-Angel: FTR, I tested it: adding a duplicate results in [this assertion](https://github.com/ninja-build/ninja/blob/20b30dac6698d119e7797b34d6ed2c4ed8f48417/src/build.cc#L100). Okay, I'm gonna make SortedVec a SortedSet then.

Hi-Angel:
Note to myself: comments are addressed, gotta make tests.

Hi-Angel:
@jhasse see, I added tests for iteration, dropping an element, deduplication, and ordering. I also removed `erase` method since it's not used anyway. I think mainly this should cover the functional, see if you want anything else.

While on it: I see CI is failing to build because of C++11 things. Do you want me to rework code to make them pass, or is it better to upgrade the CI to newer compilers? IMO in 2019 C++11 sounds pretty old, I don't see why not to support it, but oh well.

Hi-Angel:
> newer compilers

rather, I think just adding `-std=c++11` would be enough.

jonesmz:
The Ninja project won't accept code that can't be built on a list of unspecified ancient operating systems and ancient compilers.

But the general guidelines are that it has to work with the gcc version that is available from the latest Ubuntu 14.04 LTS without providing -std=c++0x on the command line.

Hi-Angel:
As you folks wish.

Lol, I never knew pre C++11 code was so terrible. Somehow, I always worked at least with c++11, the older c++, turns out, doesn't support so many common features… I already reworked it a lot, and it still doesn't pass CI. And I wonder if the last problem, with templates, will result in even more terrible rewrite…

Hi-Angel:
Hmm, for some reason the last CI problem, with template not begin compiled, I can't reproduce locally even when building with `-std=c++98 -pedantic`. It results in lots of warnings from ninja code, but everything compiles correctly.

So either the current or the older gcc is buggy. Not sure what to do about it.

Hi-Angel:
Should pass CI now. Okay, it wasn't that bad, because these changes with `std::less` in 1st commit are getting overwritten in the 2-nd commit anyway.

jonesmz:
You should consider squashing your commits into a single commit to make it easier for reviewers.

jhasse:
No need, GitHub can do that when merging the PR.

jhasse:
Still looking good so far (haven't tested performance impact yet). Needs a test, that recently modified files are build first.

Hi-Angel:
> You should consider squashing your commits into a single commit to make it easier for reviewers.

This would make it harder for reviewers. A separate logical change should go in a separate commit. Squashing everything in a single one would result in a mess, because every next line in the diff may do something different. And later then, reading logs would be easier as well.

> No need, GitHub can do that when merging the PR.

Please don't do this, if there's a regression, again, bisecting would be easier too, and one not gonna guess whether regression was introduced in "sorting by mtime commit" or because of the new data structure.

> Needs a test, that recently modified files are build first.

Thanks, nice idea, I was thinking of that too. But should do I approach that test? I only imaging creating a separate test project, building it with `ninja -nv`, and parsing the output, any better ideas? Any pointers, maybe some existing tests whose approach I could reuse?

jhasse:
> I only imaging creating a separate test project, building it with ninja -nv, and parsing the output [...]

We already have a script for this: `misc/output_test.py`. But I that's too high-level for this.

Have a look at `src/build_test.cc`.

jonesmz:
> This would make it harder for reviewers. A separate logical change should go in a separate commit. Squashing everything in a single one would result in a mess, because every next line in the diff may do something different. And later then, reading logs would be easier as well.

*shrug*. I disagree with your assessment, but this isn't the forum to have a discussion about it. I only mentioned it because you said you have changes in your first commit being replaced by changes in your second commit. Do whatever you think is best.

src/sorted_set_test.cc
```diff
@@ -0,0 +1,59 @@
+// Copyright 2019 Google Inc. All Rights Reserved.
```
jonesmz: Why are you saying that Google owns the copyright for code that Google did not write? The copyright is yours, unless you substantially derived this file from another copyrighted work.

Attributing copyright in this way without a proper contributor license agreement is bad form, and can cause a lot of issues if someone were to ever attempt to enforce the terms of the Apache license for this copyrighted code.

Hi-Angel: Should I write then `Copyright 2019 Konstantin Kharlamov. All Rights Reserved`?

jonesmz: I am not a lawyer, so you're going to have to make that determination for yourself.

But, my non-lawyer advise, is unless someone is paying you to write software (e.g., a work for hire), you are the one who owns the copyright for it.

Hi-Angel: > In this case, this entire file appears to be something you wrote yourself. So unless you copied it from someone / somewhere else, my non-lawyer expectation is that you would be the owner of the copyright for it.

Thanks. I was thinking that contributing a code to a project makes the project to own the code.

jonesmz: My understanding of international copyright law is, obviously, not perfect, but I've discussed it over the years with various lawyers during college, and also in the business world, so I feel reasonably confident in saying that copyright can potentially be handled on a per-line-of-code basis.

The default behavior of copyright is:

Copyright is automatic, and inherent to the creation of a work. No registration needed. The mere act of creating something grants you the copyright for it.
Copyright is not transferred automatically unless there is an explicit agreement otherwise. This might be because someone is paying you (work for hire) or a contributor license agreement where you've explicitly agreed that anything you submit to a project via a pull request is automatically also a full transfer of copyright, so that it becomes their property and not yours.

In this case, you are the owner of the copyright for this file. And if anyone ever breached the license for this copyrighted code (the Apache 2.0 license, specifically), you have the right to personally take them to court for violating that license.

It's important for open source projects to know who the copyright holders are, because "A project" can't sue violators of the license. Only the owners of the copyright can.

It's also important because some licenses don't allow for certain types of re-licensing without explicit permission. I'm not fully up to speed on the Apache2.0 license's limitations on re-licensing, but I know it's very tricky with GPLv2 licensed projects, involving someone needing to track down every, single, contributor, ever, and get their written approval, OR the code in question would need to be removed from the project in order to re-license it.

jhasse: Are you a lawyer?

Hi-Angel: He says he isn't, but he talked with them a lot.

jhasse: GitHub only showed me the first comment somehow ... nvm.

Hi-Angel:
FTR: I'm thinking, maybe I'll try to take a deeper stub at refactoring and constification to make sure Edges are not modified once a plan is created. This would remove a need in a hack with creating a separate data structure.

But that's probably not on this week.

src/graph.cc
atetubou: I think it is not good to use vector for sorted set.
It looks to cause O(n^2) time complexity.

Can we just remove edge from ready_ before modification and re-insert edge after that?

Hi-Angel: > Can we just remove edge from ready_ before modification and re-insert edge after that?

I don't think they're modified after they got into the `ready_`. But this can't guarantee that someone not gonna add the code that does modify them.

jhasse: std::sort is O(n*log(n)) IIRC.

src/sorted_set.h
```diff
@@ -0,0 +1,94 @@
...
+
+  bool has_element(const Elem& elem) const {
+    return std::find(data.begin(), data.end(), elem) != data.end();
+  }
+
+  void insert(const Elem& elem) {
+    if (has_element(elem))
+      return;
+    data.push_back(elem);
+    sort(data.begin(), data.end(), comp);
```
ilyapopov: This is suboptimal. `sort` may require more than O(n) operations here. Something like the following is guaranteed to take O(n) operations:
```suggestion
    std::vector<Elem>::iterator pos = std::lower_bound(data.begin(), data.end(), elem, comp);
    data.insert(pos, elem);
```
You could alse merge it with searching if element already exists, to avoid double searching.

src/sorted_set.h
```diff
@@ -0,0 +1,94 @@
...
+/// out-of-order (which can happen after a key was modified)
+template<class Elem, class Compare>
+struct SortedSet {
+  SortedSet(): comp(Compare()) {}
+  size_t size() const {
+    return data.size();
+  }
+
+  bool has_element(const Elem& elem) const {
+    return std::find(data.begin(), data.end(), elem) != data.end();
```
ilyapopov: [`std::find`](https://en.cppreference.com/w/cpp/algorithm/find) is O(n). You want [`std::binary_search`](https://en.cppreference.com/w/cpp/algorithm/binary_search) here.

src/sorted_set.h
ilyapopov: By the way, `flat_map` and `flat_set` (sorted containers) are scheduled to be included in C++20. Proposal is here https://wg21.link/p1222 See also how reference implementation does `insert` here: https://github.com/WG21-SG14/SG14/blob/d12d8acc59071cfdc44bc5d6ad66d43e1db99d0a/SG14/flat_set.h#L394-L402

mqudsi:
Does ninja have formal performance regression tests (e.g. building Chromium)? It might be a good idea to try them on this..

Hi-Angel:
I'll probably close my PR for now, because wherever I happen to do the refactoring necessary for this to land, it would probably obsolete discussions here. Also, I'm not sure when I will get back to working on this, so there's no point in keeping this PR opened.

Issue #1459, opened 2018-08-14
Feature requestion: Option to use file characteristics instead of timestamps


greened:
As noted in other issues, using timestamps to determine whether to rebuild something can be problematic.  While timestamps are convenient and relatively fast, it is often desirable to key build decisions on some characteristic intrinsic to the file itself, such as a hash of its contents.

I use git a lot and it is annoying that simply changing branches triggers rebuilds.  Ideally, I would be able to switch from my current work branch to some other branch, not touch any files and then switch back to the original branch and not have to rebuild  anything at all.  As far as I know that is not possible if the build system uses timestamps.  Using file hashes would solve this particular issue.

I understand the using file hashes or other such intrisic file characteristics could slow down Ninja so using them should be an option.

metti:
#929 has an implementation. Even though it is successfully used (in a fork) for thousands of builds daily, it was not considered for merging.

jhasse:
#929 is single-threaded and can therefore be slower like `ccache` or other solutions. Also I think that this should be a command line flag instead, so that it doesn't require changes to the build definitions.

metti:
It can't (or at least should not) be a command line flag, as then hashing would apply to all rules. Hashing e.g. all inputs of link rules is expensive and therefore not desired, while source files and their known dependencies are good candidates. For that distinction, it has to be part of the build description. In addition, to make use of the feature you should use the flag consistently all the times. Not just sometimes.

Responding to the single-threaded argument: Yes, it adds instructions to the single-thread loop. In reality that only matters if the single thread gets more work than it can work down (I.e. more rules finish than the one thread can process (depslog+hashlog+...)). Only then hashing hurts. Else the single thread loop waits for jobs to finish anyway. We never saw a busy ninja with hashing even with -j1000 experiments. (And for fast-finishing rules, hashing is anyway not interesting to safe time.)

Also consider: the hashing with murmur hash is considerably fast and even large source files take only some miliseconds to be hashed. In addition the hashing happens right after the source file (and the dependencies) have been seen by the compiler. Therefore they are usually read from filesystem cache.
As hashing happens during the build (in parallel to executed rules), the overall build time is usually not measureably affected.

Lastly, the implementation in #929 is opt-in and comes at no cost for people not using the feature (besides the if-statement).

jhasse:
> Hashing e.g. all inputs of link rules is expensive and therefore not desired, while source files and their known dependencies are good candidates.

[...]

PR #1580, opened 2019-05-24
Prevent unnecessary reconfig by not writing build.ninja to .ninja_log.

donhinton:
This is a fix to #1376.

### DIFF
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -117,8 +117,8 @@ BuildLog::LogEntry::LogEntry(const string& output, uint64_t command_hash,
     start_time(start_time), end_time(end_time), mtime(restat_mtime)
 {}
 
-BuildLog::BuildLog()
-  : log_file_(NULL), needs_recompaction_(false) {}
+BuildLog::BuildLog(const string& path)
+  : BuildFile(path), log_file_(NULL), needs_recompaction_(false) {}
 
 BuildLog::~BuildLog() {
   Close();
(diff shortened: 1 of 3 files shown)


jhasse:
As mentioned in https://github.com/ninja-build/ninja/pull/1527#issuecomment-491315057, I want to fix the bug in a different way.

donhinton:
Will your proposed fix require changes to cmake?

jhasse:
Yes, see https://gitlab.kitware.com/cmake/cmake/merge_requests/3316.

donhinton:
Thanks for the quick response.

I agree your proposal is probably a much cleaner solution, however I’m A
clang/llvm contributor, and we won’t be able to upgrade cmake, currently
using 3.4.3, for a long time, perhaps years.

If you could find an alternative that didn’t require cmake changes, I’d
really appreciate it.

Thanks again for a great tool!

jhasse:
btw: If you don't want to change any flags, you can also build the `rebuild_cache` target.

Why can you upgrade Ninja, but can't upgrade CMake?

donhinton:
I’ll try that, thanks for the hint.

I help maintain the cmake files, and I’ve been burned a few times by using
newer features that aren’t available in 3.4.3 — our buildbots are
maintained by 3rd parties and only get upgraded as needed.

Now I only use the approved version locally so I don’t accidentally
introduce a feature that will break one of the bots.  I’ll be stuck on
3.4.3 for at least another year, and even then, I doubt we’ll upgrade to
the latest and greatest.

jhasse:
I see. You'll be able to call `ninja -t restat` manually after reconfiguring which should prevent the additional reconfig :)

donhinton:
Thanks, but that’s not really a workable solution.   I need something that
works without me having to remember to run some special command.

I’m just going to use my fork which fixes the problem from now on.

Thanks again...

Mis1eader-dev:
Is this not implemented yet?
Personally I have a use case where I copy big C++ projects and perform a `sed` to replace the `./build` directory file contents pointing to the new path, and executing `ninja` on top triggers a reconfigure of CMake and essentially rebuilds the entire thing.

I assume CMake is getting triggered because of the modification time from copying the C++ project directory?

jhasse:
In your usecase it probably happens because the timestamps have changed, see #1459.

PR #1588, opened 2019-05-31
build_log: make sure to carry the old uid and gid

marcelhollerbach:
see #1302 as a normal workflow you do things like "ninja all && sudo
ninja install" for the case that the install target executes a build
target then ninja_log will be writte ntwo which will result in its
permissions to be root / root, which is a problem since you cannot
execute the workflow from above again.

This PR replaces 1362, since it appears that i cannot update 1362 anymore.


marcelhollerbach:
@nico Sooo i moved back to just using the normal codepath when on windows, the CopyFile etc. did not work as i wrote in the previous comments. My testcases have been errornous.

I don't think I can write easily a testcase for this as i would have to make the deps_log and so on so big that the codepath is hit that the file are rewritten.

marcelhollerbach:
As I already wrote in the code itself, this only fixes the issue on linux. Not on windows, it is way too hard to fetch permissions on windows. However i expect that it wont be too bad on windows, as user-switches should be rather unusal here.

src/util.h
```diff
@@ -114,6 +114,8 @@ bool Truncate(const string& path, size_t size, string* err);
 #define PATH_MAX _MAX_PATH
 #endif
 
+bool ReplaceContent(const string& old_path, const string& new_path, string* err);
```
jhasse: Please add a short doc comment.

src/util.cc
```diff
@@ -622,3 +624,51 @@ bool Truncate(const string& path, size_t size, string* err) {
...
+  //apply uid and gid again so we always stay as the uid and gid of the first caller that created a file
+  if (found_uid_gid) {
+    if (chown(path.c_str(), old_file.st_uid, old_file.st_gid)) {
+      *err = strerror(errno);
+      return false;
+    }
+  }
+#else
+  /*
+   * This will not carry over uid / gid stats
```
jhasse: Use `//` two times instead.

src/util.cc
```diff
@@ -622,3 +624,51 @@ bool Truncate(const string& path, size_t size, string* err) {
   }
   return true;
 }
+
+bool ReplaceContent(const string& path, const string& old_path, string* err) {
+#ifndef _WIN32
+  struct stat old_file;
+  bool found_uid_gid = true;
+
+  //get the stat
```
jhasse: Remove this comment (it's obvious from the code)

src/util.cc
```diff
@@ -622,3 +624,51 @@ bool Truncate(const string& path, size_t size, string* err) {
...
+  if (stat(path.c_str(), &old_file) < 0) {
+    //save if we found the stat
```
jhasse: save what?

(also add space after `//`)

src/util.cc
```diff
@@ -622,3 +624,51 @@ bool Truncate(const string& path, size_t size, string* err) {
...
+    *err = strerror(errno);
+    return false;
+  }
+
+  if (rename(old_path.c_str(), path.c_str()) < 0) {
+    *err = strerror(errno);
+    return false;
+  }
+
+  //apply uid and gid again so we always stay as the uid and gid of the first caller that created a file
```
jhasse: Add space after `//` and check line length

src/util.cc
```diff
@@ -622,3 +624,51 @@ bool Truncate(const string& path, size_t size, string* err) {
...
+  if (found_uid_gid) {
+    if (chown(path.c_str(), old_file.st_uid, old_file.st_gid)) {
+      *err = strerror(errno);
+      return false;
```
jhasse: Do we really care if this fails?

src/deps_log.cc
```diff
@@ -353,15 +356,7 @@ bool DepsLog::Recompact(const string& path, string* err) {
...
-  if (unlink(path.c_str()) < 0) {
-    *err = strerror(errno);
-    return false;
-  }
-
-  if (rename(temp_path.c_str(), path.c_str()) < 0) {
-    *err = strerror(errno);
-    return false;
-  }
+  ReplaceContent(path, temp_path, err);
```
jhasse: You're ignoring the return value.

marcelhollerbach:
Thank you for the review, afaics i fixed all problems :)

src/util.cc
marcelhollerbach: I would say yes. It would be a  rare case i hope, but still, we should keep the ownership the same.

src/util.cc
```diff
@@ -622,3 +624,50 @@ bool Truncate(const string& path, size_t size, string* err) {
   }
   return true;
 }
+
+/// ReplaceContent the content of file_dst with new_content.
```
jhasse: s/ReplaceContent/Content/

src/util.cc
```diff
@@ -622,3 +624,50 @@ bool Truncate(const string& path, size_t size, string* err) {
...
+/// The file pointed to by file_dst will keep the same uid and gid.
+/// In case of an error, false is returned and *err will carry the error message.
+bool ReplaceContent(const string& file_dst, const string& new_content, string* err) {
+#ifndef _WIN32
```
jhasse: Use `#ifdef` and swap

src/util.cc
```diff
@@ -622,3 +624,50 @@ bool Truncate(const string& path, size_t size, string* err) {
   }
   return true;
 }
+
+/// ReplaceContent the content of file_dst with new_content.
+/// The file pointed to by file_dst will keep the same uid and gid.
```
jhasse: * not on Windows

src/util.cc
jhasse: I don't know. I think a warning makes more sense, as the file content was still updated and one might not care if the permissions couldn't be preserved when he runs `ninja -t recompact`. Also we didn't care about this previously (and still don't on Windows).

src/util.cc
```diff
@@ -622,3 +624,50 @@ bool Truncate(const string& path, size_t size, string* err) {
...
+  // apply uid and gid again so we always stay as the uid and gid of the first caller that created a file
+  if (found_uid_gid) {
+    if (chown(file_dst.c_str(), old_file.st_uid, old_file.st_gid)) {
+      *err = strerror(errno);
+      return false;
+    }
+  }
+#else
+   /// This will not carry over uid / gid stats
+   /// However, this is only in windows, on linux uid / gid will be preserved
```
jhasse: Please use `//` instead of `///`.

Also explain briefly why we don't do this on Windows (you can link to your comment in the previous PR).

src/util.h
```diff
@@ -114,6 +114,11 @@ bool Truncate(const string& path, size_t size, string* err);
 #define PATH_MAX _MAX_PATH
 #endif
 
+/// ReplaceContent the content of file_dst with new_content.
```
jhasse: s/ReplaceContent/Replace/

Reference parameter names with `\a ` (see other docstrings).

src/util.cc
```diff
@@ -622,3 +624,50 @@ bool Truncate(const string& path, size_t size, string* err) {
...
+  if (!DeleteFile(file_dst.c_str())) {
+    *err = strerror(errno);
```
jhasse: I don't think `strerror` will work for WinAPI calls. Why not keep using unlink?

marcelhollerbach: unlink did not work on the CI because it appears that there is some symbol error.

marcelhollerbach:
I don't know how i should test this behavior, do you have a idea ?

src/util.cc
```diff
@@ -622,3 +624,48 @@ bool Truncate(const string& path, size_t size, string* err) {
...
+
+  if (rename(new_content.c_str(), file_dst.c_str()) < 0) {
+    *err = strerror(errno);
+    return false;
+  }
+
+  // apply uid and gid again so we always stay as the uid and gid of the first caller that created a file
+  if (found_uid_gid) {
+    if (chown(file_dst.c_str(), old_file.st_uid, old_file.st_gid)) {
+      Warning("Reapplying previous uid and gid failed with: %s uid: %d gid: %d", strerror(errno), old_file.st_uid, old_file.st_gid);
```
jhasse: Check line length (tip: use clang-format)

src/util.h
```diff
@@ -114,6 +114,12 @@ bool Truncate(const string& path, size_t size, string* err);
 #define PATH_MAX _MAX_PATH
 #endif
 
+/// ReplaceContent the content of \a file_dst with the content of \a new_content.
+/// The file pointed to by \a file_dst will keep the same uid and gid.
+/// In case of an error, false is returned and \a err will carry the error message.
+/// On Windows, the same gid / uid is not garanteed
```
jhasse: I would merge this sentence with the second one like this: On platforms other than Windows, the file pointed to by \a file_dst will keep the same uid and gid.

typo in garanteed, but that wouldn't matter then ;)

src/util.cc
```diff
@@ -622,3 +624,48 @@ bool Truncate(const string& path, size_t size, string* err) {
...
   return true;
 }
+
+/// Replace the content of file_dst with the content of new_content.
+/// The file pointed to by file_dst will keep the same uid and gid.
+/// In case of an error, false is returned and *err will carry the error message.
+bool ReplaceContent(const string& file_dst, const string& new_content, string* err) {
+#ifdef _WIN32
+  // This will not carry over uid / gid stats
+  // However, this is only in windows, on linux uid / gid will be preserved
```
jhasse: You can remove those two comments as it is now mentioned in the docstring of the function.

marcelhollerbach:
Afair travis only has one user, so we are going to walk into issues there. I will containue testing, maybe we can add some sort of temp user, However I would not wait for me on that, i got a lot of other things to do, but I wanted to see this mainissue fixed here :)

marcelhollerbach:
Okay @jhasse i don't think this is possible. Can we move on with this how it is right now ?

src/util.cc
```diff
@@ -622,3 +624,44 @@ bool Truncate(const string& path, size_t size, string* err) {
...
+    *err = strerror(errno);
+    return false;
+  }
+
+  if (rename(new_content.c_str(), file_dst.c_str()) < 0) {
+    *err = strerror(errno);
+    return false;
+  }
+
+  // apply uid and gid again so we always stay as the uid and gid of the first caller that created a file
```
jhasse: What do you mean by "first caller"?

marcelhollerbach: first caller of ninja

mathstuf: Maybe saying "ninja invocation" instead of "caller" would be clearer wording.

src/util.cc
```diff
@@ -622,3 +624,44 @@ bool Truncate(const string& path, size_t size, string* err) {
...
+    *err = strerror(errno);
+    return false;
+  }
+
+  if (rename(new_content.c_str(), file_dst.c_str()) < 0) {
+    *err = strerror(errno);
+    return false;
+  }
+
+  // apply uid and gid again so we always stay as the uid and gid of the first ninja invocation that created the file at file_dst
```
jhasse: 80 character line limit. (tip: use clang-format)

src/util.h
```diff
@@ -114,6 +114,13 @@ bool Truncate(const string& path, size_t size, string* err);
 #define PATH_MAX _MAX_PATH
 #endif
 
+/// ReplaceContent the content of \a file_dst with the content of \a new_content.
+/// On platforms other than Windows, the file pointed to by \a file_dst will keep the same uid and gid.
```
jhasse: same

marcelhollerbach:
Is there anything left ?

marcelhollerbach:
ping?

jhasse:
I didn't have time to take another look yet.

marcelhollerbach:
Ping ?

src/util.cc
```diff
@@ -622,3 +624,46 @@ bool Truncate(const string& path, size_t size, string* err) {
...
+  if (rename(new_content.c_str(), file_dst.c_str()) < 0) {
+    *err = strerror(errno);
+    return false;
+  }
+
+  // apply uid and gid again so we always stay as the uid and gid of the first
+  // ninja invocation that created the file at file_dst
+  if (found_uid_gid) {
+    if (chown(file_dst.c_str(), old_file.st_uid, old_file.st_gid)) {
+      Warning("Reapplying previous uid and gid failed with: %s uid: %d gid: %d",
```
jhasse: Do you know an example case where this might fail and how an error message might look like?

src/util.h
```diff
@@ -114,6 +114,15 @@ bool Truncate(const string& path, size_t size, string* err);
 #define PATH_MAX _MAX_PATH
 #endif
 
+/// ReplaceContent the content of \a file_dst with the content of \a
+/// new_content. On platforms other than Windows, the file pointed to by \a
+/// file_dst will keep the same uid and gid. Reason for this is that permissions
+/// on Windows are more complex than on Linux, fetching permissions on Windows
+/// would involve an none trivial amount of code to fetch users and have
+/// fallback paths
+bool ReplaceContent(const string& old_path, const string& new_path,
```
jhasse: wrong parameter names

src/util.cc
```diff
@@ -622,3 +624,46 @@ bool Truncate(const string& path, size_t size, string* err) {
   }
   return true;
 }
+
+bool ReplaceContent(const string& file_dst, const string& new_content,
+                    string* err) {
+#ifdef _WIN32
+  if (!DeleteFile(file_dst.c_str())) {
```
jhasse: I would still like to know *why* unlink fails on the CI. I'm having a similar problem in #1562.

marcelhollerbach: I sadly dont know, however, we have a codepath for windows and linux anyways, so i guess this does not really make a difference ...

src/util.cc
marcelhollerbach: I dont have an example, but i think we should tell the user one way or another. This is not returning false or causing any side effect changes to the pure nature of building.

marcelhollerbach:
Okay, I fixed the wrong parameter names, i ran ctest -vv locally just like in the configured github action. However, it does not fail for me ... And the output in github is also not really telling much..

jhasse:
Can you rebase on master? GitHub Actions now runs `./ninja_test` directly, maybe there's more output.

Also keep in mind that GitHub Actions test Linux in a CentOS 7 Docker.

marcelhollerbach:
Okay, i have spent now quite a bit debugging that, i can make it crash in centos7:

```
==6833== Invalid read of size 8
==6833==    at 0x40BEB0: DepsLog::GetDeps(Node*) (in /ninja/build/ninja_test)
==6833==    by 0x443AD1: (anonymous namespace)::DepsLogTestRecompact::Run() [clone .19034] (in /ninja/build/ninja_test)
==6833==    by 0x404A12: main (in /ninja/build/ninja_test)
==6833==  Address 0xa9ed9e0 is 0 bytes inside a block of size 32 free'd
==6833==    at 0x4C2B51D: operator delete(void*) (vg_replace_malloc.c:586)
==6833==    by 0x40BF0C: DepsLog::~DepsLog() (in /ninja/build/ninja_test)
==6833==    by 0x438B99: DepsLog::Recompact(std::string const&, std::string*) (in /ninja/build/ninja_test)
==6833==    by 0x443A86: (anonymous namespace)::DepsLogTestRecompact::Run() [clone .19034] (in /ninja/build/ninja_test)
==6833==    by 0x404A12: main (in /ninja/build/ninja_test)
==6833==  Block was alloc'd at
==6833==    at 0x4C2A593: operator new(unsigned long) (vg_replace_malloc.c:344)
==6833==    by 0x409731: DepsLog::UpdateDeps(int, DepsLog::Deps*) (in /ninja/build/ninja_test)
==6833==    by 0x42FF79: DepsLog::Load(std::string const&, State*, std::string*) (in /ninja/build/ninja_test)
==6833==    by 0x44379F: (anonymous namespace)::DepsLogTestRecompact::Run() [clone .19034] (in /ninja/build/ninja_test)
==6833==    by 0x404A12: main (in /ninja/build/ninja_test)
==6833==
==6833== Invalid read of size 8
==6833==    at 0x40BEB0: DepsLog::GetDeps(Node*) (in /ninja/build/ninja_test)
==6833==    by 0x443BE0: (anonymous namespace)::DepsLogTestRecompact::Run() [clone .19034] (in /ninja/build/ninja_test)
==6833==    by 0x404A12: main (in /ninja/build/ninja_test)
==6833==  Address 0xa9ed9f0 is 16 bytes inside a block of size 32 free'd
==6833==    at 0x4C2B51D: operator delete(void*) (vg_replace_malloc.c:586)
==6833==    by 0x40BF0C: DepsLog::~DepsLog() (in /ninja/build/ninja_test)
==6833==    by 0x438B99: DepsLog::Recompact(std::string const&, std::string*) (in /ninja/build/ninja_test)
==6833==    by 0x443A86: (anonymous namespace)::DepsLogTestRecompact::Run() [clone .19034] (in /ninja/build/ninja_test)
==6833==    by 0x404A12: main (in /ninja/build/ninja_test)
==6833==  Block was alloc'd at
==6833==    at 0x4C2A593: operator new(unsigned long) (vg_replace_malloc.c:344)
==6833==    by 0x409731: DepsLog::UpdateDeps(int, DepsLog::Deps*) (in /ninja/build/ninja_test)
==6833==    by 0x42FF79: DepsLog::Load(std::string const&, State*, std::string*) (in /ninja/build/ninja_test)
==6833==    by 0x44379F: (anonymous namespace)::DepsLogTestRecompact::Run() [clone .19034] (in /ninja/build/ninja_test)
==6833==    by 0x404A12: main (in /ninja/build/ninja_test)
==6833==

```

I don't know how i can cause this. I replaced here locally the content of ReplaceContent with the exact code that was before in Recompact, and it crashes. However, when i simply copy:

```
  const string& file_dst = path;
  const string& new_content = temp_path;

  struct stat old_file;
  bool found_uid_gid = true;

  if (stat(file_dst.c_str(), &old_file) < 0) {
    found_uid_gid = false;
  }

  if (unlink(file_dst.c_str()) < 0) {
    *err = strerror(errno);
    return false;
  }

  if (rename(new_content.c_str(), file_dst.c_str()) < 0) {
    *err = strerror(errno);
    return false;
  }

  // apply uid and gid again so we always stay as the uid and gid of the first
  // ninja invocation that created the file at file_dst
  if (found_uid_gid) {
    if (chown(file_dst.c_str(), old_file.st_uid, old_file.st_gid)) {
      Warning("Reapplying previous uid and gid failed with: %s uid: %d gid: %d",
              strerror(errno), old_file.st_uid, old_file.st_gid);
    }
  }
```
Into Recompact, then its working just fine. Do you have any ideas what can cause this ? 
Additionally: the Release profile seems to cause this.

I am really not understanding how this can all happen :(

src/util.cc
```diff
@@ -630,3 +632,46 @@ bool Truncate(const string& path, size_t size, string* err) {
...
+  }
+#else
+  struct stat old_file;
+  bool found_uid_gid = true;
+
+  if (stat(file_dst.c_str(), &old_file) < 0) {
+    found_uid_gid = false;
+  }
+
+  if (unlink(file_dst.c_str()) < 0) {
```
jonesmz, line 656: This provides a window of opportunity where the file named file_dst is not on the filesystem.

Better would be to take advantage of the posix filesystem guarentee that renaming a file into the name of an already existing file, so long as the destination file can be modified and/or deleted by the current user, is an atomic operation.

So you would want to first create the new file, and give it it's contents. Then change it's attributes, and then rename.

e.g.

```
stat
write_file
change new file attributes
rename
// no unlink
if rename fails, delete new-file before erroring out, to avoid litter.
```

src/util.cc
```diff
@@ -630,3 +632,46 @@ bool Truncate(const string& path, size_t size, string* err) {
...
+
+bool ReplaceContent(const string& file_dst, const string& new_content,
+                    string* err) {
+#ifdef _WIN32
+  if (!DeleteFile(file_dst.c_str())) {
+    *err = GetLastErrorString();
+    return false;
+  }
+
+  if (rename(new_content.c_str(), file_dst.c_str()) < 0) {
```
jonesmz, line 644: Probably this should be done via

https://docs.microsoft.com/en-us/windows/win32/fileio/deprecation-of-txf?redirectedfrom=MSDN
https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-replacefilea

to provide the same atomic replacement semantics as on posix, but also with ACL information copied over.

It also preserves the creation time, but i think updates the modification time like we would want it to.

This API's been available since Windows XP, as far as I can tell from the documentation page.

jonesmz: But if you don't want to do that, the original file shouldn't be deleted or renamed until after the original file is on disk regardless, so the unlink of the original should happen last

src/util.cc
```diff
@@ -630,3 +632,46 @@ bool Truncate(const string& path, size_t size, string* err) {
...
+    *err = strerror(errno);
+    return false;
+  }
+
+  // apply uid and gid again so we always stay as the uid and gid of the first
+  // ninja invocation that created the file at file_dst
+  if (found_uid_gid) {
+    if (chown(file_dst.c_str(), old_file.st_uid, old_file.st_gid)) {
```
jonesmz, line 669: Does this also pull over extended attributes?

jonesmz: This has time-of-check / time-of-use issues.

The new file might be replaced with an imposter file between when the `rename()` function is called, and the `chown()` function is called.

Better would be to work with a filehandle instead of a string-path.

jerome-pouiller: > Does this also pull over extended attributes?

no. It won't.

jhasse:
@marcelhollerbach It's been a few years, but I still think this should be merged. Do you have the time to solve the conflict and have another look?

jerome-pouiller:
Superseded by #2697

PR #1597, opened 2019-07-09
Die on parent process death

code-complete:
Motivation
On linux, QtCreator launches ninja as cmake build tool and when stop button is pressed in QtCreator then cmake process dies, but ninja process doesn't stop.
This patch fixes this problem under linux by subscribing on parent death event via prctl.

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -30,6 +30,11 @@
 #include <unistd.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


jhasse:
This looks like a Qt Creator or CMake bug. Either Qt Creator should kill the whole process tree or CMake should forward SIGTERM.

Issue #1499, opened 2018-11-15
Consider adding a second, "heavy" binary


nico:
ninja as-is works reasonably well for many projects, and it's reasonably small and simple.

But it doesn't work great for all use-cases, and there are many pull requests for adding "heavy" dependencies (e.g. make jobserver support #1139 / #1140, a persistent daemon #1389 / #1438, some configurable frontend thingy with a message-passing interface #1210).

One half-baked idea would be to make the core ninja code a "real" library, and then keep the regular ninja binary reasonably small and focused, and then add a second binary somewhere (contrib/monster or what have you) that links to the ninja library and adds all these features. That binary could then depend on all kinds of stuff.

I haven't thought this through, but I figured I should at least write it down :-)

bobzhang:
> One half-baked idea would be to make the core ninja code a "real" library, and then keep the regular ninja binary reasonably small and focused, and then add a second binary somewhere (contrib/monster or what have you) that links to the ninja library and adds all these features. That binary could then depend on all kinds of stuff.

This would be cool. FYI, we used ninja in the underlying bsb build system for ReasonML/BuckleScript projects, in general, we enjoy the performance of ninja, but it still feels a bit limited without a persistent mode, I understand it might introduce more deps or make it less cross platform after introducing more features, but it makes sense to export ninja as a library

jhasse:
> some configurable frontend thingy with a message-passing interface #1210

That PR doesn't add any runtime dependency at all. What exactly is it that you're concerned about there?

Because if it's Git repository/distribution size or maintainability, having a second binary would hardly help.

Leedehai:
I'd like to caution that many projects using Ninja do not expect the users to have ninja pre-installed on their machine, and the projects provide a wrapper script that download Ninja binary in the project tree when Ninja is needed. If Ninja is no longer a standalone executable, it will break those projects.

Therefore, producing a standalone executable is always needed.

EliRibble:
I'm interested in pushing this forward. My current thought it so work on making the main() function for ninja very small and pushing all the current functionality inside ninja.cc into a library boundary. I'm going to try sketching out some of these ideas in code to see if we can turn that into a discussion on the various technical tradeoffs.

johan-boule:
[...]

PR #1600, opened 2019-07-12
Add support for external frontends, take 2

EliRibble:
This is based on https://github.com/ninja-build/ninja/pull/1210 by collincross.

There are many requests for customizations to Ninja's build output cataloged in [#746](https://github.com/ninja-build/ninja/issues/746). This patch set attempts to address the underlying issue by spawning a separate frontend process and feeding serialized build status updates to it. The frontend can be extended without complicating Ninja's core logic, or can be completely swapped out for a custom frontend like implemented in [#1020.](https://github.com/ninja-build/ninja/pull/1020) The output can also be serialized to disk for later parsing, which will work well for buildbots.

The interface between Ninja and the frontend is documented in FRONTEND.md in 7th patch in the series, "Add frontend option for build status output"

This is an attempt to get external frontends landed in a way that is in harmony with ninja's policy of being conceptually simple. This means that the ninja binary will not depend on any implementation of protobufs. This PR will adds a new binary that has the ability to take a '--frontend' argument that passes protobuf messages and allows arbitrary logic for showing progress and messages to the user.


EliRibble:
This PR is going to need some more work before it has fully achieved its purpose. From the original PR, [#1210](https://github.com/ninja-build/ninja/pull/1210), there was the following post from nico:

> I thought a bit about this. As said on the mail thread, I'm not a fan of putting protobuf in ninja itself (even if it's a reimplementation). At the same time, I agree that lots of people want to tweak ninja's output in many ways, and this is a good approach to attack that problem.
> 
> What do you think about adding a second binary, ninja_protoout or ninja_buildcore or what have you, somewhere below contrib/, make that link to the ninja lib, put the proto code there, and make the ninja lib have an interface for selecting the output behavior (and core ninja would provide a class that works like the current output and the new binary would instead use the proto behavior)? That way, ninja itself stays the way it is, and people who want very custom output can use the new binary for that?

I have not yet done that work. Before I do I'd like to actually understand the requirements. This is what I believe needs to be done:

1. The ninja binary cannot have compiled into it any implementation of protobuf communication. For this PR, that means `proto.h|proto.c`.
2. The current code in this PR creates a `StatusSerializer` or a `StatusPrinter` in `real_main` based on the presence of a `--frontend` parameter. Instead we'll need do work to break up `real_main` into code that can be shared and create two different source files, one that will be used by `ninja` and will always provide a `StatusPrinter` and another that will be used by `ninja_buildcore` and can provide `StatusPrinter` or `StatusSerializer`.
3. There may be packaging work to add `ninja_buildcore` to releases.

Open questions:

Do I need to create a `contrib` directory in the root of the repository and put all protobuf-related source into it? If not, what does "somewhere below contrib/" mean?

Is there a preference in sharing the logic in `ninja.cc` between keeping most of the shared code within `ninja.cc` or moving shared code to a new source file?

jhasse:
Thanks! I've also done a rebase of this some some time ago: https://github.com/jhasse/ninja/tree/serialize Never managed to create a PR from it though. What else did you change beside rebasing?

> From the original PR, #1210, there was the following post from nico:

In the post you've quoted nico is asking questions. I don't think that the final word about this has been spoking, so I wouldn't start on any work yet.

EliRibble:
I haven't changed anything yet aside from rebasing. That's because I agree, the final word has not been spoken on how this should be done.

jonesmz:
@EliRibble Could you take a look at the discussion here? https://github.com/ninja-build/ninja/pull/1614

If that pull request was merged, would you be willing to rebase this pull request against it?

src/status.h
```diff
@@ -0,0 +1,141 @@
...
+/// completion fraction, printing updates.
+struct Status {
+  virtual void PlanHasTotalEdges(int total) = 0;
+  virtual void BuildEdgeStarted(Edge* edge) = 0;
+  virtual void BuildEdgeFinished(Edge* edge, bool success, const string& output,
+                         int* start_time, int* end_time) = 0;
+  virtual void BuildLoadDyndeps() = 0;
+  virtual void BuildStarted() = 0;
+  virtual void BuildFinished() = 0;
+
```
jonesmz: Unless you plan to somehow prevent deleting a `Status*`, you need to have a virtual destructor. Otherwise, the destructor for derived classes won't be called.

jonesmz: If you do plan to prevent deleting a `Status*`, and only allow deleting derived classes, then you should define a non-virtual destructor, and make it `protected` so that Status cannot be directly deleted.

src/status.h
```diff
@@ -0,0 +1,141 @@
...
+  virtual void PlanHasTotalEdges(int total);
+  virtual void BuildEdgeStarted(Edge* edge);
+  virtual void BuildEdgeFinished(Edge* edge, bool success, const string& output,
+                         int* start_time, int* end_time);
+  virtual void BuildLoadDyndeps();
+  virtual void BuildStarted();
+  virtual void BuildFinished();
+
+  virtual ~StatusPrinter() { }
```
jonesmz: No reason to have this. The virtual destructor that you add to `Status` is sufficient.

src/status.h
```diff
@@ -0,0 +1,141 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef NINJA_STATUS_H_
+#define NINJA_STATUS_H_
+
+#include <map>
+#include <string>
+using namespace std;
```
jonesmz: Considering the proposed change to the "using namespace std" convention, found here: https://github.com/ninja-build/ninja/pull/1616

You might consider eschewing the "using namespace std" in this newly added file.

src/ninja.cc
```diff
@@ -1299,6 +1298,8 @@ NORETURN void real_main(int argc, char** argv) {
     exit((ninja.*options.tool->func)(&options, argc, argv));
   }
 
+  Status* status = new StatusPrinter(config);
```
jonesmz: Dealing with raw pointers is prone to memory leaks.

Consider using, when compiled for C++03, std::auto_ptr, and when compiled for C++11 and newer, std::shared_ptr.

src/status.cc
```diff
@@ -250,3 +251,24 @@ void StatusPrinter::PrintStatus(Edge* edge, EdgeStatus status) {
                  force_full_command ? LinePrinter::FULL : LinePrinter::ELIDE);
 }
 
+void StatusPrinter::Warning(const char* msg, ...) {
```
jonesmz: Hrmm...

Is using va_list really the only way?

I know we can't use C++11 parameter packs, but would it be practical to have an ifdef that uses a template parameter pack when C++11, and use va_list otherwise?

src/status.h
jhasse: No need for protected, deleting Status* without a virtual dtor will result in a warning.

src/ninja.cc
jhasse: Better to avoid auto_ptr. I would keep the raw pointer for now.

jonesmz:
Consider re-arranging the patch series such that the addition of the protocol buffer implementation happens as the second-to-last patch.

src/status.cc
jhasse: #ifdef #else with two code paths isn't a good idea, because only one of them will be tested by most people.

jonesmz:
I'm sorry for bringing this up so late in the process, but I've only recently started looking at pull requests for the Ninja project.

While I don't really have any objections (beyond the minor technical things I pointed out) to the implementation of this approach, I don't think that this is the right way to accomplish the stated goal. 

My understanding is that the objective here is to make it easier for Ninja to be customized, with regards to it's output, and the formatting thereof.

I think that a much more straight forward way to accomplish this goal is to make it easier to consume ninja as a library, and offer a simple C-API that can be called by other languages via a foreign-function-interface.

The method that's being employed here, where it's *ninjas* job to explicitly execute another program to do the display work, just seems very weird to me.

On one hand, the goal is to keep ninja from growing complex when supporting complex formatting behavior, which is a great objective. But doing that by adding a protocol buffers implementation (Not a trivial data format, by any stretch) simply doesn't seem like the right strategy.

If you really don't want to support library-ifying ninja to accomplish this, an alternative proposal is to have Ninja produce a JSON output. This can then be piped to another program, without Ninja needing to be aware of the program in question, and allowing for humans to copy-paste the json output as they see fit.

JSON is trivial enough to write an encoder by hand (and a decoder, if you really want to), is flexible enough to handle basically any data type, is human readable, and does not require an external library (or a part thereof). JSON is supported naively by most web-related things, and has some form of very well established library in every programming language i've ever heard of.

Better yet, there's no need for the machinery of a protobuf .proto file to be implemented.

Ultimately I like everything about this patch set, except two things. 1) using protobuf, and 2) Piping things to a program that ninja has to execute. 

At my last job, i used Ninja in a custom build system (With ninja replacing NMake, no less. ugh) that built several million lines of code. There were multiple situations where I would have liked better programmatic access to ninja's internal state to display debugging info on the screen, but I would not have enjoyed needing to write a little sub program that would receive a protocol buffer and then dumped the info to the console. In that situation, i would have parsed the protocol buffer, and converted it to json to be dumped onto stdout, so that my actual build system could capture stdout, and understand the data.

So I'd like to ask that you consider this approach more carefully.

jhasse:
Let's not discuss Protobuf vs JSON here please.

The library approach is something to consider, requires more work though (might be worth it).

jonesmz:
The reason I brought up JSON instead of protobuf is because JSON (as an example) is human readable, which makes dumping it to the end-user's screen an acceptable thing to do.

This would allow ninja to simply print it's serialized state (in my example, JSON) to the screen.

If the user wanted to have the output formatted differently, then the JSON can be piped to another program, in the conventional way program output is handled on posix like systems.

E.g.

```
ninja --serialized-output | my_custom_frontend
```

vs.

```
ninja --frontend my_custom_frontend
```

In the first example, ninja doesn't know, or care, where its output is being sent.

in the second example, it becomes ninja's responsibility to manage yet-another-thing.


Regardless, making Ninja consumable as a library is the most robust and feature-ful way of accomplishing this goal, and even if this PR gets merged as is, i'd still like to see that happen eventually.

jhasse:
> If the user wanted to have the output formatted differently, then the JSON can be piped to another program, in the conventional way program output is handled on posix like systems.

That change doesn't require JSON, it could also work with protobuf.

The reason for using fd 3 is that jobs in the console pool can then still directly use fds 0, 1 and 2.

jonesmz:
> That change doesn't require JSON, it could also work with protobuf.

The most common things to pipe output to, e.g. grep, sed, awk, etc, do not understand protobuf. These tools work on plain text.

While it's possible for a program that speaks protobuf to receive the protobuf messages via the `|` style pipe, it's not practical to do so with any tool I've heard of. A human readable format (regardless of the format that is chosen, JSON or otherwise) is able to be used with the nearly ubiquitous text processing toolkit that many linux users use regularly, and interactively constructing a specialized custom "front end" becomes possible on the fly (Because the user can just run ninja, and then pipe it to grep and see what they get back), whereas with protobuf, it is not especially easy to get started without knowledge of how to use protobuf and all of it's related gotchas.

Even as someone who's sent patches to a custom protobuf library in the past, it's a fairly intimidating thing to get set up to use. I don't even really remember where to start with it at the moment, and i'd have to do some searching on stackoverflow before doing anything useful.

> The reason for using fd 3 is that jobs in the console pool can then still directly use fds 0, 1 and 2.

That makes sense.

While I don't know the syntax for it, i'm fairly sure it's possible to pipe-operator alternative file descriptors, isn't it?

jhasse:
```
ninja --serialized-output | protoc --decode_raw | your_plain_text_tool
```

And now please stop with the opinion pieces about protobuf. I think everyone is aware of the advantages and disadvantages, if not see https://www.google.com/search?q=protobuf+vs+json

> While I don't know the syntax for it, i'm fairly sure it's possible to pipe-operator alternative file descriptors, isn't it?

As far as I know it isn't possible.

Sarcasm:
I'm really sorry to talk about this too, but as a user I would find JSON more accessible.
For example, if I want to make a Ninja UI inside Emacs, using JSON I would just have to write the UI code.
Using protobuf would just be painful, bigger barrier to entry for what reasons?

I don't think googling "json vs protobuf" is of any help, every tools has pros and cons.
ninja is quite simple to use, its lack of external dependencies is quite convenient for building and distribution.
Adding protobuf for json reporting while a "json line" protocol could do the same job without the protobuf dependency seems weird to me.

I feel like it would be better to explain what are the protobuf benefits over json line for ninja users.
IMHO, you could put "convenience" on the side of JSON.

jonesmz:
@jhasse 

It's really rude of you to tell me what is or is not appropriate to talk about with regard to improvements to a tool I use on a regular basis.

It's dismissive, and results in a casual reader coming to the conclusion that you do not value contributions from others.

It's also inappropriate in that you're clearly missing my point. I am not *trying* to discuss protobuf versus JSON. If you dismiss people's architectural advice based on the example format that they use to demonstrate the problems with the design, as examples, then you're going to dismiss a lot of valuable feedback for the wrong reason.

What I am *trying* to point out that using a non-human readable format introduces additional hurdles that the end user of the tool has to go through, while simultaneously being a poor architectural choice.

I am using JSON as an example. I don't *really* care what the actual format is.

My objection here is that the design of this functionality is *not* the right way to approach this feature.

As a sophisticated end user, who's touched nearly every line of code in the ninja codebase in my fork, and who used Ninja in the very recent past as a core component of a multi thousand line custom build system employed on a multimillion line C++ codebase, I think what I have to say is actually pretty important.

I'm telling you, this feature is *not* useful to any project I've ever been involved with. Having ninja call a sub-program is not an appropriate action. It introduces architectural hurdles that the consumer of Ninja's output have to deal with.

Further, as a second concern, using a non human readable format, such as protobuf, is also inappropriate, because

1) It adds a significant amount of code to ninja, and nico and evmar have both regularly rejected features that added much less code, while providing much more direct ease-of-use (E.g. see the environment variable for CLI arguments discussion)
2) It adds a burden on other users of Ninja's output that alternative implementations would not incur.

You say to use `protoc`, but don't explain what I should do if there is no `protoc` on my platform. JSON, as in my example, is ubiquitous because it's so trivial to write a parser for it. `protoc` is not. So you're adding functionality that is of questionable ability to be *actually used* by your users.

I mean, hell, ninja has a copy of the `getopt` function in it's codebase because there are apparently platforms that don't have the `right` version of this function. If you're going to include a protobuffer implementation in the ninja codebase, then get rid of the internal copy of `getopt`, because you're throwing out all pretenses of multi-platform support.

In summary:

This Pull Request is:

1) Architecturally unsound
2) Directly imports a large amount of code, when previous contributions to ninja have been rejected with the amount of additional code being a significant justification for the rejection
2) Involves adding a burden to the end user of needing additional programs installed on their platform that they can't reasonably write by hand.

Using JSON (or something similar) solves problems 2 and 3.

Using JSON (or something similar) does not solve problem 1.

So, as I think is clear, I am not discussing the merits of protobuf versus JSON, I am discussing the benefit of specific implementation choices and how they effect me. It just so happens to be the case that protobuf is a poor choice, but it is certainly not the only aspect of my feedback.

src/ninja.cc
jonesmz: Please elaborate why it's better to avoid using auto_ptr?

auto_ptr is used several places in Ninja's codebase.

src/status.cc
jonesmz: The real answer here is to adopt C++11.

My usage of Ninja will only ever use the template-parameter-pack implementation. Using va_lists is not type safe, and is prone to multiple issues that template-parameter-packs resolve.

src/status.h
jonesmz: You cannot rely on a compiler to issue a diagnostic for this. Not only will some compilers simply not issue a diagnostic for this at all, but even compilers that do may not be able to issue a diagnostic for this in every situation due to AST transformations losing the context of the delete operation.

Using the protected keyword guarantees that the code will not compile if Status* is deleted directly.

Rely on official ISO C++ standard behavior. Not implementation whims of the compiler vendor.

jhasse: > but even compilers that do may not be able to issue a diagnostic for this in every situation due to AST transformations losing the context of the delete operation.

Can you give an example?

jonesmz:
Note:

Aside from the commit that adds the protobuf code, and the last commit that employs the protobuf code to serialize the state into a FD, I'm perfectly happy with all of the changes from this PR.

I'd be happy to see those changes split out into their own PR, which I would be willing to very carefully review and help test.

jonesmz:
Further, you yourself said on the mailing list that you don't think it's a good idea to support this kind of behavior in the Ninja codebase: https://groups.google.com/forum/#!topic/ninja-build/49UVk6WLqU0

The real answer to these desires is to make Ninja a real library: https://github.com/ninja-build/ninja/issues/1499

src/status.h
jonesmz: Of *this specific* situation? No.

But I've seen first hand several times where GCC, or MSVC, would properly warn about a problem with a particular piece of code, and then a change to another function that happened to call the function being warned about resulted in the warning going away.

The only conclusion that I was able to come to when I investigated further is that the compilers decision to inline some aspect of the functions in question resulted in the diagnostic not triggering due to the specific implementation of the diagnostic not being aware of the inlining behavior that was happening.

Regardless, my point is that diagnostics are not standard. Relying on them (e.g. with -Werror) to ensure a particular usage of a part of the code is not allowed is not what the diagnostics are intended for, but there is an explicit, standard defined, mechanism that can be used to ensure this interface class cannot be mis-used.

Ninja should prefer the explicitly standard defined mechanisms over trusting that the compiler vendor implemented a diagnostic, or implemented it for all situations.

jhasse: I disagree.

jonesmz: You... Disagree about what, specifically?

jhasse: > Ninja should prefer the explicitly standard defined mechanisms over trusting that the compiler vendor implemented a diagnostic, or implemented it for all situations.

At least in this particular case.

src/status.cc
jonesmz: Justify your position please.

You are ignoring the standard mechanism to prevent bad behavior, which costs 2 lines of code, in favor of relying on arbitrary functionality offered by some compilers, some times, with no guarantee of working, and no guarantee that it'll continue working in future releases.

jhasse: ???
I did:

> because only one of them will be tested by most people.

jonesmz: I apologize. This is not the section that this comment was intended to be added to.

I don't understand how this happened.

src/status.h
jonesmz: This is where the comment that ended up in the other section was intended to go.

> Justify your position please.

> You are ignoring the standard mechanism to prevent bad behavior, which costs 2 lines of code, in favor of relying on arbitrary functionality offered by some compilers, some times, with no guarantee of working, and no guarantee that it'll continue working in future releases.

jhasse: > some times,

How do you know that?

jonesmz: Its not a diagnostic that is mandated by the standard. The burden of proof that this diagnostic will be available is on you, not me.

jhasse: See GCC's or Clang's compiler manual.

jonesmz: So?

What about MSVC? ICC? Other compilers that I don't know about.

With just two lines of code. We can solve this problem in a manner that is guaranteed by the standard.

Why are you against that?

jhasse: I never said that I'm "against" it, just that there is no need.

jhasse:
@jonesmz

Sorry, I definitely should have phrased that differently in a not-so-dismissive way.

The reason why I asked to stop the anti-protobuf arguments was NOT because I don't value them. It's because I already know them. The same goes with the other points you've mentioned. I'm well aware and I never said that I don't agree with you - this PR won't be merged any time soon, so no need to panic or object.

@Sarcasm

> I feel like it would be better to explain what are the protobuf benefits over json line for ninja users.

I think the general pro-protobuf arguments apply.

jonesmz:
@jhasse 

> this PR won't be merged any time soon, so no need to panic or object.

Ok. Since this PR is one of the main justifications for delaying this PR #1614, it sounds like there is no longer any reason to wait to merge #1614.

Please go ahead and merge #1614, so that we can get closer to solving #1499 

I will personally resolve merge conflicts on any other PR that you ask me to for a period of 30 days after #1614 is merged, including this one.

jonesmz:
@jhasse 

> this PR won't be merged any time soon, so no need to panic or object.

Ok. Since this PR is one of the main justifications for delaying this PR #1614, it sounds like there is no longer any reason to wait to merge #1614.

Please go ahead and merge #1614, so that we can get closer to solving #1499 

I will personally resolve merge conflicts on any other PR that you ask me to for a period of 30 days after #1614 is merged, including this one.

colincross:
I'm the original author of these changes from #1210.  Not sure why this needed to be split into a new pull request, and I'd prefer that you leave me as the author in Git when making minor changes to my patches.

As for JSON vs. protobuf, I would much rather have used JSON, but JSON is actually terrible at passing arbitrary data.  The data must be valid UTF8, which precludes some arbitrary binary sequences that might be found in the output of rules unless you base64 encode everything, which is both inefficient and loses the human-readability advantages of JSON.  I originally wrote this with UBJSON instead of protobuf, but that loses most of the advantages of JSON without providing the advantages of protobuf (library support in multiple languages, easy cross-version compatibility).  I ended up having to write rules on how to rev the schema without breaking compatibility with existing frontend, and there were no good tools to enforce the schema.  So I went with protobuf instead.

I'm happy to continue working on this PR if there is a path forward, but it is merged as-is to the fork of ninja used in the Android build system and has been working great for us there.  It has enabled us to store multiple output streams with [different levels](https://ci.android.com/builds/submitted/5780832/aosp_arm64-eng/latest/view/logs/build.log) of [logging](https://ci.android.com/builds/submitted/5780832/aosp_arm64-eng/latest/logs/verbose.log.gz), collect the error messages into a separate file, and produce an accurate [build trace](https://ci.android.com/builds/submitted/5780832/aosp_arm64-eng/latest/logs/build.trace.gz) that can be loaded in chrome://tracing. Recently we have turned on local build output using a status table instead of a single status line that shows the description of the 8 longest currently running edges.  There are more features planned including streamed build progress from our CI and colorized logs output from CI.

jonesmz:
@colincross 

Would you consider opening a PR that has all the changes from this PR, excluding the protobuf specific changes?

I think that 

- Turn BuildStatus into an interface
- Add unique IDs to edges
- Put builder output through status interface
- Move edge time from Status to Builder
- Simplify running edges status

Are all positive changes, as is.

If those get merged, your work patching the protobuf functionality into ninja going forward decreases, and the protobuf changes can be considered in isolation.

Sarcasm:
> JSON is actually terrible at passing arbitrary data. The data must be valid UTF8, which precludes some arbitrary binary sequences that might be found in the output of rules unless you base64 encode everything, which is both inefficient and loses the human-readability advantages of JSON.

I agree it's technically possible to have arbitrary byte sequences, that contains binary data, and ninja should support it, but I don't think everything needs to be base64 encoded.

Looking at the src/frontend.proto, I see the following `string` fields (correct me if I'm wrong):

    EdgeStarted#inputs
    EdgeStarted#outputs
    EdgeStarted#desc
    EdgeStarted#command
    EdgeFinished#output
    Message#message

I'm not sure all theses can contain binary data, but if ninja had to support if for all these fields, it should be possible to just add something like `<field>_format` with a value of `base64` or something like that, e.g.:

```json
{
  "type": "EdgeStarted",
  "inputs": ["Zm9vLGNwcA=="],
  "inputs_type": "base64",
  "outputs": ["foo.o"],     # outputs_format can be skipped if it's utf-8
}
```

JSON has way more support than protobuf and is supported by default by a lots of programming languages or environments (Python standard library, Emacs Lisp, frameworks like Qt, ...).
While ninja can content itself to write a light version of the protobuf serialization and add the new features it needs bit by bit, I expect the protobuf consumers will often try to support all of protobuf unless the ninja subset is stable and publicly documented.

The C++ Module folks have a similar issue to deal with non-utf8 text and it looks like JSON is not a showstopper for that http://www.open-std.org/pipermail/tooling/2019-March/000553.html, I expect most real-life files will be human readable, so it seems a good compromise.

I'm not sure I want to write this, but really my Emacs integration example is interesting.
With a protobuf API, I would have to build either:
1. a protobuf deserializer in Emacs Lisp (because no one exists), I suspect it would take more code and time than to make the ninja UI itself
2. or depend on an external tool that to do the conversion to an intermediate format

While with JSON, there is built-in support, with efficient built-in support.
If it's just a simple integration just for my comfort, I can even skip the handling of non-utf-8 data.
I could easily test it with fake data that I write by hand.

---

Another thing, regarding the frontend that is invoked by ninja.
I also expected more an API like @jonesmz  feel like it would be easier to have an API like this:

    ninja --serialized-output

With output going to stdout or stderr.
And if user want confidence that the data is not mangled with other outputs, make it configurable, e.g.:

    ninja --serialized-output --serialized-output-fd=3

That way, it seems easier to integrate in a tool like Emacs, because if I wanted to integrate the ninja build inside my Emacs session, with the current design, it looks like I would to write a frontend script, that forward the data to Emacs, so I would need 3 processes, while Emacs could maybe just invoke ninja with the rights params and consume the additional reporting stream.

---

Sorry to bring Emacs up, but I feel like editor integration is a reasonable example of a custom frontend, and right now the API does not seem easy to support that.

The other idea of a frontend I have in mind, is to make a python frontend that would make a D-Bus progress notification, so I can start the build in a workspace, go to another workspace, e.g. in my browser, and still track compilation progress.

EliRibble:
@colincross 

> I'm the original author of these changes from #1210. Not sure why this needed to be split into a new pull request, and I'd prefer that you leave me as the author in Git when making minor changes to my patches.

My apologies, I didn't realize that was something that could be done in Github. I wanted to try to move your ideas forward without requiring you to do additional work. I tried to point out as much as I could that the work is almost entirely yours. If there's something you'd like me to do to make this right please let me know.

EliRibble:
As far as the discussion here goes. I tend to agree with the idea that making ninja a library rather than making it start up a separate process to act as its frontend makes sense. I'll going to start looking in to what it would take to do that. I also think this PR could be modified to remove the (controversial) protobuf work to get the benefit of some of the changes while the discussion continues. I'll try to make that happen.

jhasse:
> My apologies, I didn't realize that was something that could be done in Github.

Not really GitHub, but normally Git will leave the original author of the commit when you rebase. It will look like this in GitHub: https://github.com/jhasse/ninja/commits/serialize "colincross authored and jhasse committed"

src/ninja.cc
jonesmz: ?

EliRibble:
Created subset PR: https://github.com/ninja-build/ninja/pull/1631

src/ninja.cc
jonesmz: ?

XVilka:
There are now some conflicts that should be resolved. Please rebase.

kaspar030:
> Another thing, regarding the frontend that is invoked by ninja.
> I also expected more an API like @jonesmz feel like it would be easier to have an API like this:
> 
> ```
> ninja --serialized-output
> ```
> [...]
> 
> That way, it seems easier to integrate in a tool like Emacs, because if I wanted to integrate the ninja build inside my Emacs session, with the current design, it looks like I would to write a frontend script, that forward the data to Emacs, so I would need 3 processes, while Emacs could maybe just invoke ninja with the rights params and consume the additional reporting stream.

Just as a datapoint, I agree with this.
I'm currently writing a frontend that "transparently" calls out to ninja. Currently I cannot now what ninja did. if it exited without error, I can just assume that everything has been built, and if it fails, ... I guess the need for the functionality in general is clear. :)

So with the solution this PR proposes, my frontend would call ninja with another tool (which is a "frontend" in ninja's terminology) as parameter which ninja executes, which would somehow communicate back to my frontend. That sounds clumsy, error prone and not trivial to get right in a portable way.

Just having ninja's output in a structured way that's easy to parse would be much easier to integrate.

I don't care if it is json or protobufs.

amilcarlucas:
ping

Issue #115, opened 2011-09-15
Estimate time left on build


evmar:
Using the previously-logged timing info, estimate the remaining time on a build.

Parallelization may make this difficult to impossible.

PR #1602, opened 2019-07-12
Add remaining time estimate to Ninja status.

jpakkane:
Creates a new token `%h` (for "human readable", more suitable letters were already taken) that displays an estimate on how long the current build will last.

It's meant to be informative rather than accurate. When there is a lot of time remaining it will look like this:

    1h 15m

and when there is only a small amount of work left it looks like this:

    53s

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -205,6 +205,7 @@ Several placeholders are available:
 `%c`:: Current rate of finished edges per second (average over builds
 specified by `-j` or its default)
 `%e`:: Elapsed time in seconds.  _(Available since Ninja 1.2.)_
+`%h`:: Estimated time remaining.  _(Available since Ninja 1.10.)_
 `%%`:: A plain `%` character.
 
 The default progress status is `"[%f/%t] "` (note the trailing space
--- a/src/build.cc
+++ b/src/build.cc
@@ -274,6 +274,28 @@ string BuildStatus::FormatProgressStatus(
         break;
       }
 
+      case 'h': {
+        double ratio_finished = double(finished_edges_) / total_edges_;
+        if (overall_rate_.Elapsed() < 1.0 || ratio_finished < 0.01) {
+          snprintf(buf, sizeof(buf), "--m --s");
+        } else {
+          double time_remaining =
+              (1.0 / ratio_finished - 1) * overall_rate_.Elapsed();
+          int hours = int(time_remaining) / 3600;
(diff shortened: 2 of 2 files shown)


jonesmz:
Does this pull request address historical data from previous builds? 

I only took a quick look, but from what I can tell, this change only tries to offer an estimate based on the time that items have taken in the current run, which I don't think is accurate at all.

Better would be to offer no estimate at all for the first build, and then subsequently offer estimates based on how long each task has taken in the past.

jpakkane:
> Does this pull request address historical data from previous builds?

No.

> which I don't think is accurate at all.

This MR is _not meant to be_ to be super accurate. It is to give a rough estimate to the user. Simply getting to know whether the build will take 5 seconds, a minute, 5 minutes or 30 minutes is useful on its own. It is better than having no estimates at all, which is the current situation.

> Better would be to offer no estimate at all for the first build, and then subsequently offer estimates based on how long each task has taken in the past

That can be done but it will take a _lot_ more code. Doing it right takes on the order of hundreds of lines of code whereas this MR is around 20. Also the exact estimation algorithm can be changed later so if someone wants to improve the accuracy, they can do that later in a separate merge request.

jonesmz:
> This MR is not meant to be to be super accurate. 

I take it you meant "PR"? I'm not sure what an MR would be.

> It is to give a rough estimate to the user. Simply getting to know whether the build will take 5 seconds, a minute, 5 minutes or 30 minutes is useful on its own. It is better than having no estimates at all, which is the current situation.

An estimate that is known to not be correlated with the real result in any meaningful fashion is worse than no estimate at all.

Some projects have 99 cpp files that all compile in under 1 minute, but then one cpp file takes 30 minutes. From what I can tell, your PR would claim that the last cpp file will finish in 1 minute, based on the completion rate of the other files, being wrong by, in this case, 2900%.

> That can be done but it will take a lot more code. Doing it right takes on the order of hundreds of lines of code whereas this MR is around 20. Also the exact estimation algorithm can be changed later so if someone wants to improve the accuracy, they can do that later in a separate merge request.

I'm not the maintainer of this project, so I can't say one way or another whether this will be merged. That being said, my opinion is that this code should not be merged in this state.

The only way to provide meaningful estimates is to know how long things have taken in the past, using the existing ninja log database to provide that information, and then provide the estimate based on the previous data (perhaps averaged over time)

jonesmz:
https://github.com/ninja-build/ninja/issues/115

jhasse:
Closing in favor of #1963.

PR #1604, opened 2019-07-17
Enable ninja to use > 64 CPUs on Windows

### LINKED ISSUE #1603: ninja not making use of more than 64 CPUs
When ninja runs on a Windows system with more than 64 logical processors, its default value for the number of available processors (when not given a -j value) is below the actual number of processors.

This is due to Windows Processor Groups (see: https://docs.microsoft.com/en-us/windows/desktop/procthread/processor-groups), the way systems with more than 64 processors are handled by Windows. The system's processors are logically divided into groups, and each group is treated as a single entity.

For example, a 72-logical-processor machine will be split into two groups of 36 processors each on Windows. On such a system, ninja identifies only 36 processors available for use, thereby not taking full advantage of the system's 72 available processors.

jessemckenna:
This incorporates a fix to ninja's calculation of the number of processors on Windows systems with more than 64 cores.

Windows systems with more than 64 logical processors divide the processors into groups, with most applications using only one group. Previously, ninja was using a Windows API call that only counted the number of processors in the first processor group, which resulted in only partial utilization of high-processing-power systems when no -j value was specified.

When building Chrome on a 72-core Windows system, this change has the following effect:

Before:
ninja detects 36 cores (i.e. one processor group only)
Average time to build Chrome over 10 builds: 5628 s

After:
ninja detects all 72 cores
Average time to build Chrome over 10 builds: 3215 s (57% of the time needed before)

This fixes #1603 .

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -351,7 +351,7 @@ else:
     except:
         pass
     if platform.is_mingw():
-        cflags += ['-D_WIN32_WINNT=0x0501']
+        cflags += ['-D_WIN32_WINNT=0x0601']
     ldflags = ['-L$builddir']
     if platform.uses_usr_local():
         cflags.append('-I/usr/local/include')
--- a/src/util.cc
+++ b/src/util.cc
@@ -481,9 +481,7 @@ string StripAnsiEscapeCodes(const string& in) {
 
 int GetProcessorCount() {
 #ifdef _WIN32
-  SYSTEM_INFO info;
-  GetNativeSystemInfo(&info);
-  return info.dwNumberOfProcessors;
+  return GetActiveProcessorCount(ALL_PROCESSOR_GROUPS);
 #else
 #ifdef CPU_COUNT
   // The number of exposed processors might not represent the actual number of


atetubou:
You need to have -D_WIN32_WINNT=0x0601 in cygwin build?

jessemckenna:
Thanks so much @atetubou for the helpful suggestion! : )

jonesmz:
@jessemckenna Can you explain more about the change to _WIN32_WINNT ? What does that change mean in terms of the windows platforms that Ninja will execute on?

@jhasse Does ninja target a specific windows platform? E.g. is Ninja still supporting windows XP? Vista? 8? 8.1?

There are also various server versions...

Regardless, aside from the potential to lose support for ancient versions of windows, this pull request looks straight forward and obvious to me.

jhasse:
> What does that change mean in terms of the windows platforms that Ninja will execute on?

It means that Ninja requires Windows 7 (Windows Server 2008) now.

> Does ninja target a specific windows platform? E.g. is Ninja still supporting windows XP? Vista? 8? 8.1?

At the moment: All of them. After this PR: Only 8 and 8.1.

For this PR to be merged, someone with a >64 core machine needs to test it. Also anyone with an unusual Windows setup (e.g. 2 CPUs) could give it a try.

jonesmz:
I have a dual socket CPU, but it only has 24 logical cores total.

Unfortunately, it does not have windows on it, so testing this change would not be practical unless there's no one else who can test it.

jessemckenna:
If losing support for older versions of Windows is not an option, the alternative would be for Windows users with > 64 processors to explicitly pass a -j value that reflects their actual number of processors. This could perhaps be reflected in the documentation somehow.

Re: testing, I believe there is a way to manually set the system group size (see: https://docs.microsoft.com/en-us/windows-hardware/drivers/devtest/boot-parameters-to-test-drivers-for-multiple-processor-group-support#create-multiple-processor-groups-by-changing-the-group-size). This change could be tested by setting group size to (number of processors / 2) to create two processor groups. I believe the ninja will then detect only half the system processors, i.e. one processor group.

jonesmz:
Personally I don't think anyone should be supporting Windows XP or Vista. Vista reached end of "Extended Life" support on April 11, 2017.

Windows 7 reached total end-of-life in January 2020.

I think supporting a single end-of-life'd version of Windows is more than generous.

Note that since NInja tends to release only about once a year, if this change is merged into master now, it's unlikely to be released before Windows 7 reaches official end of life.

atetubou:
Maybe better to keep fallback to GetNativeSystemInfo so that ninja can run on older os for a while?

jhasse:
> Windows 7 reaches total end-of-life in January 2020.

Windows 7 would still be supported.

> Maybe better to keep fallback to GetNativeSystemInfo so that ninja can run on older os for a while?

Do you still need XP/Vista support at Google? If not, I don't see who would honestly.

jonesmz:
> Windows 7 would still be supported.

Right. I said that supporting one out-of-life version of windows is generous enough, much less multiple out-of-life versions.

Since this PR would still allow Ninja to run on Windows 7, which would be end-of-life by the time it was included in an official release, that seems like a sufficiently long support lifetime.

jhasse:
> which would be end-of-life by the time it was included in an official release

That's just speculation.

jonesmz:
> That's just speculation.

It's an educated prediction. Development of Ninja happens slowly. January 2020 is only about 5 months away. Unless you have plans to the contrary which haven't been communicated, it's reasonable to assume that Ninja won't see an additional release by then.

atetubou:
> > Maybe better to keep fallback to GetNativeSystemInfo so that ninja can run on older os for a while?
> 
> Do you still need XP/Vista support at Google? If not, I don't see who would honestly.

No, we don't use XP/Vista anymore. So I'm fine with this patch.

PR #1609, opened 2019-07-26
ManifestParser: Avoid calling CanonicalizePath for known nodes

domidimi:
We have been calculating the canonical paths of nodes even if we have
seen them already. As CanonicalizePath is a hot spot it was visible in
the performance measurements. Avoid that by calculating the canonical
path only if the node is not known already.

On a 420MB build.ninja file reduces the noop load time from 5.5s to 5.0s

Signed-off-by: Dimitar Yordanov <dimitar.yordanov@sap.com>


domidimi:
Improvement could depend on the content of the build.ninja e.g. in the way it was generated.

domidimi:
The second commit is more intrusive, not quite elegant and requires c++11. I think further improvements are possible. Would be nice to share your opinion. I see there is also a different pull request #1208 , which seems to have similarities to what is proposed here.

jonesmz:
It's unlikely that C++11 code will be accepted by Ninja.

See discussion here: https://github.com/ninja-build/ninja/issues/1403

Among other places.

domidimi:
Should be doable also without c++11.

src/manifest_parser.cc
```diff
@@ -408,14 +406,14 @@ bool ManifestParser::ParseFileInclude(bool new_scope, string* err) {
     return false;
   string path = eval.Evaluate(env_);
 
-  ManifestParser subparser(state_, file_reader_, options_);
+  ManifestParser *subparser = new ManifestParser(state_, file_reader_, options_);
```
domidimi: needs to be fixed.

domidimi: but if we follow the ninja approach we leave it to the OS to free?

src/build_test.cc
```diff
@@ -2175,7 +2175,7 @@ TEST_F(BuildWithDepsLogTest, DepFileOKDepsLog) {
 
     Edge* edge = state.edges_.back();
 
-    state.GetNode("bar.h", 0)->MarkDirty();  // Mark bar.h as missing.
+    state.GetNode(std::string("bar.h"), 0)->MarkDirty();  // Mark bar.h as missing.
```
jonesmz: Why is this an improvement?

The c-style string will be converted to an std::string if needed, and not, if not.

Why insert an explicit conversion?

src/eval_env.h
```diff
@@ -91,7 +172,7 @@ struct BindingEnv : public Env {
   const Rule* LookupRuleCurrentScope(const string& rule_name);
   const map<string, const Rule*>& GetRules() const;
 
-  void AddBinding(const string& key, const string& val);
+  void AddBinding(string&& key, string&& val);
```
jonesmz: This is bad form.

If you want the function to take ownership of the string, pass it by value, not by rvalue-reference.

rvalue-reference parameters is only appropriate for "destructive" operations, and even that's debatable.

See further discussion here: https://github.com/redboltz/mqtt_cpp/issues/289#issuecomment-505193666

src/build_test.cc
domidimi: There is no intention to improve the tests. In order not to change the tests we have to remove the second GetNode candidate or add one that explicitly accepts a char *

src/manifest_parser.cc
domidimi: > Overall, I think this changes way too many things all at the same time.
> 
> I recommend breaking these changes into smaller pieces, and submitting them individually.
You could maybe  review/accept/reject the first commit in the pull request, which is smaller?

src/build_test.cc
jonesmz: Why doesn't it already accept a char*? it does currently.

Because of your change to only accept string&&, in a variety of functions?

I strongly disagree with changing the parameter to those functions to string&&.

src/eval_env.h
domidimi: > If you want the function to take ownership of the string, pass it by value, not by rvalue-reference.
> See further discussion here: redboltz/mqtt_cpp#289 (comment)

You're referring to item 41 (and 27?) from "Effective Modern C++"? But I want to accept rvalues only so that the compiler will also tell me when passing any lvalues as those will produce a copy. And passing by value would have an extra cost.

> rvalue-reference parameters is only appropriate for "destructive" operations, and even that's debatable.

your personal opinion?

domidimi: And please have a look at the code.

jonesmz: I had to purchase a copy of the book to figure out what you were referring to.

Apparently Scott had the same idea I did.

Yes. I am referring to item 41 and 27.

What you are doing here is undesirable. It makes the code more complex. And doesn't actually solve and performance problems.


You've changed way too many things at once for me to be able to effectively review this pull request.

If you revert the changes to string parameters so that they require rvalue references, I would be happy to review the remaining code.

And no, that's not my opinion. Rvalue references are for "moving" an object, and moving is "destructive" (not destructor) to the original object. Perhaps my wording was weird, but my meaning wasn't derived from an opinion.

jhasse:
Can you split the improvements into separate PRs? I also doubt that the string change actually improves performance. Did you test that?

domidimi:
I'll remove the second patch and push only the first in this PR. If I find time later, I'll look into changing the second patch to not require C++11 support. Anyways, the main point of the second patch was to reduce the number of string allocations, which are also visible in a performance profile. I tested both patches manually with a big build.ninja file. Sorry, I did not create benchmark tests.

jonesmz:
Aside from my one comment about re-evaluating the "path" variable, this PR now appears to be an improvement.

PR #1612, opened 2019-08-02
Expand the continuous integration coverage to more platforms / operating systems

jonesmz:


jhasse:
Nice! The Ubuntu 16.04 configuration is now tested twice on Travis and AppVeyor, right?

I would also like to reduce the number of CI services in the future and move to Azure Pipelines, see #1526.

jonesmz:
When I was first implementing this change, I found that the travis 16.04 builds behaved slightly differently than on AppVeyor, but I have to admit I don't remember the specifics, as it was about 6 months ago.

That's the reason why I kept Ubuntu16.04 on both CI services.

Personally I'd rather have more CI (So long as it continues working, of course) than less. Subtle differences in the way platforms are configured can result in surprising outcomes.

I was not able to get Azure Pipelines working, actually. But I didn't give it a huge amount of time before giving up.

jhasse:
After this PR no CI configuration tests with re2c missing. We should still do that.

jonesmz:
Could you elaborate on the problem? I'm not sure I understand your phrasing.

jonesmz:
Oh.

You're saying that none of the build types see how the Ninja code behaves if the re2c package isn't installed, and therefore, the code that's bundled with ninja is used as is, instead of regenerated.

I disagree that that should be a supported situation.

Nevertheless, I can try to adjust the CI changes to trigger that.

jonesmz:
@jhasse 

If I removed the parts of this PR that ensure that re2c is installed on the platform, is that sufficient for you to merge this change?

That's essentially what the current CI relies on to test the code that's directly included in the repository.

jhasse:
Yes, also please remove Ubuntu 16.04 from one of the CIs.

appveyor.yml
```diff
@@ -21,14 +37,16 @@ for:
       ./ninja all\n
       ./ninja_test 2>&1\n
       ./misc/ninja_syntax_test.py 2>&1\n\"@"
-  -
-    matrix:
+
+  - matrix:
```
jhasse: Is this necessary?

jonesmz: As in, will it break something if the change isn't made? No, it won't break anything if that change is not made.

However, it is necessary for my own sanity.

You can edit it back out if you like after the PR is merged.

jonesmz:
> Yes, also please remove Ubuntu 16.04 from one of the CIs.

I disagree with this, but I did it anyway.

jonesmz:
Two targets, one for Windows, one for Linux, now don't regenerate the re2c derived code.

I've also removed Ubuntu 16.04 from AppVeyor.

Please inform me if you need any other changes.

jhasse:
Thanks :)

PR #1620, opened 2019-08-05
Fix appveyor

goshhhy:
this should fix the sed command in appveyor.yml - leaving in draft mode until confirmed


goshhhy:
sorry, i made a mess with this attempt at a merge request, please ignore it

PR #1624, opened 2019-08-06
Recommend MD over MMD for header dependencies.

bungeman:
The MMD flag will silently omit includes found through pointy brackets or system include paths. This can lead to issues not only when system headers change, but any paths included through the isystem flag. Because the isystem flag implicitly turns off warnings as errors it has often come to be used as a "not my code" flag used with local third party dependencies which may be frequently updated or changed for debugging. As a result, it is far safer to default to MD (which includes all include dependencies) in this example.

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -569,10 +569,10 @@ Use it like in the following example:
 ----
 rule cc
   depfile = $out.d
-  command = gcc -MMD -MF $out.d [other gcc flags here]
+  command = gcc -MD -MF $out.d [other gcc flags here]
 ----
 
-The `-MMD` flag to `gcc` tells it to output header dependencies, and
+The `-MD` flag to `gcc` tells it to output header dependencies, and
 the `-MF` flag tells it where to write them.
 
 deps


nico:
Do you have numbers to back up that this is perf-neutral?

bungeman:
Not sure what you mean. Not only do I not have numbers to "back up" that this is perf-neutral, I expect that this may slow down deps checking quite a bit in general. The issue is that MMD is simply wrong since it tells the compiler to not report all of the dependencies. I've personally run into this issue on at least one occasion (our gn build rules just used MMD as described in the documentation here) where an update to some third party headers brought in through -isystem (in order to suppress the warnings in them) did not cause expected targets to be rebuilt (fixed by moving to MD). This change to the documentation is to avoid leading users by default into incorrect incremental builds. If you really want you could add after this change a "for advanced users only and only if full deps checking is too slow and I pinky swear that nothing dragged in through -isystem will ever change, you can use MMD" that's fine, but I expect that's actually a vanishingly small set of use cases.

jhasse:
Using `-isystem` to suppress warnings isn't uncommon, I think this change in the doc makes sense.

PR #1626, opened 2019-08-07
List all clang-format options instead of relying on an upstream formatting style.

jonesmz:
This allows for the project to individually tweak formatting options in the future.

The options are as close as I possibly could get to the *actual* coding style of the project, instead of the Google style.

I compared the number of differences with this shell script;

```
git checkout src ; clang-format -i src/*.h src/*.cc ;  git checkout src/getopt.h src/lexer.cc src/depfile_parser.cc src/*test*.cc ; git diff | grep -e "+" -e "-" | wc -l
```

And then compared the final result by hand to ensure nothing particularly egregious was happening.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -16,10 +16,141 @@
 # Still use your best judgement for formatting decisions: clang-format
 # sometimes makes strange choices.
 
-BasedOnStyle: Google
+# Based on the Clang-7.0 documentation
+
+Language: Cpp
+Standard: Cpp03
+DisableFormat: false
+
+AccessModifierOffset: -2
+
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Left
+AlignOperands: true
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: false
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
 AllowShortFunctionsOnASingleLine: Inline
 AllowShortIfStatementsOnASingleLine: false
 AllowShortLoopsOnASingleLine: false
+
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: false
+AlwaysBreakTemplateDeclarations: Yes
+
+BinPackArguments: true
+BinPackParameters: true
+
+BreakBeforeBraces: Custom
+BraceWrapping:
+  AfterClass: true
+  AfterControlStatement: false
+  AfterEnum: false
+  AfterFunction: false
+  AfterNamespace: false
+#  AfterObjCDeclaration
+  AfterStruct: false
+  AfterUnion: false
+  AfterExternBlock: false
+  BeforeCatch: false
+  BeforeElse: false
+  IndentBraces: false
+  SplitEmptyFunction: false
+  SplitEmptyRecord: false
+  SplitEmptyNamespace: false
+
+#BreakAfterJavaFieldAnnotations
+
+BreakBeforeBinaryOperators: None
+BreakBeforeTernaryOperators: false
+BreakConstructorInitializers: BeforeColon
+BreakInheritanceList: BeforeColon
+BreakStringLiterals: true
+
+ColumnLimit: 80
+
+CommentPragmas: ''
+
+CompactNamespaces: true
 ConstructorInitializerAllOnOneLineOrOnePerLine: false
+ConstructorInitializerIndentWidth: 4
+ContinuationIndentWidth: 4
+
 Cpp11BracedListStyle: false
+
+DerivePointerAlignment: false
+
+FixNamespaceComments: false
+
+ForEachMacros: ['']
+
+IncludeBlocks: Preserve
+
+IncludeCategories:
+  - Regex:           '^<.*'
+    Priority:        1
+  - Regex:           '^".*'
+    Priority:        2
+
+IncludeIsMainRegex: "(_test)?$"
+
 IndentCaseLabels: false
+IndentPPDirectives: None
+IndentWidth: 2
+IndentWrappedFunctionNames: true
+
+#JavaScriptQuotes
+#JavaScriptWrapImports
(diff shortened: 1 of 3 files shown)


jhasse:
I actually prefer relying on the Google style. No problem in tweaking formatting options in the future, we just need to add them.

jonesmz:
The Google style doesn't accurately represent the actual formatting of the Ninja project.

Also, where the built-in styles actually defined? I was not able to locate them anywhere.

jhasse:
> The Google style doesn't accurately represent the actual formatting of the Ninja project.

IMHO it's better not to introduce a "Ninja-style".

jonesmz:
Sure. I don't care what style the project has. I care that the style is applied consistently.

jonesmz:
If you don't want this change then close the PR.

Otherwise, tell me how I can improve the clang-formatting situation.

jhasse:
Closing then. Don't worry too much about formatting.

jonesmz:
If we're not worrying about formatting, then please don't ask me to change the formatting of pull requests that I make in the future.

I'm willing to do the work to get the entire code base formatted properly, but I'm not at all interested in having my PRs nitpicked.

Either all of the code base, or none of it.

jhasse:
No, sorry. There's something in-between.

jonesmz:
You misunderstand.

If you're not going to let me either:

* Fix the entire codebase, to properly comply with the formatting rules that the project claims to want enforced
* Fix the formatting rules to match the codebase

Then I'm going to ignore any future requests to change formatting of code that I submit PRs for, because I see changing individual lines of code at a time as *significantly* more work than just fixing the codebase.

I'm willing to do the grunt work for getting the formatting under control. I'm also willing to do it incrementally.

If we can come to an agreement to get the project's formatting under control, then while the process for getting the formatting working (which may well take many months, because we apparently can't ever introduce merge conflicts in multiple-years-old PRs) is underway, I'll put up with nitpicking of PRs.

But if we don't have an active effort going on to get the formatting under control, then I'm not going to bother, because as you said, "Don't worry too much about formatting."

jhasse:
> that the project claims to want enforced

Where did the project claim that? `.clang-format` says: "This isn't meant to be authoritative, but it's good enough to be useful. Still use your best judgement for formatting decisions"

jonesmz:
The comment in .clang-format is non-actionable. It's lacks meaning to contributors, because all bets are off on whether or not code the contributor submits a PR for will be accepted with, or without, or despite, clang-format being run on the lines that were changed.

> Where did the project claim that?

You do, whenever you want to enforce bits and pieces of the partially specified clang-format rules by requiring PRs conform to some, but not all, of the formatting rules.

jhasse:
I think you're referring to my comment in the using-namespace-std PR? The 80 column limit is explicitly mentioned here: https://github.com/ninja-build/ninja/blob/6e02ebc4b5947095fa64b5bf814f14dd6f6f6a9a/HACKING.md#coding-guidelines

jonesmz:
Yes, it is.

It's also one of the lines of the clang-format file.

I want to make the project actually compatible with flagrant use of clang-format.

Point me in the right direction.

jhasse:
The project doesn't want to be though.

jonesmz:
Why?

What's wrong with using clang-format?

jonesmz:
Also, lets be honest here.

When you say:

> The project doesn't want to be though.

You mean *you* don't want the Ninja codebase to be able to be formatted by clang-format automatically.

You're the only maintainer of this project at this time.

Justify why having the project set up so that people can have faith that running clang-format will "do the right thing" is not what you want.

jhasse:
No, I mean the author of those lines in `.clang-format`.

jonesmz:
> No, I mean the author of those lines in .clang-format.

That comment was written > 5 years ago.

It's irrelevant.

Both because:

1. The person who wrote it no longer contributes to the project
2. Clang-format has become significantly more sophisticated.

It's clear (to me) that the original intent behind the comment was that clang-format *at that time* was not able to properly capture the comment-author's preferred style. I see no reason why clang-format can't be employed today with significantly better, or even ideal, results.

> Justify why having the project set up so that people can have faith that running clang-format will "do the right thing" is not what you want.

jhasse:
It was the policy for the last 5 years, I don't want to change it.

One disadvantage is, that you'll need `// clang-format off` comments for some special parts of the code.

I also prefer not to touch existing lines, which formatting the whole code certainly would.

jonesmz:
> It was the policy for the last 5 years, I don't want to change it.

Why?

What kind of policy is:

> # This isn't meant to be authoritative, but it's good enough to be useful.
> # Still use your best judgement for formatting decisions: clang-format
> # sometimes makes strange choices.

It's not a policy.

Your position is "We shouldn't change anything because someone observed 5 years ago that the tool in question might do weird things".

That's not acceptable.

> One disadvantage is, that you'll need // clang-format off comments for some special parts of the code.

Aside from the two specific places that I added them to avoid clang-format drastically changing two large blocks of code, I disagree with your observation.

Those two places were:

1. A large switch case, which I actually thought was improved by clang-format, but predicted would be objected to by you due to the number of lines that were changed all at once. (For what it's worth, that switch case is unweildly anyway, and I'd like to see it broken down into smaller parts).
2. A large string, which only need to be guarded from clang-format because the project isn't using C++11, which provides raw string literals, which is by far superior to the way the string is currently handled.

So I don't think this is correct at all, clang-format acceptably formats everything in the ninja codebase, aside from that one single string, which would be better as either:

- A stand alone file, read in as needed for the test
- A C++11 raw string literal


> I also prefer not to touch existing lines, which formatting the whole code certainly would.

Why?

jhasse:
To avoid merge conflicts and preserve line history.

jonesmz:
> To avoid merge conflicts and preserve line history.

Which is why I said we can conduct this process over a period of multiple months.

Personally I don't think there's any value to be found in either avoiding merge conflicts, or preserving line history, but you clearly do, so I'm willing to work with you on this so that everyone wins.

- Merge conflicts can simply be resolved, either by the original PR author, or by someone else.
- By line history, I assume you mean things like git blame, and similar. While formatting changes can make those slightly more complicated, ultimately there's a lot of negative value in forbidding progress for the sake of such a minor activity.

PR #1627, opened 2019-08-08
Escape backslashes in rsp_file on Windows

vleschuk:
Ninja generates RSP files on Windows with paths in the following format:

path\to\file1.obj
path\to\file2.obj
...

At linkage state linker is called with paths missing '\':

pathtofile1.obj

This patch fixes the situation.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -19,6 +19,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <functional>
+#include <sstream>
 
 #ifdef _WIN32
 #include <fcntl.h>
@@ -72,6 +73,18 @@ bool DryRunCommandRunner::WaitForCommand(Result* result) {
    result->edge = finished_.front();
    finished_.pop();
    return true;
+}
+
+void EscapeSlashes(string& str) {
+#ifdef _WIN32
+  ostringstream ss;
+  for(size_t i = 0; i < str.size(); ++i) {
+    ss << str[i];
+    if (str[i] == '\\')
+      ss << '\\';
+  }
+  str = ss.str();
+#endif // _WIN32
 }
 
 }  // namespace
@@ -916,6 +929,7 @@ bool Builder::StartEdge(Edge* edge, string* err) {
   string rspfile = edge->GetUnescapedRspfile();
   if (!rspfile.empty()) {
     string content = edge->GetBinding("rspfile_content");
+    EscapeSlashes(content);
     if (!disk_interface_->WriteFile(rspfile, content))
       return false;
   }


vleschuk:
Similar problem was reported in #1262.

atetubou:
I think we use quotation for rspfile_content.
e.g.
```
rspfile_content = "-imsvcC:\src\..."
```

vleschuk:
> I think we use quotation for rspfile_content.
> e.g.
> 
> ```
> rspfile_content = "-imsvcC:\src\..."
> ```

Do you mean rspfile_content in *.ninja files? I faced the issue with CMake-generated ninja files and there were no rspfile_content fields, for example:

```
  build bin\mylib.dll: very_long_list_of_objects
  LANGUAGE_COMPILE_FLAGS = -tM -Od -v
  LINK_FLAGS = -tM -lS:1048576 -lSc:4098 -lH:1048576 -lHc:8192 -v
  LINK_LIBRARIES = iphlpapi.lib import32.lib
  OBJECT_DIR = Foundation\CMakeFiles\Foundation.dir
  POST_BUILD = cd .
  PRE_LINK = cd .
  RESTAT = 1
  TARGET_FILE = bin\PocoFoundationd.dll
  TARGET_IMPLIB = lib\PocoFoundationd.lib
  TARGET_PDB = PocoFoundation.dll.dbg
  pool = BCC32LinkPool
  RSP_FILE = CMakeFiles\Foundation.rsp
```

Here is example of Foundation.rsp (190 lines) generated by release version of Ninja:
```
Foundation\CMakeFiles\Foundation.dir\src\ASCIIEncoding.cpp.obj
Foundation\CMakeFiles\Foundation.dir\src\AbstractObserver.cpp.obj
Foundation\CMakeFiles\Foundation.dir\src\ActiveDispatcher.cpp.obj
Foundation\CMakeFiles\Foundation.dir\src\ArchiveStrategy.cpp.obj
Foundation\CMakeFiles\Foundation.dir\src\Ascii.cpp.obj
Foundation\CMakeFiles\Foundation.dir\src\AsyncChannel.cpp.obj
.....
```
Which results in linkage error:
```
Fatal: Unable to open file 'FOUNDATIONCMAKEFILESFOUNDATION.DIRSRCASCIIENCODING.CPP.OBJ'
```

With the patch applied the contents of rsp file looks like:
```
Foundation\\CMakeFiles\\Foundation.dir\\src\\ASCIIEncoding.cpp.obj
Foundation\\CMakeFiles\\Foundation.dir\\src\\AbstractObserver.cpp.obj
Foundation\\CMakeFiles\\Foundation.dir\\src\\ActiveDispatcher.cpp.obj
Foundation\\CMakeFiles\\Foundation.dir\\src\\ArchiveStrategy.cpp.obj
Foundation\\CMakeFiles\\Foundation.dir\\src\\Ascii.cpp.obj
Foundation\\CMakeFiles\\Foundation.dir\\src\\AsyncChannel.cpp.obj
...
```

And link process is sucessful.

atetubou:
Do you use link.exe?
Also I'm wondering how build rule looks like for your dll build.
I'm afraid this breaks currently successfull build.

vleschuk:
I use linker from embarcadero toolchain, also verified with link.exe from
visual studio. And it passed all tests on GitHub ci.

Best Regards
Victor

atetubou:
I think rspfile_content should be used as-is and it is better to change cmake to generate escaped path when embarcadero toolchain is used in that case.
Some ninja user may rely on rspfile_content containing \ without escape.

vleschuk:
In my case cmake didn't generate rspfile_content with visual studio
toolchain too. Just rsp_file option.

Best Regards
Victor

vleschuk:
Does anybody need this?

atetubou:
I think your ninja.build should have rspfile_content around build rule.
And if that is not properly escaped, you need to change cmake or cmakefile instead of changing ninja.

nico:
I agree, this looks like a cmake bug.

(But lots of people use cmake + ninja successfully, so maybe something is strange in your environment? Or maybe most people don't use cmake + ninja + linker rsp files?)

YanzhouYu:
Hi, I also met this problem while building with ninja on Windows, I tried with your change and it works! Will the changes be merged into ninja master branch for official release?

vleschuk:
No it will not be merged into master 'cause it may brake other builds and it looks like cmake problem.

PR #1628, opened 2019-08-08
Use 'deps' with 'depfile'

bungeman:
Without 'deps' stating the format of the 'depfile', the 'depfile' will be silently ignored. The 'depfile' example before this change is broken in that it will not actually store the deps in 'depfile' due to the lack of 'deps'.

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -568,8 +568,9 @@ Use it like in the following example:
 
 ----
 rule cc
-  depfile = $out.d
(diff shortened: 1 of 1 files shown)


bungeman:
So the original comment on this commit is incorrect. However, the change itself is probably still recommended, since by default it is better to store the deps in the deps log than in scattered deps files.

jonesmz:
You can amend your comment

git commit --amend
git push -f

PR #1629, opened 2019-08-08
Error on depfile without deps.

bungeman:
If a rule specifies a depfile but no deps (to specify the format) the depfile is ignored. Make this an error instead of silently ignoring the depfile.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -943,7 +943,7 @@ bool Builder::FinishCommand(CommandRunner::Result* result, string* err) {
   string deps_type = edge->GetBinding("deps");
   const string deps_prefix = edge->GetBinding("msvc_deps_prefix");
   if (deps_type.empty()) {
-    if (!edge->GetBindingBool("depfile")) {
+    if (edge->GetBindingBool("depfile")) {
       *err = string("edge with depfile but no deps makes no sense");
       return false;
     }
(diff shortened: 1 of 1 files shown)


bungeman:
This is equal parts pull request and bug report. This is one way to fix the issue, the other way to fix this is to see if there is a depfile specified and if deps isn't specified assume it must be gcc. However, this seems to go against the idea of everything being explicit.

jonesmz:
I agree that this should be an error.

This problem bit me in the past, where a typo caused my ninja file generator to not specify the deps type, and I got miscompiles as a result.

bungeman:
So I go to update the tests and am wondering how they are possibly working. It turns out if there are no deps in the .ninja_deps for a file (there won't be since the depfile is being ignored) the actual depfile will be looked for and that will be assumed to be in gcc format (since msvc isn't generally persisted to disk). It turns out that if the depfile isn't being parsed into the .ninja_deps file it isn't deleted, so the next build it should be read (but only because the format wasn't stated). Then the -tdeps tool only dumps dependencies from the .ninja_deps file and not all the deps of all the targets, apparently because IsDepsEntryLiveFor returns false if 'deps' isn't set on the edge. Now I'm super confused about what the right behavior is.

bungeman:
So I think I've convinced myself that this particular request probably isn't something that can be accepted since existing users may be depending on the interesting behavior of falling back to looking for the depfile directly. While documented to do so, the deps tool returning only logged deps is surprising. Maybe what I want is a change to 'gn' to always output 'deps = gcc' when setting a depfile.

PR #1630, opened 2019-08-09
Enable build on IBM i platform

ThePrez:
Python on the IBM i platform actually runs inside an AIX variant named "PASE". The system self-identifies as "OS400" (an unfortunate history of names, perhaps). 

There is no perfstat libraries available for IBM i, and IBM i has no good way to determine load averages. This patch allows builds without load average enforcement, which is better than a failing build.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -97,6 +97,9 @@ class Platform(object):
     def is_aix(self):
         return self._platform == 'aix'
 
+    def is_aix_os400_variant(self):
+        return self._platform == 'aix' and os.uname().sysname.startswith('OS400')
+
     def uses_usr_local(self):
         return self._platform in ('freebsd', 'openbsd', 'bitrig', 'dragonfly', 'netbsd')
 
@@ -536,7 +539,7 @@ if platform.is_msvc():
 else:
     libs.append('-lninja')
 
-if platform.is_aix():
+if platform.is_aix() and not platform.is_aix_os400_variant():
     libs.append('-lperfstat')
 
 all_targets = []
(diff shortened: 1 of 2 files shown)


configure.py
```diff
@@ -97,6 +97,9 @@ def is_solaris(self):
     def is_aix(self):
         return self._platform == 'aix'
 
+    def is_aix_os400_variant(self):
```
kadler: `is_pase` would be a better name, I suspect

configure.py
```diff
@@ -97,6 +97,9 @@ def is_solaris(self):
...
+        return self._platform == 'aix' and os.uname().sysname.startswith('OS400')
```
kadler: Please check for `self._platform` is 'os400' first, since at some point we'll fix Python to not lie about what platform it's on.

configure.py
ThePrez: Want to keep os400, so a compromise of is_os400_pase() was chosen

src/util.cc
```diff
@@ -562,6 +562,10 @@ double GetLoadAverage() {
 
   return posix_compatible_load;
 }
+#elif defined (__PASE__)
```
jhasse: Remove space before (

src/util.cc
```diff
@@ -562,6 +562,10 @@ double GetLoadAverage() {
...
+double GetLoadAverage() {
+    return -0.0f;
```
jhasse: Use 2 space indent.

ThePrez: Fixed. Sorry!

src/util.cc
ThePrez: Fixed. Sorry!

PR #1614, opened 2019-08-02
Remove 'using namespace std' from header files, properly namespace all std symbols

jonesmz:
I wish to use ninja as a library in another application.

I am aware that the ninja codebase does not attempt to offer any kind of stable API, and I'm not looking for Ninja to do so in the future.

This pull request, if merged, would resolve #1613

My motivations for using the ninja code in a library situation include, but are not limited to, wanting to be able to parse the existing dependency database, and also the ninja log file, to drive some application specific behavior that ninja itself wouldn't benefit from.

There are a variety of other situations where supporting ninja in a library format would also make sense.

E.g. consider a situation where a program written in an interpreted language wanted to use ninja to execute some tasks. That application might want to modify the build graph while executing parts of it based on external information.

Right now, Ninja can't be used in that way, but I strongly believe that from both an architectural point of view, as well as a performance standpoint, moving the Ninja codebase toward a model where it's easier to use as a library (even if that's never officially supported!!!) will do nothing but benefit Ninja. Not the least reason of which is the ease of future maintenance of the code.

There are plenty more reasons, but I don't need to belabor the point.

Regardless, the problem that I continually have, is that it's not practical to directly #include headers from ninja into my existing project. I already *have* data types named map, and string, that live inside their own namespaces. But because ninja has "using namespace std", if I try to use "map" or "string" inside my namespace, but without the namespace indicator, I get ambiguity on which "map" or "string" is actually being used.

I know that Ninja uses "using namespace std" in it's code intentionally. I, of course, disagree with this, as have many others on the mailing list and on github.

So this pull request is a compromise.

Each implementation file can continue using, or not using, "using namespace std" as it sees fit, while header files will use "std::" when referring to types from the std namespace. This makes it significantly less difficult for downstream projects (Not only my own project of course, any downstream project at all) to use Ninja's codebase for their projects.

PR #1631, opened 2019-08-09
Status interface

EliRibble:
This is a subset of PR #1600 which adds a new frontend. This is designed to avoid the controversy around what data format is used for the new frontend and whether ninja creates the subprocess. This change will form the base of making a ninja library by allowing library clients to define their own status reporter class.


jonesmz:
@EliRibble 

Are you able to rebase the patch series to avoid the merge conflict?

EliRibble:
Working on it right now.

src/build.cc
```diff
@@ -252,28 +239,27 @@ string BuildStatus::FormatProgressStatus(
...
+        current_rate_.UpdateRate(finished_edges_, time_millis_);
         SnprintfRate(current_rate_.rate(), buf, "%.1f");
         out += buf;
         break;
 
         // Percentage
-      case 'p':
-        percent = (100 * finished_edges_) / total_edges_;
+      case 'p': {
+        int percent = (100 * finished_edges_) / total_edges_;
```
jonesmz: Not mission critical, but personally I'd drop a comment near the top of this function that indicates that integer division is, in fact, the intended type of division.

src/build.h
```diff
@@ -227,6 +226,13 @@ struct Builder {
                     const string& deps_prefix, vector<Node*>* deps_nodes,
                     string* err);
 
+  /// Map of running edge to time the edge started running.
+  typedef map<Edge*, int> RunningEdgeMap;
```
jonesmz: Similar to my comment about "EdgePtrSet", I would personally think it an improvement to name this type "RunningEdgePtrMap", or "RunningEdgePtrMap_t"

src/build.h
```diff
@@ -239,9 +245,9 @@ struct Builder {
 struct BuildStatus {
   explicit BuildStatus(const BuildConfig& config);
   void PlanHasTotalEdges(int total);
-  void BuildEdgeStarted(Edge* edge);
-  void BuildEdgeFinished(Edge* edge, bool success, const string& output,
-                         int* start_time, int* end_time);
+  void BuildEdgeStarted(Edge* edge, int64_t start_time_millis);
+  void BuildEdgeFinished(Edge* edge, int64_t end_time_millis, bool success,
+                         const string& output);
```
jonesmz: In the spirit of https://github.com/ninja-build/ninja/pull/1614

Consider using `const std::string& output` instead.

src/build.cc
```diff
@@ -382,7 +382,7 @@ void Plan::EdgeWanted(Edge* edge) {
 Edge* Plan::FindWork() {
   if (ready_.empty())
     return NULL;
-  set<Edge*>::iterator e = ready_.begin();
+  EdgeSet::iterator e = ready_.begin();
```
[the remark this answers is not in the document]
jonesmz: I'd also recommend that typedefs have a suffix, "_t", as a clue to the reader that the type in question is not, itself, a concrete type, but is instead an alias to something else. 

E.g. it could very well be that "EdgeSet" or "EdgePtrSet" is a completely independent class.

But the _t suffix would hint that it's probably a template type, most likely std::set, and so all of the associated functions would be available.

src/build.h
```diff
@@ -219,7 +219,7 @@ struct Builder {
 #else
   unique_ptr<CommandRunner> command_runner_;  // auto_ptr was removed in C++17.
 #endif
-  BuildStatus* status_;
+  Status* status_;
```
jonesmz: I would prefer to see the lifetime of this managed properly.

Is this the only place that has a pointer to the Status object?

If so, can we wrap it in auto_ptr / unique_ptr, just like the CommandRunner above?

src/ninja.cc
```diff
@@ -1294,6 +1296,8 @@ NORETURN void real_main(int argc, char** argv) {
         kDepfileDistinctTargetLinesActionError;
   }
 
+  Status* status = new StatusPrinter(config);
```
jonesmz: Hrmm, in C++11, this would be better as `std::make_unique` or `std::make_shared`

But we're using C++98 / C++03, and I don't really have a better suggestion than doing this for now

src/status.cc
```diff
@@ -0,0 +1,243 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
```
jonesmz: @colincross Are you employed by google?

src/status.cc
```diff
@@ -0,0 +1,243 @@
...
+
+  if (edge->use_console() || printer_.is_smart_terminal())
+    PrintStatus(edge, start_time_millis);
+
+  if (edge->use_console())
+    printer_.SetConsoleLocked(true);
+}
+
+void StatusPrinter::BuildEdgeFinished(Edge* edge, int64_t end_time_millis,
+                                      bool success, const string& output) {
```
jonesmz: This file is new, so lets not use "using namespace std".

Can you properly prefix std lib types with `std::` ?

src/status.h
```diff
@@ -0,0 +1,108 @@
...
+#include "build.h"
+#include "line_printer.h"
+
+/// Abstract interface to object that tracks the status of a build:
+/// completion fraction, printing updates.
+struct Status {
+  virtual void PlanHasTotalEdges(int total) = 0;
+  virtual void BuildEdgeStarted(Edge* edge, int64_t start_time_millis) = 0;
+  virtual void BuildEdgeFinished(Edge* edge, int64_t end_time_millis,
+                                 bool success, const string& output) = 0;
```
jonesmz: `std::`

src/status.h
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+  private:
+    double rate_;
+    const size_t N;
+    queue<double> times_;
+    int last_update_;
+  };
+
+  mutable SlidingRateInfo current_rate_;
```
jonesmz: Why make this mutable?

Can the reason be documented?

src/build.cc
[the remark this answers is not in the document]
jhasse: Please avoid _t suffix. If you want to know if an expression is a typedef, the IDE can show you.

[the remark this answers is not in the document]
jonesmz: What IDE?

You keep making assumptions about the capabilities of the people working with the Ninja codebase.

E.g. I don't use an IDE that provides this kind of information. My reasons for not doing so are irrelevant, but one of them is that I find intellisense functionality is typically unable to provide this information at best, or worse -- provides incorrect information. For example, I've had horrible experiences with Microsoft Visual Studio's Intellisense in past projects, which is one of the major reasons why I stopped using "fat" IDEs like MSVC altogether.

Other people, such as those who are posting patches for AIX, or various IBM related systems, may not have an IDE available at all. I wouldn't know, actually.

The google style guide makes no statements about using or not using _t. https://google.github.io/styleguide/cppguide.html

The actual C++ standard uses _t *everywhere* to indicate that things are typedefs.

I think it's strongly preferable to ensure that the person working with code knows whether they're working with a typedef, or not.

If you want to continue making assumptions about the capabilities of contributors to the Ninja codebase, then put your assumptions in the contributor guidelines.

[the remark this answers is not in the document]
jhasse: > The names of all types — classes, structs, **type aliases**, enums, and type template parameters — have the same naming convention. Type names should start with a capital letter and have a capital letter for each new word. No underscores.

https://google.github.io/styleguide/cppguide.html#Type_Names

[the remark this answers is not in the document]
jonesmz: I stand corrected.

I think their advice is dangerous and misguided, but nevertheless, I failed to correctly identify that they do talk about it.

src/state.h
```diff
@@ -62,7 +63,7 @@ struct Pool {
   void DelayEdge(Edge* edge);
 
   /// Pool will add zero or more edges to the ready_queue
-  void RetrieveReadyEdges(set<Edge*>* ready_queue);
+  void RetrieveReadyEdges(EdgeSet* ready_queue);
```
jonesmz: Should this really be a pointer-to-container?

Wouldn't a reference be a more appropriate choice?

jhasse: https://google.github.io/styleguide/cppguide.html#Reference_Arguments

jonesmz: Alright.

src/status.h
EliRibble: I don't know, maybe @colincross  can chime in.

src/build.h
EliRibble: RunningEdgePtrMap_t is out due to the google naming convention.
I don't care much between RunningEdgeMap and RunningEdgePtrMap. If anyone else has strong feelings on this please chime in, otherwise I'll leave it as-is to avoid extra work.

jhasse: Given that this isn't new code, but just moved, I would say renaming this doesn't need to be part of this commit. The typedef will be less needed anyway, when we can use `auto` in the future.

src/status.h
EliRibble: Okay, since @colincross hasn't chimed in, I'm going to remove it and see if anything breaks.

src/status.cc
EliRibble: Looks like all of the files have a copyright header. I think this just maintains consistency with existing implementation files. I was unable to find anything in the source indicating a requirement to sign an agreement around copyright, or how to specify copyright on new files, so I'm not sure what the team maintaining ninja expect in this regard.

jonesmz: The official project rules are that the copyright line says copyright Google. This is completely wrong unless the person who wrote the file is either employed by Google, or signs a copyright transfer agreement. 

You're not going to get a different answer from the official project maintainers. So you'll have to make a judgment call 

Since @colincross submitted the original commit for this file, its not clear to me that changing it to something else would be appropriate either.

Its a crappy project rule that results in this weird outcome.

src/status.h
colincross: This was mutable in the original, this change just moves it to a new file.

src/status.cc
colincross: Sorry, missed this.  Yes, I contributed this as a Google employee.

jhasse:
Can you rebase on master?

EliRibble:
Absolutely, will do in the next few hours.

EliRibble:
I've been working on this rebase as I can. It's taken quite a bit longer
than I expected. There's lots of great feedback on the PR. In doing the
rebase I'm starting to realize how meandering the many commits are. It may
be better for me to start over and create a PR that does the same work in
far fewer commits. This will make for less working for reviewers.

jhasse:
Did you have a look at my rebase? https://github.com/jhasse/ninja/commits/serialize I've recently done it for 1.10, the tests don't compile yet though.

jhasse:
Just the first commit: #1866

jhasse:
Are you still working on this? If so, please rebase and reopen.

EliRibble:
I'm no longer working on this - as mentioned in the discussion at https://github.com/ninja-build/ninja/issues/1499 - it seems the maintainers are not interested since they would not engage in conversation around it.

PR #1637, opened 2019-08-20
Improve error messages when ninja commands fail on Windows.

beevik:
Leading spaces cause the Win32 CreateProcessA function to
generate an ERROR_INVALID_PARAMETER error. This change
strips leading spaces from commands to prevent the error.

Also, output the command that caused CreateProcessA to result
in an error. This will help users track down the cause of a
ninja build failure.

### DIFF
--- a/src/subprocess-win32.cc
+++ b/src/subprocess-win32.cc
@@ -124,12 +124,20 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
       buf_ = "CreateProcess failed: The system cannot find the file "
           "specified.\n";
       return true;
-    } else if (error == ERROR_INVALID_PARAMETER) {
-      // This generally means that the command line was too long. Give extra
-      // context for this case.
-      Win32Fatal("CreateProcess", "is the command line too long?");
     } else {
-      Win32Fatal("CreateProcess");    // pass all other errors to Win32Fatal
+      fprintf(stderr, "\nCreateProcess failed. Command attempted:\n\"%s\"\n",
+              command.c_str());
+      const char* hint = NULL;
+      // ERROR_INVALID_PARAMETER means the command line was formatted
+      // incorrectly. This can be caused by a command line being too long or
+      // leading whitespace in the command. Give extra context for this case.
+      if (error == ERROR_INVALID_PARAMETER) {
+        if (command.length() > 0 && (command[0] == ' ' || command[0] == '\t'))
+          hint = "command contains leading whitespace";
+        else
+          hint = "is the command line too long?";
+      }
+      Win32Fatal("CreateProcess", hint);
     }
   }
 


atetubou:
I think stripping white space is not the thing ninja needs to do.
Why not fix build.ninja instead?

beevik:
Suppose your `build.ninja` file has an entry like this:

```
command = $pre_link gcc -o $out @$out.rsp $all_ldflags  $libs  $post_build
```

Then suppose that `$pre_link` is undefined.  This leads to a space character before the `gcc` command.  The leading space character causes `CreateProcessA` to fail.

It's true that the problem can be fixed by changing the `build.ninja` file, but it took me a lot of work to track down the cause.  Ninja didn't tell me why the command failed.  It just said "CreateProcess failed (is the command line too long?)".

It also seems strange that ninja would even allow a command with leading whitespace.  It feels like that should be reported as an error, or the whitespace should be stripped.  In the interest of user-friendliness, my pull request stripped the whitespace.

atetubou:
Even in such case, I think you need to fix build.ninja instead.
Also this breaks build running binary like '    .exe'?

Error message improvement looks good to me though.

beevik:
The example I gave is the one that built me.  Since it was in the `link` rule, it prevented me from building the executable.

If you're not willing to allow a correction of the white-space, can we at least include an error message if the command includes leading white space?  That would have prevented me from wasting a couple hours building ninja and trying to understand why it failed in the first place.

It's still concerning to me that some `build.ninja` files will end up working on one platform but not another.  Unless they test on Windows, other people will end up shipping ninja files that work on linux and Mac without realizing they are broken on Windows.

jonesmz:
I think that for the whitespace stripping to be accepted, it would have to be demonstrated that all possible build rules which have leading whitespace will always be broken.

Otherwise, if ninja strips the whitespace, it would prevent a build rule that intentionally had leading whitespace from functioning.

beevik:
>Otherwise, if ninja strips the whitespace, it would prevent a build rule that intentionally had leading whitespace from functioning.

Out of curiosity, how would you create a rule that has leading whitespace?  According to [the documentation](https://ninja-build.org/manual.html):

>In a name = value statement, whitespace at the beginning of a value is always stripped.

The problem with leading whitespace in my case is arising because of a combination of two effects:
1. The `command` value starts with a substituted variable name, and
2. The substituted variable is not defined.

Consider the original example:
```
rule link
  command = $pre_link gcc -o $out @$out.rsp $all_ldflags $libs $post_build
```
Because `$pre_link` is not defined, this leads to a `command` value with leading whitespace, which seems to violate the expectation described by the documentation.  `CreateProcessA` doesn't like the command, since it's trying to  run `" gcc"` (with a leading space) instead of `"gcc"`.

On the other hand, if I do this:
```
rule link
  command = $pre_link gcc -o $out @$out.rsp $all_ldflags $libs $post_build
  pre_link =
```
then ninja does what the documentation describes and strips the leading whitespace from the `command` value.

I'm currently working around this issue by defining `pre-link` to be the empty string.

jonesmz:
Ah.

Thank you for explaining more.

I now agree with you that stripping the white space would be the correct behavior.

beevik:
@jonesmz It turns out that I was wrong about my second example:
```
rule link
  command = $pre_link gcc -o $out @$out.rsp $all_ldflags $libs $post_build
  pre_link =
```
This work-around doesn't actually solve the problem.  Even if `pre_link` is defined as the empty string, you still end up with a space in front of the `gcc` command.  (The reason it was working for me is that I was still running my own version of ninja that stripped the leading whitespace.)

Of course, stripping leading whitespace from a command will allow the example above to work.  But since my explanation above was not 100% correct, I don't know if this affects your conclusion that leading white-space stripping should still be performed.

jonesmz:
@beevik I should point out that the only person who regularly interacts with this Ninja project on github, who also has the access-permissions to accept merge requests, is @jhasse

So you don't have to convince me one way or another.

But I don't believe my re-formed opinion has changed.

As you pointed out, the ninja documentation does clearly state that leading whitespace will be stripped, so it's not *supposed* to be possible to construct a build rule that has leading whitespace, even if there is a loop-hole that allows it to happen in the current code.

As a user of Ninja, it is not my expectation that an empty variable will result in the leading whitespace of the fully calculated build rule not being fully stripped.

atetubou:
I still don't think stripping whitespace after variable expansion is correct.
The document does not say exactly about the behavior when first variable is empty, but I think we can interpret that stripping is only happen before variable expansion.

Also, stripping in subprocess-win32.cc may remove intententional whitespace which can be given by '$ '?

beevik:
@atetubou wrote:
> Also, stripping in subprocess-win32.cc may remove intententional whitespace which can be given by '$ '?

This is a good point.  My change would have stripped out that whitespace, inappropriately.

I think I'll just keep my pull request as is (without the white-space stripping).

PR #1644, opened 2019-08-30
Various continuous integration improvements

jonesmz:
Generic: 
- Fix misc/ci.py for windows hosts

Travis CI:
- Add Ubuntu Bionic
- Add more XCode targets
- Configure homebrew using some environment variables that skip several very slow, unnecessary, steps.

Appveyor:
- Ensure misc/ci.py runs on windows hosts
- Run re2c install on mingw target in the dedicated "install" step, instead of during the build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,16 +18,28 @@ matrix:
     - os: linux
       dist: xenial
       compiler: clang
+    - os: linux
+      dist: bionic
+      compiler: gcc
+    - os: linux
+      dist: bionic
+      compiler: clang
     - os: osx
       osx_image: xcode9.4
     - os: osx
       osx_image: xcode10
     - os: osx
       osx_image: xcode10.1
+    - os: osx
+      osx_image: xcode10.2
+    - os: osx
+      osx_image: xcode10.3
+    - os: osx
+      osx_image: xcode11
 sudo: false
 language: cpp
 before_install:
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install re2c             ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then HOMEBREW_NO_AUTO_UPDATE=1 HOMEBREW_NO_INSTALL_CLEANUP=1 brew install re2c ; fi
   - if [[ "$TRAVIS_OS_NAME" == "windows" ]]; then choco install re2c python ; fi
 script:
   - ./misc/ci.py
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,24 +21,27 @@ matrix:
       MSYSTEM: MSVC
 
 for:
-  -
-    matrix:
+  - matrix:
       only:
         - MSYSTEM: MINGW64
+    install:
+      ps: "C:\\msys64\\usr\\bin\\bash -lc \"pacman -S --quiet --noconfirm --needed re2c 2>&1\""
     build_script:
       ps: "C:\\msys64\\usr\\bin\\bash -lc @\"\n
-      pacman -S --quiet --noconfirm --needed re2c 2>&1\n
+      ./misc/ci.py 2>&1\n
       ./configure.py --bootstrap --platform mingw 2>&1\n
       ./ninja all\n
       ./ninja_test 2>&1\n
       ./misc/ninja_syntax_test.py 2>&1\n\"@"
-  -
-    matrix:
+
+  - matrix:
       only:
         - MSYSTEM: MSVC
     build_script:
     - cmd: >-
         call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvars64.bat"
+
+        python misc/ci.py
 
         python configure.py --bootstrap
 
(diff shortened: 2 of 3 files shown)


appveyor.yml
```diff
@@ -21,25 +21,28 @@ matrix:
       MSYSTEM: MSVC
 
 for:
-  -
-    matrix:
+  - matrix:
       only:
         - MSYSTEM: MINGW64
+    install:
```
jhasse: Any advantage of splitting it up?

appveyor.yml
```diff
@@ -21,25 +21,28 @@ matrix:
...
+      ps: "C:\\msys64\\usr\\bin\\bash -lc \"pacman -S --quiet --noconfirm --needed re2c 2>&1\""
     build_script:
       ps: "C:\\msys64\\usr\\bin\\bash -lc @\"\n
-      pacman -S --quiet --noconfirm --needed re2c 2>&1\n
+      ./misc/ci.py 2>&1\n
```
jhasse: It actually would be best to run this script only once. It shouldn't find anything new on different platforms.

appveyor.yml
jonesmz: AppVeyor provides (slightly) better error output when prerequisite installations are conducted in the `install` section.

It also matches the way I handle appveyor.yaml in my fork of Ninja, which has several other CI targets on appveyor, which all install re2c (I remove the generated code that is stored in git).

appveyor.yml
jonesmz: misc/ci.py is part of the ninja source code, it should be exercised on all platforms where it's easy to do so.

You're correct that if it fails on one platform, it should fail on all platforms, but there are, or might be in the future, downstream forks of Ninja which may not use the same CI targets that Ninja does. (e.g., may only configure a Travis CI account, or an AppVeyor account, but not both).

This change ensures that misc/ci.py is always working, even for downstream projects that only use a subset of the CI configurations.

.travis.yml
```diff
@@ -18,16 +18,28 @@ matrix:
...
+      osx_image: xcode10.2
+    - os: osx
+      osx_image: xcode10.3
+    - os: osx
+      osx_image: xcode11
 sudo: false
 language: cpp
 before_install:
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install re2c             ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then HOMEBREW_NO_AUTO_UPDATE=1 HOMEBREW_NO_INSTALL_CLEANUP=1 brew install re2c ; fi
```
jonesmz: Note: TravisCI does have a specific syntax for installing prereqs that's supposed to behave "optimally" with homebrew.

It doesn't work, and the error output that it gives does not hint at any way to resolve the problem.

So while I'd rather use the dedicated syntax, this is what works, so it's what I'm proposing to be merged.

misc/ci.py
```diff
@@ -3,11 +3,15 @@
 import os
 
 ignores = [
+	'ninja_deps',
 	'.git/',
 	'misc/afl-fuzz-tokens/',
-	'ninja_deps',
 	'src/depfile_parser.cc',
 	'src/lexer.cc',
+	'.git\\',
```
jonesmz: I would have preferred to have the python code here dynamically translate the path separators... but I do not know python very well, so doing it "properly" was beyond the level of effort that I wanted to put into this.

appveyor.yml
```diff
@@ -21,25 +21,28 @@ matrix:
...
       ./misc/ninja_syntax_test.py 2>&1\n\"@"
-  -
-    matrix:
+
+  - matrix:
       only:
         - MSYSTEM: MSVC
     build_script:
     - cmd: >-
         call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvars64.bat"
```
jonesmz: Unfortunately, there's no way to remove the extra newlines in the file.

For reasons unknown, appveyor interprets lines on the NEXT line as part of the previous line (without whitespace between them...?)

jonesmz:
By my measurements, this change (despite adding additional targets) reduces the overall time to run continuous integration by approximately 5 minutes.

appveyor.yml
jonesmz: Supplemental information 

The `install` section for the MINGW target looks identical to how it's done in the `build_script` section, because it is.

But AppVeyor has special handling for the other package managers that they support, when used in the `install` section.

That's not obvious in this PR because the Ninja project is only using the downloaded re2c package on the MINGW target (relying on the bundled pre-generated code for the others), but using the `install` section is significantly nicer looking for the other AppVeyor platform types.

jhasse:
> By my measurements, this change (despite adding additional targets) reduces the overall time to run continuous integration by approximately 5 minutes.

Why is that?

jonesmz:
> Why is that?

The additional flags to the homebrew installation of re2c. They cause homebrew to skip several very expensive, but completely unnecessary (for our situation) operations.

.travis.yml
jhasse: I see, thanks for the info.

jhasse:
We should only do the ci.py changes properly (use `os.path.seperator()` or something like that). 

Can you remove the change and the calls to the script from the Windows builds for now?

jhasse:
Now that we have GitHub Actions (Homebrew is fast there by default) , I want to phase out the other CI set-ups.

PR #1647, opened 2019-09-08
Small constifications

Hi-Angel:
So, as part of rework that turned out necessary in pull 1577, I'm slowly constifying functions. There's a lot more left, but it's unfortunately a complex work that I do in free time, so I decided to pull in the few changes I already have. Immutability is always good to have anyway.

### DIFF
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -406,16 +406,16 @@ std::string Edge::GetBinding(const std::string& key) const {
   return env.LookupVariable(key);
 }
 
-bool Edge::GetBindingBool(const string& key) {
+bool Edge::GetBindingBool(const string& key) const {
   return !GetBinding(key).empty();
 }
 
-string Edge::GetUnescapedDepfile() {
+string Edge::GetUnescapedDepfile() const {
   EdgeEnv env(this, EdgeEnv::kDoNotEscape);
   return env.LookupVariable("depfile");
 }
 
-string Edge::GetUnescapedDyndep() {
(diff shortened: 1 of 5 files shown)


src/graph.cc
```diff
@@ -406,16 +406,16 @@ std::string Edge::GetBinding(const std::string& key) const {
   return env.LookupVariable(key);
 }
 
-bool Edge::GetBindingBool(const string& key) {
+bool Edge::GetBindingBool(const string& key) const {
   return !GetBinding(key).empty();
```
jonesmz: Can this function be changed to accept the "StringPiece" class instead?

Hi-Angel: I don't think so: `GetBinding()` internally passes a string reference over to `EdgeEnv::LookupVariable()`, which in turn exploits string API, e.g. it compares the string with other strings.

Hi-Angel:
Thanks!

src/graph.cc
jonesmz: Ah.

Guess we'll need to wait until we have access to std::less<> then.

Hi-Angel:
@jhasse btw, I'm wondering: why did you collapse the commits? If a commit introduced a regression on some platform, it would be much easier to bisect it because multiple commits is less code than one commit. So you don't need to detangle from one large commit which `const`s you need to undo to fix the problem.

jhasse:
Because the commits were relatively small. But in general I agree with you, I might have made the wrong call there, sorry.

PR #1649, opened 2019-09-19
handle EAGAIN when spawning new processes

wbbradley:
In Linux it is possible for various syscalls to return EAGAIN when the
system is too busy to handle more load. This is sad. Currently ninja
will just stop in the event of an EAGAIN error. This change allows ninja
to backoff a bit and continue trying to get work done. Note that we have
seen this issue crop up when various parallel build systems are running
concurrently on a single machine.

### DIFF
--- a/src/subprocess-posix.cc
+++ b/src/subprocess-posix.cc
@@ -29,6 +29,18 @@ extern char** environ;
 
 #include "util.h"
 
+#define HANDLE_EAGAIN(x) ({ \
+  __typeof__(x) eintr_wrapper_result; \
+  do { \
+    eintr_wrapper_result = (x); \
+    if (eintr_wrapper_result == -1 && errno == EAGAIN) \
+      sleep(1); \
+    else \
+      break; \
+  } while (1); \
+  eintr_wrapper_result; \
+  })
+
 Subprocess::Subprocess(bool use_console) : fd_(-1), pid_(-1),
                                            use_console_(use_console) {
 }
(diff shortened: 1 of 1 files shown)


wbbradley:
My change needs to be rethought. brb.

src/subprocess-posix.cc
```diff
@@ -29,6 +29,18 @@ extern char** environ;
 
 #include "util.h"
 
+#define HANDLE_EAGAIN(x) ({ \
```
jonesmz: This explicitly uses a non-standard GCC language extension, which is not available on all compilers that might be used for subprocess-posix.cc, and circumvents the compiler's decision making for inlining the function call (or not) as it thinks is best.

Why not make this a function, instead?

Or, since you're only using this macro in exactly one place, just put the code there directly.

src/subprocess-posix.cc
```diff
@@ -29,6 +29,18 @@ extern char** environ;
...
+  __typeof__(x) eintr_wrapper_result; \
+  do { \
+    eintr_wrapper_result = (x); \
+    if (eintr_wrapper_result == -1 && errno == EAGAIN) \
+      sleep(1); \
```
jonesmz: Instead of sleeping, is there any practical work that Ninja might do somewhere else before revisiting the call to posix_spawn ?

src/subprocess-posix.cc
jonesmz: As an alternative thought on the matter, this is basically what C++11 lambda's are for (immediately invoked function expression). I would be more enthusiastic about this change if it used a C++11 lambda instead of a non-standard GCC extension.

However, you're going to need to convince the project maintainers to use C++11 or newer if you want to go that route.

wbbradley: I'll push it into a separate function.

wbbradley:
Sorry, first time using Github pull requests in a while. I've opened a new version of the PR here https://github.com/ninja-build/ninja/pull/1650 which incorporates feedback so far.

src/subprocess-posix.cc
wbbradley: When this condition occurs it is generally a kernel-level race condition for locks. It may be possible to continue to do more work in user-space, but this fix is intended to prevent the ninja process from simply giving up and exiting, which it currently does. If possible, I'd like to avoid trying to solve both of those problems at the same time.

PR #1650, opened 2019-09-19
handle EAGAIN when spawning new processes

wbbradley:
In Linux it is possible for various syscalls to return EAGAIN when the
system is too busy to handle more load. This is sad. Currently ninja
will just stop in the event of an EAGAIN error. This change allows ninja
to backoff a bit and continue trying to get work done. Note that we have
seen this issue crop up when various parallel build systems are running
concurrently on a single machine.

### DIFF
--- a/src/subprocess-posix.cc
+++ b/src/subprocess-posix.cc
@@ -29,6 +29,19 @@ extern char** environ;
 
 #include "util.h"
 
+int HandleEAGAINPosixSpawn(pid_t* pid, const char* path,
+                           const posix_spawn_file_actions_t* file_actions,
+                           const posix_spawnattr_t* attrp, char* const argv[],
+                           char* const envp[]) {
+  while (true) {
+    int ret = posix_spawn(pid, path, file_actions, attrp, argv, envp);
+    if (ret == -1 && errno == EAGAIN)
+      sleep(1);
+    else
+      return ret;
+  }
+}
+
 Subprocess::Subprocess(bool use_console) : fd_(-1), pid_(-1),
                                            use_console_(use_console) {
 }
@@ -111,8 +124,8 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
     Fatal("posix_spawnattr_setflags: %s", strerror(err));
 
   const char* spawned_args[] = { "/bin/sh", "-c", command.c_str(), NULL };
-  err = posix_spawn(&pid_, "/bin/sh", &action, &attr,
-        const_cast<char**>(spawned_args), environ);
+  err = HandleEAGAINPosixSpawn(&pid_, "/bin/sh", &action, &attr,
+                               const_cast<char**>(spawned_args), environ);
   if (err != 0)
     Fatal("posix_spawn: %s", strerror(err));
 


src/subprocess-posix.cc
```diff
@@ -29,6 +29,19 @@ extern char** environ;
...
 #include "util.h"
 
+int HandleEAGAINPosixSpawn(pid_t* pid, const char* path,
+                           const posix_spawn_file_actions_t* file_actions,
+                           const posix_spawnattr_t* attrp, char* const argv[],
+                           char* const envp[]) {
+  while (true) {
+    int ret = posix_spawn(pid, path, file_actions, attrp, argv, envp);
+    if (ret == -1 && errno == EAGAIN)
+      sleep(1);
```
nico: I'm not sure silently being slow is better than failing quickly. This happens if you oversubscribe your build box (like it happened on that cros bug where this is from); a better fix to me seems to fix the bot instead.

I'd kind of buy the patch without the sleep, for spec compat reasons, but imho if you run into a situation where you need this, something already went wrong earlier.

jhasse: At the very least some kind of logging (warning) should happen.

jonesmz: I'd point out that running the loop indefinitely is also not a great choice. At worst, put a limit of 1000 retries or something, so it doesn't just continue failing forever.

wbbradley: I think the right answer here is not obvious. Clearly everyone has a good point. However, unless someone is intimately familiar with the semantics of posix_spawn(), they will expect it to block until it is complete. Performance vs. Correctness is another question. Again, this comes down to "what is the user expectation when they run ninja?" Do the expect ninja to be mindful of their system resources and not retry upon any sort of system-level hiccup (thus pushing the design in Linux up to the user) or do they just want the build to push through, waiting until it has a chance to run?

How would folks feel if I plumbed a ninja command-line option (and/or env var) down here so that users could elect one behavior or another?

jonesmz: I think that for purposes of this PR, it might be better for you to make the policy be a parameter to the Subprocess::Start() function, and default it to the current behavior.

Then once the Ninja project figures out how it wants to handle environment variables, and/or commandline options, it can (potentially) be added later.

I don't think that adding a new cli parameter or environment variable will be accepted at this time.

jonesmz: Consider also consulting the behavior that other implementations of "Spawn a process" use. E.g. Boost.Process (https://github.com/boostorg/process/blob/develop/include/boost/process/detail/posix/executor.hpp)

I'm not saying that I have an answer for you, simply that consulting other implementations may give Ninja some wisdom on potential pitfalls.

src/subprocess-posix.cc
```diff
@@ -29,6 +29,19 @@ extern char** environ;
 
 #include "util.h"
 
+int HandleEAGAINPosixSpawn(pid_t* pid, const char* path,
+                           const posix_spawn_file_actions_t* file_actions,
+                           const posix_spawnattr_t* attrp, char* const argv[],
+                           char* const envp[]) {
+  while (true) {
+    int ret = posix_spawn(pid, path, file_actions, attrp, argv, envp);
+    if (ret == -1 && errno == EAGAIN)
```
jonesmz: Something to be careful of:

posix_spawn returns the same error codes as fork and vfork (Since it calls one of those under the hood).

From: http://man7.org/linux/man-pages/man2/fork.2.html

> 
>        EAGAIN A system-imposed limit on the number of threads was
>               encountered.  There are a number of limits that may trigger
>               this error:
> 
>               *  the RLIMIT_NPROC soft resource limit (set via
>                  setrlimit(2)), which limits the number of processes and
>                  threads for a real user ID, was reached;
> 
>               *  the kernel's system-wide limit on the number of processes
>                  and threads, /proc/sys/kernel/threads-max, was reached (see
>                  proc(5));
> 
>               *  the maximum number of PIDs, /proc/sys/kernel/pid_max, was
>                  reached (see proc(5)); or
> 
>               *  the PID limit (pids.max) imposed by the cgroup "process
>                  number" (PIDs) controller was reached.
> 

So EAGAIN doesn't happen if you create processes too *quickly*. It happens when the number of processes exceeds some limit.

I don't know that simply looping will actually address the problem you want to address.

A more comprehensive solution to this issue is

1) Have a group of tasks that are supposed to be "alive" according to the dependency graph. These tasks can be in the "waiting to launch" state, the "launched" state, and the "finished" state.
2) When a new task is added to that group, attempt to launch it immediately, if possible.
3) When a task enters the finished state, attempt to launch other tasks that are in the "waiting to launch" state, in addition to the other current activities that Ninja would do (e.g. add the next set of tasks to the list of child processes)
4) Periodically (some user and/or system specified time period) attempt to launch one or more of the "waiting to launch" tasks.
5) Tasks attempting to transition from "waiting to launch" to "launched" only stay in the "waiting to launch" state if the error code from posix_spawn indicates that attempting to launch the task again later can possibly succeed.

wbbradley: Yeah, I had started to think about whether it would make any sense to methodically lower the number of concurrent processes allowed whenever EAGAIN is specified, altering the maximum parallelism ninja feels like it can utilize. This would require the command runner code to have a more flexible return code that would effectively re-queue a task to be run whenever a process "resource" is available. That breaks some nice encapsulation in the Edge/Node code. And, it is indeed more complex. At some point the rubber would have to meet the road and we'd have to decide whether ninja should be allowed to keep trying to fork or not. Going to have to think on it a bit. I'm open to more ideas, keep 'em coming... Again, thanks for the feedback.

src/subprocess-posix.cc
jhasse: I would expect the build to push through, waiting until it has a chance to run. IMHO your patch is fine, it should just not do this silently, but warn the user.

wbbradley:
I'm no longer working on a project where Ninja is a dependency. Going to go ahead and close this due to it dropping in priority for me. Feel free to reopen/reassign as necessary. Thanks for your consideration!

Issue #1418, opened 2018-04-07
Edges with output have wrong status line after #1405


colincross:
#1405 updates the status line to show the oldest running edge instead of the most recently finished edge, but does so before the edge output has been printed.  For example, this file:
```
rule echo
  command = sleep $delay && echo $out
  description = echo $out

build a: echo
  delay = 3
build b: echo
  delay = 2
build c: echo
  delay = 1
```

Produces:
```
[1/3] echo a
c
[2/3] echo a
b
[3/3] echo a
a
```

When it should produce:
```
[1/3] echo c
c
[2/3] echo b
b
[3/3] echo a
a
```

jhasse:
Thanks for the detailed description! I've used your test case to add an automated test for this to Travis CI: https://github.com/ninja-build/ninja/blob/3841023a44621e8d8133e9eb6e79e26481dc0bc3/misc/output_test.py#L34

PR #1651, opened 2019-09-20
Fix test_issue_1418 to pass on 1-core VM

bmwiedemann:
the previous assert would fail because on a 1-core VM, the 3 outputs
were produced sequentially from top to bottom

This PR was done while working on reproducible builds for openSUSE.


jhasse:
Or pass `-j3` to ninja?

bmwiedemann:
Updated it. This is related to #1418

PR #1654, opened 2019-09-24
Ensure substring indexing never goes negative

kstrafe:
With widths lower than 4, the ElideMiddle function would crash because
its substring access would wrap around and attempt to access the max
size_t value. This patch fixes that.

This situation occasionally occurs on i3 with tmux, where  a new window may make the tmux window sub-3 in width. Ninja then crashes.

### DIFF
--- a/src/util_test.cc
+++ b/src/util_test.cc
@@ -420,6 +420,10 @@ TEST(ElideMiddle, NothingToElide) {
   string input = "Nothing to elide in this short string.";
   EXPECT_EQ(input, ElideMiddle(input, 80));
   EXPECT_EQ(input, ElideMiddle(input, 38));
+  EXPECT_EQ("", ElideMiddle(input, 0));
+  EXPECT_EQ(".", ElideMiddle(input, 1));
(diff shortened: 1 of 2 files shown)


jhasse:
Shouldn't we return

```
"" if width == 0
"." if width == 1
".." if width == 2
"..." if width == 3
```

?

src/util_test.cc
```diff
@@ -420,6 +420,9 @@ TEST(ElideMiddle, NothingToElide) {
   string input = "Nothing to elide in this short string.";
   EXPECT_EQ(input, ElideMiddle(input, 80));
   EXPECT_EQ(input, ElideMiddle(input, 38));
+  for (int idx = 0; idx < 5; ++idx) {
+    EXPECT_EQ("...", ElideMiddle(input, idx));
```
jhasse: Why does it return "..." for idx == 4?

BourgondAries:
@jhasse it now returns according to the above, can we merge?

jhasse:
Looks good. The above was a question though, I'm not 100% sure if that is the correct behavior.

PR #1662, opened 2019-10-25
Log version

kasper93:
This changes allow to handle new log format in other tools that were broken by cfd0bd3.

Thanks,
Kacper

### DIFF
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -384,7 +384,8 @@ bool BuildLog::Recompact(const string& path, const BuildLogUser& user,
     return false;
   }
 
-  if (fprintf(f, kFileSignature, kCurrentVersion) < 0) {
+  if (fprintf(f, kFileSignature, kCurrentVersion) < 0 ||
+      fprintf(f, kFileColumnLabels) < 0) {
     *err = strerror(errno);
     fclose(f);
     return false;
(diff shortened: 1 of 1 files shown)


nico:
@jhasse Maybe we should just revert cfd0bd3007b291df505f8c45083453310142d681 instead?

jhasse:
Which tools where broken by cfd0bd3? I thought it was a backwards compatible change since it only adds a comment.

kasper93:
> Which tools where broken by cfd0bd3?

Specifically ninjatracing, see https://github.com/nico/ninjatracing/pull/17 

> I thought it was a backwards compatible change since it only adds a comment.

I agree, but in the same time, as we can see not everyone skip `#` lines. I proposed one way to handle the fallout after this change. I don't see much gain from having this header line, but in the same time I don't care as long as everything works as it should.

jhasse:
Yeah, I think it would be a good idea for everyone to skip # lines.

I don't like bumping the version for such a minor change. Revert or fix in ninjatracing?

kasper93:
Actually I am more inclined to reverting the change. Quick search shows that it would need fixing also things like https://chromium.googlesource.com/chromium/tools/depot_tools.git/+/master/post_build_ninja_summary.py

It is not a problem to fix that, but still someone will need to do that :)

jhasse:
Done. Thanks for finding this.

Sorry @chagui

kasper93:
@nico: I see you merged my PR on ninjatracing. Please revert it there too in this case as it is not valid change anymore.

Issue #12, opened 2011-03-08
A problem of inline.sh


iwadon:
Hello,
I found a problem when run bootstrap.sh on Mac OS X and FreeBSD.

```
% ./bootstrap.sh 
Building ninja manually...
In file included from src/browse.cc:20:
src/../build/browse_py.h:249: error: \x used with no following hex digits
src/../build/browse_py.h:249: error: \x used with no following hex digits
(many same lines...)
```

Because od command prints many blank characters.

```
% cat src/browse.py | od -t x1 -A n -v | head -3
           23  21  2f  75  73  72  2f  62  69  6e  2f  70  79  74  68  6f
           6e  0a  23  0a  23  20  43  6f  70  79  72  69  67  68  74  20
           32  30  30  31  20  47  6f  6f  67  6c  65  20  49  6e  63  2e
```

I hope my gist help you.
https://gist.github.com/860677

Issue #11, opened 2011-02-27
File creation in unit test not tested resulting in Segmentation fault


sebbu2:
When I tried to run ninja_test from a directory where I don't have write access, I got a segmentation fault as showed lower, instead of the standard "stat(xx[truncated]xx): File name too long" error message then Success

Actual Error Log :
[----------] 3 tests from DiskInterfaceTest
[ RUN      ] DiskInterfaceTest.Stat
Segmentation fault

Issue #10, opened 2011-02-24
Handle variable expansion with upper-case characters


i-am-orestis:
**EDIT**

Hi martine, thanks for the response.

I have included a test case for variables with uppercase characters, named OneVariableUpperCase. It is basically a copy-paste of the OneVariable test with the variable name changed.

Here is the new patch which includes both the changes to the parser and the test case.

```
diff -Naur ninja/src/eval_env.cc ninja-new/src/eval_env.cc
--- ninja/src/eval_env.cc   2011-02-21 00:14:12.000000000 +0200
+++ ninja-new/src/eval_env.cc   2011-02-26 03:29:29.000000000 +0200
@@ -56,7 +56,7 @@
     } else {
       for (end = start; end < input.size(); ++end) {
         char c = input[end];
-        if (!(('a' <= c && c <= 'z') || ('0' <= c && c <= '9') || c == '_'))
+        if (!(('a' <= c && c <= 'z') || ('A' <= c && c <= 'Z') || ('0' <= c && c <= '9') || c == '_'))
           break;
       }
       if (end == start) {
diff -Naur ninja/src/ninja_test.cc ninja-new/src/ninja_test.cc
--- ninja/src/ninja_test.cc 2011-02-21 00:14:12.000000000 +0200
+++ ninja-new/src/ninja_test.cc 2011-02-26 04:35:08.000000000 +0200
@@ -80,6 +80,17 @@
   env.vars["var"] = "there";
   EXPECT_EQ("hi there", str.Evaluate(&env));
 }
+TEST(EvalString, OneVariableUpperCase) {
+  EvalString str;
+  string err;
+  EXPECT_TRUE(str.Parse("hi $VaR", &err));
+  EXPECT_EQ("", err);
+  EXPECT_EQ("hi $VaR", str.unparsed());
+  TestEnv env;
+  EXPECT_EQ("hi ", str.Evaluate(&env));
+  env.vars["VaR"] = "there";
+  EXPECT_EQ("hi there", str.Evaluate(&env));
+}
 TEST(EvalString, Error) {
   EvalString str;
   string err;
```

**ORIGINAL**

Although the definition of variables with upper-case characters is recognized by the parser,
the expansion (usage) of these variables, when not enclosed in { .. }, is not supported.

The problem is located in eval_env.cc, which uses the statement "if (!(('a' <= c && c <= 'z') || ('0' <= c && c <= '9') || c == '_'))" to detect errors with variable names.

[...]

PR #1670, opened 2019-11-08
Fix BuildLogTest.MultiTargetEdge crash problem …

xianglin1006:


jhasse:
Which crash problem?

xianglin1006:
> Which crash problem?

#0  0x00007ffff753dc37 in __GI_raise (sig=sig@entry=6)
    at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007ffff7541028 in __GI_abort () at abort.c:89
#2  0x00007ffff797ffe5 in __gnu_debug::_Error_formatter::_M_error() const ()
   from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x000000000042b434 in std::__debug::vector<Node*, std::allocator<Node*> >::operator[] (
    this=0x798fd0, __n=0) at /usr/include/c++/4.8/debug/vector:353
#4  0x00000000004bb7d3 in EdgeEnv::LookupVariable (this=0x7fffffffe200, var="in")
    at src/graph.cc:348
#5  0x00000000004b487a in EvalString::Evaluate (this=0x7989e8, env=0x7fffffffe200)
    at src/eval_env.cc:105
#6  0x00000000004b4744 in BindingEnv::LookupWithFallback (this=0x794ad8, var="command",
    eval=0x7989e8, env=0x7fffffffe200) at src/eval_env.cc:91
#7  0x00000000004bbb53 in EdgeEnv::LookupVariable (this=0x7fffffffe200, var="command")
    at src/graph.cc:375
#8  0x00000000004bbda5 in Edge::GetBinding (this=0x798fc0, key="command") at src/graph.cc:410
#9  0x00000000004bbc95 in Edge::EvaluateCommand (this=0x798fc0, incl_rsp_file=true)
    at src/graph.cc:399
#10 0x000000000049e9fd in BuildLog::RecordCommand (this=0x7fffffffe410, edge=0x798fc0,
    start_time=21, end_time=22, mtime=0) at src/build_log.cc:158
#11 0x00000000004074d7 in (anonymous namespace)::BuildLogTestMultiTargetEdge::Run (
    this=0x794a00) at src/build_log_test.cc:252
#12 0x00000000004789ce in main (argc=0, argv=0x7fffffffe5d0) at src/ninja_test.cc:151

jhasse:
When does it happen? Ideally this should be reproducible by CI.

xianglin1006:
> When does it happen? Ideally this should be reproducible by CI.

It happenes every time I run ./ninja_test which is compiled by ninja ninja_test.

jhasse:
Can't reproduce, sorry.

xianglin1006:
> Can't reproduce, sorry.

Do you run it in C++11 std or C++98? I found the program crash when run in C++98, it doesn't happen in C++11.

image

For test case MultiTargetEdge ,, there is no input for edge cat, so the sizeof edge_->inputs_ is 0, and addressing it in C++98 will crash

jhasse:
Ah I see! I don't think there's a way around it because ` std::vector::data()` isn't available in C++98.

jhasse:
I think we should add `/Zc:__cplusplus` to configure.py instead.

xianglin1006:
> I think we should add `/Zc:__cplusplus` to configure.py instead.

Yeah,  we could fix it by adding `/Zc:__cplusplus`. But maybe it's better to check whether 'explicit_deps_count ' is greater than zero and provide backwards compatibility to C++98.

jhasse:
I prefer leaving C++98 behind (it's also just happens in debug mode).

xianglin1006:
OK.

PR #1671, opened 2019-11-08
Ninja UTF8 conversion: Using native windows API for conversion.

jlonnberg:
WIP for issue #1195.
Updated version using windows native API for converting between wide and utf8 for increased speed. Fixed buildissue for mingw and added needed defines to the cmake script. 

Metrics for building chromium
`
Ninja Master
metric           	count 	avg (us) 	total (ms)
.ninja parse     	4849  	1931.3  	9364.8
canonicalize str 	1889079	0.0     	58.5
canonicalize path	1985429	0.0     	42.8
lookup node      	1932008	0.0     	36.7
.ninja_log load  	1     	1504.0  	1.5
.ninja_deps load 	1     	1314.0  	1.3
node stat        	7363  	285.0   	2098.4
depfile load     	22    	184.6   	4.1
StartEdge        	823   	5840.9  	4807.1
FinishCommand    	822   	3970.5  	3263.8
CLParser::Parse  	753   	618.3   	465.6

Ninja UTF8-version
metric              	count 	avg (us) 	total (ms)
.ninja parse        	4849  	2104.7  	10205.6
canonicalize str    	1889079	0.0     	90.5
canonicalize path   	1985429	0.0     	60.8
lookup node         	1933885	0.1     	169.9
.ninja_log load     	1     	2000.0  	2.0
.ninja_log recompact	1     	14878.0 	14.9
node stat           	8748  	260.8   	2281.1
.ninja_deps load    	1     	3960.0  	4.0
depfile load        	22    	209.7   	4.6
StartEdge           	823   	5909.6  	4863.6
FinishCommand       	822   	908.7   	746.9
CLParser::Parse     	753   	615.2   	463.2
`

PR #1678, opened 2019-11-13
Use std::move() when statting all files in a directory on windows, to reduce string copies

jonesmz:


jonesmz:
To elaborate on why this is a big improvement:

```
if (ffd.cFileName[0] == '.' && ffd.cFileName[1] == '.') {
```
Skips an allocation and copy if the result wouldn't be used.

```
stamps->emplace(std::move(lowername), TimeStampFromFileTime(ffd.ftLastWriteTime));
```
instead of
```
stamps->insert(make_pair(lowername,
                             TimeStampFromFileTime(ffd.ftLastWriteTime)));
```
Skips *two* allocations and copies.

1. emplace() instead of insert(make_pair()) skips copying the resulting std::pair, which contains a string, which results in an allocation and copy.
2. std::move() on lowername skips an allocation and copy for the std::string that's emplaced into the map.

nico:
Do you have any benchmark data on how much this saves?

jonesmz:
I do not have a windows computer, and this is windows specific code, so I'm unable to run benchmarks for this change.

I would rather see this change implemented directly in https://github.com/ninja-build/ninja/pull/1671, but implemented it in this PR both so it doesn't have to wait for the other UTF-8 changes, and also to stand as an isolated example.

jhasse:
As this change doesn't involve the WinAPI directly, you can test with WINE. Here's how you would do that on Fedora Linux for example:

```
mingw64-cmake -H. -Bbuild-mingw -DCMAKE_BUILD_TYPE=Release -GNinja
ninja -Cbuild-mingw
./misc/measure.py wine build-mingw/ninja.exe -C path/to/big/project/with/no/work/to/do
```

jonesmz:
I don't really see how benchmarking this code is going to do any good.

As I detailed in a previous comment, this change postpones a potentially unneeded allocation+copy, as well as uses rvalue-reference / move to construct the map node, instead of the allocate+copy of std::string.

It should be impossible for this to be slower.

jhasse:
> It should be impossible for this to be slower.

Sure, but we wonder if it's really a "big improvement". Sometimes the optimizer does magic things, therefore it's good to check any performance optimization and not optimize prematurely.

jonesmz:
The C++ standard forbids compilers from implicitly casting variables to rvalue-references when passing those variables to functions as parameters.

The only situations where implicit cast to rvalue-references are allowed is around return-value-optimzation and copy-elision and the like.

src/disk_interface.cc
```diff
@@ -116,15 +116,14 @@ bool StatAllFilesInDir(const string& dir, map<string, TimeStamp>* stamps,
     return false;
   }
   do {
-    string lowername = ffd.cFileName;
-    if (lowername == "..") {
+    if (ffd.cFileName[0] == '.' && ffd.cFileName[1] == '.') {
```
atetubou, line 119: no need to check length of `ffd.cFileName`?

atetubou: Sorry, I think this is fine. Ignore this.

src/disk_interface.cc
```diff
@@ -116,15 +116,14 @@ bool StatAllFilesInDir(const string& dir, map<string, TimeStamp>* stamps,
...
       // Seems to just copy the timestamp for ".." from ".", which is wrong.
       // This is the case at least on NTFS under Windows 7.
       continue;
     }
-    transform(lowername.begin(), lowername.end(), lowername.begin(), ::tolower);
-    stamps->insert(make_pair(lowername,
-                             TimeStampFromFileTime(ffd.ftLastWriteTime)));
+    std::string lowername = ffd.cFileName;
+    std::transform(lowername.begin(), lowername.end(), lowername.begin(), ::tolower);
+    stamps->emplace(std::move(lowername), TimeStampFromFileTime(ffd.ftLastWriteTime));
```
digit-google, line 126: FYI, `emplace()` checks for duplicates _after_ creating a temporary `value_type`, so this change does not avoid a string copy after all. A solution is to use the C++11 equivalent of `try_emplace()` (available since C++17), such as:

```
// C++11 equivalent to stamps->try_emplace(std::move(lowername), TimeStamp(..))
stamps->emplace(std::piecewise_construct,
                std::forward_as_tuple(std::move(lowername)),
                std::forward_as_tuple(TimeStampFromFileTime(ffd.ftLastWriteTime)));
```

Another potential optimization on line 186 to avoid a double-lookup + string copy on insertion is:

```
auto ret = cache_.emplace(std::piecewise_construct,
                          std::forward_as_tuple(std::move(dir_lowercase)));
Cache::iterator ci = ret.first;
if (ret.second) {
  if (!Stat(...)) {
    cache_.erase(ci);
    return -1;
  }
}
```

Hope this helps!

PR #1679, opened 2019-11-14
Rework CMakeLists to use CMake generator expressions

jonesmz:


jhasse:
After reading CMake's documentation, I still don't understand what's the advantage of e.g.

```cmake
target_compile_definitions(libninja PRIVATE $<$<CXX_COMPILER_ID:MSVC>:_WIN32_WINNT=0x0601>)
```

over

```cmake
if(MINGW)
target_compile_definitions(libninja PRIVATE _WIN32_WINNT=0x0601)
endif()
```

(in this particial case it even looks wrong to me, since MSVC != MINGW)

jhasse:
As I understand it now the generator expressions are useful, if a value can change during build time (e.g. Debug vs. Release when generating Visual Studio projects). The platform (Windows) can't change though, therefore this would only be a different syntax. I prefer to use the if though, as it's easier to understand.

(Also MinGW != MSVC so I think this would break MinGW builds with CMake)

jonesmz:
This PR did a lot more than just replace "if" with generator expressions.

Among other things, it made sure to use the "target_compiler_definitions" function, instead of modifying CXX_FLAGS directly.

jhasse:
That sounds like a good change, although I would prefer to use `add_compile_definitions` so that we can keep multiple targets (and don't have to re-add the flags if we add new target later).

PR #1680, opened 2019-11-14
Fix invalid preprocessor #if

jonesmz:


jhasse:
Should not make difference since `_WIN32` is defined to 1, right?

jonesmz:
Causes warnings on my compiler (gcc 9.2)

jhasse:
Interesting

jonesmz:
If the value isn't defined, using #if is invalid (needs to be #ifdef). So gcc warns, and says that it assumes the value must be zero.

Issue #1533, opened 2019-02-22
Ninja 1.9.0, CentOS 7: GLIBCXX symbols not found


abthompson:
**Background**
We build/package our deliverable in a CentOS 7 Docker container that has the minimum-required dependencies for our build. This includes Ninja 1.8.2, which we download directly from GitHub during the image building process:

```bash
# Download the archive
curl -LO .../ninja-linux.zip

# Unarchive it to a location that should already be in the PATH
unzip -d /usr/local/bin ninja-linux.zip
```

I'm investigating upgrading our dependencies (CMake, etc.) and noticed that Ninja is now up to 1.9.0. Before updating my image, I wanted to make sure it wouldn't have any issues building our code, so I spawned a barebones CentOS 7 container, downloaded/unarchived the official Linux binary archive, and ran `ninja --version` to try to verify I'm running 1.9.0.

```bash
# Download the archive
curl -LO .../ninja-linux.zip

# Working locally for demonstration purposes
unzip ninja-linux.zip

# The unarchived file was not executable
chmod +x ninja
```

The following documents/demonstrates the issues I ran into.

**Notes**
- I ran the following commands in the same barebones CentOS 7 Docker container
- I renamed the `ninja` executables according to their respective release version

**Checking the `ninja` version**
```
$ ./ninja-1.8.2 --version
1.8.2

$ ./ninja-1.9.0 --version
./ninja-1.9.0: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by ./ninja-1.9.0)
./ninja-1.9.0: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by ./ninja-1.9.0)
```
It seems the official binary was built with symbols that aren't found in the system-provided GCC 4.8.5 `libstdc++.so.6`. Determined via:
```
$ ldd -v ninja-1.9.0
./ninja-1.9.0: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by ./ninja-1.9.0)
./ninja-1.9.0: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by ./ninja-1.9.0)
        linux-vdso.so.1 =>  (0x00007ffffd727000)
        libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007ffb8c86d000)
        libc.so.6 => /lib64/libc.so.6 (0x00007ffb8c4ac000)
[...]

PR #1681, opened 2019-11-14
Add GitHub Actions workflow for building release binaries

nightlark:
This adds a GitHub Actions workflow that builds Ninja release binaries for Windows, Linux, and macOS on each commit and stores them as artifacts. For published releases on GitHub, the compiled release binary archives are uploaded as assets.

Currently CMake is used instead of the Python script bootstrapping process, because it kept the steps more similar across platforms (using configure.py on Windows requires extra steps with vcvarsall.bat to make so that the MSVC compiler is found). If configure.py is preferred it should be fairly easy to switch.

Another issue that was mentioned is using the ubuntu-latest image (LTS 18.04), and wanting a binary that works with older versions of Linux. Two ideas to address this are static linking with system libraries, or switching the build to use a Docker image that has an older version of glibc and stdc++. These should also address issue #1533, which using Ubuntu 16.04 as the build image might not.

### DIFF
--- a/.github/workflows/release-ninja-binaries.yml
+++ b/.github/workflows/release-ninja-binaries.yml
@@ -40,6 +40,10 @@ jobs:
         cmake -DCMAKE_BUILD_TYPE=Release ..
         cmake --build . --parallel --config Release --target ninja
         
+    - name: Strip Linux binary
+      if: matrix.os == 'ubuntu-latest'
+      run: cd build && strip ninja
+        
     - name: Create ninja archive
       shell: bash
       env:
--- a/.github/workflows/release-ninja-binaries.yml
+++ b/.github/workflows/release-ninja-binaries.yml
@@ -1,6 +1,9 @@
 name: Release Ninja Binaries
 
-on: [push]
+on:
+  push:
+  release:
+    types: published
 
 jobs:
   build:
@@ -37,7 +40,7 @@ jobs:
         cmake -DCMAKE_BUILD_TYPE=Release ..
         cmake --build . --parallel --config Release --target ninja
         
-    - name: Create artifact
+    - name: Create ninja archive
       shell: bash
       env:
         ZIP_NAME: ${{ matrix.zip_name }}
@@ -52,4 +55,13 @@ jobs:
         name: ninja-binary-archives
         path: artifact
         
-        
+    - name: Upload release asset
+      if: github.event.action == 'published'
+      uses: actions/upload-release-asset@v1.0.1
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        upload_url: ${{ github.event.release.upload_url }}
+        asset_path: ./artifact/${{ matrix.zip_name }}.zip
+        asset_name: ${{ matrix.zip_name }}.zip
+        asset_content_type: application/zip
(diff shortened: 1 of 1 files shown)


.github/workflows/release-ninja-binaries.yml
```diff
@@ -0,0 +1,71 @@
...
+      uses: actions/upload-artifact@v1
+      with:
+        name: ninja-binary-archives
+        path: artifact
+
+    - name: Upload release asset
+      if: github.event.action == 'published'
+      uses: actions/upload-release-asset@v1.0.1
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
jhasse: Do I have to set this up somehow?

nightlark: The nice thing with GitHub Actions is that nothing needs to be done to set up the token (I think they create a token as needed with expiration time of a few hours).

Commits directly to the repository use a token that has read/write access. For security, PRs from forked repositories have a token with read-only access.

nightlark:
I added a trigger for pull requests that should make it show up under checks for PRs from forked repositories.

Testing it with some forks I created, the checks seem to show up as expected in the PRs when adding a new workflow. I think it's most likely just a permissions thing, since the workflow is being added in a PR from a fork without collaborator/write access to this repository.

Some of the test PRs between forks:
https://github.com/nightlark/ninja/pull/3
https://github.com/ryan-mast/ninja/pull/1

jhasse:
Ah I see.

Can you switch to Ubuntu 16.04? Better than nothing.

We also need to link Visual Studio's C++ runtime statically on Windows.

nightlark:
Mm yea, I'll see if I can get those changes made within the next couple of days.

jhasse:
I've merged this PR for now so that we can see the result on this repository in action.

(Thank you very much btw!)

nightlark:
It seems to be working for forks, based on open PRs like #1674 that have had commits since the merge.

A quick update on creating Linux binaries -- I think I've figured out how to build a binary that will take care of issue #1533 (CentOS 7 GLIBCXX symbols not found), and that should work with glibc-based Linux distros released since around 2011.

mathstuf:
Using CentOS 6 was our solution before we ended up needing other newer support (CentOS 7 is suitable for us since the older Debian/Ubuntu releases it "jumped" over aren't supported by that project either).

jhasse:
Automatic uploading of release assets didn't work: https://github.com/ninja-build/ninja/releases/tag/v1.10.0
@nightlark Do you have an idea why?

nightlark:
I’d guess that it was just a v1.10.0 tag that was created, rather than publishing a release like was done for v1.9.0? If you create/publish a release for the v1.10.0 tag it should trigger the build and upload the assets.
 
It should be fairly straightforward (though simpler when all the builds are in the same workflow file) to change the action to trigger on a `v*` tag, and create a release for uploading the assets to.

jhasse:
I've created a release now: https://github.com/ninja-build/ninja/releases/tag/v1.10.0

Maybe I should have created the release before pushing the tag? Or created the tag and the release in one go with GitHub's UI?

jhasse:
Ah it worked. I'm starting to understand.

> It should be fairly straightforward (though simpler when all the builds are in the same workflow file) to change the action to trigger when a v* tag gets added, and create a release for uploading the assets to.

Yes, that sounds like a good idea :)

PR #1685, opened 2019-11-16
Add restat tool which recalculates all mtimes in the build log

### LINKED ISSUE #1376: Ninja unnecessarily runs CMake just after it was run manually

jhasse:
This could then be used to fix #1376.

@mathstuf

### DIFF
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -242,14 +242,14 @@ struct LineReader {
   char* line_end_;
 };
 
-bool BuildLog::Load(const string& path, string* err) {
+LoadStatus BuildLog::Load(const string& path, string* err) {
   METRIC_RECORD(".ninja_log load");
   FILE* file = fopen(path.c_str(), "r");
   if (!file) {
     if (errno == ENOENT)
-      return true;
+      return LOAD_NOT_FOUND;
     *err = strerror(errno);
-    return false;
+    return LOAD_ERROR;
   }
 
   int log_version = 0;
@@ -270,7 +270,7 @@ bool BuildLog::Load(const string& path, string* err) {
         unlink(path.c_str());
         // Don't report this as a failure.  An empty build log will cause
         // us to rebuild the outputs anyway.
-        return true;
+        return LOAD_SUCCESS;
       }
     }
 
@@ -340,7 +340,7 @@ bool BuildLog::Load(const string& path, string* err) {
   fclose(file);
 
   if (!line_start) {
-    return true; // file was empty
+    return LOAD_SUCCESS; // file was empty
   }
 
   // Decide whether it's time to rebuild the log:
@@ -355,7 +355,7 @@ bool BuildLog::Load(const string& path, string* err) {
     needs_recompaction_ = true;
   }
 
-  return true;
+  return LOAD_SUCCESS;
 }
 
 BuildLog::LogEntry* BuildLog::LookupByOutput(const string& path) {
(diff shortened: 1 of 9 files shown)


src/build_log.cc
```diff
@@ -418,3 +419,50 @@ bool BuildLog::Recompact(const string& path, const BuildLogUser& user,
 
   return true;
 }
+
+bool BuildLog::Restat(const std::string& path,
```
jonesmz: path should be a StringPiece

src/build_log.cc
```diff
@@ -418,3 +419,50 @@ bool BuildLog::Recompact(const string& path, const BuildLogUser& user,
...
+
+    if (!WriteEntry(f, *i->second)) {
+      *err = strerror(errno);
+      fclose(f);
+      return false;
+    }
+  }
+
+  fclose(f);
+  if (unlink(path.str_) < 0) {
```
mathstuf: Like #1572, shouldn't this tool run on an not-run build still succeed? Alternatively, does `ninja` document the file names that will be used for this tool so that tools can know whether to call this tool or not? What happens if run in a build meant for `subninja` inclusion?

jhasse: Good questions. I will add some documentation.

As for running it on a clean build: CMake knows if it is run on a clean build directory, right? It could not call `ninja -t restat` then. But that wouldn't help for re-running cmake before if one hasn't called ninja yet ... hm ...

Do you plan on displaying the output of `ninja -t restat`? I'm thinking about ignoring a non-existing build_log (i.e. return 0), but still display a warning.

mathstuf: > CMake knows if it is run on a clean build directory, right?

It knows if it initialized it, but that doesn't correlate with whether this tool will work or not as you mention. Someone could have removed `.ninja_log` manually too.

> Do you plan on displaying the output of ninja -t restat?

I think it is a plain `fork`/`exec`, so its output is tied to CMake's stdout/stderr. I don't know that we want to suppress any output; this isn't the only reason ninja would ever want to output something.

jhasse: If Ninja output a warning but returned 0 if there's no log, would that work for you? Or would you even prefer no warning so that the output is clean?

mathstuf: I forsee that someone would complain to CMake about it (especially if CTest ends up catching it and marking it as a warning on CDash). I think the best I could do is point them to this discussion and say that this is what ninja does and CMake can't blanket silence it because errors would also get suppressed (CMake doing the filtering would be…unfortunate especially since every ninja generator would have to do the same thing. Factor in any tool internationalization (unlikely here, but ninja is not a singleton implementation) and it's basically intractable).

jhasse:
@mathstuf Now both restat and recompact only require a `build.ninja` file :)

mathstuf:
Thanks! I'll rebase #1572 and use the new enum once this is in.

hjmjohnson:
Is there anything I can do to help this along?  I think it will improve development time for those of us using CMake to generate the ninja build files.

jhasse:
Sure: Testing is always appreciated :)

I will merge this in the coming days if there are no objections.

PR #1689, opened 2019-11-22
Small code improvement: add const to LookupPool method

hyperxor:
I have noticed that `LookupPool` method of `State` class can be `const`. Also it's used in `assert` in `AddPool` method and this change will emphasize that there are no side effects.

### DIFF
--- a/src/state.cc
+++ b/src/state.cc
@@ -84,8 +84,8 @@ void State::AddPool(Pool* pool) {
   pools_[pool->name()] = pool;
 }
 
-Pool* State::LookupPool(const string& pool_name) {
-  map<string, Pool*>::iterator i = pools_.find(pool_name);
+Pool* State::LookupPool(const string& pool_name) const {
+  map<string, Pool*>::const_iterator i = pools_.find(pool_name);
   if (i == pools_.end())
     return NULL;
   return i->second;
--- a/src/state.h
+++ b/src/state.h
@@ -90,7 +90,7 @@ struct State {
   State();
 
   void AddPool(Pool* pool);
-  Pool* LookupPool(const string& pool_name);
+  Pool* LookupPool(const string& pool_name) const;
 
   Edge* AddEdge(const Rule* rule);
 


jonesmz:
This looks excellent, thank you.

jhasse:
Isn't the pool part of the state and therefore a const method should also return `const Pool*`?

jonesmz:
A member function that performs a lookup, and returns a handle to an internal state "thing" doesn't need to return a handle to a non-mutable "thing".

I would be happy to see the returned pointer point to a `const Pool` instead of a non-const `Pool`, but I don't think that it's an automatic necessity, especially if the returned Pool is currently used in a non-const way. (Seems to be a recurring problem with the Ninja codebase, doing weird things with Const.)

jhasse:
> A member function that performs a lookup, and returns a handle to an internal state "thing" doesn't need to return a handle to a non-mutable "thing".

Why not? Isn't that what std containers do too, if they don't contain pointers? Also for pointers there's [propagate_const](https://en.cppreference.com/w/cpp/experimental/propagate_const) for exactly this reason.

jonesmz:
I am simply saying that the language does not require or imply that a const member function must return a pointer-to-const. Semantically the constness of the member function is unrelated to what it is returning.

It is good API design in many cases, especially since when a function is used in a const-context the so-returned pointer-to-member might only be accessible as a const-object anyway. So I don't object to enforcing that rule at all.

Was just trying to say it isn't a language level requirement.

jhasse:
I see.

In this case, I would rather have the const method return a pointer-to-const, too.

jonesmz:
You shouldn't reject an improvement simply because it does not go far enough.

The ninja codebase does silly things all over the place. Like const-casting StringPiece.

This is an improvement. You should accept it.

jhasse:
IMHO it's not an improvement though.

jonesmz:
Why?

Its a lookup function that returns a pointer. The lookup function does not modify the object that it belongs to. Therefore it should be const function.

This has consequences for the compiler being able to reason about the result of calling the function, such as skipping a second invocation when no non-const operations happen on the object.

jhasse:
> This has consequences for the compiler being able to reason about the result of calling the function, such as skipping a second invocation when no non-const operations happen on the object.

I doubt that this is legal. Can you give a minimal reproducible example where the compiler does this?

jonesmz:
This is only practically achievable if you have LTO turned on, or are using inline-functions, and as such it's pretty challenging to provide an example on Compiler Explorer that isn't entirely contrived, given it only deals with a single translation unit and can't LTO between different translation units.

The extension __attribute((const)) and __attribute((pure)) which both GCC and Clang implement is much more effective at accomplishing this than the simple "const" keyword.
([[gnu::pure]], and [[gnu::const]], in C++11)

https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html

The basic conceptual framework is that if the compiler can prove that between two invocations of a function, the outcome would not change, it does not need to omit machine code to trigger the second invocation. The const keyword is not the enabler for the behavior, but it helps keep a program structured in such a way that it is more likely for the compiler to be able to do this.

jhasse:
I would be fine with an entirely contrived example. I can also compile stuff locally, no need for an Compiler Explorer example.

PR #1696, opened 2019-12-02
Remove hash_map.h in favor of using std::tr1::unordered_map with the default algorithm

jonesmz:


src/build_log.h
```diff
@@ -19,7 +19,9 @@
 #include <stdio.h>
 using namespace std;
 
-#include "hash_map.h"
+#include "unordered_map.h"
+
```
jhasse: why this empty line?

src/build_log_test.cc
```diff
@@ -12,8 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <algorithm>
+#include <cstring>
 #include "build_log.h"
-
```
jhasse: Don't remove this line. `#include "build_log.h"` is special and should come first.

src/unordered_map.h
```diff
@@ -0,0 +1,62 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef NINJA_UNORDERED_MAP_H_
+#define NINJA_UNORDERED_MAP_H_
+
+#include "string_piece.h"
+
+#if __APPLE__ || (__cplusplus >= 201103L) || (_MSC_VER >= 1900)
```
jhasse: Should only be `__cplusplus >= 201103L`.

src/util_test.cc
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <algorithm>
```
jhasse: Should be after `util.h`, see https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

src/unordered_map.h
jonesmz: No, that's not sufficient.

AppleClang doesn't ship with tr1 headers, regardless of the standard.

MSVC appears to have the same problem.

jhasse: This if branch is for the non-tr1 version though?

jonesmz: That's correct.

If we only check for C++11 to determine whether to include <unordered_map>, then AppleClang and MSVC (even when compiling for C++98) will fail to find <tr1/unordered_map> in the #else below, because they do not ship that header.

jhasse: Shouldn't both default to >= C++11 anyway though?

jonesmz: I don't know.

If I take away either of __APPLE__ or (_MSC_VER >= 1900), it won't build.

If I leave them in it does.

I'm not really interested in putting a lot of time into this, because I consider supporting anything older than C++11 a waste of time.

I'm submitting this PR to Ninja because I use this code (the non tr1 version) in my fork. I benefit by having a smaller set of patches to update against Ninja master.

If you have specific insights into how to improve this, I'm all ears, but I don't have either a Mac or a Windows computer, so it's already an uphill battle to figure out why it does or doesn't build when I change the flags.

My recommendation is to accept the change, and let other contributors point out problems if there are any. Ideally, with continuous integration tests that reproduce the failure automatically.

jhasse: I'll have a look when I have the time.

nico:
Did you do any perf measurements for this? IIRC last I checked things were a bit slower with unordered_map.

jonesmz:
I will investigate doing a performance comparison.

src/unordered_map.h
jhasse: > If I take away either of __APPLE__ or (_MSC_VER >= 1900), it won't build.

Regarding MSVC: Does it work with the CMake build? Because it should after 71c2cf3905a66f68409edde4745cb69ed967aefb.

jonesmz: I used the GitHub Actions continuous integration to determine if things built.

jhasse:
> I will investigate doing a performance comparison.

Any news?

jonesmz:
Upon reviewing the code in this PR again, I won't be attempting any performance measurements.

The intention of this PR is to use a single "name" for the unordered_map, instead of the various compiler-specific names for the same concept.

While it's possible that each compiler-specific implementation of hash_map / unordered_map has minor functionality differences compared to the tr1::unordered_map implementation, I believe any such differences are outweighed by the benefits of having a standardized implementation of the container type being used.

On new enough versions of the compilers, the compiler-specific versions of hash_map / unordered_map is already not available at all, anyway, and the tr1::unordered_map, or std::unordered_map is already being used by the existing code.

E.g. on any compiler where (__cplusplus >= 201103L), std::unordered_map is used, and on MSVC newer than (_MSC_VER >= 1900), even if being built with c++98 or c++03 mode.

As I continue advocating, Ninja should simply drop pre-C++11 compilers entirely, and be done with it.

jhasse:
No, this PR removes Ninja's own hash map in favor of the std one. That's the performance comparision we're interested in.

jhasse:
(because of the custom hash function used)

jonesmz:
Ninja doesn't implement a hash_map / unordered_map.

It's implementing it's own hash, specifically MurmurHash2.

If that's what you're concerned about performance for, then I can use MurmurHash2 with the std::tr1::unordered_map.

jonesmz:
If I keep the murmur hash implementation, will that make this PR more acceptable?

jhasse:
Yes, although I'm not sure what you want to actually change then.

jhasse:
Feel free to reopen if you have another look at this.

PR #1697, opened 2019-12-02
Have Github Actions also run tests

jonesmz:


.github/workflows/release-ninja-binaries.yml
```diff
@@ -40,6 +40,7 @@ jobs:
         mkdir build && cd build
         cmake -DCMAKE_BUILD_TYPE=Release ..
         cmake --build . --parallel --config Release --target ninja
+        ctest -vv
```
jhasse: Will this print the output if a test fails?

.github/workflows/release-ninja-binaries.yml
```diff
@@ -40,6 +40,7 @@ jobs:
         mkdir build && cd build
         cmake -DCMAKE_BUILD_TYPE=Release ..
         cmake --build . --parallel --config Release --target ninja
+        cmake --build . --parallel --config Release --target ninja_test
```
jhasse: Why not remove `--target ninja` from the previous command? Should also be faster.

.github/workflows/release-ninja-binaries.yml
jonesmz: It should, yes.

.github/workflows/release-ninja-binaries.yml
jonesmz: Alright.

PR #1698, opened 2019-12-02
Work around test framework verbosity issue

jonesmz:
ctest knows how to get diagnostic information from a bunch of different unit test frameworks, but apparently not the custom one ninja is using.

This change should ensure we have full test output.


jhasse:
Maybe we should just call `ninja_test` directly but in a separate step, so that one can also see output if it succeeds.

jonesmz:
I personally don't see any benefit to seeing test output on a success. 

I prefer to use ctest because it allows the CMakeLists.txt file to configure the full list of test programs to execute, such that more than one can be run by ctest.

But if you want to change the github actions script to run ninja_test directly, feel free.

jhasse:
c6148b0c3377a7d12b679f99f35e09cc7f36e940

PR #1699, opened 2019-12-03
Added CPack packaging

KOLANICH:


jonesmz:
Ninja is not currently suitable for use as a library. I don't see how changing libninja from a static library to a shared library is an improvement.

KOLANICH:
It allows to update 2 parts separately.

KOLANICH:
@jhasse

jhasse:
What is the need for CPack? Isn't there a debian package for ninja already? I doubt that they are using CPack.

> It allows to update 2 parts separately.

Wouldn't both files be in the same package anyway?

Maybe creating an issue first would be a good idea. This adds a lot of code to the CMakeLists.txt, I'm a bit hesitant in merging without having a use for it. Sorry for closing this PR for now.

KOLANICH:
>What is the need for CPack? Isn't there a debian package for ninja already? I doubt that they are using CPack.

I don't care how they use it. It doesn't matter. What matters is that an end user should be able to fetch this repo and build the bleeding edge package and install it. Without any `debhelper`s.

>Wouldn't both files be in the same package anyway?

It is adjustable using `SEPARATE_LIBNINJA`.

PR #1702, opened 2019-12-03
Ignore Qt Creator project files

jonesmz:


.gitignore
```diff
@@ -36,3 +36,6 @@ TAGS
 # Visual Studio Code project files
 /.vscode/
 /.ccls-cache/
+
+# Qt Creator project files
+CMakeLists.txt.user
```
jhasse: Should only appear at the root level, right? So please add `/` in the front.

jonesmz: Done.

Issue #1692, opened 2019-11-25
GitHub Actions: Statically link Visual C++ runtime


jhasse:
This will result in a bigger binary but no dependency on the VCRuntime installer.

nightlark:
Is pulling in a copy of the [ucm cmake macros](https://github.com/onqtam/ucm) something you're okay with, or would you prefer the flags get added to targets manually?

jhasse:
Haven't heard of ucm yet, sounds good :)

jhasse:
How do you want to pull it in though? `ExternalProject` would be preffered (no extra code, still works without Git).

nightlark:
My initial thought was stick a copy of ucm.cmake in a cmake modules folder (`set (CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${PROJECT_SOURCE_DIR}/cmake/")` and then include it with `include(ucm)`. But if `ExternalProject` is preferred, I think that could also work.

jhasse:
On second thought ... might just be simpler to copy the flags from https://github.com/onqtam/ucm#macro-ucm_set_runtimestatic-dynamic

Or just from our own configure.py.

jhasse:
I've tried:

```cmake
set(CMAKE_CXX_FLAGS_RELEASE "/MT /O2 /Ob2 /DNDEBUG")
```

But this results in

```
LINK : warning LNK4098: defaultlib 'MSVCRT' conflicts with use of other libs; use /NODEFAULTLIB:library 
```

when linking. Any ideas?

jonesmz:
Isn't /MT for linking against the multithreaded library? I thought Ninja was single threaded.

nightlark:
/MT makes it use the static version of the run time library; the [MSVC compiler options reference](https://docs.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library?view=vs-2019) lists the available options (static/dynamic and debug/nodebug).

The warning probably means something is still getting linked with the dynamic runtime. I'll see if I can find some time tonight to look at what's going on with static linking with MSVC.

nightlark:
Well I added the same set command, and didn't see any linker warnings about MSVCRT -- maybe it's a difference in where the flags were set. I'll cleanup and push a copy of what I have.

PR #1705, opened 2019-12-07
Add system runtime static linking

nightlark:
Re: Issue #1692

This adds a `NINJA_STATIC_SYSTEM_LIBS` CMake option that can be used to add the flags needed for static linking with the system runtime libraries with VC++ on Windows and gcc/clang (using libstdc++) on other platforms. It should handle adding flags for multi-configuration and single build type CMake generators correctly.

By default the static linking option is off so users building and installing a copy of ninja from source on their systems will link with the shared system libraries and (in theory) be able to get whatever benefits come from updating the system libraries.

The builds on GitHub Actions enable the static linking option for Windows and Linux builds (checking the new Windows binary with `dumpbin /dependents` shows none of the VC++ runtime libraries, the Linux binaries have a significantly smaller number of system library dependencies).

There's also a minor switch from matrix.os to runner.os which should be more reliable for checking the OS type than using image names.

Some future areas for improving the compatibility of released binaries:
- This brings the Linux binaries close to working on more systems, the next step will be switching over to a CentOS 6 Docker image for building the Linux release binary.
- Apple doesn't support static linking with system libraries, so the option is not enabled for macOS builds (I have some other ideas to try). Gatekeeper on 10.15 is a challenge.

### DIFF
--- a/.github/workflows/release-ninja-binaries.yml
+++ b/.github/workflows/release-ninja-binaries.yml
@@ -15,10 +15,13 @@ jobs:
         include:
           - os: ubuntu-latest
             zip_name: ninja-linux
+            extra_cmake_flags: '-DNINJA_STATIC_SYSTEM_LIBS=ON'
           - os: macOS-latest
             zip_name: ninja-mac
+            extra_cmake_flags: ''
           - os: windows-latest
             zip_name: ninja-win
+            extra_cmake_flags: '-DNINJA_STATIC_SYSTEM_LIBS=ON'
 
     steps:
     - uses: actions/checkout@v1
@@ -38,7 +41,7 @@ jobs:
       shell: bash
       run: |
         mkdir build && cd build
-        cmake -DCMAKE_BUILD_TYPE=Release ..
+        cmake -DCMAKE_BUILD_TYPE=Release ${{ matrix.extra_cmake_flags }} ..
         cmake --build . --parallel --config Release
         ctest -vv
 
--- a/.github/workflows/release-ninja-binaries.yml
+++ b/.github/workflows/release-ninja-binaries.yml
@@ -28,13 +28,13 @@ jobs:
 
     # Install OS specific dependencies
     - name: Install Linux dependencies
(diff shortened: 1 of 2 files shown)


debugzxcv:
for cmake >= 3.15, [check this ](https://cmake.org/cmake/help/latest/prop_tgt/MSVC_RUNTIME_LIBRARY.html)

jhasse:
Thanks for the PR! I had to split macOS and Windows up, since `find` stopped working on Windows. I've implemented the CMAKE_MSVC_RUNTIME_LIBRARY version in #1719.

We don't statically link the runtime on macOS and Linux, at least on Linux we now (#1712) build on CentOS 7. Let's see if we need to revisit this again later. As it looks now all is set for the (automatic) release of 1.10.

nightlark:
I think I'd agree that for now it is okay for the 1.10 release (will check out the actions workflow setup later) and is probably a bit better than the binaries from the 1.9 release. Setting the minimum CMake version to 3.15 is great -- no need to support old ways of doing things. :)

Recently I've been working on a portable pip package for a library, and some of the relevant things found for creating portable binary release packages for future ninja changes are:

- Windows binaries should be okay
- Static linking with system runtimes on macOS isn't possible, which presents its own challenges for portable binaries.
- Thoughts are mixed for whether static linking libstdc++/libgcc is needed for portable Linux binaries -- I think the answer to what the minimum required compiler version supported is determines if it is the best way to go (fortunately ninja doesn't have to deal with being loaded as a library).

The current set of libraries linked to by the Linux ninja release binary are:
```
  NEEDED               libstdc++.so.6
  NEEDED               libm.so.6
  NEEDED               libgcc_s.so.1
  NEEDED               libc.so.6
```

Given that ninja isn't useful without gcc/clang installed I think we can count on at least some version of libstdc++/libgcc being installed for now.

jonesmz:
Ninja may evolve into a loadable library. So please be mindful of that future possibility.

PR #1711, opened 2019-12-17
Adds the inputs tool to show all recursive inputs to given targets.

Daniel-Norman:
Test: ./ninja -t inputs -i -d ninja

### DIFF
--- a/src/graph.h
+++ b/src/graph.h
@@ -43,7 +43,9 @@ struct Node {
         dirty_(false),
         dyndep_pending_(false),
         in_edge_(NULL),
-        id_(-1) {}
+        id_(-1),
+        inputs_checked_(false),
+        used_as_dep_(false) {}
 
   /// Return false on error.
   bool Stat(DiskInterface* disk_interface, string* err);
@@ -103,6 +105,12 @@ struct Node {
 
   void Dump(const char* prefix="") const;
 
+  /// Used in the inputs debug tool.
+  bool InputsChecked() const { return inputs_checked_; }
+  void MarkInputsChecked() { inputs_checked_ = true; }
+  bool UsedAsDep() const { return used_as_dep_; }
+  void MarkUsedAsDep() { used_as_dep_ = true; }
+
 private:
   string path_;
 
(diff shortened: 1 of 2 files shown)


naarcini:
+1, I was about to put up a similar PR. I'll leave some comments since I also would like this change :)

src/ninja.cc
```diff
@@ -348,6 +349,96 @@ int NinjaMain::ToolGraph(const Options* options, int argc, char* argv[]) {
...
+  if (!CollectTargetsFromArgs(argc, argv, &nodes, &err)) {
+    Error("%s", err.c_str());
+    return 1;
+  }
+
+  for (vector<Node*>::iterator i = nodes.begin(); i != nodes.end(); ++i) {
+    Node* node = *i;
+    if (Edge* edge = node->in_edge()) {
+      node->MarkInputsChecked();
+      ToolInputsProcessEdge(edge, &deps_log_, leaf_only, include_deps);
```
naarcini: What does this approach get you that edge->GetBinding("in_newline") does not? See `EdgeEnv::LookupVariable` for more details: https://github.com/ninja-build/ninja/blob/ed154b90522eaa47f255f416e882de9f9771aadf/src/graph.cc#L341

I'm genuinely curious since I haven't quite been able to figure out where exactly these contents are populated. I do know that GetBinding() is what the commands tool uses, though, so it seemed promising.

src/graph.h
```diff
@@ -103,6 +105,12 @@ struct Node {
 
   void Dump(const char* prefix="") const;
 
+  /// Used in the inputs debug tool.
+  bool InputsChecked() const { return inputs_checked_; }
+  void MarkInputsChecked() { inputs_checked_ = true; }
+  bool UsedAsDep() const { return used_as_dep_; }
+  void MarkUsedAsDep() { used_as_dep_ = true; }
```
naarcini: Does any of this state need to be cleared when the tool is done running? I'm not sure what gets cached in ninja.

src/ninja.cc
```diff
@@ -348,6 +349,96 @@ int NinjaMain::ToolGraph(const Options* options, int argc, char* argv[]) {
...
+  // Print all of this node's deps from the deps log. This often includes files
+  // that are not known by the node's input edge.
+  if (deps_log->IsDepsEntryLiveFor(node)) {
+    DepsLog::Deps* deps = deps_log->GetDeps(node);
+    if (deps != NULL) {
+      for (int i = 0; i < deps->node_count; ++i) {
+        Node* dep = deps->nodes[i];
+        if (!dep->UsedAsDep()) {
+          dep->MarkUsedAsDep();
+          printf("%s\n", dep->path().c_str());
```
naarcini: `puts` will append a newline character for you:
http://www.cplusplus.com/reference/cstdio/puts/

jhasse:
Closing in favor of #1730.

PR #1720, opened 2020-01-19
Simplify Contributor on-loading with Gitpod

JesterOrNot:
image
Hi :smile: This setup will allow new contributors to start contributing at the push of a button with the project already built and the tools installed try it

[Open in Gitpod](https://gitpod.io/#https://github.com/JesterOrNot/ninja)

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,2 +1,5 @@
 tasks:
-  - init: cmake -Bbuild-cmake -H. && cmake --build build-cmake
+  - init: >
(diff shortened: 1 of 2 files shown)


jhasse:
Hi :) This does sound useful for some people, but having it in the main repository would be some kind of endorsement of Gitpod. I've only heard of the service now, haven't had the time to read the terms of service.

This may be a good idea for the wiki first and if it gets popular, we might revisit adding it to Git. Sorry.

PR #1721, opened 2020-01-22
recompact: no need for loading build.ninja

mathstuf:
Follows the same pattern as d986e4db5630cf1c5547e69b5556f006f7d3444a for
restat.

---
Cc: @jhasse

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -967,7 +967,7 @@ const Tool* ChooseTool(const string& tool_name) {
     { "compdb",  "dump JSON compilation database to stdout",
       Tool::RUN_AFTER_LOAD, &NinjaMain::ToolCompilationDatabase },
     { "recompact",  "recompacts ninja-internal data structures",
-      Tool::RUN_AFTER_LOAD, &NinjaMain::ToolRecompact },
+      Tool::RUN_AFTER_FLAGS, &NinjaMain::ToolRecompact },
     { "restat",  "restats all outputs in the build log",
       Tool::RUN_AFTER_FLAGS, &NinjaMain::ToolRestat },
     { "rules",  "list all rules",


jhasse:
How sure are you that this is okay? I'm not that confident with the recompact tool.

Maybe there was a reason @nico ran it after loading build.ninja in 8de46a9685f562b75319bb48263e8c787e86d48b.

mathstuf:
What information is used from the `build.ninja` in recompaction?

mathstuf:
Found it. I see is `OpenDepsLog` uses `state_`. This ends up getting used to compute a checksum in `DepsLog::Load`. Seems to be about canonicalizing paths as well. Is that ignorable for recompaction?

mathstuf:
Hmm. I have a build tree which does go through the `state`-using codepath, but isn't hitting the asserts with either `AFTER_FLAGS` or `AFTER_LOAD`. Maybe some other ninja generator makes things that tickle them?

jhasse:
Needs some further investigation.

For now, as @bradking said:

> CMake can work without the change if needed.  I don't think this needs to block the 1.10 release for Ninja.

mathstuf:
Agreed. I'll have CMake avoid it when we don't have a single `build.ninja` being used.

@KyleFromKitware Though I guess this means that if we have a multi-config build where ninja does a recompaction during one configuration's build, ninja doesn't know about the other manifests and may have side effects similar to making `-t recompact` not load a manifest at all. Something to keep an eye out for.

bradking:
While testing this locally I found that not loading the manifest causes `recompact` to drop almost everything from the deps log and results in a full rebuild.

mathstuf:
In light of that behavior, this is not something we can do. Closing.

PR #1725, opened 2020-01-28
Clarify order-only dependencies documentation

marc-h38:
Around 2011 and after "a lot of soul-searching" Evan Martin made an
important change with respect to rebuilding the "dependencies of
order-only dependencies":
 https://groups.google.com/forum/m/#!msg/ninja-build/lSiVYhWEIyM/qZtVh5k8AgAJ

Document this significant change and the fact that it removed this
feature:
https://stackoverflow.com/questions/31271493/prevent-order-only-prerequisite-from-being-rebuilt-if-out-of-date

Also clarify the depfile bootstrap example.

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -955,16 +955,22 @@ changes to the script should cause the output to rebuild.
 Note that dependencies as loaded through depfiles have slightly different
 semantics, as described in the <<ref_rule,rule reference>>.
 
-3. _Order-only dependencies_, expressed with the syntax +|| _dep1_
-   _dep2_+ on the end of a build line.  When these are out of date, the
-   output is not rebuilt until they are built, but changes in order-only
-   dependencies alone do not cause the output to be rebuilt.
+3. _Order-only dependencies_, expressed with the syntax
+   +build output: ... || _dep1_ _dep2_+ on the end of a build line.
+   When these are out of date, +output+ is not rebuilt until they
+   are built, but changes in order-only dependencies _alone_ do not cause
+   +output+ to be rebuilt.
 +
 Order-only dependencies can be useful for bootstrapping dependencies
-that are only discovered during build time: for example, to generate a
-header file before starting a subsequent compilation step.  (Once the
-header is used in compilation, a generated dependency file will then
-express the implicit dependency.)
+that are only discovered during build time: for example, to generate
+header files that the compiler may or may not need depending on the
+project's configuration. Once the compiler can run, a generated
+dependency file will then express real headers usage with implicit
+dependencies.
++
+Compared with an implicit dependency, the order-only dependency example
+above can save rebuilding +output+. However it does not save rebuilding
+out of date +dep1+ and +dep2+.
 
 File paths are compared as is, which means that an absolute path and a
 relative path, pointing to the same file, are considered different by Ninja.


marc-h38:
Ping @jhasse , others? Before even looking at my English mistakes, I'd love confirmation that this small addition is factually correct.

jhasse:
Haven't had the time to have a closer look at the issue, sorry.

marc-h38:
> Haven't had the time to have a closer look at the issue, sorry.

This "Do we rebuild order-only dependencies?" issue is subtle but only because it lacks documentation, which this small PR hopefully fixes.

Anyone else more familiar with order-only dependencies who could help?

marc-h38:
@scivision maybe?

scivision:
sorry I'm not familiar with that part of ninja

marc-h38:
@jhasse, others, ping?

I submitted this one year ago after many hours of research and experimentation. This is a small, pure documentation  no code change, documenting a logical change in ninja performed in 2011 (!) This clarifies a feature that is routinely used by (at least) meson and CMake and exposed to their users and that seems to generate frequent confusion, e.g.: https://github.com/mesonbuild/meson/issues/6391
I got exactly zero feedback :-(

Cc:   @jpakkane, @bradking, @mathstuff

marc-h38:
Ooops, I meant @mathstuf not @MathStuff, apologies to the latter, I hope you can somehow unsubscribe (although the spam level has been fairly low here :-D)

mathstuf:
Oof, that's an org/team. No idea how that works. Didn't know I had a doppelgänger either :) .

I can pass the buck too as I'm not in a right mind to untangle dependency description prose at the moment :) . @RedBeard0531?

doc/manual.asciidoc
```diff
@@ -955,16 +955,22 @@ changes to the script should cause the output to rebuild.
...
-header is used in compilation, a generated dependency file will then
-express the implicit dependency.)
+that are only discovered during build time: for example, to generate
+header files that the compiler may or may not need depending on the
+project's configuration. Once the compiler can run, a generated
+dependency file will then express real headers usage with implicit
+dependencies.
++
+Compared with an implicit dependency, the order-only dependency example
+above can save rebuilding +output+. However it does not save rebuilding
```
jhasse, line 972: Because with an implicit dependency output might be built before deps1 and deps2?

doc/manual.asciidoc
```diff
@@ -955,16 +955,22 @@ changes to the script should cause the output to rebuild.
...
+out of date +dep1+ and +dep2+.
```
jhasse, line 973: What do you mean by that? In general I'm a bit confused what "to save rebuilding" means.

jhasse (commented):
Sorry for never getting back to this. I found the time, but still don't understand the issue 100%.

marc-h38 (commented):
Thanks for taking a look at this. The very first thing is to make sure you fully understand the change Evan made in 2011: https://groups.google.com/g/ninja-build/c/lSiVYhWEIyM/m/KgDyvZ2DG7cJ
Shorter explanation at the end of this comment.

I've read that this 2011 change solved some elaborate, real world use cases. I'm still not sure which ones but in the most basic order-only example, that 2011 change is really counter-intuitive. The purpose of this PR is to help users realize this counter-intuitive behaviour is by design and not a bug. I noticed the ninja documentation is generally quite "dry", more of a reference manual than a tutorial so I tried to change and add as little as possible but maybe I went "too far" and didn't add enough which could be why I puzzled you?

Even before looking at the 2011 change, the documentation could some actual order-only example that goes beyond merely showing the syntax and hints at a real use case. For instance something like this:

```ninja
rule compile
  command = touch $out
  # shorter, real dependencies based on actual #includes
  # depfile = $out.d

rule generate
   command = touch $out

# We don't know yet which headers are needed by this object.
# Ask for all of them for bootstrap reasons but do not re-compile
# for any, possibly unused header.
build object:                  compile || all_generated_headers

build all_generated_headers:  generate |  all_template_headers
```
This example is runnable.

Once order-only dependencies are a bit clearer, documenting the 2011 change would be much easier.

Try not to add too much repetition to https://groups.google.com/g/ninja-build/c/lSiVYhWEIyM/m/KgDyvZ2DG7cJ, the counter-intuitive 2011 behavior this PR tries to document is:

1. Change `all_template_headers`
2. Ask ninja for `object`
3. `object` is deemed up to date and not rebuilt (OK) _and_ `all_generated_headers` get rebuilt: counter-intuitive.

doc/manual.asciidoc
jhasse: What do you mean then?

jhasse:
I see. I think I fully understand it now.

marc-h38:
So would you like me to make a longer version of this change? With a generated headers example maybe?

Without an example I found this feature extremely abstract and hard to understand. I wasted time trying to use it for things it wasn't suitable for and the 2011 change made things even more confusing.

jhasse:
Yes that would be great :)

zdave:
I spent quite a while today trying to understand order-only dependencies, and agree that the current documentation isn't completely clear. That said, I don't think this change as it stands now is completely clear either. An example would be helpful, but I think the documentation should really just be more explicit about the behaviour.

AIUI, the following rules fully cover the behaviour of _regular_ dependencies:

1. A node is wanted if any of its dependents are wanted, or it was directly requested on the command line.
2. A node is dirty if any of its dependencies are dirty or were modified more recently than the node itself.
3. If a node is built (which happens if it is wanted _and_ dirty), it is always built after its dependencies.

The only difference between regular dependencies and order-only dependencies is that the second rule does not apply to order-only dependencies. The documentation makes it clear that rule 2 does not apply, and it also makes it clear that rule 3 still does apply. What it doesn't make clear is that rule 1 also still applies.

The first rule is really too strong for the header-generation use case of order-only dependencies. This is what leads to the unintuitive behaviour where, in the example given above, `all_generated_headers` is rebuilt even though it is not needed for `object`.

That said, I don't think the rules prior to the 2011 change (#125) were strong enough for the header-generation use case; AIUI, `all_generated_headers` would have only been rebuilt if the headers did not yet exist, or there was an explicit dependency on a header from a `depfile`. As mentioned on the mailing list, this is insufficient in the case where a source file is modified to include a header that it didn't include before.

A more suitable rule for the header-generation use case would be "a node is wanted if any of its regular dependents are wanted, or any of its order-only dependents are wanted _and_ dirty". This rule would not be suitable for other use cases however, eg the shared library use case mentioned on the mailing list.

evmar:
That is a great analysis!  I am trying to refresh my memory here, but here is another way to frame the same thing:

- When building, all transitive dependencies of the command line targets are brought up to date, in dependency order.
- A target is up to date when it is newer than its (non-order-only) dependencies.

It's still not great wording, but maybe you get my direction: you can frame it as global properties about the graph rather than per-node computations and I think it seems more cohesive.  (To be comprehensive the above also needs an asterisk about how "newer" may not be mtime-newer due to how restat works...)

doc/manual.asciidoc
```diff
@@ -955,16 +955,22 @@ changes to the script should cause the output to rebuild.
 Note that dependencies as loaded through depfiles have slightly different
 semantics, as described in the <<ref_rule,rule reference>>.
 
-3. _Order-only dependencies_, expressed with the syntax +|| _dep1_
-   _dep2_+ on the end of a build line.  When these are out of date, the
-   output is not rebuilt until they are built, but changes in order-only
-   dependencies alone do not cause the output to be rebuilt.
+3. _Order-only dependencies_, expressed with the syntax
+   +build output: ... || _dep1_ _dep2_+ on the end of a build line.
```
eli-schwartz, line 959: This is no longer the end of a build line -- it's the whole line.

PR #1727, opened 2020-01-29
Update disk_interface.cc

dexter4life:
Notice that FindFirstFileExA in windows needs a directory structure in the form of C:\somepath\some_sub_directory but instead, we get /libpath:C:\somepath\some_sub_directory. This causes FindFirstFileExA  to return INVALID_HANDLE_VALUE


atetubou:
This sounds like you use ninja wrongly?

dexter4life:
So how to use it?

PR #1729, opened 2020-02-01
WIP: ninja-as-library-draft-1

EliRibble:
This is my take on structuring ninja as a library. I'm hoping it'll be the starting point of a discussion so I can found out what other things I may need to do in order to make the maintainers happy. I'm sure I violated half a dozen architectural decisions out of pure ignorance.

Feel free to beat up heavily on the PR. The goal is to create something the maintainers are happy with that meets the need.


configure.py
```diff
@@ -710,4 +710,8 @@ def has_re2c():
     if options.verbose:
         rebuild_args.append('-v')
 
-    subprocess.check_call(rebuild_args)
```
jonesmz: Not sure that it would be worth updating configure.py to support ninja as a library, the CMake integration seems to be the favored way forward.

src/public/version.h
```diff
@@ -0,0 +1,25 @@
+// Copyright 2019 Google Inc. All Rights Reserved.
```
jonesmz: I'm skeptical that the data of the copyright should be updated to 2019 unless the code is also being modified instead of just cut-and-pasted.

src/public/version.h
```diff
@@ -0,0 +1,25 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef NINJA_PUBLIC_VERSION_H_
+#define NINJA_PUBLIC_VERSION_H_
+
+namespace ninja {
```
jonesmz: I don't have a problem with putting the version symbol in the namespace ninja, but you might get a minor pushback from doing this "early" in the patch series, since putting symbols inside the ninja namespace appears to be its own commit later on.

src/public/version.h
```diff
@@ -0,0 +1,25 @@
...
+/// The version number of the current Ninja release.  This will always
+/// be "git" on trunk.
+extern const char* kNinjaVersion;
```
jonesmz: I know that this is already a char* in trunk, but it might be worth doing something else if things are already being done.

```cpp
struct VersionInfo
{
    major
    minor
    patch
    hotfix
    vcs-commit-id
}
```
?

It's not like it's complicated to pre-process the code via the continuous integration system.

Maybe a discussion for another PR though.

src/version.h
```diff
@@ -16,19 +16,15 @@
 #define NINJA_VERSION_H_
 
 #include <string>
-using namespace std;
```
jonesmz: You could rebase on https://github.com/ninja-build/ninja/pull/1614  to have this done for you.

src/version.h
```diff
@@ -16,19 +16,15 @@
...
 
 namespace ninja {
-/// The version number of the current Ninja release.  This will always
-/// be "git" on trunk.
-extern const char* kNinjaVersion;
 
 /// Parse the major/minor components of a version string.
-void ParseVersion(const string& version, int* major, int* minor);
+void ParseVersion(const std::string& version, int* major, int* minor);
```
jonesmz: string_piece?

src/version.h
```diff
@@ -16,19 +16,15 @@
...
 
 /// Check whether \a version is compatible with the current Ninja version,
 /// aborting if not.
-void CheckNinjaVersion(const string& required_version);
+void CheckNinjaVersion(const std::string& required_version);
```
jonesmz: string_piece?

src/public/version.h
jonesmz: Ooops, got my commits out of order in my head. Never mind this.

src/public/build_config.h
```diff
@@ -0,0 +1,40 @@
...
+    VERBOSE
+  };
+  Verbosity verbosity;
+  bool dry_run;
+  int parallelism;
+  int failures_allowed;
+  /// The maximum load average we must not exceed. A negative value
+  /// means that we do not have any limit.
+  double max_load_average;
+  DepfileParserOptions depfile_parser_options;
```
jonesmz: Bad struct packing.

should be

```
  /// The maximum load average we must not exceed. A negative value
  /// means that we do not have any limit.
  double max_load_average;
  int parallelism;
  int failures_allowed;
  Verbosity verbosity;
  DepfileParserOptions depfile_parser_options;
  bool dry_run;
```

src/public/depfile_parser_options.h
```diff
@@ -0,0 +1,31 @@
...
+
+#ifndef NINJA_PUBLIC_DEPFILE_PARSER_OPTIONS_H_
+#define NINJA_PUBLIC_DEPFILE_PARSER_OPTIONS_H_
+
+enum DepfileDistinctTargetLinesAction {
+  kDepfileDistinctTargetLinesActionWarn,
+  kDepfileDistinctTargetLinesActionError,
+};
+
+struct DepfileParserOptions {
```
jonesmz: Why is this a struct instead of just an enum?

src/depfile_parser_test.cc
```diff
@@ -16,10 +16,11 @@
 
 #include "test.h"
 
+namespace ninja {
 struct DepfileParserTest : public testing::Test {
   bool Parse(const char* input, string* err);
 
-  DepfileParser parser_;
+  ninja::DepfileParser parser_;
```
jonesmz: Shouldn't need the ninja:: prefix?

src/disk_interface.cc
```diff
@@ -240,7 +241,7 @@ bool RealDiskInterface::WriteFile(const string& path, const string& contents) {
 }
 
 bool RealDiskInterface::MakeDir(const string& path) {
-  if (::MakeDir(path) < 0) {
+  if (ninja::MakeDir(path) < 0) {
```
jonesmz: Is the ninja:: necessary? Isn't simply "MakeDir" sufficient?

src/ninja.cc
```diff
@@ -56,6 +56,10 @@ int MSVCHelperMain(int argc, char** argv);
 void CreateWin32MiniDump(_EXCEPTION_POINTERS* pep);
 #endif
 
+// Use namespace ninja for now until we can carve out all
```
jonesmz: Honestly I'd rather there not be any symbols at all that are outside of the ninja namespace.

src/public/tools.h
```diff
@@ -0,0 +1,31 @@
...
+namespace ninja {
+
+class Node;
+class State;
+
+// Given a path to a node return the node that is
+// most likely a match for that path. This accounts
+// for users mispelling parts of the path and gives
+// the node nearest what the user requested.
+Node* SpellcheckNode(State* state, const std::string& path);
```
jonesmz: string_piece?

src/tools.cc
```diff
@@ -0,0 +1,39 @@
...
+
+namespace ninja {
+
+Node* SpellcheckNode(State* state, const std::string& path) {
+  const bool kAllowReplacements = true;
+  const int kMaxValidEditDistance = 3;
+
+  int min_distance = kMaxValidEditDistance + 1;
+  Node* result = NULL;
+  for (State::Paths::iterator i = state->paths_.begin(); i != state->paths_.end(); ++i) {
```
jonesmz: Seems like this function would be better as a call to std::min with a comparitor passed in (e.g. a C++11 lambda)

src/tools.cc
```diff
@@ -36,4 +36,53 @@ Node* SpellcheckNode(State* state, const std::string& path) {
...
 
+Node* CollectTarget(State* state, const char* cpath, std::string* err) {
+  std::string path = cpath;
+  uint64_t slash_bits;
+  if (!CanonicalizePath(&path, &slash_bits, err))
+    return NULL;
+
+  // Special syntax: "foo.cc^" means "the first output of foo.cc".
+  bool first_dependent = false;
+  if (!path.empty() && path[path.size() - 1] == '^') {
```
jonesmz: Seems fishy, what about when path.size() == 1?

src/ninja.cc
```diff
@@ -109,10 +109,14 @@ void ExitNow() {
 /// to poke into these, so store them as fields on an object.
 struct NinjaMain : public BuildLogUser {
   NinjaMain(const char* ninja_command, const BuildConfig& config) :
+      disk_interface_(new RealDiskInterface),
       ninja_command_(ninja_command), config_(config),
       start_time_millis_(GetTimeMillis()),
       state_(new State()) {}
 
+  /// Functions for accesssing the disk.
```
jonesmz: typo on "accessing"

src/public/logger.h
```diff
@@ -0,0 +1,46 @@
...
+// limitations under the License.
+#ifndef NINJA_PUBLIC_LOGGER_H_
+#define NINJA_PUBLIC_LOGGER_H_
+
+#include <iostream>
+#include <string>
+
+namespace ninja {
+
+class Logger {
```
jonesmz: This opens the question on whether to support compile-time polymorphism (templates) or run-time polymorphism (virtual member functions).

It's clear to me that the current codebase of Ninja is very runtime oriented on things like interfaces, and converting large parts of the code to templates would meet a lot of resistance.

Of course, compile-time polymorphism can be (depending on implementation details) strictly more flexible than run-time polymorphism, as you can pass an abstract base class as the type to instantiate the template with.

src/public/logger.h
```diff
@@ -0,0 +1,46 @@
...
+    ERROR = 0,
+    WARNING = 1
+  };
+
+  virtual void OnMessage(Level level, const std::string& message) = 0;
+};
+
+class LoggerBasic : Logger {
+public:
+  virtual void OnMessage(Level level, const std::string& message) {
```
jonesmz: string_piece

src/state.cc
```diff
@@ -68,9 +69,13 @@ Pool State::kDefaultPool("", 0);
 Pool State::kConsolePool("console", 1);
 const Rule State::kPhonyRule("phony");
 
-State::State() :
-  build_log_(new BuildLog()),
-  deps_log_(new DepsLog()) {
+State::State() : State(NULL) {}
```
jonesmz: Allow a default constructed state?

Do you think people will construct the state by parts?

src/state.cc
```diff
@@ -140,6 +145,28 @@ bool State::AddDefault(StringPiece path, string* err) {
   return true;
 }
 
+bool State::IsPathDead(StringPiece s) const {
+  Node* n = LookupNode(s);
+  if (!n || !n->in_edge())
```
jonesmz: if() {
}

src/state.cc
jonesmz: Wouldn't State::State(Logger* logger = NULL) be sufficient?

src/ninja.cc
```diff
@@ -37,15 +37,12 @@
 #include "public/tools.h"
```
jonesmz: Not a comment on the way you implemented this change, just a comment on the concept.

I could see a benefit to having the build system set up to allow each of these tools to be their own binary. That would allow people who are compiling the Ninja project to choose a multi-binary, or a set of stand alone binaries (and both options would also allow dynamic or static linking against the library-fied libninja)

src/logger.cc
```diff
@@ -0,0 +1,22 @@
+#include "public/logger.h"
```
jonesmz: Obviously nothing to worry about this early in the patch series, but something to consider for future improvements would be support for alternative logging systems.

Theres a lot of ways it can be approached. Macros are an "easy" approach, where ninja defines a macro collection that gets used in specific well defined ways, and library users can define their own.

Template arguments passed in to replace the logging functions at compile time.

Lots of options.

src/public/ui.h
```diff
@@ -18,6 +18,18 @@
...
+#ifdef _WIN32
+#include "win32port.h"
+#else
+#include <stdint.h>
+#endif
+
+#ifdef _MSC_VER
+#define NORETURN __declspec(noreturn)
+#else
+#define NORETURN __attribute__((noreturn))
```
jonesmz: This'll probably need to be it's own header if someone adds a check for C++11's attribute syntax

src/public/ui.h
```diff
@@ -29,6 +41,12 @@ const char* Warning();
 /// Returns a Tool, or NULL if Ninja should exit.
 const Tool* ChooseTool(const std::string& tool_name);
 
+/// Execute ninja as the main ninja binary would
+/// Does not return, prefering to exit() directly
+/// to avoid potentially expensive cleanup when  destructuring
+/// Ninja's state.
+NORETURN void Execute(int argc, char** argv);
```
jonesmz: Of course this is already the existing behavior, but I'd wager that the vast majority of potential library users aren't going to be amused if a shared library that they load makes their application terminate.

src/public/execution.h
```diff
@@ -0,0 +1,72 @@
...
+    /// Whether a depfile with multiple targets on separate lines should
+    /// warn or print an error.
+    bool depfile_distinct_target_lines_should_err;
+  };
+
+  Execution();
+  Execution(Options options);
+
+  /// Get access to the underlying disk interface
+  RealDiskInterface* DiskInterface();
```
jonesmz: Shouldn't this return a pointer to the abstract base class?

src/public/logger.h
```diff
@@ -22,7 +22,8 @@ class Logger {
 public:
   enum Level {
     ERROR = 0,
-    WARNING = 1
+    WARNING = 1,
+    INFO = 2
```
jonesmz: At least in the applications I work with, i tend to use these log levels.

debug, detailed, normal, warning, fatal

src/public/execution.h
```diff
@@ -27,25 +28,47 @@ struct Tool;
...
-    /// Whether duplicate rules for one target should warn or print an error.
-    bool dupe_edges_should_err;
-  
-    /// Whether phony cycles should warn or print an error.
-    bool phony_cycle_should_err;
-  
-    /// Whether a depfile with multiple targets on separate lines should
-    /// warn or print an error.
-    bool depfile_distinct_target_lines_should_err;
   };
```
jonesmz: struct packing
```
    /// Build file to load.
    const char* input_file;

    /// The tool to use
    const Tool* tool_;

    /// Directory to change into before running.
    const char* working_dir;

    /// The number of failures allowed before terminating the build.
    int failures_allowed;

    /// The level of parallelism to use during the build
    int parallelism;

    /// The maximum load to allow.
    float max_load_average;

    /// Whether phony cycles should warn or print an error.
    bool phony_cycle_should_err;

    /// True to include verbose logging. Default is false.
    bool verbose;

    // Whether a depfile with multiple targets on separate lines should
    /// warn or print an error.
    bool depfile_distinct_target_lines_should_err;

    /// Whether or not this is a dry-run. IE, it should just
    /// show what would be performed without taking any action.
    bool dry_run;

    /// Whether duplicate rules for one target should warn or print an error.
    bool dupe_edges_should_err;
```

src/clean.cc
```diff
@@ -238,21 +238,21 @@ int Cleaner::CleanRule(const char* rule) {
   return status_;
 }
 
-int Cleaner::CleanRules(int rule_count, char* rules[]) {
+int Cleaner::CleanRules(const std::vector<std::string>& rules) {
```
jonesmz: could this instead be a pair or iterators to vector?

Or a template function that takes generic iterators?

src/execution.cc
```diff
@@ -51,7 +51,73 @@ int GuessParallelism() {
   }
 }
 
+Node* TargetNameToNode(const State* state, const std::string& path, std::string* err) {
+  uint64_t slash_bits;
+  std::string canonical_path = path.c_str();
```
jonesmz: This requires that the constructor for canonical_path run strlen on the provided char*.

Better to simply do

```cpp
std::string canonical_path = path;
```

src/execution.cc
```diff
@@ -261,19 +261,20 @@ int Execution::Graph() {
...
     return 1;
   }
 
   DyndepLoader dyndep_loader(state_, state_->disk_interface_);
 
-  for (int i = 0; i < argc; ++i) {
+  for (size_t i = 0; i < options_.targets.size(); ++i) {
     string err;
-    Node* node = ui::CollectTarget(state_, argv[i], &err);
+    std::string target_name = options_.targets[i];
```
jonesmz: Copy necessary here?

src/browse.h
```diff
@@ -20,11 +20,11 @@ struct State;
...
 /// \a ninja_command is the command used to invoke ninja.
-/// \a args are the number of arguments to be passed to the Python script.
-/// \a argv are arguments to be passed to the Python script.
+/// \a input_file is the path to the build.ninja file.
+/// \a initial_target is the target name to start the browser at, or NULL.
 /// This function does not return if it runs successfully.
-void RunBrowsePython(State* state, const char* ninja_command,
-                     const char* input_file, int argc, char* argv[]);
+void RunBrowsePython(const char* ninja_command,
+                     const char* input_file, const char* initial_target);
```
jonesmz: string_piece?

src/ui.cc
```diff
@@ -217,11 +217,11 @@ NORETURN void Execute(int argc, char** argv) {
   setvbuf(stdout, NULL, _IOLBF, BUFSIZ);
 
   ninja::Execution::Options options;
+  const char* ninja_command = argv[0];
```
jonesmz: Better to copy? argv is mutable, after all.

src/execution.cc
```diff
@@ -118,12 +118,13 @@ bool TargetNamesToNodes(const State* state, const std::vector<std::string>& name
 
 }  // namespace
 
-Execution::Execution() : Execution(NULL, Options()) {}
-
+Execution::Execution() : Execution(NULL, Options(), std::make_unique<LoggerBasic>()) {}
 Execution::Execution(const char* ninja_command, Options options) :
+  Execution(ninja_command, options, std::make_unique<LoggerBasic>()) {}
```
jonesmz: You're almost certainly going to get pushback from the ninja maintainers to use std::unique_ptr on anything that isn't *already* using it, because there are some unspecified systems that the Ninja maintainers want to preserve backwards compatibility for, though they won't explain which platforms those systems have, or really talk about it in a meaningful way.

The places where unique_ptr is allowed are the places where the code in question represents some kind of stand-alone tool, so requiring C++11 or newer on that new code was considered OK.

src/state.cc
```diff
@@ -69,10 +69,10 @@ Pool State::kDefaultPool("", 0);
 Pool State::kConsolePool("console", 1);
 const Rule State::kPhonyRule("phony");
 
-State::State() : State(NULL) {}
+State::State() : State(std::make_unique<LoggerBasic>()) {}
```
jonesmz: Doesn't this change the semantics of the default constructor to change it from having no logger, to having a default logger?

src/ui.cc
```diff
@@ -417,6 +419,35 @@ int ReadFlagsClean(int* argc, char*** argv, Execution::Options* options) {
   return ReadTargets(argc, argv, options);
 }
 
+int ReadFlagsCommands(int* argc, char*** argv, Execution::Options* options) {
+  // Step back argv to include 'clean' so that getopt will
+  // work correctly since it starts reading at position 1.
+  ++(*argc);
+  --(*argv);
```
jonesmz: Yikes...

src/execution.cc
```diff
@@ -38,6 +43,40 @@
 namespace ninja {
 namespace {
 
+void EncodeJSONString(const char *str) {
+  while (*str) {
+    if (*str == '"' || *str == '\\')
+      putchar('\\');
```
jonesmz: if() {
}

src/msvc_helper_main-win32.cc
```diff
@@ -25,19 +25,8 @@
...
-"usage: ninja -t msvc [options] -- cl.exe /showIncludes /otherArgs\n"
-"options:\n"
-"  -e ENVFILE load environment block from ENVFILE as environment\n"
-"  -o FILE    write output dependency information to FILE.d\n"
-"  -p STRING  localized prefix of msvc's /showIncludes output\n"
-         );
-}
-
-void PushPathIntoEnvironment(const string& env_block) {
+void PushPathIntoEnvironment(const std::string& env_block) {
```
jonesmz: string_piece?

src/execution.cc
```diff
@@ -25,7 +25,6 @@
 #include <getopt.h>
```
jonesmz: Typo in commit message: impelementation

src/public/ui.h
```diff
@@ -56,8 +56,10 @@ struct Tool {
   Implementation implementation;
 };
 
+/// Structure for returning the information parsed
+/// from the commandline arguments.
 struct ParsedFlags {
-  const Tool* tool;
+  const Tool* tool = NULL;
```
jonesmz: Can't. Default values for struct members is a C++11 feature.

jonesmz: Plus... if you're going to C++11-ize it, might as well use `nullptr`

src/ui.cc
```diff
@@ -423,6 +423,8 @@ int ReadFlags(int* argc, char*** argv,
       return ReadFlagsClean(argc, argv, &flags->options);
     } else if (strcmp(optarg, "commands") == 0) {
       return ReadFlagsCommands(argc, argv, &flags->options);
+    } else if (strcmp(optarg, "compdb") == 0) {
```
jonesmz: Is this fixed in the latest release?

Maybe make a separate PR if not?

src/util.cc
```diff
@@ -29,7 +29,6 @@
 #include <stdarg.h>
```
jonesmz: Commit title is a bit hard to understand.

Perhaps instead:

> Stop using 'using namespace std;' in util.

src/version.cc
```diff
@@ -21,18 +21,18 @@
 namespace ninja {
 const char* kNinjaVersion = "1.9.0.git";
 
-void ParseVersion(const string& version, int* major, int* minor) {
+void ParseVersion(const std::string& version, int* major, int* minor) {
```
jonesmz: string_piece?

src/debug_flags.cc
```diff
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-bool g_explaining = false;
```
jonesmz: Per your commit message, one would argue that a library shouldn't have any global state at all, so multiple instances can be loaded and executed in the same application.

src/execution.cc
```diff
@@ -202,15 +202,16 @@ void PrintToolTargetsList(const vector<Node*>& nodes, int depth, int indent) {
```
jonesmz: I really wish we could just bite the bullet and use C++11.

9 years seems long enough to support legacy systems for a tool like this...

src/browse.cc
```diff
@@ -33,7 +33,7 @@ namespace ninja {
 void LogError(Logger* logger, const char* prefix) {
```
jonesmz: Commit message typo.

src/execution.cc
```diff
@@ -732,11 +753,13 @@ bool Execution::DoBuild() {
 void Execution::DumpMetrics() {
   g_metrics->Report();
 
-  printf("\n");
+  logger_->Info("\n");
   int count = (int)state_->paths_.size();
```
jonesmz: Seems strange to cast to int here, since we're going to get converted to double in a moment just below.

src/execution.cc
```diff
@@ -42,13 +42,15 @@
...
 namespace {
 
-void EncodeJSONString(const char *str) {
+std::string EncodeJSONString(const char *str) {
+  std::ostringstream buffer;
   while (*str) {
     if (*str == '"' || *str == '\\')
-      putchar('\\');
-    putchar(*str);
+      buffer << '\\';
```
jonesmz: if() {
}

src/logger.cc
```diff
@@ -9,17 +9,37 @@ const char kLogInfo[] = "ninja: ";
 const char kLogWarning[] = "ninja: warning: ";
```
jonesmz: I recommend taking a look at the Boost logging framework, which handles this very differently by allowing you to have multiple different logger destinations addressed by parameters to the logger being written to

configure.py
```diff
@@ -394,7 +394,7 @@ def binary(name):
 # Search for generated headers relative to build dir.
```
jonesmz: If we're going to move includes around like this, i personally have a strong preference for keeping public headers separated entirely from the internal headers.

ninja.git/
  - src/
    - *.cc
    - *.h
  - inc/ninja
    - *.h

This allows for package managers to simply "cp -r inc/ /usr/inc/" in order to install the public API.

src/status.h
```diff
@@ -109,11 +109,11 @@ struct StatusPrinter : Status {
         rate_ = times_.size() / ((times_.back() - times_.front()) / 1e3);
     }
 
-  private:
-    double rate_;
-    const size_t N;
-    queue<double> times_;
-    int last_update_;
+    private:
```
jonesmz: This is pretty subjective.

Personally I prefer it the old way :/

src/clean.cc
```diff
@@ -39,19 +39,19 @@ Cleaner::Cleaner(State* state,
     status_(0) {
```
jonesmz: If you want to just rip the bandaid off, you could instead rebase on this commit : https://github.com/splinter-build/splinter/tree/1-std-namespace

src/logger.cc
EliRibble: I'm fine looking in to that, I was under the impression that the maintainers of ninja would strongly prefer not to add any large new dependencies to the project. I was shooting for something relatively simple to implement that makes the default output behavior for ninja the same.

src/status.h
EliRibble: Agreed, totally subjective.

src/logger.cc
jonesmz: I wasn't implying that you should directly use the Boost logging framework.
Just to look at it as a source of inspiration.

EliRibble:
Okay, I'm back from vacation and have some bandwidth again to look into resolving these comments. I'm thinking at this point what I'm going to do is to create a new branch based on the latest ninja/master and try to get to the same endpoint with far fewer commits so that the path is clearer. I'm going to leave this PR open until I can create the new PR. I'll be working to incorporate the comments on this PR and to continue discussion where it makes sense.

src/logger.cc
EliRibble: I've never used the boost logging framework before, but I've now read some of the documentation after your suggestion. The logging framework is quite large, complex, and configurable. That's neat, but I don't think we want to build something like that entirely for ninja. Could you discuss one or two high points from boost logging you think I should incoporate?

src/ui.cc
EliRibble: Sorry, I'm probably just missing something obvious since I don't normally use getopt - what's bad about this?

jonesmz: Manipulating argc and argv is just gross in general, that's all.

Not saying that your code is wrong.

src/logger.cc
jonesmz: The most important aspects of it for the code that I'm writing is the following:

1. Multiple backends, so I can dynamically target log files, console output, or a logging server on the network, as I feel like at runtime.
2. Dynamic variables captured with each log statement, which can include the current functions name, or the current timestamp, or what have you. 
3. Easy to replace the functionality using macros. E.g. I can provide my own logging statements that customize what gets logged, by defining a macro that i call.

src/clean.cc
EliRibble: Interesting. Very interesting. I think I might do that. Are you the one who started the splinter project? I couldn't find any explanation in the code for splinter about who started it and why, other than that it seems to be C++17.

src/logger.cc
EliRibble: Thanks, that's helpful. That sounds like an enormous amount of work considering that all I originally wanted from the logging system was a way for library users to capture messages themselves and everyone else to continue to get information on stdout. If you were writing this how long do you think it would take you?

src/clean.cc
jonesmz: Yes. I'm the only contributor of code beyond a PR or two that I've merged.

There's another person on that project that did a bunch of project-management stuff, but I haven't heard from them in quite a while.

jonesmz: I update the set of branches in that github repo periodically when ninja updates things.

src/logger.cc
jonesmz: Honestly, if I were doing it for the Ninja project, I would just define some macro, e.g. NINJA_LOG_STREAM, or what have you, and only define the macro to the ninja project's preferred default if it doesn't already have a definition.

That allows consumers of the Ninja project to inject their own log implementation if they want to.

jonesmz: You could get a little bit more robust with

#define NINJA_LOG_STREAM(Severity) /**/

to allow for custom log implementations to capture the severity identifier as well, but it's not strictly necessary.

jonesmz: Two completely different way that this could be handled, if using macros isn't the right approach.

Dynamic polymorphism: Have Ninja provide an abstract base class type (e.g. an Interface) that can be inherited from, and then have the Ninja library accept an instance of this type when the library is instantiated. Any logging calls happen through this interface, using virtual functions.

Static polymorphism: Have the ninja codebase accept the logging object to use as a template parameter. This requires significantly more changes to the ninja codebase than the Dynamic polymorphism version though.

Implementing Dynamic polymorphism is a step on the road to static polymorphism regardless, so I'd start with that and then if the performance impact of the virtual function calls is too much,  investigate the static approach.

jonesmz: If the API for either version of the logging concepts above accept a `char (&)[]` (reference to character array) and length, we can use the __PRETTY_FUNCTION__ identifier (or similar) to still log the function name that's calling the logging code.

jhasse:
too many conflicts now, closing

PR #1730, opened 2020-02-03
Add 'inputs' tool to print out all inputs for a set of targets

naarcini:
Per discussion on ninja-build, this change introduces a ninja tool to enumerate all inputs for a set of targets:
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/ninja-build/vPM6COsFslY

This works almost identically to the `commands` tool except that it gets the `in_newline` binding instead of evaluating a command.

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -257,6 +257,10 @@ than the _depth_ mode.
 executed in order, may be used to rebuild those targets, assuming that all
 output files are out of date.
 
+`inputs`:: given a list of targets, print a list of all inputs which are used
+to rebuild those targets.
+_Available since Ninja 1.10._
+
 `clean`:: remove built files. By default it removes all built files
 except for those created by the generator.  Adding the `-g` flag also
 removes built files created by the generator (see <<ref_rule,the rule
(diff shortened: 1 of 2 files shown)


jonesmz:
Is this substantially different than https://github.com/ninja-build/ninja/pull/1711 ? If so, could you document pros and cons?

naarcini:
> Is this substantially different than #1711 ? If so, could you document pros and cons?

Great question! The approach in that PR specifically tries to look for code files only. In my case, I look at all inputs that would be found via the `commands` tool. This'll include additional artifacts such as ".stamp" files.

For my change vs. the the one you mention:

Pros:
- Less invasive. We never leave `ninja.cc` to introduce new hooks in other code. This mostly relies on existing code that figured out inputs already.
- More complete. We're getting all inputs that the tooling found and will use to actually run a build. I've tested this in Chromium and I find I get more code files in my version.

Cons:
- Potentially noisy if all someone wants are code files.

I can add this to the PR description if you'd like.

spectras:
Hello!
Is there any chance to see either this or #1711 merged at some point?

This feature is would make the life of monorepos much, much easier (each test job could query ninja and intersect its inputs with the list of changed files to know whether it can skip this run).

jhasse:
Yes. I'm preferring this over #1711 as it's less invasive.

naarcini:
Whoops, misclicked

PR #1731, opened 2020-02-05
CMake: fix object library usage

yegorich:
Object libraries cannot be use in target_link_libraries() command
as they are no normal binary files like *.a or *so but a collection
of object files.

See add_library() definition for details.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,8 +84,7 @@ target_compile_definitions(libninja PRIVATE _WIN32_WINNT=0x0601 __USE_MINGW_ANSI
 endif()
 
 # Main executable is library plus main() function.
-add_executable(ninja src/ninja.cc)
-target_link_libraries(ninja PRIVATE libninja libninja-re2c)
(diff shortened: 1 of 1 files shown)


jhasse:
> Object libraries cannot be use in target_link_libraries()

Why not? It works, doesn't it?

yegorich:
ninja will be used in [Buildroot](https://buildroot.org/). AFAIK it is not possible to compile the newest CMake version on all autobuilder hosts that may have older GCC. So for now the minimal required host CMake version is 3.10.

If you remove the make_policy(SET CMP0091 NEW) and allow CMake 3.10, it is possible to successfully build ninja with this PR.

Seems like in the newer version CMake allows to use object libraries as ordinary libraries. As without this patch CMake 3.10 throws the following error:

CMake Error at CMakeLists.txt:87 (target_link_libraries):
  Target "libninja" of type OBJECT_LIBRARY may not be linked into another
  target.  One may link only to STATIC or SHARED libraries, or to executables
  with the ENABLE_EXPORTS property set.

jhasse:
To keep it simple, I would like to only support the newest CMake versions. For platforms where CMake 3.15 won't compile we still support building via `configure.py`. That should work with Python 2 even.

PR #1738, opened 2020-02-08
Remove unnecessary string copy in MakeDirs method

hyperxor:
Removing unnecessary `dir` string copy in `MakeDirs` method and adding `const` for improving code.

### DIFF
--- a/src/disk_interface.cc
+++ b/src/disk_interface.cc
@@ -133,7 +133,7 @@ bool StatAllFilesInDir(const string& dir, map<string, TimeStamp>* stamps,
 // DiskInterface ---------------------------------------------------------------
 
 bool DiskInterface::MakeDirs(const string& path) {
-  const string& dir = DirName(path);
+  const std::string& dir = DirName(path);
   if (dir.empty())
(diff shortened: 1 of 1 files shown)


src/disk_interface.cc
```diff
@@ -132,11 +133,11 @@ bool StatAllFilesInDir(const string& dir, map<string, TimeStamp>* stamps,
 // DiskInterface ---------------------------------------------------------------
 
 bool DiskInterface::MakeDirs(const string& path) {
-  string dir = DirName(path);
+  const std::string& dir = DirName(path);
```
jhasse: `DirName` doesn't return a const &, therefore this won't avoid the copy (it actually introduces an indirection when optimizations are turned off). The only way to avoid the copy would be to use `std::string_view`, but we'll have to wait for C++17 for that.

hyperxor: @jhasse Yes, you are totally right!

jonesmz: Why can't StringPiece be used in place of std::string_view?

PR #1742, opened 2020-02-18
CMake: modernize, add metadata

scivision:
* add metadata to project()
* cmake_policy(set ... NEW) is implied for these policies by cmake_minimum_required(VERSION 3.15)
* string(APPEND ... helps avoid typos
* make tab indents everywhere in CMakeLists.txt as this seems to be the existing Ninja project practice

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,9 +15,9 @@ endif()
 
 if(MSVC)
 	set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
-	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /W4 /GR- /Zc:__cplusplus")
+	string(APPEND CMAKE_CXX_FLAGS " /W4 /GR- /Zc:__cplusplus")
 else()
-	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-deprecated -fdiagnostics-color")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1,9 +1,10 @@
 cmake_minimum_required(VERSION 3.15)
-cmake_policy(SET CMP0091 NEW)
-project(ninja)
+project(ninja
+LANGUAGES CXX
```
jonesmz: Indentation

jonesmz:
I've not heard of the cmake string() function.

Is that newish? Or am I just out of the loop?

jonesmz:
This change looks like a good improvement to me. We should merge it.

scivision:
string(APPEND ...) was added in CMake 3.4

jhasse:
What's the use-case for the metadata?

Regarding tab-indents: While it is used in CMakeLists.txt it might have been a mistake as the rest of the code base uses 2-spaces indent.

scivision:
A primary use for the "description" and "homepage_url" are by [CPack](https://cmake.org/cmake/help/latest/module/CPack.html), which can be used to generate binary / source installers for various platforms.

jhasse:
Not sure if I'd want to add CPack support.

Can you make a PR with only

> cmake_policy(set ... NEW) is implied for these policies by cmake_minimum_required(VERSION 3.15)

and

>string(APPEND ... helps avoid typos

?

jhasse:
Thanks :)

PR #1743, opened 2020-02-18
allow Intel compiler to build Ninja

scivision:
These changes allow Intel compiler to work along with existing GNU, Clang or MSVC.

* cmake: only set gnu-like flags for gnu-like compilers
* src/utils.h: handle intel compiler in preprocessor macro selection

Ninja is using C++17 syntax (e.g. fallthrough), so set C++17 standard for non-MSVC compilers. MSVC is stricter about syntax when specifying C++17 and will fail to build Ninja, so we don't yet specify C++17 for MSVC.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,10 +13,12 @@ if(CMAKE_BUILD_TYPE MATCHES "Release")
 	endif()
 endif()
 
+set(_gnulike GNU Clang)
+
 if(MSVC)
 	set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
 	string(APPEND CMAKE_CXX_FLAGS " /W4 /GR- /Zc:__cplusplus")
-else()
+elseif(CMAKE_CXX_COMPILER_ID IN_LIST _gnulike)
 	string(APPEND CMAKE_CXX_FLAGS " -Wno-deprecated -fdiagnostics-color")
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,10 @@ elseif(CMAKE_CXX_COMPILER_ID IN_LIST _gnulike)
 	string(APPEND CMAKE_CXX_FLAGS " -Wno-deprecated -fdiagnostics-color")
 endif()
 
+if(NOT MSVC)  # MSVC is too strict right now for Ninja code with C++17
+	set(CMAKE_CXX_STANDARD 17)
+endif()
+
 find_program(RE2C re2c)
 if(RE2C)
 	# the depfile parser and ninja lexers are generated using re2c.
--- a/src/util.h
+++ b/src/util.h
@@ -35,7 +35,7 @@ using namespace std;
 NORETURN void Fatal(const char* msg, ...);
 
 // Have a generic fall-through for different versions of C/C++.
-#if defined(__cplusplus) && __cplusplus >= 201703L
+#if defined(__cplusplus) && __cplusplus >= 201703L || defined(__INTEL_COMPILER)
 #define NINJA_FALLTHROUGH [[fallthrough]]
 #elif defined(__cplusplus) && __cplusplus >= 201103L && defined(__clang__)
 #define NINJA_FALLTHROUGH [[clang::fallthrough]]


CMakeLists.txt
```diff
@@ -15,10 +15,12 @@ if(CMAKE_BUILD_TYPE MATCHES "Release")
 	endif()
 endif()
 
+set(_gnulike GNU Clang)
```
jonesmz: Is this variable necessary if it's only used in one place?

src/util.h
```diff
@@ -40,9 +40,9 @@ NORETURN void Fatal(const char* msg, ...);
 #elif defined(__cplusplus) && __cplusplus >= 201103L && defined(__clang__)
 #define NINJA_FALLTHROUGH [[clang::fallthrough]]
 #elif defined(__cplusplus) && __cplusplus >= 201103L && defined(__GNUC__) && \
-    __GNUC__ >= 7
+    __GNUC__ >= 7 && !defined(__INTEL_COMPILER)
```
jonesmz: No version of the Intel compiler supports this attribute, but they claim to be a gcc? That seems... suboptimal, and personally I'd rather not see Ninja cluttered even more with special rules for every compiler.

Is there a version of the Intel compiler that would support the C++17 version of the fallthrough attribute?

Is there a specific macro that can be checked for support for the attribute, without consideration of compiler vender?

jonesmz:
I don't see how, but I don't suppose it would be possible to improve the continuous integration to also build with the Intel compiler, would it?

src/util.h
scivision: Since Intel 13.0, Intel defines __GNUC__ to signify ABI compatibility.
What I didn't try is seeing if simply "fallthrough" can be used in general with current compilers.
Yes, now I specify that C++17 standard should be used, since not all compilers default to C++17. However, MSVC C++17 is too strict for Ninja at the moment, so I let MSVC stay at default.

scivision:
Where CI with Intel compilers is desired, what I've seen is to make a private CI with Github Actions or Appveyor etc. that's triggered like the rest of the CI. What seems to be typically done instead is periodically someone with Intel compilers runs the test suite.

CMakeLists.txt
scivision: yes that's how CMake needs to have it if not using a Regex

src/util.h
scivision: Intel [defines](https://software.intel.com/en-us/cpp-compiler-developer-guide-and-reference-additional-predefined-macros) `__cplusplus=1` instead of a specific C++ revision level

scivision: Intel does not like the first option `#define NINJA_FALLTHROUGH [[fallthrough]]` so it's not useful to have

```cxx
#if defined(__cplusplus) && __cplusplus >= 201703L || defined(__INTEL_COMPILER)
```

scivision: This is the value Intel icpc is happy with:

```cxx
#elif defined(__GNUC__) && __GNUC__ >= 7 // gcc 7 or Intel compiler
#define NINJA_FALLTHROUGH __attribute__ ((fallthrough))
```

scivision: That's a bit odd since Intel [docs](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0188r1.pdf) say that `[[fallthrough]]` is supported...

scivision: I fixed the macro `[[fallthrough]]` logic to be cleaner. From the docs noted just above, I think this is one of the cleanest ways to do this.

src/util.h
```diff
@@ -35,7 +35,7 @@ using namespace std;
 NORETURN void Fatal(const char* msg, ...);
 
 // Have a generic fall-through for different versions of C/C++.
-#if defined(__cplusplus) && __cplusplus >= 201703L
+#if defined(__cplusplus) && __cplusplus >= 201703L || defined(__INTEL_COMPILER)
```
jhasse: Shouldn't the Intel compiler set `__cplusplus`?

scivision: * Intel [defines](https://software.intel.com/en-us/cpp-compiler-developer-guide-and-reference-additional-predefined-macros) `__cplusplus=1` instead of a specific C++ revision level.
* Intel [fallthrough support](https://software.intel.com/en-us/articles/c17-features-supported-by-intel-c-compiler) is on all [currently supported versions](https://software.intel.com/en-us/articles/intel-parallel-studio-xe-supported-and-unsupported-product-versions) (2018 version expected to be dropped from support ~ March 2020)

jhasse: > Intel defines __cplusplus=1 instead of a specific C++ revision level.

Well, they aren't complying with the standard then. I don't want to include any workarounds for the Intel compiler (even if it's just a ifdef), sorry.

jonesmz: @jhasse 

You want to support 10 plus year old versions of GCC, but you won't support a compiler released less than 2 years ago because of a single line change to an ifdef?

That's messed up.

jhasse: @jonesmz Can you try to be more friendly in your comments? You often don't add much to a discussion other than saying how incompetent I am (or the "ninja project" is) apparently. It hurts and doesn't help your case.

scivision:
Thanks--I was being too careless with the options. Ninja has a large user / developer base, making it more important to be conservative and careful with changes.

src/util.h
jonesmz: If you feel that I consider you or the Ninja project in general to be incompetent, that is not the case, and I'm sad that you have that impression. I don't believe that I've ever said anything with the intention of implying that I think anyone here is incompetent, and would be happy to be educated on bad wording choices.

I am unhappy that the project has so many undocumented or under-documented polices, and I strongly disagree with several of the ones that have been documented, but that doesn't imply that I think anyone is incompetent. What I do think is that the project is not consistent with it's own stated policies and goals.

This particular PR is one example of the inconsistencies. Either the project should support ancient and/or questionably standards complaint compilers, or it shouldn't. The Intel compiler in question may not be standards compliant in this specific way, but at least the version of the compiler being discussed is new-ish and still under official support from the vendor.

Further, the cost of compatibility was only one line of code, in an existing #ifdef clause no less. (at the time this PR was still opened anyway, I see that @scivision has found a different way to implement support for the Intel compiler).

On the other hand, we have a version of getopt from 2011 bundled into the code and patched to support AIX and it's weird requirements, and we're using std::auto_ptr (Widely considered to be absolute garbage, and one of the few library features actually removed from C++, ever) all over the place because we still support C++03 because of an unknown list of compilers and platforms that no one will name.

I'm not saying that supporting older platforms is bad. I'm saying that if we're going to support older platforms, with all of the compromises that doing so requires, then rejecting compatibility with newer platforms is inconsistent. Either support everything that requires a little bit of tweaking, or don't support anything that requires non-standards compliant code. 

This project is fundamental infrastructure for a very large portion of world-wide computer technology, and is a tool that I use daily in my professional work. I have a strong interest in seeing it improved, and I've supplied code, or extensive code-review, on every issue that I have with the Ninja project, at no small time cost, in order to do so.

jhasse: > I don't believe that I've ever said anything with the intention of implying that I think anyone here is incompetent, and would be happy to be educated on bad wording choices.

Try to avoid rhetorical questions and "*shrug*".

> What I do think is that the project is not consistent with it's own stated policies and goals.

Most inconsistencies come from different people making their own independent decisions.

> On the other hand, we have a version of getopt from 2011 bundled into the code and patched to support AIX and it's weird requirements

AIX isn't the only reason for getopt, I think Windows among other platforms requires it, too.

> I'm not saying that supporting older platforms is bad. I'm saying that if we're going to support older platforms, with all of the compromises that doing so requires, then rejecting compatibility with newer platforms is inconsistent.

I wouldn't consider the Intel compiler a "platform". Furthermore we certainly *do* support newer platforms, e.g. we'll use unique_ptr instead of auto_ptr on newer compilers.

> [...] or don't support anything that requires non-standards compliant code.

Extreme rules like this hardly every work out. You could say "try to avoid supporting anything that requires non-standards compliant code whenever possible" or "minimize usage of non-standard compliant code".

PR #1746, opened 2020-02-19
travis-ci: add powerpc arch

scivision:
add IBM PowerPC arch on Travis-CI


jhasse:
I don't think this is worth the extra build time. Can you think of something this would catch?

scivision:
not in particular, it's not necessary but easy to access if a future need arises.

PR #1748, opened 2020-03-04
enable Intel, IBM XL, and PGI compilers without header/source changes

scivision:
The goal in the PR was to enable:

* Intel compiler `icpc` and `icl` on Linux, Mac, Windows. It turns out that Intel Mac/Linux compiler requires CMake command line options `cmake -DCMAKE_CXX_FLAGS=-std=c++17` 
* PGI compiler `pgc++` on Linux
* IBM Power XL compiler `xlc++`

while:

1. avoid modifying Ninja C++ source
2. avoid compiler specific CMake `if()` 
3. remain compatible with older compilers

## Results

* all `ninja_test` pass on PGI, IBM XL and Intel
* some benchmarks run faster as built by Intel compiler
* works with or without `re2c` 

#### PGI 19.10

```sh
CXX=pgc++ CC=pgcc cmake -DCMAKE_BUILD_TYPE=Release
```

* canon_perftest: min 77ms  max 79ms  avg 78.0ms
* clparser_perftest: Parse 32768 times in 2069ms avg 63.1us

#### Intel 19.1

```sh
CXX=icpc CC=icc cmake -DCMAKE_BUILD_TYPE=Release
```

* canon_perftest: min 78ms  max 79ms  avg 78.8ms
* clparser_perftest: Parse 65536 times in 3435ms avg 52.4us

#### Gfortran 7.4.0

```sh
CXX=g++ CC=gcc cmake -DCMAKE_BUILD_TYPE=Release
```

* canon_perftest:  min 93ms  max 94ms  avg 93.4ms
* clparser_perftest: Parse 65536 times in 3453ms avg 52.7us

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,7 @@
 cmake_minimum_required(VERSION 3.15)
 project(ninja)
 
+# --- optional link-time optimization
 if(CMAKE_BUILD_TYPE MATCHES "Release")
 	include(CheckIPOSupported)
 	check_ipo_supported(RESULT lto_supported OUTPUT error)
@@ -13,13 +14,23 @@ if(CMAKE_BUILD_TYPE MATCHES "Release")
 	endif()
 endif()
 
+# --- compiler flags
 if(MSVC)
 	set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
 	string(APPEND CMAKE_CXX_FLAGS " /W4 /GR- /Zc:__cplusplus")
 else()
-	string(APPEND CMAKE_CXX_FLAGS " -Wno-deprecated -fdiagnostics-color")
+	include(CheckCXXCompilerFlag)
+	check_cxx_compiler_flag(-Wno-deprecated flag_no_deprecated)
+	if(flag_no_deprecated)
+		string(APPEND CMAKE_CXX_FLAGS " -Wno-deprecated")
+	endif()
+	check_cxx_compiler_flag(-fdiagnostics-color flag_color_diag)
+	if(flag_color_diag)
+		string(APPEND CMAKE_CXX_FLAGS " -fdiagnostics-color")
+	endif()
 endif()
 
+# --- optional re2c
 find_program(RE2C re2c)
 if(RE2C)
 	# the depfile parser and ninja lexers are generated using re2c.


jonesmz:
Can you elaborate on what MSVC is too strict about?

CMakeLists.txt
```diff
@@ -13,13 +13,19 @@ if(CMAKE_BUILD_TYPE MATCHES "Release")
 	endif()
 endif()
 
+set(_gnulike GNU Clang)
```
jhasse: This is missing a lot of GNU like compilers (e.g. `AppleClang`). Is there a Modern CMake way to check if a flag is supported maybe?

CMakeLists.txt
```diff
@@ -13,13 +13,19 @@ if(CMAKE_BUILD_TYPE MATCHES "Release")
...
 if(MSVC)
 	set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
 	string(APPEND CMAKE_CXX_FLAGS " /W4 /GR- /Zc:__cplusplus")
-else()
+elseif(CMAKE_CXX_COMPILER_ID IN_LIST _gnulike)
 	string(APPEND CMAKE_CXX_FLAGS " -Wno-deprecated -fdiagnostics-color")
 endif()
 
+if(NOT MSVC)  # MSVC is too strict right now for Ninja code with C++17
+	set(CMAKE_CXX_STANDARD 17)
```
jhasse: IIRC the Modern CMake way to do this is `target_compile_features` with `cxx_std_17`.

I also wonders what happens on compilers that don't support C++17?

CMakeLists.txt
scivision: Excellent, at the least, these two options could be implemented with complementary logic--enable them *only* when a known-incompatible compiler (Intel icpc/icl) is being used.

CMakeLists.txt
scivision: Yes that would be better way to do it. 

CMake by default is supposed to gracefully degrade for older compilers. There are naturally corner cases not anticipated. Like the option above, it may be safe to enable this only when known to be needed -- e.g. for Intel icl/icpc

scivision:
Great feedback--as more projects use C++17, Ninja makes an interesting test case as it touches so many platform/compiler combinations. 
If it seems good, I will plan to change this PR to manipulate these compiler options only as known-necessary (for Intel icl/icpc at least). This should help avoid causing problems with older compilers that do work today.

jhasse:
I want to keep the compiler specific ifs in CMakeLists.txt to a minimum, maybe it's possible to check if a compiler supports the flags we want to add?

CMakeLists.txt
scivision: in eb77f3fc4 this is done now with [check_cxx_compiler_flag](https://cmake.org/cmake/help/latest/module/CheckCXXCompilerFlag.html)

CMakeLists.txt
scivision: It turns out that target_compile_features() breaks older compilers like GCC 4.8 since that function strictly requires C++17, while the code supports graceful fallback and so instead we use 8cef6933d1dbd
[CXX_STANDARD](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html) property which also decays gracefully.

scivision:
96e6a80a58 makes MSVC work in C++17 mode. Looking into this, it appears the use of `using namespace std` in some of the Ninja files triggers this MSVC bug, and this workaround is from Microsoft.

scivision:
Are there any current Ninja benchmarks of interest beyond those included in this repo? At first glance, I found that Intel compiler can make the Ninja benchmarks in this repo run up to 2x to 5x faster than GCC. 

I suppose even if the real-world Ninja speedup wasn't that dramatic, there are projects that would appreciate even 10's of percent speedup.

CMakeLists.txt
scivision: ...however, there are still remaining bugs invoked by *carte blanche* setting of C++17 standard due to `using namespace std` in Ninja code. So in the end a compiler-specific if(Intel) was the least-change way of doing this.

jhasse:
I like 2, but 3 adds a compiler specific if. IMHO when someone really wants to compile with the Intel compiler he can enable C++17 via `CXXFLAGS` himself.

CMakeLists.txt
jonesmz: > IIRC the Modern CMake way to do this is target_compile_features with cxx_std_17.

SET(CMAKE_CXX_STANDARD 17) if you want it to apply to all target in the project.

jonesmz: > carte blanche setting of C++17 standard due to using namespace std in Ninja code.

@scivision 

Does this PR resolve that problem? Or would it also need to remove "using namespace std" in implementation files as well ? https://github.com/ninja-build/ninja/pull/1614

scivision:
I took out the intel-specific if(). Indeed, `cmake -DCMAKE_CXX_FLAGS=c++17` works for Intel compilers.

CMakeLists.txt
scivision: I tried #1614 with MSVC 16.5.2 and the error still existed:

```
error C2872: 'byte': ambiguous symbol
```

jhasse:
Thanks!

CMakeLists.txt
jonesmz: Ah, ok, thanks for the information.

scivision:
Thanks for persisting through the PR process on this. In the end I realized and agreed that we should not introduce build logic that supports incomplete and/or incorrect compiler vendor behavior.

PR #1759, opened 2020-03-31
Remove redundant test

jonesmz:


jhasse:
Originally OneStep2 made in2 dirty while OneStep made cat1 dirty. They became somewhat identical in 94f2fd0acff696458a0e08727e11060a5e02ea5a. Not sure what we should do here.

jhasse:
Until we figure this out, we should not remove the test. Feel free to open an issue though.

Issue #1764, opened 2020-04-18
ninja v1.10.0 fails to run on macOS 10.12


past-due:
ninja v1.10.0 is failing to run on macOS 10.12 (on our CI setup), with the following error:
```
dyld: lazy symbol binding failed: Symbol not found: ____chkstk_darwin
Referenced from: /<path>/vcpkg/downloads/tools/ninja-1.10.0-osx/ninja
Expected in: /usr/lib/libSystem.B.dylib
```

It appears as though the deployment target isn't being explicitly specified in the GitHub Actions macOS workflow.
https://cmake.org/cmake/help/v3.12/envvar/MACOSX_DEPLOYMENT_TARGET.html

PR coming shortly.

PR #1765, opened 2020-04-18
macOS workflow: specify the minimum version of macOS supported

past-due:
Specifying `MACOSX_DEPLOYMENT_TARGET` is then [used by CMake](https://cmake.org/cmake/help/latest/envvar/MACOSX_DEPLOYMENT_TARGET.html) to set the default value for [CMAKE_OSX_DEPLOYMENT_TARGET](https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_DEPLOYMENT_TARGET.html)

Should fix: #1764

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -18,6 +18,8 @@ jobs:
 
     - name: Build ninja
       shell: bash
+      env:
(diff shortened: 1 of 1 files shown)


jhasse:
Thank you very much! This was long overdue.

PR #1767, opened 2020-04-21
configure.py Windows more reliable detect MinGW

scivision:
Before this, the general Windows user not already in an MSVC prompt gets a vague FileNotFoundError if they run Ninja's configure.py.

This PR retains the original default of looking for MSVC first, adding the behavior of next looking for MinGW if MSVC not found and **gives a clear error message if neither is found**.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,6 +20,9 @@ matrix:
     - image: Ubuntu1804
       MSYSTEM: MSVC
 
+init:
+- cmd: set PATH=C:\Python36;%PATH%
+
 for:
   -
     matrix:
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -20,6 +20,9 @@ matrix:
     - image: Ubuntu1804
       MSYSTEM: MSVC
 
+init:
+- cmd: set PATH=C:\Python36;%PATH%
```
jhasse, line 24: why?

scivision: The shutil.which requires at least Python 3.5. I just put the latest supported Python 3.6. 
Does Ninja officially support a minimum Python release? Then there would be a different way to do this fix depending on minimum Python version.

jhasse: Nothing really official. As Python 2.7 has worked with configure.py in the past, I think people expect that to continue to work for the time being.

jhasse:
Old Python support is more important for now, especially since there's also CMake to build.

PR #1768, opened 2020-04-24
Run more CI tests on Ubuntu 20.04

jhasse:


.github/workflows/test.yml
```diff
@@ -0,0 +1,29 @@
...
+    types: published
+
+jobs:
+  build:
+    runs-on: [ubuntu-latest]
+    steps:
+    - uses: actions/checkout@v1
+    - name: Install dependencies
+      run: |
+        sudo apt install -y python3-pytest
```
scivision, line 16: I have intermittent failures with `apt` on actions if not doing `apt update` first.
However, in this case, I would consider doing `pip install pytest` instaed

.github/workflows/test.yml
```diff
@@ -0,0 +1,29 @@
...
+        curl -L -O https://github.com/Kitware/CMake/releases/download/v3.16.2/cmake-3.16.2-Linux-x86_64.sh
```
scivision, line 17: might consider `pip install cmake` instead

.github/workflows/test.yml
```diff
@@ -0,0 +1,29 @@
...
+        chmod +x cmake-3.16.2-Linux-x86_64.sh
+        ./cmake-3.16.2-Linux-x86_64.sh --skip-license --prefix=/usr/local
+    - name: Build ninja
+      shell: bash
+      run: |
+        cmake -Bbuild-cmake -H. -DCMAKE_BUILD_TYPE=Debug -GNinja
+        cd build-cmake
```
scivision, line 24: line 24-25 can be replaced by `cmake --build build-cmake`

.github/workflows/test.yml
```diff
@@ -0,0 +1,29 @@
...
+        ninja
+    - name: Unit tests
+      run: ./build-cmake/ninja_test
```
scivision, line 27: to do this using CMake, consider

```sh
cd build-cmake
ctest
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,29 @@
...
+    - name: Python tests
+      run: cd build-cmake && pytest-3 --color=yes ..
```
scivision, line 29: I think this can just be `run: pytest`

.github/workflows/test.yml
```diff
@@ -0,0 +1,29 @@
...
+  push:
+  release:
+    types: published
+
+jobs:
+  build:
+    runs-on: [ubuntu-latest]
+    steps:
+    - uses: actions/checkout@v1
+    - name: Install dependencies
```
scivision, line 14: consider inserting here:

```yml
    - uses: actions/setup-python@v1
      with:
        python-version: '3.x'
```

to use python in Actions in a friendlier/easier/faster way

.github/workflows/test.yml
```diff
@@ -0,0 +1,29 @@
...
+  pull_request:
+  push:
+  release:
+    types: published
+
+jobs:
+  build:
+    runs-on: [ubuntu-latest]
+    steps:
+    - uses: actions/checkout@v1
```
scivision, line 13:
```yml
- uses: actions/checkout@v2
```
is the latest recommendation

.github/workflows/test.yml
jhasse: Why?

.github/workflows/test.yml
jhasse: Can I install a specific CMake version with pip?

jhasse: Can I install a specific CMake version with pip?

.github/workflows/test.yml
jhasse: What's the advantage though?

.github/workflows/test.yml
jhasse: Why would that be faster?

.github/workflows/test.yml
scivision: Re: using `pip` instead of `apt` -- one gets the latest Pytest release, but the real motivation here is to have the Actions image startup and run the test faster, by avoiding slower `apt update`. Just like for a Linux laptop, once in a while `apt update` is needed because the distro changed something such that `apt install` won't work until `apt  update`.  This type of problem can persist at least for hours until Actions staff update their Ubuntu image cache.

.github/workflows/test.yml
scivision: Yes you can pick from the CMake versions at: https://pypi.org/project/cmake/#history

`pip install cmake==3.16.2` and so on

.github/workflows/test.yml
scivision: The advantage is, besides not having to remember which test names to type, is that under certain conditions, especially on Windows, a test may silently fail. CTest ensures the failure is noisy on the console:

```
ninja\build> .\ninja_test.exe

ninja\build> ctest
Test project C:/Users/micha/code_other/ninja/build
    Start 1: NinjaTest
1/1 Test #1: NinjaTest ........................Exit code 0xc0000139
***Exception:   0.01 sec
```

.github/workflows/test.yml
scivision: Instead of `apt install` out of date python packages, "uses python" is instant, and enables `pip install` which is typically faster and gives the latest version (or allows selecting a specific version)

jhasse: Not using "uses python" is also instant, since ubuntu-latest ships Python 3 ;)

.github/workflows/test.yml
jhasse: So `.\ninja_test.exe` did fail? GitHub Actions checks for exit code != 0.

.github/workflows/test.yml
jhasse: Neat! Thanks :)

.github/workflows/test.yml
scivision: This would be to show that there wasn't some test unintentionally excluded in the CI. For example, if someone added a test in CMakeLists but didn't check that the CI script also explicitly calls that test. If CTest is used in a CMake CI script, then one knows all intended tests are run.

It's just a general "best practice" when using CMake, for most situations

jhasse: https://github.com/ninja-build/ninja/runs/673809404?check_suite_focus=true showed that ctest doesn't give enough output in some cases, while calling the test binary directly does: https://github.com/ninja-build/ninja/pull/1588/checks?check_run_id=673891970
I'd also want the output, even if the tests succeed.

Given that we only have one test binary (and probably won't add another one), I'll pass on ctest.

.github/workflows/test.yml
jhasse: Unfortunately I need apt anyway for clang-tidy and other packages.

.github/workflows/test.yml
scivision: OK that's cool -- for ctest, it wants uppercase `ctest -V`  or `ctest -VV`

jhasse: Oh. Why didn't it complain?

jhasse:
Had to revert due to a GitHub Actions change :(

(note to self: check for https://clang.llvm.org/extra/clang-tidy/checks/google-global-names-in-headers.html)

PR #1770, opened 2020-04-30
cmake: print feature summary

scivision:
this prints the results of the CMake configuration at the end, helping users see if they omitted desired options.

For example, before this PR, if the user forgets to specify "Release" built type, not only do they get -O0, they also silently omit LTO. This may remind them to specify Release.

Example:

```
> cmake -B build
...
-- The following features have been enabled:

 * re2c, lexer generator

-- The following features have been disabled:

 * LTO, Link time / interprocess optimization

-- Configuring done
-- Generating done
```

Or 

```
> cmake -B build
...
CMake Warning at CMakeLists.txt:43 (message):
  re2c was not found; changes to src/*.in.cc will not affect your build.


-- The following features have been disabled:

 * LTO, Link time / interprocess optimization
 * re2c, lexer generator

-- Configuring done
-- Generating done

```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,10 +7,7 @@ if(CMAKE_BUILD_TYPE MATCHES "Release")
 	check_ipo_supported(RESULT lto_supported OUTPUT error)
 
 	if(lto_supported)
-		message(STATUS "IPO / LTO enabled")
 		set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
-	else()
-		message(STATUS "IPO / LTO not supported: <${error}>")
 	endif()
 endif()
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -138,3 +135,10 @@ enable_testing()
 add_test(NinjaTest ninja_test)
 
 install(TARGETS ninja DESTINATION bin)
+
+include(FeatureSummary)
+
+add_feature_info(LTO lto_supported "Link time / interprocess optimization  ${error}")
```
jhasse, line 141: ${error} doesn't seem to work now?

scivision: I think ${error} prints as expected; if I manually do

```cmake
set(error "hello")
```

and then

```sh
cmake -B build
```

```
-- The following features have been disabled:

 * LTO, Link time / interprocess optimization  hello
 * re2c, lexer generator
```

Some compilers that don't have LTO like PGI also don't have any "error" output

jhasse:
Thanks for the PR. But IMHO CMake's feature summary is rather ugly. I prefer the way it's currently displayed (LTO enabled / not supported and a warning for re2c).

PR #1771, opened 2020-04-30
bugfix: MinGW in CMake like configure.py bootstrap

### LINKED ISSUE #1888: MinGW: -static flag needed in CMake like configure.py
I just confirmed that this issue still exists with current MSYS2/MinGW/GCC 10.2.0.

## Symptoms

If a MSYS2 user does not have MinGW DLL on their Windows PATH, ninja.exe will fail to run, when **built from Ninja's CMakeLists.txt**. 
However, a ninja.exe built from Ninja's configure.py --bootstrap does run. By inspection, the configure.py adds a "--static" flag to MinGW building ninja.exe, while CMakeLists.txt does not.

## Why this is important

Some Ninja users may not wish to always have MinGW on their path--apparently this was recognized for the configure.py build method but was apparently missed for those using CMake to build ninja.exe. All we are looking for is parity between CMakeLists.txt and configure.py Ninja build method for MinGW/MSYS2

scivision:
without this `-static` flag, MinGW ninja build ninja.exe has
error code 0xc0000139 even with gdb, instant crash

I noticed this flag was in configure.py, so I added it to CMake for MinGW and it fixed the problem for MinGW / MSYS2 built Ninja.

fixes #1888

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,10 @@ else()
 	endif()
 endif()
 
+if(MINGW)
+	string(APPEND CMAKE_CXX_FLAGS " -static")
+endif()
+
 # --- optional re2c
 find_program(RE2C re2c)
 if(RE2C)


jhasse:
> without this -static flag, MinGW ninja build ninja.exe has error code 0xc0000139 even with gdb, instant crash

Any idea why?

scivision:
This flag was added by https://github.com/ninja-build/ninja/commit/7e3e51f60711d73731f50fc31808f90234613ad8

I think it's because otherwise users would need to have the MinGW DLLs in their PATH.

jhasse:
I think flags like that should be added via `-DCMAKE_CXX_FLAGS`. It's off by default, too, which has its reasons I guess. If gdb crashes with shared runtime, I think that's a bug (in gdb / MinGW).

Issue #1693, opened 2019-11-26
Invalid volume label syntax with MinGW GCC 10


expnkx:
![image](https://user-images.githubusercontent.com/57303909/69611973-dfd9f580-0ffc-11ea-814f-b709f394166e.png)

jhasse:
Can you try with #1671?

jhasse:
Wait, no the problem seems to be `d/:`. No idea where that invalid volume label syntax comes from.

expnkx:
> Wait, no the problem seems to be `d/:`. No idea where that invalid volume label syntax comes from.

How to try #1671?
"Wait, no the problem seems to be d/:. No idea where that invalid volume label syntax comes from."
I do not know. I will upload my MinGW GCC and you can test it. It is so sad ninja is the only tool which can be used on windows for GCC, however, it breaks as well :(

jhasse:
Where did you download MinGW?

I'm using [MSYS2](http://www.msys2.org/) which works with mingw-w64 and Ninja.

expnkx:
> Where did you download MinGW?
> 
> I'm using [MSYS2](http://www.msys2.org/) which works with mingw-w64 and Ninja.

No. Your version is too old. Mine is GCC 10.0
https://bitbucket.org/ejsvifq_mabmip/mingw-gcc-mcfgthread-pacman/src/master/

expnkx:
![image](https://user-images.githubusercontent.com/57303909/69612705-427fc100-0ffe-11ea-8269-879e12ef48bb.png)

jhasse:
So it works with 9 but stops working with 10?

expnkx:
> So it works with 9 but stops working with 10?

Yes

expnkx:
I will upload a full version so you can test it. Just use cmake + ninja to build ninja and you will see the problem.

jhasse:
btw: You can try PRs by downloading the artifact. For #1671 that would be: https://github.com/ninja-build/ninja/runs/320687337

expnkx:
> btw: You can try PRs by downloading the artifact. For #1671 that would be: https://github.com/ninja-build/ninja/runs/320687337

No use

expnkx:
I have uploaded the latest version. You can unzip and try it.

https://bitbucket.org/ejsvifq_mabmip/mingw-gcc/src/master/

expnkx:
@jhasse @evmar Can you fix the problem??

jhasse:
[...]

PR #1774, opened 2020-05-07
bugfix: Process escaped colon in GCC depfiles. 

zero9178:
Since this commit https://github.com/gcc-mirror/gcc/commit/d7b6aee8cd346328b6f22fc415ffc28b9cf8a285 in GCC which has been included in the now released GCC 10 version, colons are escaped with backslashes just like hashtags https://github.com/gcc-mirror/gcc/blob/c7100843831147a034fe37d231c54ac53ceace45/libcpp/mkdeps.c#L158. This has lead to issues with GCC on Windows as colons are used in paths but would also lead to issues on other platforms if targets or file names contain colons. This patch fixes this issue using similar code as with hashtags. 

Fixes https://github.com/ninja-build/ninja/issues/1693


src/depfile_parser_test.cc
```diff
@@ -132,13 +132,10 @@ TEST_F(DepfileParserTest, Escapes) {
   // Put backslashes before a variety of characters, see which ones make
   // it through.
   string err;
-  EXPECT_TRUE(Parse(
-"\\!\\@\\#$$\\%\\^\\&\\[\\]\\\\:",
-      &err));
+  EXPECT_TRUE(Parse("\\!\\@\\#$$\\%\\^\\&\\[\\]\\\\::", &err));
```
jhasse, line 135: Hm ... why change this input? Maybe add another test instead.

src/depfile_parser_test.cc
```diff
@@ -132,13 +132,10 @@ TEST_F(DepfileParserTest, Escapes) {
...
   ASSERT_EQ("", err);
   ASSERT_EQ(1u, parser_.outs_.size());
-  EXPECT_EQ("\\!\\@#$\\%\\^\\&\\[\\]\\\\",
-            parser_.outs_[0].AsString());
+  EXPECT_EQ("\\!\\@#$\\%\\^\\&\\[\\]\\:", parser_.outs_[0].AsString());
```
jhasse, line 138: Can you keep this on 2 lines?

src/depfile_parser_test.cc
zero9178: I believed this test to be quite fitting as it's the only test I could find that deals with character contains. It also already tested for correct escaping of # which is treated the same way as : in the linked GCC source code.

zero9178:
It might break something if output from Clang or pre 10 GCC were to somehow output a backslash followed by a colon. The only way I could imagine that happening is if a users file name contains that sequence or a users file starts with a colon and the directory separator used is a backslash. Normal Colons are unaffected

I am not quite sure which version would fit either. The above case is probably extremely rare or simply shouldn't be supported. But using 1.11.0 might just be safer

lhmouse:
> It might break something if output from Clang or pre 10 GCC were to somehow output a backslash followed by a colon. 

On Windows filenames cannot contain colons. You may create such names with syscalls directly, but most WinAPIs will get confused.

jhasse:
What about non-Windows platforms though?

src/depfile_parser_test.cc
jhasse: Sure, but : is a bit special and I like having the same input to see exactly how the parser changed. By changing this test I had to double check what exactly is going on.

Also: Doesn't the double-colon have a special meaning?

Please add a new test which explicitly checks with a real world scenario like `C\:/msys64/mingw32/i686-w64-mingw32/include/stdio.h`.

zero9178:
Mac doesn't allow using colons in filenames either. Only platform that allows Colons is Linux (when not using FAT32 too mind you). As it is not compatible with Mac, Windows nor FAT32 I would think that a **Backslash followed by a colon** in a filename is an incredibly rare or probably non existent use case. If the filename contains a colon that does not have a backslash in front of it it will continue to work after this patch.

I would prefer GCC 10 to work on Windows (not only MinGW but also cross compilers such as the GCC provided by ARM for it's Microcontrollers) than for people who use a \: in their filenames

src/depfile_parser_test.cc
zero9178: You're right.

Added a test that uses two paths from depfiles of a project of mine that has once been created by GCC 10 and once by Clang

ctrlcctrlv:
Maybe _you_ would prefer that, but you're fixing a bug by knowingly introducing another bug in a way that's overbroad. If you really think there's "probably" no file named e.g. `\:.png` on Earth, I'd like to know how you came to this conclusion. Because I can think of several things that might cause one to exist.

The proper fix is, of course, (assuming Ninja is even the right place to fix this and not GCC itself), to modify the parser to act differently on Win32 than on other platforms. At the very least, you ought to only match PCRE `^[A-Za-z]\\:` as Win32 does not allow drive letters above Z and `\:` isn't a problem if not at the beginning.

lhmouse:
Note on Windows GCC outputs _forward_ slashes as path separators. You just interpret `\:` as a plain colon. No more than that. Not Windows-specific.

zero9178:
As GCC does still generate valid make syntax and projects still compile with make, ninja would need to adapt to the new behaviour. If GCC were to introduce a patch reverting it from generating \: for : I would happily close this pull request.

Regarding people with `\:.png`: Those people are currently not able to compile their project using ninja and GCC 10 either. A depfile created by GCC 9 for  `\:.png` would simply be  `\:.png:` while GCC 10 would create a depfile listing  `\\:.png:` as output instead. The GCC change is done on every platform, it just affects Windows people the most. 

See https://imgur.com/NUwqtMX for the different outputs in GCC 9 and GCC 10.

jhasse:
> A depfile created by GCC 9 for `\:.png` would simply be `\:.png:` [...]

And therefore not work with (GNU) Make?

zero9178:
Correct, as GNU Make would parse it as `:.png` and not find the file. I am guessing that the intent of the GCC patch was to fix that issue. If I try to run make on the depfile created by GCC 9 shown in the above screenshot you can see that it parses the filename as `:test.c` instead of `\\test.c`. Funnily GCC 10 errors with some other error that I have no clue about lol

image

src/depfile_parser_test.cc
```diff
@@ -133,7 +133,7 @@ TEST_F(DepfileParserTest, Escapes) {
   // it through.
   string err;
   EXPECT_TRUE(Parse(
-"\\!\\@\\#$$\\%\\^\\&\\[\\]\\\\:",
+"\\!\\@\\#$$\\%\\^\\&\\[\\]\\\\ :",
```
jhasse, line 136: Not sure about the space. Shouldn't : at the end be special and therefore not need any escaping?

zero9178: Just added a new commit. I used lookahead to make sure that the next character is not whitespace nor EOF. As whitespace in side of filenames are escaped this should work I believe. Sadly not an expert on re2c

zero9178:
Would it be okay for a test to call Parse multiple times? Also seems like their is a bug in older versions of re2c that I'll have to find a workaround for

src/depfile_parser_test.cc
```diff
@@ -157,6 +157,23 @@ TEST_F(DepfileParserTest, EscapedColons)
   EXPECT_EQ("c:\\gcc\\x86_64-w64-mingw32\\include\\stddef.h",parser_.ins_[0].AsString());
 }
 
+TEST_F(DepfileParserTest,EscapedTargetColon)
```
jhasse, line 160: Add a space after ,

jhasse:
We can bump the minimum required re2c version if needed.

src/depfile_parser_test.cc
```diff
@@ -142,6 +142,38 @@ TEST_F(DepfileParserTest, Escapes) {
   ASSERT_EQ(0u, parser_.ins_.size());
 }
 
+TEST_F(DepfileParserTest, EscapedColons)
+{
+  string err;
```
jhasse, line 147: Please add std::

src/depfile_parser_test.cc
```diff
@@ -142,6 +142,38 @@ TEST_F(DepfileParserTest, Escapes) {
...
+  // Tests for correct parsing of depfiles produced on Windows by both Clang, GCC pre 10 and GCC 10
+  EXPECT_TRUE(Parse(
+"c\\:\\gcc\\x86_64-w64-mingw32\\include\\stddef.o: \\\n"
+" c:\\gcc\\x86_64-w64-mingw32\\include\\stddef.h \n"
+      ,&err));
```
jhasse, line 152: space

zero9178:
> We can bump the minimum required re2c version if needed.

I found a workaround so I think there shouldn't be a need

src/depfile_parser_test.cc
```diff
@@ -142,6 +142,38 @@ TEST_F(DepfileParserTest, Escapes) {
...
+{
+  std::string err;
+  // Tests for correct parsing of depfiles produced on Windows by both Clang, GCC pre 10 and GCC 10
+  EXPECT_TRUE(Parse(
+"c\\:\\gcc\\x86_64-w64-mingw32\\include\\stddef.o: \\\n"
+" c:\\gcc\\x86_64-w64-mingw32\\include\\stddef.h \n",
+      &err));
+  ASSERT_EQ("", err);
+  ASSERT_EQ(1u, parser_.outs_.size());
+  EXPECT_EQ("c:\\gcc\\x86_64-w64-mingw32\\include\\stddef.o",parser_.outs_[0].AsString());
```
jhasse, line 155: space after ,

src/depfile_parser_test.cc
```diff
@@ -142,6 +142,38 @@ TEST_F(DepfileParserTest, Escapes) {
...
+  ASSERT_EQ(1u, parser_.ins_.size());
+  EXPECT_EQ("c:\\gcc\\x86_64-w64-mingw32\\include\\stddef.h",parser_.ins_[0].AsString());
```
jhasse, line 157: space after ,

src/depfile_parser_test.cc
```diff
@@ -142,6 +142,38 @@ TEST_F(DepfileParserTest, Escapes) {
...
+{
+  std::string err;
+  EXPECT_TRUE(Parse(
+"foo1\\: x\n"
+"foo1\\:\n"
+"foo1\\:\r\n"
+"foo1\\:\t\n"
+"foo1\\:",
+      &err));
+  ASSERT_EQ("",err);
```
jhasse, line 170: space after ,

src/depfile_parser_test.cc
```diff
@@ -142,6 +142,38 @@ TEST_F(DepfileParserTest, Escapes) {
...
+  ASSERT_EQ(1u,parser_.outs_.size());
```
jhasse, line 171: space after ,

src/depfile_parser_test.cc
```diff
@@ -142,6 +142,38 @@ TEST_F(DepfileParserTest, Escapes) {
...
+  EXPECT_EQ("foo1\\",parser_.outs_[0].AsString());
```
jhasse, line 172: space after ,

src/depfile_parser_test.cc
```diff
@@ -142,6 +142,38 @@ TEST_F(DepfileParserTest, Escapes) {
...
+  ASSERT_EQ(1u,parser_.ins_.size());
```
jhasse, line 173: space after ,

src/depfile_parser_test.cc
```diff
@@ -142,6 +142,38 @@ TEST_F(DepfileParserTest, Escapes) {
...
+  EXPECT_EQ("x",parser_.ins_[0].AsString());
```
jhasse, line 174: space after ,

Issue #1482, opened 2018-10-18
Environment variable for limiting the number of JOBS [Again]


kobalicek:
Is it possible to finally add the support for an environment variable that would limit the maximum number of jobs ninja can spawn?

I would suggest having an optional environment variable `NINJA_NUM_JOBS`, which would be used as an initial value when provided. This variable will be parsed before command line arguments that could of course override it.

I'm not asking for "additional flags" support [https://github.com/ninja-build/ninja/pull/1399], just a way to override the initial guess of the number of jobs as I understand the concerns behind `NINJA_FLAGS` and I understand that it might never get accepted.

I'm proposing a small change like this:

```c++
int GuessParallelism() {
  int n = GetProcessorCount();
  switch (n) {
    case 0: 
    case 1: n = 2; break;
    case 2: n = 3; break;
    default: n += 2; break;
  }

  const char* s = getenv("NINJA_NUM_JOBS"); // or any better name.
  if (!s)
    return n;

  char* end;
  int nUserDefined = strtol(s, &end, 10);

  // If the environment value is invalid or is greater than our initial guess, drop it.
  if (*end != 0 || nUserDefined <= 0 || nUserDefined > n)
    return n;

  // Otherwise use the sane value provided by the user.
  return nUserDefined;
}
```

jhasse:
Have you thought about writing a wrapper script which you put on your `PATH` named `ninja` which calls the real `ninja` with the amount of jobs you want?

kobalicek:
"Anyone can make the simple complicated."

BTW all of these were already discussed in other issues [I read them]. I really don't understand why there is such refusal regarding this feature as it's just a useful one. Can I workaround it? Yeah, but is it a solution? No.

jhasse:
> Yeah, but is it a solution? No.

Why not?

kobalicek:
You cannot see a difference between having a possibility to override a default value and between passing a default value to every invocation of ninja?

jhasse:
> You cannot see a difference between having a possibility to override a default value and between passing a default value to every invocation of ninja?

I can see the difference.

> I'm wondering why such refusal and whether it has some reason.

The reason for me would be that this can already solved by creating a wrapper script, so there's no need to implement this in Ninja itself.

kobalicek:
[...]

PR #1776, opened 2020-05-12
src/ninja.cc: add NINJAFLAGS environment variable support

phkrl:
Supported arguments: -j, -l
This allows to specify system-wide ninja performance constraints to prevent it from consuming too much RAM (and using swap) when building large projects.
A simple example is building a project like mesa on a device with 6 cores and only 4GB RAM especially using some extra build script (like Arch PKGBUILD)

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -1252,7 +1252,7 @@ int ExceptionFilter(unsigned int code, struct _EXCEPTION_POINTERS *ep) {
 
 #endif  // _MSC_VER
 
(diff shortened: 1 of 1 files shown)


jonesmz:
Related: https://github.com/ninja-build/ninja/issues/1482

phkrl:
> will `export NINJAFLAGS=-j10; ninja & ninja &` use 10 or 20 cores?

Each process will get its own copy of NINJAFLAGS and won't know about the other one, so it will be 20 cores

PR #1777, opened 2020-05-13
RFC: graph: batch stat() system calls using io_uring on Linux

MaxKellermann:
This adds an optional dependency on liburing
(https://github.com/axboe/liburing) to use Linux's new io_uring API
which can batch I/O system calls, reducing overhead.

The new source "uring.cc" implements class BulkStat as a wrapper for
liburing, plugged into DependencyScan::RecomputeDirty() which walks
the graph and batches most of the necessary stat() calls before
RecomputeDirty() needs them.

This requires Linux kernel 5.6 or newer.

This patch is not yet integrated well; for example, it bypasses the
DiskInterface, and to avoid breaking the unit tests, this patch adds a
IsReal() method to DiskInterface so we can disable the io_uring code
in the unit tests.

### DIFF
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -28,11 +28,65 @@
 #include "state.h"
 #include "util.h"
 
+#ifdef HAVE_LIBURING
+#include "uring.h"
+#endif
+
 bool Node::Stat(DiskInterface* disk_interface, string* err) {
   return (mtime_ = disk_interface->Stat(path_, err)) != -1;
 }
 
+#ifdef HAVE_LIBURING
+
+void Node::BulkStatCallback(TimeStamp t, const char*, void* data) {
+  reinterpret_cast<Node*>(data)->mtime_ = t;
+}
+
+inline bool Node::BulkStatIfNecessary(BulkStat &bulk_stat) {
+  if (stat_attempted_)
+    return false;
+
+  bulk_stat.Queue(path_.c_str(), this);
+  stat_attempted_ = true;
+  return true;
+}
+
+static void RecursiveBulkStatInner(BulkStat &bulk_stat, Node *node) {
+  if (!node->BulkStatIfNecessary(bulk_stat))
+    return;
+
+  Edge* edge = node->in_edge();
+  if (!edge)
+    return;
+
+  for (vector<Node*>::iterator i = edge->inputs_.begin();
+       i != edge->inputs_.end(); ++i)
+    RecursiveBulkStatInner(bulk_stat, *i);
+
+  for (vector<Node*>::iterator i = edge->outputs_.begin();
+       i != edge->outputs_.end(); ++i)
+    RecursiveBulkStatInner(bulk_stat, *i);
+}
+
+static void RecursiveBulkStat(BulkStat &bulk_stat, Node *node) {
+  if (bulk_stat.IsAvailable()) {
+    bulk_stat.SetCallback(Node::BulkStatCallback);
+    RecursiveBulkStatInner(bulk_stat, node);
+    bulk_stat.Wait();
+  }
+}
+
+#endif
+
 bool DependencyScan::RecomputeDirty(Node* node, string* err) {
+#ifdef HAVE_LIBURING
+  if (disk_interface_->IsReal()) {
+    RealDiskInterface &real_disk_interface = *(RealDiskInterface *)disk_interface_;
+    BulkStat &bulk_stat = real_disk_interface.GetBulkStat();
+    RecursiveBulkStat(bulk_stat, node);
+  }
+#endif
+
   vector<Node*> stack;
   return RecomputeDirty(node, &stack, err);
 }
--- a/src/uring.cc
+++ b/src/uring.cc
@@ -1,0 +1,145 @@
+// Copyright 2020 Max Kellermann. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "uring.h"
+
+#include <assert.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <stdlib.h>
+#include <string.h>
+
+BulkStat *global_bulk_stat;
+
+BulkStat::BulkStat() : callback_(NULL), n_pending(0) {
+  if (io_uring_queue_init(1024, &ring, 0) < 0) {
+    ring.ring_fd = -1;
+    return;
+  }
+
+  struct io_uring_probe* probe = io_uring_get_probe_ring(&ring);
+  if (!probe) {
+    Close();
+    return;
+  }
+
+  bool supported = io_uring_opcode_supported(probe, IORING_OP_STATX);
+  free(probe);
+  if (!supported) {
+    Close();
+    return;
+  }
+}
+
+void BulkStat::Queue(const char* path, void* data) {
+  assert(callback_);
+
+  if (!IsAvailable()) {
+    callback_(-1, "io_uring is not available", data);
+    return;
+  }
+
+  struct io_uring_sqe* s = io_uring_get_sqe(&ring);
+  if (!s) {
+    Wait();
+
+    s = io_uring_get_sqe(&ring);
+    if (!s) {
+      callback_(-1, "io_uring_get_sqe() failed", data);
+      return;
+    }
+  }
+
+  std::pair<void*, struct statx> value;
+  value.first = data;
+  PendingMap::iterator i = pending.insert(value).first;
+
+  io_uring_prep_statx(s, AT_FDCWD, path, AT_STATX_FORCE_SYNC,
+                      STATX_MTIME, &i->second);
+  io_uring_sqe_set_data(s, data);
+
+  ++n_pending;
+}
+
+void BulkStat::Wait() {
+  assert(callback_);
+
+  if (pending.empty())
+    return;
+
+  assert(IsAvailable());
+
+  int error = io_uring_submit_and_wait(&ring, n_pending);
+  if (error < 0) {
+    // something went terribly wrong, shouldn't happen
+    FailAll(strerror(-error));
+    Close();
+    return;
+  }
+
+  while (!pending.empty()) {
+    struct io_uring_cqe *cqe;
+    error = io_uring_wait_cqe(&ring, &cqe);
+    if (error < 0) {
+      if (error == -EAGAIN)
+        break;
+
+      // something went terribly wrong, shouldn't happen
+      FailAll(strerror(-error));
+      Close();
+      return;
+    }
+
+    void* data = io_uring_cqe_get_data(cqe);
+    assert(data);
+    PendingMap::const_iterator i = pending.find(data);
+    assert(i != pending.end());
+
+    if (cqe->res == 0) {
+      TimeStamp t = 0;
+      if (i->second.stx_mask & STATX_MTIME) {
+        t = i->second.stx_mtime.tv_sec;
+        if (t == 0)
+          // Some users (Flatpak) set mtime to 0, this should be
+          // harmless and avoids conflicting with our return
+          // value of 0 meaning that it doesn't exist.
+          t = 1;
+      }
+
+      callback_(t, NULL, data);
+    } else {
+      callback_(-1, strerror(-cqe->res), data);
+    }
+
+    pending.erase(i);
+    --n_pending;
+    io_uring_cqe_seen(&ring, cqe);
+  }
+}
+
+void BulkStat::FailAll(const char* error) {
+  assert(callback_);
+
+  for (PendingMap::const_iterator i = pending.begin(); i != pending.end(); ++i)
+    callback_(-1, error, i->first);
+
+  pending.clear();
+}
+
+void BulkStat::Close() {
+  if (ring.ring_fd >= 0) {
+    io_uring_queue_exit(&ring);
+    ring.ring_fd = -1;
+  }
+}
--- a/src/uring.h
+++ b/src/uring.h
@@ -1,0 +1,71 @@
+// Copyright 2020 Max Kellermann. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef NINJA_URING_H_
+#define NINJA_URING_H_
+
+#include "timestamp.h"
+
+#include <map>
+
+#include <liburing.h>
+#include <sys/stat.h>
+
+/// Helper class for submitting bulk statx() calls to the Linux kernel
+/// via io_uring
+class BulkStat {
+  typedef void (*Callback)(TimeStamp t, const char* error, void* data);
+
+public:
+  BulkStat();
+
+  ~BulkStat() {
+    Close();
+  }
+
+  /// Is io_uring/statx available?  This requires Linux 5.6 or newer.
+  bool IsAvailable() const {
+    return ring.ring_fd >= 0;
+  }
+
+  void SetCallback(Callback callback) {
+    callback_ = callback;
+  }
+
+  /// Queue a statx() call on the given path.  The callback will be
+  /// invoked later with the given "data" pointer.
+  void Queue(const char* path, void* data);
+
+  /// Wait for all queued operations to complete.  After returning,
+  /// all pending callbacks have been invoked.
+  void Wait();
+
+private:
+  void FailAll(const char* error);
+
+  void Close();
+
+  Callback callback_;
+
+  struct io_uring ring;
+
+  typedef std::map<void* , struct statx> PendingMap;
+  PendingMap pending;
+
+  PendingMap::size_type n_pending;
+};
+
+extern BulkStat *global_bulk_stat;
+
+#endif  // NINJA_URING_H_
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,6 +48,17 @@ else()
 endif()
 target_include_directories(libninja-re2c PRIVATE src)
 
+# --- optional liburing for io_uring support
+if(CMAKE_SYSTEM_NAME STREQUAL Linux)
+  find_package(PkgConfig)
+  if(PKG_CONFIG_FOUND)
+    pkg_check_modules(LIBURING liburing)
+    if(LIBURING_FOUND)
+      add_compile_definitions(HAVE_LIBURING)
+    endif()
+  endif()
+endif()
+
 # Core source files all build into ninja library.
 add_library(libninja OBJECT
 	src/build_log.cc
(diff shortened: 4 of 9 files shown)


jonesmz:
Just my two-cents, but I feel like adding async programming to the mix like this should be applied in a more uniform way. I'm personally interested in seeing Ninja adopt the async model of programming from e.g. https://think-async.com/Asio/AsioStandalone.html

However, to meaningfully take advantage of async programming style, the project needs to bite the bullet and accept C++11 fully.

MaxKellermann:
@jonesmz agree; but this PR attempts to introduce the advantages of io_uring in the least intrusive way.

src/uring.cc
```diff
@@ -0,0 +1,129 @@
+// Copyright 2011 Google Inc. All Rights Reserved.
```
jhasse, line 1: 2011?

src/uring.cc
```diff
@@ -0,0 +1,129 @@
...
+
+#include "uring.h"
+
+#include <assert.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <stdlib.h>
+#include <string.h>
+
+BulkStat::BulkStat(Callback callback):callback_(callback), n_pending(0) {
```
jhasse, line 23: Please add spaces around :

src/uring.cc
```diff
@@ -0,0 +1,129 @@
...
+    Wait();
+
+    s = io_uring_get_sqe(&ring);
+    if (!s) {
+      callback_(-1, "io_uring_get_sqe() failed", data);
+      return;
+    }
+  }
+
+  std::pair<void *, struct statx> value;
```
jhasse, line 60: Please align * with void (i.e. no space)

src/uring.cc
```diff
@@ -0,0 +1,129 @@
...
+        if (t == 0)
+          // Some users (Flatpak) set mtime to 0, this should be
+          // harmless and avoids conflicting with our return
+          // value of 0 meaning that it doesn't exist.
+          t = 1;
+      }
+
+      callback_(t, NULL, data);
+    } else
+      callback_(-1, strerror(-cqe->res), data);
```
jhasse, line 116: Please add {} here

src/uring.cc
```diff
@@ -0,0 +1,129 @@
...
+void BulkStat::Wait() {
+  if (pending.empty())
+    return;
+
+  assert(IsAvailable());
+
+  int error = io_uring_submit_and_wait(&ring, n_pending);
+  if (error < 0) {
+    // something went terribly wrong, shouldn't happen
+    FailAll(strerror(-error));
```
jhasse, line 80: Why `-`?

src/uring.h
```diff
@@ -0,0 +1,70 @@
+// Copyright 2011 Google Inc. All Rights Reserved.
```
jhasse, line 1: 2011?

src/uring.h
```diff
@@ -0,0 +1,70 @@
...
+  void Queue(const char *path, void *data);
+
+  /// Wait for all queued operations to complete.  After returning,
+  /// all pending callbacks have been invoked.
+  void Wait();
+
+private:
+  void FailAll(const char *error);
+
+  void Close() {
```
jhasse, line 53: Please move to cpp file.

src/uring.h
```diff
@@ -0,0 +1,70 @@
...
+
+#include <map>
+
+#include <liburing.h>
+#include <sys/stat.h>
+
+/// Helper class for submitting bulk statx() calls to the Linux kernel
+/// via io_uring
+class BulkStat {
+  typedef void (*Callback)(TimeStamp t, const char *error, void *data);
```
jhasse, line 28: Please move down after `private:`

src/graph.cc
```diff
@@ -28,11 +28,63 @@
...
+    RecursiveBulkStat(bulk_stat, node);
+    bulk_stat.Wait();
+  }
+}
+
+#endif
+
 bool DependencyScan::RecomputeDirty(Node* node, string* err) {
+#ifdef HAVE_LIBURING
+  if (disk_interface_->IsReal() && false)
```
jhasse, line 84: ` && false`?

src/graph.cc
```diff
@@ -28,11 +28,63 @@
...
+#endif
+
 bool Node::Stat(DiskInterface* disk_interface, string* err) {
   return (mtime_ = disk_interface->Stat(path_, err)) != -1;
 }
 
+#ifdef HAVE_LIBURING
+
+void Node::BulkStatCallback(TimeStamp t, const char *, void *data) {
+  Node *node = (Node *)data;
```
jhasse, line 42: Please use `reinterpret_cast`, maybe even like this:
```cpp
reinterpret_cast<Node&>(*data).mtime_ = t;
```

src/uring.cc
MaxKellermann: Because liburing returns negative errno values on error.

src/uring.h
MaxKellermann: Can't. It's used by the constructor, and the typedef must be defined before it's used.

src/graph.cc
MaxKellermann: Whoops. A leftover from a debugging session, which I accidently committed. Good catch.

src/graph.cc
MaxKellermann: Until now, I thought `reinterpret_cast` was new in C++11 so it can't be used in Ninja, but I looked it up and I was wrong. Some aspects of that operator are new in C++11, but the operator is older.

MaxKellermann:
Thanks @jhasse, I rebased and force-pushed with fixes for most of your review comments.

src/uring.h
jhasse: ah right, my bad.

src/uring.cc
jhasse: I see. Where can I find the documentation of liburing / `io_uring_submit_and_wait`? (I've only found some higher-level documentation about it)

MaxKellermann: Unfortunately, liburing is underdocumented, but this is some example code using that function: https://github.com/axboe/liburing/blob/fe500488190ff39716346ee1c0fe66bde300d0fb/examples/ucontext-cp.c#L245-L249

jonesmz:
There's also this document that gives a good overview of io_uring: https://unixism.net/loti/

MaxKellermann:
@jhasse are there any open questions/issues with my code? Do you need anything?

jhasse:
I need more time ;)

Do you have any performance numbers to back up the improvement?

I'm also probably not using Ninja for such large projects that I would notice this change. Please forgive me that I'm therefore not that invested in pushing this change forward.

MaxKellermann:
> Do you have any performance numbers to back up the improvement?

Comparing `ninja -n` on Kodi; a large project with 1383 build targets and 6025 `stat()` calls.
With this code, only 1790 `stat()` calls remain.

Comparing old and new ninja using `/usr/bin/time`:
```
0.19user 0.08system 0:01.11elapsed 25%CPU (0avgtext+0avgdata 15292maxresident)k
55632inputs+0outputs (8major+3336minor)pagefaults 0swaps

0.20user 0.03system 0:00.75elapsed 31%CPU (0avgtext+0avgdata 15312maxresident)k
11896inputs+0outputs (8major+3351minor)pagefaults 0swaps
```
Clearly, disk input is much lower, and wallclock time is 30% lower. Less time is spent in the kernel (which accounts for the reduced system call overhead).

Now with `perf stat` (for more accurate raw CPU usage):
```
302.344640      task-clock:u (msec)

230.167888      task-clock:u (msec)
```

Top system calls with `perf stat -e 'syscalls:sys_enter_*'`:
```
             6,014      syscalls:sys_enter_newstatu                                   
             2,796      syscalls:sys_enter_writeu                                   
             2,796      syscalls:sys_enter_ioctlu                                   
             1,226      syscalls:sys_enter_openatu                                   
                74      syscalls:sys_enter_readu                                    
                67      syscalls:sys_enter_brku                                     
                23      syscalls:sys_enter_mmapu                                    
                11      syscalls:sys_enter_newfstatu                                   
                 8      syscalls:sys_enter_closeu                                   
                 7      syscalls:sys_enter_mprotectu                                   

             2,768      syscalls:sys_enter_writeu                                   
             2,768      syscalls:sys_enter_ioctlu                                   
             1,778      syscalls:sys_enter_newstatu                                   
             1,227      syscalls:sys_enter_openatu                                   
                75      syscalls:sys_enter_readu                                    
                67      syscalls:sys_enter_brku                                     
                29      syscalls:sys_enter_mmapu                                    
                12      syscalls:sys_enter_newfstatu                                   
                 9      syscalls:sys_enter_closeu                                   
                 8      syscalls:sys_enter_mprotectu                                   
```
These results are obvious, but we could save even more system calls.

I have force-pushed a rebased version of this PR with another optimization (initializing io_uring only once, as field of `class RealDiskInterface`).

jhasse:
We definitely need to cover the io_uring code with the same unit tests - the `IsReal()` workaround is not something we should do (I bet there are many bugs waiting to happen due to the code duplication).

I think we need to get rid of the `VirtualFileSystem` and rather implement all the test functionality for the real file system (also need to make sure everything is inside `ScopedTempDir`). Filesystem access can still be verified / mocked, but with real files in the background. That looks feasible to me, haven't tried it yet though.

After that io_uring can be integrated and tested. If you have any other ideas or disagree, feel free to reopen this PR or open a new issue where we can discuss the io_uring plan in general.

MaxKellermann:
I agree that `IsReal()` is a kludge (and my very first post here mentioned that already), and I was hoping you'd give me some hints how to do it better. The current ninja code base is not well suited for integrating such a system call batcher, and needs some refactoring.
I'm not very comfortable with the Ninja code base because C++03 is mandatory, not even C++11 is allowed (a 9 year old standard). If Ninja would switch to modern C++, I'd be glad to help with refactoring, but working on C++03 code doesn't feel like a joyful investment of my free time.
I don't agree with closing this PR, but that's your decision and your way of managing this project.

jhasse:
> [...] and I was hoping you'd give me some hints how to do it better. The current ninja code base is not well suited for integrating such a system call batcher, and needs some refactoring.

I agree: Refactoring the code is the way to do this better. But in order to do this, more changes need to happen. One of them is switching to modern C++ as you said. Not going to happen any time soon as there's already too many TODOs stacked up.

>  I don't agree with closing this PR, but that's your decision and your way of managing this project.

There are so many open PRs and since this one is unmergable with the IsReal kludge, I think the RFC about io_uring is better suited as an issue. Hope you understand.

MaxKellermann:
> One of them is switching to modern C++ as you said.

The first step is trivial: change the documentation to allow C++17 (or at least C++11) instead of C++03. That would set the stage for more changes to change the actual code.

Would you accept a PR with this trivial change? Up to now, I couldn't imagine you would, because it sounded like it was an explicit decision not to upgrade to modern C++.

jhasse:
> That would set the stage for more changes to change the actual code.

And that's exactly what no one has the time to do / review.

So no, not at this time. But feel free to implement any changes like io_uring with C++17 as it will take a long time to review it anyway.

MaxKellermann:
> And that's exactly what no one has the time to do / review.

Right now, nobody takes the time to modernize the Ninja code base because doing so is not allowed according to the documentation. If modern C++ were allowed, you would certainly see code refactoring PRs from me. That about the "no one has the time to do".

"No one has the time to review" is the real problem here. A project where the maintainer has no time to write code is bad enough already - but if he doesn't even have time to review code written by others, then the project is seriously defunct.

(Please don't feel offended - I don't mean to criticize you here. I know how it feels to have not enough time for all the things one wants to do, really. I just try to figure what the situation you described means for the Ninja project.)

> So no, not at this time. But feel free to implement any changes like io_uring with C++17 as it will take a long time to review it anyway.

Nah. This requires lots of refactoring, before I can even begin with improving the io_uring changes. If there's little chance for the refactoring ever to be done (because nobody has time to review those refactoring patches), any time spent on io_uring would be wasted.

jhasse:
> Right now, nobody takes the time to modernize the Ninja code base because doing so is not allowed according to the documentation.

That isn't true, because it has already been done: https://github.com/splinter-build/splinter

> If modern C++ were allowed, you would certainly see code refactoring PRs from me. That about the "no one has the time to do".

I believe you, I would even do it myself - modernizing code is fun! What I meant with "no one has the time to do" was more "fixing stuff that breaks", "writing unit tests to ensure everything still works", "fixing merge conflicts with open PRs", etc. The not so fun part ;)

> "No one has the time to review" is the real problem here.

You're right, I was a bit vague there, sorry. Reviewing and fixing the fallout are the real problems, not actually doing it.

> A project where the maintainer has no time to write code is bad enough already - but if he doesn't even have time to review code written by others, then the project is seriously defunct.

I _do_ write code and I also have time to review code written by others. Not much though (it's my free time, too).

jonesmz:
> Right now, nobody takes the time to modernize the Ninja code base because doing so is not allowed according to the documentation. 

I bring the ninja codebase up to C++17, including a 95% finished port to using std::filesystem, here: https://github.com/splinter-build/splinter

The next patch to get merged is here: https://github.com/ninja-build/ninja/pull/1862

After that, my remaining C++98 compatible change, which is optional, would be to require re-generating re2c files every build: https://github.com/splinter-build/splinter/tree/2-ignore-autogenerated

After that for C++11, I have :

- range based for loops
- nullptr instead of NULL
- override / final keywords on virtual methods
- use of = default for constructors where appropriate
- use of emplace_back

Then for C++17:
- string_view
- more efficient string concatenation than just using operator+ (requires C++17 constexpr-if for my implementation, C++14 could work as well, I suppose)
- use of structured bindings where appropriate
- use of if-init where appropriate
- std::filesystem working for Linux, but not for other platforms
- a very not-working attempt at using std::error_code instead of out-parameters to indicate error.

Notably, I have not tried to do any major re-architecting in my patches until the std::filesystem change. So while I "modernize" in the sense of replacing code with C++17 equivalents, it's not modernized to the point it would be if the project were C++17 from the beginning.

> "No one has the time to review" is the real problem here.

I am available to review PRs, but am not a 'Collaborator', so have no ability to merge things. I can't imagine I'll ever be made a Collaborator given my stated disagreement with project policy on many points.

For example:
- I would close most of the existing PRs. E.g. https://github.com/ninja-build/ninja/pull/581 has had merge conflicts for years, and no activity from anyone but myself since 2018.
- I would change the minimum C++ language support level to C++17. I don't personally care about platforms that are 10 years out of date. People who still need older platforms to work would, if C++17 were made the minimum, be welcome to either fork the project to support their platform, use a competitor, like samurai, or use an older release.
- I would merge most of the patches that I have in my "splinter-build" fork into ninja, though with suitable codereview of course.

jhasse:
std::filesystem doesn't do async IO, not sure if it's worth it, if we need to change the code in the future *again* to use asio. That might not even be the future of IO in C++, as there are some design deficits last I heard. Did someone had a look at https://github.com/facebookexperimental/libunifex btw?

> I would close most of the existing PRs. E.g. #581 has had merge conflicts for years, and no activity from anyone but myself since 2018.

I have closed many PRs due to inactivity. Missed that one though, will close it right now.

jonesmz:
> std::filesystem doesn't do async IO, not sure if it's worth it, if we need to change the code in the future _again_ to use asio.

You're certainly right that std::filesystem doesn't support asynchronous file stat queries, nor does it provide a mechanism to batch multiple file stat queries. 

My objective behind using std::filesystem in Ninja / Splinter right now is simply that we're already not doing async / batched calls for file statting, so std::filesystem offers a platform-agnostic unified interface for that.

If an async / batched file statting mechanism is adopted, either std::filesystem becomes an invalid approach, or some adaptation between std::filesystem and the async / batched file statting tool needs to be put in place.

I personally have not looked into libunifex, I will try to evaluate it sometime in the next few weeks.

My understanding is that the sender / receiver model that is being considered for C++23 (tentatively?) is compatible with Boost.Asio plus or minus a few API changes, so I'm not too worried about the underlying primitives.

MaxKellermann:
> > Right now, nobody takes the time to modernize the Ninja code base because doing so is not allowed according to the documentation.
> 
> I bring the ninja codebase up to C++17

Is your project a full fork, i.e. have you given up on merging back, and will you do releases eventually?
Will you keep runtime compatibility with Ninja? Your fork sounds much more interesting to contribute to.

> including a 95% finished port to using std::filesystem, here: https://github.com/splinter-build/splinter

I havn't quite decided yet if I like std::filesystem or not, but I lean towards not liking it. By its core design, it doesn't and cannot and will never support things that are important to me, like `O_PATH` descriptors. It carries a lot of runtime bloat, for example its `path` class always allocates memory - there's no way to use a string that's already owned by somebody else.
It's an API that tries to implement the lowest common denominator of all filesystem APIs, at the cost of ignoring all system-specific optimizations.

> The next patch to get merged is here: #1862 

Oh-oh, that this PR got closed (and the reason for closing it) is a big red flag for me, for never wanting to contribute to Ninja again. What a shame, that contributions get thrown away for this lame reason!

> I would be interested in seeing io_uring used in Ninja and/or Splinter, and would seriously consider adopting io_uring in my fork as one of the patches, but I do want to offer caution that I'm not personally sure that io_uring is the right API for a project like Ninja to use. Personally I'm more interested in adopting https://think-async.com/Asio/AsioStandalone.html as the underlying asynchronous abstraction, and possibly something akin to boost-process for child-process management.

I've given up on Asio. Recently, I've migrated one of my projects away from it. It's heavily over-engineered, and some very basic things are broken, for example, you cannot safely cancel a timer. I got annoying crashes that were nearly impossible to fix. And none of my projects needs the multi-threaded aspect, and removing multi-threading from the equations allows many optimizations. No, I never want to use Asio again.

jonesmz:
> Is your project a full fork, i.e. have you given up on merging back, and will you do releases eventually?
> Will you keep runtime compatibility with Ninja? Your fork sounds much more interesting to contribute to.

At the moment, it is not a full fork. My stated goal is to eventually get all of my patches merged into Ninja.

My workflow is to periodically (every couple of months), rebase each patch in that project onto Ninja master.

I have no objection to doing a full fork in the future, it's just not my goal at the moment. If I did a full fork, I would intend to maintain backwards compatibility with Ninja where reasonable.

> I havn't quite decided yet if I like std::filesystem or not, but I lean towards not liking it. By its core design, it doesn't and cannot and will never support things that are important to me, like O_PATH descriptors. It carries a lot of runtime bloat, for example its path class always allocates memory - there's no way to use a string that's already owned by somebody else.
> It's an API that tries to implement the lowest common denominator of all filesystem APIs, at the cost of ignoring all system-specific optimizations.

Yea, it has lots of problems. No disagreement there. Note that the std::filesystem implementation in that patch isn't even fully working, and the way that I've structured the project allows me to drop that patch without any consequences to the other improvements I've been collecting.

> I've given up on Asio. Recently, I've migrated one of my projects away from it. It's heavily over-engineered, and some very basic things are broken, for example, you cannot safely cancel a timer. I got annoying crashes that were nearly impossible to fix. And none of my projects needs the multi-threaded aspect, and removing multi-threading from the equations allows many optimizations. No, I never want to use Asio again.

My hope is that the C++ standards committee can reach a consensus for what the async primitives in C++ will look like officially for C++23, and then I can find a library that implements something that looks like what C++23 will have. If that's not ASIO, I have no problem. I just want something that's close to what the standardized thing will look like, and I've so far only got experience with ASIO, and it's roughly what the Networking TS is, so that's what I've been using.

My intention was not to say "Anything that's not based on Boost.Asio is bad!" it was to say "I'm not sure that directly consuming the io_uring API is the right approach. I would hope to instead use some higher level API that abstracts away some of the gritty details and also allows for sexier handling of child processes at the same time, similar to (but not exactly like, since it has it's own horrible issues) how Boost.Process abstracts these platform specific details."

jonesmz:
> It carries a lot of runtime bloat, for example its path class always allocates memory - there's no way to use a string that's already owned by somebody else.

Note also that currently Ninja doesn't do a great job of handling strings. StringPiece is certainly better than std::string everywhere, but there are multiple places where the data that StringPiece points to is mutated by some function taking a StringPiece, so there's some issues with the current implementation of that.

jonesmz:
> If modern C++ were allowed, you would certainly see code refactoring PRs from me. That about the "no one has the time to do".

@MaxKellermann Ninja uses C++11 now, if you didn't know. https://github.com/ninja-build/ninja/pull/2089

MaxKellermann:
That's good, but doesn't change anything with the rejection of this PR.

PR #1780, opened 2020-05-18
Delay actually opening log files until the first write, fix #1724

### LINKED ISSUE #1724: Log files held open when re-generating build manifest
Ninja holds its log files open in RebuildManifest.  On Windows, this locks the files such that the generator cannot use any `ninja -t *` tools to update them after generation (e.g. cleandead, recompact, or restat).  This leads to problems like [CMake Issue 20274](https://gitlab.kitware.com/cmake/cmake/issues/20274).

Ideally Ninja should close the log files while running the generator (any rule with `generator = 1`) and re-open them again afterward.

jhasse:
Calling DepsLog/BuildLog::OpenForWrite will now only save the file path. The actual opening of the file (moved to OpenForWriteIfNeeded) happens right before the first write attempt.

This is needed so that the files aren't held open when the generator runs (i.e. RebuildManifest) as it may call tools like recompact which won't be able to open the file on Windows (see #1724).

The disadvantage is that now the error reporting happens at a later time and will be reported as a failed write, not a failed open.

### DIFF
--- a/src/deps_log.cc
+++ b/src/deps_log.cc
@@ -49,34 +49,9 @@ bool DepsLog::OpenForWrite(const string& path, string* err) {
       return false;
   }
 
-  file_ = fopen(path.c_str(), "ab");
-  if (!file_) {
-    *err = strerror(errno);
-    return false;
-  }
-  // Set the buffer size to this and flush the file buffer after every record
-  // to make sure records aren't written partially.
-  setvbuf(file_, NULL, _IOFBF, kMaxRecordSize + 1);
-  SetCloseOnExec(fileno(file_));
-
-  // Opening a file in append mode doesn't set the file pointer to the file's
-  // end on Windows. Do that explicitly.
-  fseek(file_, 0, SEEK_END);
-
-  if (ftell(file_) == 0) {
-    if (fwrite(kFileSignature, sizeof(kFileSignature) - 1, 1, file_) < 1) {
-      *err = strerror(errno);
-      return false;
-    }
-    if (fwrite(&kCurrentVersion, 4, 1, file_) < 1) {
-      *err = strerror(errno);
-      return false;
-    }
-  }
-  if (fflush(file_) != 0) {
-    *err = strerror(errno);
-    return false;
-  }
+  assert(!file_);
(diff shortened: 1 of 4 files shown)


jhasse:
@bradking My first idea on how to fix this. Haven't tested with CMake 3.17 on Windows yet.

bradking:
Cc: @mathstuf  

@jhasse thanks.  I tested this locally by commenting out CMake's workaround ([CMake commit ccaa0bccc4](https://gitlab.kitware.com/cmake/cmake/commit/ccaa0bccc42a31f7515c6f85a0601ac0ad006e96)).  Without this fix I can reproduce the problem.  With this fix the problem goes away!

jhasse:
Nice!

Do you have an idea on how to (unit) test this change?

mathstuf:
> Do you have an idea on how to (unit) test this change?

The simple solution requires going through the actual disk, but then the test is actually platform-specific. The disk interface could learn to "lock" open files (on all platforms so that everyone can actually test this) and then you could poke it in a synth generator step.

src/deps_log.cc
```diff
@@ -396,6 +375,9 @@ bool DepsLog::RecordId(Node* node) {
     errno = ERANGE;
     return false;
   }
+
+  OpenForWriteIfNeeded();
```
jhasse, line 379: should check return value

src/deps_log.cc
```diff
@@ -132,6 +107,9 @@ bool DepsLog::RecordDeps(Node* node, TimeStamp mtime,
...
```
jhasse, line 111: should check return value

jhasse:
> The simple solution requires going through the actual disk, but then the test is actually platform-specific.

I think that would be okay, as we run all tests on all platforms.

> The disk interface could learn to "lock" open files (on all platforms so that everyone can actually test this) and then you could poke it in a synth generator step.

That won't work as this code isn't using the disk interface but actual fopen.

jhasse:
I've tried to come up with a good test, but failed so far. I think CMake's testing is fine for now and should catch if we regress here.

PR #1783, opened 2020-06-03
clang-tidy fixes

neheb:
ran through several clang-tidy fixes.

I tried to keep C++98 compatibility. Locally, it builds fine with -DCMAKE_CXX_STANDARD=98

edit: also, every commit was ran with git clang-format.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -77,12 +77,9 @@ bool DryRunCommandRunner::WaitForCommand(Result* result) {
 }  // namespace
 
 BuildStatus::BuildStatus(const BuildConfig& config)
-    : config_(config),
-      start_time_millis_(GetTimeMillis()),
-      started_edges_(0), finished_edges_(0), total_edges_(0),
-      progress_status_format_(NULL),
-      overall_rate_(), current_rate_(config.parallelism) {
-
+    : config_(config), start_time_millis_(GetTimeMillis()), started_edges_(0),
+      finished_edges_(0), total_edges_(0), progress_status_format_(NULL),
+      current_rate_(config.parallelism) {
   // Don't do anything fancy in verbose mode.
   if (config_.verbosity != BuildConfig::NORMAL)
     printer_.set_smart_terminal(false);
--- a/src/build.cc
+++ b/src/build.cc
@@ -14,10 +14,10 @@
 
 #include "build.h"
 
-#include <assert.h>
-#include <errno.h>
-#include <stdio.h>
-#include <stdlib.h>
+#include <cassert>
+#include <cerrno>
+#include <cstdio>
+#include <cstdlib>
 #include <functional>
 
 #ifdef _WIN32
--- a/src/build.cc
+++ b/src/build.cc
@@ -1064,8 +1064,7 @@ bool Builder::ExtractDeps(CommandRunner::Result* result,
       // complexity in IncludesNormalize::Relativize.
       deps_nodes->push_back(state_->GetNode(*i, ~0u));
     }
-  } else
-  if (deps_type == "gcc") {
+  } else if (deps_type == "gcc") {
     string depfile = result->edge->GetUnescapedDepfile();
     if (depfile.empty()) {
       *err = string("edge with deps=gcc but no depfile makes no sense");
--- a/src/build.cc
+++ b/src/build.cc
@@ -95,7 +95,7 @@ void BuildStatus::PlanHasTotalEdges(int total) {
 
 void BuildStatus::BuildEdgeStarted(const Edge* edge) {
   assert(running_edges_.find(edge) == running_edges_.end());
-  int start_time = (int)(GetTimeMillis() - start_time_millis_);
+  int start_time = static_cast<int>(GetTimeMillis() - start_time_millis_);
   running_edges_.insert(make_pair(edge, start_time));
   ++started_edges_;
 
@@ -117,7 +117,7 @@ void BuildStatus::BuildEdgeFinished(Edge* edge,
 
   RunningEdgeMap::iterator i = running_edges_.find(edge);
   *start_time = i->second;
-  *end_time = (int)(now - start_time_millis_);
+  *end_time = static_cast<int>(now - start_time_millis_);
   running_edges_.erase(i);
 
   if (edge->use_console())
@@ -653,13 +653,13 @@ void Plan::UnmarkDependents(const Node* node, set<Node*>* dependents) {
 }
 
 void Plan::Dump() const {
-  printf("pending: %d\n", (int)want_.size());
+  printf("pending: %d\n", static_cast<int>(want_.size()));
   for (map<Edge*, Want>::const_iterator e = want_.begin(); e != want_.end(); ++e) {
     if (e->second != kWantNothing)
       printf("want ");
     e->first->Dump();
   }
-  printf("ready: %d\n", (int)ready_.size());
+  printf("ready: %d\n", static_cast<int>(ready_.size()));
 }
 
 struct RealCommandRunner : public CommandRunner {
@@ -691,9 +691,9 @@ void RealCommandRunner::Abort() {
 bool RealCommandRunner::CanRunMore() const {
   size_t subproc_number =
       subprocs_.running_.size() + subprocs_.finished_.size();
-  return (int)subproc_number < config_.parallelism
-    && ((subprocs_.running_.empty() || config_.max_load_average <= 0.0f)
-        || GetLoadAverage() < config_.max_load_average);
+  return static_cast<int>(subproc_number) < config_.parallelism &&
+         ((subprocs_.running_.empty() || config_.max_load_average <= 0.0f) ||
+          GetLoadAverage() < config_.max_load_average);
 }
 
 bool RealCommandRunner::StartCommand(Edge* edge) {
--- a/src/build.cc
+++ b/src/build.cc
@@ -1033,7 +1033,7 @@ bool Builder::FinishCommand(CommandRunner::Result* result, string* err) {
   }
 
   if (!deps_type.empty() && !config_.dry_run) {
-    assert(edge->outputs_.size() >= 1 && "should have been rejected by parser");
+    assert(!edge->outputs_.empty() && "should have been rejected by parser");
     for (std::vector<Node*>::const_iterator o = edge->outputs_.begin();
          o != edge->outputs_.end(); ++o) {
       TimeStamp deps_mtime = disk_interface_->Stat((*o)->path(), err);
--- a/src/build.h
+++ b/src/build.h
@@ -190,7 +190,7 @@ struct Builder {
 
   /// Add a target to the build, scanning dependencies.
   /// @return false on error.
-  bool AddTarget(Node* target, string* err);
+  bool AddTarget(Node* node, string* err);
 
   /// Returns true if the build targets are already up to date.
   bool AlreadyUpToDate() const;
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -65,7 +65,7 @@ uint64_t MurmurHash64A(const void* key, size_t len) {
   const uint64_t m = BIG_CONSTANT(0xc6a4a7935bd1e995);
   const int r = 47;
   uint64_t h = seed ^ (len * m);
-  const unsigned char* data = (const unsigned char*)key;
+  const unsigned char* data = static_cast<const unsigned char*>(key);
   while (len >= 8) {
     uint64_t k;
     memcpy(&k, data, sizeof k);
@@ -214,7 +214,8 @@ struct LineReader {
       line_start_ = line_end_ + 1;
     }
 
-    line_end_ = (char*)memchr(line_start_, '\n', buf_end_ - line_start_);
+    line_end_ =
+        static_cast<char*>(memchr(line_start_, '\n', buf_end_ - line_start_));
     if (!line_end_) {
       // No newline. Move rest of data to start of buffer, fill rest.
       size_t already_consumed = line_start_ - buf_;
@@ -224,7 +225,8 @@ struct LineReader {
       size_t read = fread(buf_ + size_rest, 1, sizeof(buf_) - size_rest, file_);
       buf_end_ = buf_ + size_rest + read;
       line_start_ = buf_;
-      line_end_ = (char*)memchr(line_start_, '\n', buf_end_ - line_start_);
+      line_end_ =
+          static_cast<char*>(memchr(line_start_, '\n', buf_end_ - line_start_));
     }
 
     *line_start = line_start_;
@@ -281,7 +283,8 @@ LoadStatus BuildLog::Load(const string& path, string* err) {
     const char kFieldSeparator = '\t';
 
     char* start = line_start;
-    char* end = (char*)memchr(start, kFieldSeparator, line_end - start);
+    char* end =
+        static_cast<char*>(memchr(start, kFieldSeparator, line_end - start));
     if (!end)
       continue;
     *end = 0;
@@ -292,21 +295,21 @@ LoadStatus BuildLog::Load(const string& path, string* err) {
     start_time = atoi(start);
     start = end + 1;
 
-    end = (char*)memchr(start, kFieldSeparator, line_end - start);
+    end = static_cast<char*>(memchr(start, kFieldSeparator, line_end - start));
     if (!end)
       continue;
     *end = 0;
     end_time = atoi(start);
     start = end + 1;
 
-    end = (char*)memchr(start, kFieldSeparator, line_end - start);
+    end = static_cast<char*>(memchr(start, kFieldSeparator, line_end - start));
     if (!end)
       continue;
     *end = 0;
     restat_mtime = strtoll(start, NULL, 10);
     start = end + 1;
 
-    end = (char*)memchr(start, kFieldSeparator, line_end - start);
+    end = static_cast<char*>(memchr(start, kFieldSeparator, line_end - start));
     if (!end)
       continue;
     string output = string(start, end - start);
@@ -330,7 +333,7 @@ LoadStatus BuildLog::Load(const string& path, string* err) {
     entry->mtime = restat_mtime;
     if (log_version >= 5) {
       char c = *end; *end = '\0';
-      entry->command_hash = (uint64_t)strtoull(start, NULL, 16);
+      entry->command_hash = static_cast<uint64_t>(strtoull(start, NULL, 16));
       *end = c;
     } else {
       entry->command_hash = LogEntry::HashCommand(StringPiece(start,
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -21,11 +21,12 @@
 #endif
 
 #include "build_log.h"
-#include "disk_interface.h"
 
-#include <errno.h>
-#include <stdlib.h>
-#include <string.h>
+#include <cerrno>
+#include <cstdlib>
+#include <cstring>
+
+#include "disk_interface.h"
 
 #ifndef _WIN32
 #include <inttypes.h>
--- a/src/build_log_perftest.cc
+++ b/src/build_log_perftest.cc
@@ -123,7 +123,7 @@ int main() {
       fprintf(stderr, "Failed to read test data: %s\n", err.c_str());
       return 1;
     }
-    int delta = (int)(GetTimeMillis() - start);
+    int delta = static_cast<int>(GetTimeMillis() - start);
     printf("%dms\n", delta);
     times.push_back(delta);
   }
--- a/src/build_log_perftest.cc
+++ b/src/build_log_perftest.cc
@@ -12,15 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <stdio.h>
-#include <stdlib.h>
+#include <cstdio>
+#include <cstdlib>
 
 #include "build_log.h"
 #include "graph.h"
 #include "manifest_parser.h"
+#include "metrics.h"
 #include "state.h"
 #include "util.h"
-#include "metrics.h"
 
 #ifndef _WIN32
 #include <unistd.h>
--- a/src/clparser.cc
+++ b/src/clparser.cc
@@ -15,8 +15,8 @@
 #include "clparser.h"
 
 #include <algorithm>
-#include <assert.h>
-#include <string.h>
+#include <cassert>
+#include <cstring>
 
 #include "metrics.h"
 #include "string_piece_util.h"
--- a/src/clparser.cc
+++ b/src/clparser.cc
@@ -45,8 +45,8 @@ string CLParser::FilterShowIncludes(const string& line,
   const char* in = line.c_str();
   const char* end = in + line.size();
   const string& prefix = deps_prefix.empty() ? kDepsPrefixEnglish : deps_prefix;
-  if (end - in > (int)prefix.size() &&
-      memcmp(in, prefix.c_str(), (int)prefix.size()) == 0) {
+  if (end - in > static_cast<int>(prefix.size()) &&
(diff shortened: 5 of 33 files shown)


src/build.cc
```diff
@@ -656,13 +653,13 @@ void Plan::UnmarkDependents(const Node* node, set<Node*>* dependents) {
 }
 
 void Plan::Dump() const {
-  printf("pending: %d\n", (int)want_.size());
+  printf("pending: %d\n", static_cast<int>(want_.size()));
```
jonesmz, line 656: Change the print string for size_t instead of casting to int. You're both casting from unsigned to signed, as well as reducing the resolution.

src/build.cc
```diff
@@ -694,9 +691,9 @@ void RealCommandRunner::Abort() {
 bool RealCommandRunner::CanRunMore() const {
   size_t subproc_number =
       subprocs_.running_.size() + subprocs_.finished_.size();
-  return (int)subproc_number < config_.parallelism
-    && ((subprocs_.running_.empty() || config_.max_load_average <= 0.0f)
-        || GetLoadAverage() < config_.max_load_average);
+  return static_cast<int>(subproc_number) < config_.parallelism &&
+         ((subprocs_.running_.empty() || config_.max_load_average <= 0.0f) ||
+          GetLoadAverage() < config_.max_load_average);
```
jonesmz, line 696: the boolean operators would be easier to read if they were on the left hand side.

src/build.cc
```diff
@@ -694,9 +691,9 @@ void RealCommandRunner::Abort() {
 bool RealCommandRunner::CanRunMore() const {
   size_t subproc_number =
       subprocs_.running_.size() + subprocs_.finished_.size();
-  return (int)subproc_number < config_.parallelism
-    && ((subprocs_.running_.empty() || config_.max_load_average <= 0.0f)
-        || GetLoadAverage() < config_.max_load_average);
+  return static_cast<int>(subproc_number) < config_.parallelism &&
```
jonesmz, line 694: Better to convert config_.parallelism to size_t than to cast subproc_number to int.

src/build.h
```diff
@@ -190,7 +190,7 @@ struct Builder {
 
   /// Add a target to the build, scanning dependencies.
   /// @return false on error.
-  bool AddTarget(Node* target, string* err);
+  bool AddTarget(Node* node, string* err);
```
jonesmz, line 193: Why?

src/build_log.cc
```diff
@@ -65,7 +66,7 @@ uint64_t MurmurHash64A(const void* key, size_t len) {
   const uint64_t m = BIG_CONSTANT(0xc6a4a7935bd1e995);
   const int r = 47;
   uint64_t h = seed ^ (len * m);
-  const unsigned char* data = (const unsigned char*)key;
+  const unsigned char* data = static_cast<const unsigned char*>(key);
```
jonesmz, line 69: Don't static cast a pointer. Reinterpret cast it.

src/build_log.cc
```diff
@@ -214,7 +215,8 @@ struct LineReader {
       line_start_ = line_end_ + 1;
     }
 
-    line_end_ = (char*)memchr(line_start_, '\n', buf_end_ - line_start_);
+    line_end_ =
+        static_cast<char*>(memchr(line_start_, '\n', buf_end_ - line_start_));
```
jonesmz, line 219: reinterpret_cast

src/build_log.cc
```diff
@@ -224,7 +226,8 @@ struct LineReader {
       size_t read = fread(buf_ + size_rest, 1, sizeof(buf_) - size_rest, file_);
       buf_end_ = buf_ + size_rest + read;
       line_start_ = buf_;
-      line_end_ = (char*)memchr(line_start_, '\n', buf_end_ - line_start_);
+      line_end_ =
+          static_cast<char*>(memchr(line_start_, '\n', buf_end_ - line_start_));
```
jonesmz, line 230: reinterpret_cast

src/build_log.cc
```diff
@@ -330,7 +334,7 @@ LoadStatus BuildLog::Load(const string& path, string* err) {
     entry->mtime = restat_mtime;
     if (log_version >= 5) {
       char c = *end; *end = '\0';
-      entry->command_hash = (uint64_t)strtoull(start, NULL, 16);
+      entry->command_hash = static_cast<uint64_t>(strtoull(start, NULL, 16));
```
jonesmz, line 337: why is this cast needed? it's already returning unsigned long long int.

src/build_log_perftest.cc
```diff
@@ -123,7 +123,7 @@ int main() {
       fprintf(stderr, "Failed to read test data: %s\n", err.c_str());
       return 1;
     }
-    int delta = (int)(GetTimeMillis() - start);
+    int delta = static_cast<int>(GetTimeMillis() - start);
```
jonesmz, line 126: This does not appear to be the right thing to do here.

src/clparser.cc
```diff
@@ -45,8 +45,8 @@ string CLParser::FilterShowIncludes(const string& line,
   const char* in = line.c_str();
   const char* end = in + line.size();
   const string& prefix = deps_prefix.empty() ? kDepsPrefixEnglish : deps_prefix;
-  if (end - in > (int)prefix.size() &&
-      memcmp(in, prefix.c_str(), (int)prefix.size()) == 0) {
+  if (end - in > static_cast<int>(prefix.size()) &&
```
jonesmz, line 48: This is not the right thing to do.

decltype(pointer - pointer) == std::ptrdiff_t.

Put parenthesise around (end - in) to avoid operator precedence issues.

src/clparser.cc
```diff
@@ -45,8 +45,8 @@ string CLParser::FilterShowIncludes(const string& line,
...
+      memcmp(in, prefix.c_str(), static_cast<int>(prefix.size())) == 0) {
```
jonesmz, line 49: The third parameter for memcmp is size_t... this should have never been a cast to int, wtf?

src/clparser_perftest.cc
```diff
@@ -146,7 +146,7 @@ int main(int argc, char* argv[]) {
     int64_t end = GetTimeMillis();
 
     if (end - start > 2000) {
-      int delta_ms = (int)(end - start);
+      int delta_ms = static_cast<int>(end - start);
```
jonesmz, line 149: int64_t - int64_t is not int. Please ensure the result is stored as the proper accuracy.

src/depfile_parser_perftest.cc
```diff
@@ -48,8 +48,8 @@ int main(int argc, char* argv[]) {
       int64_t end = GetTimeMillis();
 
       if (end - start > 100) {
-        int delta = (int)(end - start);
-        float time = delta*1000 / (float)limit;
+        int delta = static_cast<int>(end - start);
+        float time = delta * 1000 / static_cast<float>(limit);
```
jonesmz, line 52: I believe (but may be mistaken) that it's also necessary to cast delta, and 1000, to float to ensure the compiler doesn't do stupid things.

src/deps_log.cc
```diff
@@ -83,7 +83,7 @@ bool DepsLog::OpenForWrite(const string& path, string* err) {
 bool DepsLog::RecordDeps(Node* node, TimeStamp mtime,
                          const vector<Node*>& nodes) {
   return RecordDeps(node, mtime, nodes.size(),
-                    nodes.empty() ? NULL : (Node**)&nodes.front());
+                    nodes.empty() ? NULL : const_cast<Node**>(&nodes.front()));
```
jonesmz, line 86: Woah.

Const casting is a huge code smell. This is not the right thing to do.

src/deps_log.cc
```diff
@@ -225,8 +225,9 @@ LoadStatus DepsLog::Load(const string& path, State* state, string* err) {
       int* deps_data = reinterpret_cast<int*>(buf);
       int out_id = deps_data[0];
       TimeStamp mtime;
-      mtime = (TimeStamp)(((uint64_t)(unsigned int)deps_data[2] << 32) |
-                          (uint64_t)(unsigned int)deps_data[1]);
+      mtime = static_cast<TimeStamp>(
+          static_cast<uint64_t>(static_cast<unsigned int>(deps_data[2])) << 32 |
+          static_cast<uint64_t>(static_cast<unsigned int>(deps_data[1])));
```
jonesmz, line 230: This whole series of logic is awful.

Can it be cleaned up, or at the very least, documented to explain what the intention is?

src/deps_log.cc
```diff
@@ -333,7 +334,7 @@ bool DepsLog::Recompact(const string& path, string* err) {
     (*i)->set_id(-1);
 
   // Write out all deps again.
-  for (int old_id = 0; old_id < (int)deps_.size(); ++old_id) {
+  for (int old_id = 0; old_id < static_cast<int>(deps_.size()); ++old_id) {
```
jonesmz, line 337: Don't cast to int. Change the type of old_id to size_t.

src/deps_log.cc
```diff
@@ -399,13 +400,13 @@ bool DepsLog::RecordId(Node* node) {
   if (fwrite(&size, 4, 1, file_) < 1)
     return false;
   if (fwrite(node->path().data(), path_size, 1, file_) < 1) {
-    assert(node->path().size() > 0);
+    assert(!node->path().empty());
     return false;
   }
   if (padding && fwrite("\0\0", padding, 1, file_) < 1)
     return false;
   int id = nodes_.size();
```
jonesmz, line 408: This should be a size_t to begin with.

src/deps_log_test.cc
```diff
@@ -71,7 +71,7 @@ TEST_F(DepsLogTest, WriteRead) {
   ASSERT_EQ("", err);
 
   ASSERT_EQ(log1.nodes().size(), log2.nodes().size());
-  for (int i = 0; i < (int)log1.nodes().size(); ++i) {
+  for (int i = 0; i < static_cast<int>(log1.nodes().size()); ++i) {
```
jonesmz, line 74: Don't cast. Change to size_t instead.

src/deps_log_test.cc
```diff
@@ -139,7 +139,7 @@ TEST_F(DepsLogTest, DoubleEntry) {
 
     struct stat st;
     ASSERT_EQ(0, stat(kTestFilename, &st));
-    file_size = (int)st.st_size;
+    file_size = static_cast<int>(st.st_size);
```
jonesmz, line 142: Don't cast. Change file_size to size_t instead.

src/graph.cc
```diff
@@ -637,7 +638,7 @@ bool ImplicitDepLoader::LoadDepsFromLog(Edge* edge, string* err) {
 vector<Node*>::iterator ImplicitDepLoader::PreallocateSpace(Edge* edge,
                                                             int count) {
   edge->inputs_.insert(edge->inputs_.end() - edge->order_only_deps_,
-                       (size_t)count, 0);
+                       static_cast<size_t>(count), 0);
```
jonesmz, line 641: don't case. Change the parameter type instead.

src/manifest_parser.cc
```diff
@@ -327,16 +324,16 @@ bool ManifestParser::ParseEdge(string* err) {
...
       }
+      if (!quiet_) {
+        Warning(
+            "multiple rules generate %s. "
+            "builds involving this target will not be correct; "
+            "continuing anyway [-w dupbuild=warn]",
+            path.c_str());
+      }
+      if (e - i <= static_cast<size_t>(implicit_outs))
+        --implicit_outs;
```
jonesmz, line 336: Don't cast. Change the type of implicit_outs instead.

src/manifest_parser_perftest.cc
```diff
@@ -106,7 +105,7 @@ int main(int argc, char* argv[]) {
   for (int i = 0; i < kNumRepetitions; ++i) {
     int64_t start = GetTimeMillis();
     int optimization_guard = LoadManifests(measure_command_evaluation);
-    int delta = (int)(GetTimeMillis() - start);
+    int delta = static_cast<int>(GetTimeMillis() - start);
```
jonesmz, line 108: Don't cast, change type of delta instead.

src/metrics.cc
```diff
@@ -38,7 +38,7 @@ int64_t HighResTimer() {
   timeval tv;
   if (gettimeofday(&tv, NULL) < 0)
     Fatal("gettimeofday: %s", strerror(errno));
-  return (int64_t)tv.tv_sec * 1000*1000 + tv.tv_usec;
+  return static_cast<int64_t>(tv.tv_sec) * 1000 * 1000 + tv.tv_usec;
```
jonesmz, line 41: The type of 1000 should be explicitly set with ULL to avoid weird compiler bullshit.

src/metrics.cc
```diff
@@ -102,16 +102,16 @@ void Metrics::Report() {
   int width = 0;
   for (vector<Metric*>::iterator i = metrics_.begin();
        i != metrics_.end(); ++i) {
-    width = max((int)(*i)->name.size(), width);
+    width = max(static_cast<int>((*i)->name.size()), width);
```
jonesmz, line 105: Don't cast, change the type of width.

src/ninja.cc
```diff
@@ -297,21 +296,19 @@ Node* NinjaMain::CollectTarget(const char* cpath, string* err) {
       node = edge->outputs_[0];
     }
     return node;
+  }
+  *err = "unknown target '" + Node::PathDecanonicalized(path, slash_bits) + "'";
```
jonesmz, line 300: This is very inefficient string concatenation.

src/ninja.cc
```diff
@@ -371,7 +368,7 @@ int NinjaMain::ToolQuery(const Options* options, int argc, char* argv[]) {
         }
       }
       printf("  input: %s\n", edge->rule_->name().c_str());
-      for (int in = 0; in < (int)edge->inputs_.size(); in++) {
+      for (int in = 0; in < static_cast<int>(edge->inputs_.size()); in++) {
```
jonesmz, line 371: Don't cast. Change type.

src/ninja.cc
```diff
@@ -1059,34 +1062,43 @@ bool WarningEnable(const string& name, Options* options) {
 "  phonycycle={err,warn}  phony build statement references itself\n"
     );
     return false;
-  } else if (name == "dupbuild=err") {
+  }
+
+  if (name == "dupbuild=err") {
```
jonesmz, line 1067: I don't see replacing if-else's with stand-alone if's as an improvement.

jonesmz: If you really wanted to do that, consider instead making an enum, for each string contents allowed, and then make a function that does
```
enum myenum string_to_my_enum(std::string const& str)
{
    if(str == "string_1") return enum_1;
    if(str == "string_2") return enum_2;
    if(str == "string_3") return enum_3;
    if(str == "string_4") return enum_4;
    return enum_unknown;
}
```

and then use a switch-case in this function.

src/ninja.cc
```diff
@@ -1169,10 +1181,10 @@ void NinjaMain::DumpMetrics() {
   g_metrics->Report();
 
   printf("\n");
-  int count = (int)state_.paths_.size();
-  int buckets = (int)state_.paths_.bucket_count();
+  int count = static_cast<int>(state_.paths_.size());
+  int buckets = static_cast<int>(state_.paths_.bucket_count());
```
jonesmz, line 1185: Unnecessary type casting.

src/util.cc
```diff
@@ -152,7 +153,9 @@ bool CanonicalizePath(char* path, size_t* len, uint64_t* slash_bits,
         // '.' component; eliminate.
         src += 2;
         continue;
-      } else if (src[1] == '.' && (src + 2 == end || IsPathSeparator(src[2]))) {
+      }
+
+      if (src[1] == '.' && (src + 2 == end || IsPathSeparator(src[2]))) {
```
jonesmz, line 158: Parenthesis for explicit operator precedence.

jonesmz:
See https://github.com/splinter-build/splinter for a set of patches that overlap with these. Though the intention and direction of that fork is very different than the PR, some of the same parts of the code are changed in the same way as this PR.

neheb:
All of my changes are clang-tidy cleanups, most automated. I can cleanup the code in separate commits if desired.

As far as that fork is concerned, is that going to be merged here? This uses C++98 AFAIK (no idea why).

neheb:
I like https://github.com/splinter-build/splinter/commit/89d25314b54b9427686adbc2f232d8f908038cf5

clang-tidy warns against this as well.

jonesmz:
> I like [splinter-build@89d2531](https://github.com/splinter-build/splinter/commit/89d25314b54b9427686adbc2f232d8f908038cf5)
> 
> clang-tidy warns against this as well.

Part one of the commit you referenced is submitted as a PR to ninja here: https://github.com/ninja-build/ninja/pull/1614

src/deps_log.cc
neheb: No idea what this code does.

neheb:
That's unfortunate. This PR despite being smaller, might suffer the same fate.

jonesmz:
It's almost certainly what will happen, yes. Sorry.

There are three maintainers for the Ninja project. Of them, only one has been active for many years but is not able to keep up with the influx of pull requests.

jonesmz:
> As far as that fork is concerned, is that going to be merged here? This uses C++98 AFAIK (no idea why).

I missed this question before. Sorry.

It's my intention that every one of the commits that are currently in the Splinter-Build repository will be merged into the Ninja repository. Assuming that Ninja's maintainers are willing to merge them.

I have made no architectural changes. The Splinter project is intended to be a place for C++ language feature enhancements to the Ninja codebase.

There is some recent discussion in one of the Splinter issues about working on some architectural / feature divergence. But if that's done, it would be handled such that the language-feature enhancement commits are kept separate from the feature changes to make upstreaming easier.

jonesmz:
> This uses C++98 AFAIK (no idea why).

See here for discussion about that: https://github.com/ninja-build/ninja/issues/1403

neheb:
That's also unfortunate.

On the bright side, CentOS 6 is EOL on November 30, 2020. CentOS 7 supports C++11 just fine.

https://github.com/ninja-build/ninja/issues/1403#issuecomment-378789918

is hilarious. Some of those cast changes that you pointed to are actual bugs and can prevent the compiler from properly optimizing the code. Especially in for loops.

src/deps_log.cc
jonesmz: Sure. That's fair. Semantically you're not changing anything, so I guess it's no worse than before.

jhasse:
Looks good, but it will result in lots of conflicts with all the other PRs. Maybe we can tackle this one warning at a time?

Also have a look at #1768 where I introduce a .clang-tidy config and let CI check it.

neheb:
So split up each commit as a separate PR?

jhasse:
Yes. And please expect that some of those PRs will be opened for quite a long time. There's no one paid to work on Ninja (Google even still uses v1.8.2 for Chromium and doesn't care that much about master) and changes like these do result in merge conflicts, ergo more work. There are lot of PRs for features people have been waiting YEARS to get merged. [The milestones](https://github.com/ninja-build/ninja/milestones) can give you an idea of which things are next on the priority list.

src/clparser.cc
jhasse: > Put parenthesise around (end - in) to avoid operator precedence issues.

That would just be a readability improvement (debatable), as the operator precedence is well defined.

src/build.h
neheb: clang-tidy warns against that with readability-named-parameter

jhasse: Not every clang-tidy check is a good idea. I think readability-named-parameter should make an exception for this case (parameter name is lower-case version of type). The Google C++ Style Guide section this check is based also doesn't mention this explicitly.

jonesmz: I'm very opposed to omitting the parameter name. I think that will result in substantially more confusion.

src/build_log.cc
jonesmz: Should still be reinterpret_cast for consistancy.

src/clparser.cc
jonesmz: I've seen enough situations nowhere people have screwed it up that I'd say better safe than sorry.

src/depfile_parser_perftest.cc
jonesmz: I'm not doubting. The change I suggested should be made regardless of whether the compiler is doing the correct thing. The only doubt is whether the compiler is already doing the correct thing.

src/build_log.cc
jonesmz: I'm not understanding.

key is "const void *".

So reinterpret_cast<const unsigned char *>(const void *) should be fine, shouldn't it?

neheb: Ah never mind. Thought it was for a different change.

neheb:
Placed the small commits in separate PRs.

src/build_log.cc
jhasse: I would rather avoid reinterpret_cast whenever possible. When someone changes the type of key, they should get a compile error.

src/clparser.cc
jhasse: how can you screw up mixing `+` and `>`?

src/clparser_perftest.cc
jhasse: Because there are lots of other more important things to improve in Ninja.

jhasse:
Thanks, I'll close this one than.

neheb:
Well, the big ones are still here. I guess I'll keep them local.

jhasse:
Ups, sorry!

neheb:
:). I’ll send at a later time.

src/build_log.cc
jonesmz: How about we change the type right now then? If the parameter is changed to char* it can't be messed up.

src/clparser_perftest.cc
jhasse: I would merge the static_cast change without changing the type to int64_t, all I'm saying.

src/clparser.cc
jhasse: I had a boss who messed with parenthesis around `-`. What now?

src/build_log.cc
jhasse: Either that or the static_cast.

src/clparser.cc
jonesmz: Then we're at an impasse, and you're going to use your authority as the maintainer of the project to make a decision, and I'll put parenthesis around things in my fork.

src/build_log.cc
jonesmz: Sure. Either way.

I agree that static_cast has the property of ensuring a compiler error if the type being casted from can't be casted with static_cast. That's a beneficial outcome.

Personally I prefer the parameter to this function change type, but I'm not that concerned about it, so whatever way works.

src/build.cc
jonesmz: Ok, Automated formatting is desirable. That's fine with me.

src/clparser_perftest.cc
jonesmz: Ok. static_cast<int> is better than (int), that I agree with.

I would still prefer that if we're going to change anything at all, we do it "right" immediately though.

src/deps_log.cc
jonesmz: Could be.

Makes me wonder if it should be a call to ntoh or hton instead?

src/depfile_parser_perftest.cc
jhasse: > The only doubt is whether the compiler is already doing the correct thing.

That's what I meant. Changing the C style casts to static_cast is a complete orthogonal issue.

PR #1784, opened 2020-06-03
[clang-tidy] check empty instead of size

neheb:
Found with readability-container-size-empty


jhasse:
Oh, there are a lot of instances of this warning in our test code: https://github.com/ninja-build/ninja/runs/738280592
We want to use `ASSERT_EQ("", err);` though, as this will display errs content when it isn't empty.

PR #1785, opened 2020-06-03
[clang-tidy] remove pointless c_str()

neheb:
Found with readability-redundant-string-cstr

### DIFF
--- a/src/disk_interface_test.cc
+++ b/src/disk_interface_test.cc
@@ -190,7 +190,7 @@ TEST_F(DiskInterfaceTest, ReadFile) {
 
 TEST_F(DiskInterfaceTest, MakeDirs) {
   string path = "path/with/double//slash/";
-  EXPECT_TRUE(disk_.MakeDirs(path.c_str()));
+  EXPECT_TRUE(disk_.MakeDirs(path));
   FILE* f = fopen((path + "a_file").c_str(), "w");
   EXPECT_TRUE(f);
   EXPECT_EQ(0, fclose(f));


neheb:
No idea if legit...

jonesmz:
It is.

disk_ is of type "RealDiskInterface".

Which is declared as.

```cpp
/// Interface for accessing the disk.
///
/// Abstract so it can be mocked out for tests.  The real implementation
/// is RealDiskInterface.
struct DiskInterface: public FileReader {
  /// stat() a file, returning the mtime, or 0 if missing and -1 on
  /// other errors.
  virtual TimeStamp Stat(const string& path, string* err) const = 0;

  /// Create a directory, returning false on failure.
  virtual bool MakeDir(const string& path) = 0;

  /// Create a file, with the specified name and contents
  /// Returns true on success, false on failure
  virtual bool WriteFile(const string& path, const string& contents) = 0;

  /// Remove the file named @a path. It behaves like 'rm -f path' so no errors
  /// are reported if it does not exists.
  /// @returns 0 if the file has been removed,
  ///          1 if the file does not exist, and
  ///          -1 if an error occurs.
  virtual int RemoveFile(const string& path) = 0;

  /// Create all the parent directories for path; like mkdir -p
  /// `basename path`.
  bool MakeDirs(const string& path);
};

/// Implementation of DiskInterface that actually hits the disk.
struct RealDiskInterface : public DiskInterface {
  RealDiskInterface()
#ifdef _WIN32
                      : use_cache_(false)
#endif
                      {}
  virtual ~RealDiskInterface() {}
  virtual TimeStamp Stat(const string& path, string* err) const;
  virtual bool MakeDir(const string& path);
  virtual bool WriteFile(const string& path, const string& contents);
  virtual Status ReadFile(const string& path, string* contents, string* err);
  virtual int RemoveFile(const string& path);

  /// Whether stat information can be cached.  Only has an effect on Windows.
  void AllowStatCache(bool allow);

 private:
#ifdef _WIN32
  /// Whether stat information can be cached.
  bool use_cache_;

  typedef map<string, TimeStamp> DirCache;
  // TODO: Neither a map nor a hashmap seems ideal here.  If the statcache
  // works out, come up with a better data structure.
  typedef map<string, DirCache> Cache;
  mutable Cache cache_;
#endif
};
```

So what you're doing here is avoiding the unnecessary creation of a temporary std::string, when calling `bool MakeDirs(const string& path);`

Which is the right thing to do.

neheb:
I meant the test failure.

jonesmz:
```

  File "./misc/output_test.py", line 83, in test_issue_1214

AssertionError: '[0/1] echo a\x1b[K' != '[1/1] echo a\x1b[K\n\x1b[31mred\x1b[0m\n'

- [0/1] echo a

+ [1/1] echo a

red
```

It appears that this may be caused by an embedded null.

What may be happening here is that you've inadvertently exposed a bug in the underlying code cause by shitty string management.

jonesmz:
Can do force-push the commit to re-trigger the test? We can see if it happens twice.

jonesmz:
I don't see how this specific change to the code could possibly have that outcome.

So it may be that the failure is non-deterministic and has nothing to do with your change.

neheb:
Looks like it.

PR #1788, opened 2020-06-03
[clang-tidy] do not use else after return

neheb:
Found with readability-else-after-return


jhasse:
While it's nice, it touches a lot of lines. And there are people who argue that it's actually *less* readable. Therefore I think we won't add this check yet.

PR #1789, opened 2020-06-03
[clang-tidy] fix small false positive

neheb:
else if has to be on the same line it seems.

Found with readability-misleading-indentation


jonesmz:
This should actually work just fine, but I'm definitely in support of the readability enhancement that this change makes.

neheb:
Yeah it works fine. This just fixes a clang warning.

neheb:
Rebased.

PR #1793, opened 2020-06-03
[clang-tidy] change string to reference

neheb:
Found with performance-unnecessary-copy-initialization

### DIFF
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -469,7 +469,7 @@ bool Edge::maybe_phonycycle_diagnostic() const {
 
 // static
 string Node::PathDecanonicalized(const string& path, uint64_t slash_bits) {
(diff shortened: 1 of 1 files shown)


neheb:
Seems broken for windows.

jonesmz:
This change would also run afoul of https://github.com/ninja-build/ninja/pull/1671

jhasse:
```cpp
#ifdef _WIN32
  std::string result = path;
  uint64_t mask = 1;
  for (char* c = &result[0]; (c = strchr(c, '/')) != NULL;) {
    if (slash_bits & mask)
      *c = '\\';
    c++;
    mask <<= 1;
  }
  return result;
#else
  return path;
#endif
```

PR #1797, opened 2020-06-08
Rule supports logfile attribute

xxthunder:


xxthunder:
Any feedback highly appreciated. 😀

nico:
As discussed on the "New feature: Log per target output to disk?" thread on ninja-dev: You can do this by adding a pipe to your command. Since it's easy-ish to do this outside of ninja and there weren't any other requests for this over ninja's lifetime, it seems better to not make ninja bigger for this.

jhasse:
For the record this is the thread: https://groups.google.com/d/msg/ninja-build/RPP7Uy2juvo/G-6yd7szBQAJ

I would see the need for this if it was something ninja could do automatically (e.g. a dry run which can be used to see the warnings for a given target again). But if it's necessary to change the build manifest, I would use the pipe solution instead.

Let's continue the discussion in the mailing list thread.

puetzk:
Just for the sake of a vote: this is definitely a feature I would use if it was present and easy to turn, but the big win (and reason to build it in) would be if there was a way (e.g. `--verbose=noincremental`) to have ninja read these log files back in, and print the console output again from rules that were in the graph, but up-to-date.

Having warnings not be masked if their command is skipped by an incremental build isn't a must-have, but it would be a nice quality-of-life feature from time to time. It's easy enough for a wrapper script to *create* the files, but harder to get that output streamed back into the IDE integration from targets that didn't run.

PR #1800, opened 2020-06-12
Add validation nodes to ninja

colincross:
A common problem in the Android build is inserting rules that perform
some sort of error checking that doesn't produce any artifacts needed
by the build, for example static analysis tools.  There are a few
patterns currently used, both of which have downsides.

The first is to have a rule that depends on all of the static analysis
results.  This ensures they run, but requires running static analysis
over everything, and not just the active parts of the build graph.

The second is to insert the static analysis rule into the build graph
between the artifact producing rule and anything that depends on it,
often copying the artifact as the output of the static analysis rule.
This increases the critical path of the build, often reducing
parallelism.  In the case of copying the artifact, it also wastes
disk space.

This patch adds "validation nodes" to edges in Ninja.  A build
statement can specify validation nodes using "|@" in the edge
inputs.  The validation nodes are not used as an input to the edge
(the edge can run before the validation node is ready), but are
added to the initial nodes of the build graph whenever the edge
is part of the build graph.  The edge that outputs the validation
node can depend on the output of the edge that is being validated
if desired.


colincross:
The Linux check failure looks unrelated to my change, and it's hard to tell if the Windows one is related.

Linux: 
```error: p7zip-16.02-10.el7.x86_64.rpm: not an rpm package (or package manifest): ```

Windows: 
```1/1 Test #1: NinjaTest ........................***Failed    1.03 sec```

**Comment by jhasse** [comment]
https://mirrorservice.org has some difficulties, hopefully it will be back up soon.

As for Windows: Please check again after rebasing on master, you should now see the output of ninja_test.exe.

**Comment by colincross** [comment]
I think the Windows failures were due to non-deterministic build order, apparently the Windows allocation order is less deterministic than Linux.  This probably would be fixed by https://github.com/ninja-build/ninja/pull/1210/commits/3ca0e1b6c9f6106e983c0806eb92ad96dc8c3b37, but I just removed the assertions that depend on build order.

**Comment by colincross** [comment]
Any comments on at least the approach or syntax?  I'd like to merge this to the Android ninja fork, and it would be great to agree on syntax and semantics if it gets merged upstream too.

**Comment by jhasse** [comment]
Looks good to me. Was `@` chosen arbitrary though?

The documentation could use some of the wording of the commit message.

File: src/graph.cc [hunk 1]
```diff
@@ -33,12 +34,44 @@ bool Node::Stat(DiskInterface* disk_interface, string* err) {
...
+  std::deque<Node*> nodes;
+  nodes.push_back(initial_node);
+
+  // RecomputeNodeDirty might return new validation nodes that need to be
+  // checked for dirty state, keep a queue of nodes to visit.
+  while (!nodes.empty()) {
+    Node *node = nodes.front();
+    nodes.pop_front();
+
+    stack.clear();
```
jhasse, line 57: I would move the declarations here. (Or do you think the performance difference will be noticeable?)

src/graph.cc
```diff
@@ -33,12 +34,44 @@ bool Node::Stat(DiskInterface* disk_interface, string* err) {
...
 }
 
-bool DependencyScan::RecomputeDirty(Node* node, vector<Node*>* stack,
+bool DependencyScan::RecomputeDirty(Node* initial_node,
+                                    std::vector<Node*>* validation_nodes,
                                     string* err) {
+  std::vector<Node*> stack;
+  std::vector<Node*> new_validation_nodes;
+
+  std::deque<Node*> nodes;
```
jhasse, line 48: `std::deque<Node*> nodes(1, initial_node);`

src/graph.cc
```diff
@@ -33,12 +34,44 @@ bool Node::Stat(DiskInterface* disk_interface, string* err) {
...
+  nodes.push_back(initial_node);
+
+  // RecomputeNodeDirty might return new validation nodes that need to be
+  // checked for dirty state, keep a queue of nodes to visit.
+  while (!nodes.empty()) {
+    Node *node = nodes.front();
```
jhasse, line 54: Please align the * with the type.

src/build.cc
```diff
@@ -784,16 +784,23 @@ Node* Builder::AddTarget(const string& name, string* err) {
...
-  if (!scan_.RecomputeDirty(node, err))
+  std::vector<Node*> validation_nodes;
+  if (!scan_.RecomputeDirty(node, &validation_nodes, err))
     return false;
 
   if (Edge* in_edge = node->in_edge()) {
-    if (in_edge->outputs_ready())
-      return true;  // Nothing to do.
+    if (!in_edge->outputs_ready())
+      if (!plan_.AddTarget(node, err))
```
jhasse, line 793: merge with previous if using `&&`

src/build.cc
```diff
@@ -784,16 +784,23 @@ Node* Builder::AddTarget(const string& name, string* err) {
...
+        return false;
+  }
+  for (std::vector<Node*>::iterator n = validation_nodes.begin();
+       n != validation_nodes.end(); ++n) {
+    if (Edge* in_edge = (*n)->in_edge()) {
+      if (!in_edge->outputs_ready())
+        if (!plan_.AddTarget(*n, err))
```
jhasse, line 800: merge with previous ifs using `&&`

src/graph.h
```diff
@@ -274,13 +280,24 @@ struct DependencyScan {
         dep_loader_(state, deps_log, disk_interface, depfile_parser_options),
         dyndep_loader_(state, disk_interface) {}
 
-  /// Update the |dirty_| state of the given node by inspecting its input edge.
+  /// A wrapper for tests that don't care about validation nodes.
```
jhasse, line 283: might be better to adjusts the tests than to create a method just for tests. Not sure though.

colincross:
`@` was chosen pretty arbitrarily, I wanted to make sure it wasn't going to conflict with any other syntax if it wasn't going upstream.  `|||` would fit more with the current style, but these aren't really inputs so a syntax that differentiates them from the other inputs seemed worthwhile.

src/graph.cc
colincross: In the Android tree we might have thousands of active validation nodes if we use them for Android Lint, Clang tidy and various other checks.  That would mean thousands of times through the loop, and each time through the loop might need to grow the stack a few times.  Reusing the existing stack seems worthwhile.

src/graph.h
colincross: I think you're right, the existence of this wrapper caused me to miss the DynDeps call to RecomputeDirty, which needs to handle newly discovered validation nodes somehow.  I'll update all the tests to pass a NULL validation_nodes, and make it error if it finds validation nodes in that case.

jhasse:
@mathstuf @bradking I think this might be useful for CMake's clang-tidy integration. What do you think?

bradking:
> might be useful for CMake's clang-tidy integration

Currently CMake's Ninja generator runs such lints as part of the same build statement that compiles the corresponding source.  No artifact copying is needed, but the validation is not executed concurrently (though maybe CMake could do that internally).  It's basically treated like part of the compile rule.  The validation nodes proposed here may offer an alternative, though more work would be needed for CTest's dashboard client mode to be able to capture validation failures, which are currently reported like compiler warnings.

> The edge that outputs the validation node can depend on the output of the edge that is being validated.

I might go the other way: the edge that outputs the node being validated could have a validation dependency on the validation output.  That way running `ninja foo.c.o` would both compile and run validation, and consumers of the object files do not need to be aware of the validation step.  That's purely a choice of the generator though.

doc/manual.asciidoc
```diff
@@ -969,6 +971,31 @@ express the implicit dependency.)
...
+dependency of one of the targets specified on the command line or a
+default target.
+
+Validations are added to the build graph regardless of whether the output
+files of the build statement are dirty are not, and the dirty state of
+the build statement that outputs the file being used as a validation
+has no effect on the dirty state of the build statement that requested it.
+
+A build edge can list another build edge as a validation even if the second
+edge depends on the first.
```
bradking, line 989: Is this "circular" case covered in the test suite?

Also please add a test case with a cycle among the validation dependencies themselves.

src/build.cc
```diff
@@ -783,16 +796,28 @@ Node* Builder::AddTarget(const string& name, string* err) {
...
-    if (in_edge->outputs_ready())
-      return true;  // Nothing to do.
+    if (!in_edge->outputs_ready() &&
+        !plan_.AddTarget(target, err)) {
+      return false;
+    }
   }
 
-  if (!plan_.AddTarget(target, err))
-    return false;
```
bradking, line 795: The inversion of the `outputs_ready` condition and return value look like refactoring in preparation for adding the validation nodes below.  Please split that refactoring into a preceding commit (still in this PR) with its own explanation.

src/build_test.cc
```diff
@@ -3301,3 +3335,178 @@ TEST_F(BuildTest, DyndepTwoLevelDiscoveredDirty) {
...
+  err.clear();
+  command_runner_.commands_ran_.clear();
+  state_.Reset();
+  EXPECT_TRUE(builder_.AddTarget("out", &err));
+  ASSERT_EQ("", err);
+
+  EXPECT_TRUE(builder_.Build(&err));
+  EXPECT_EQ("", err);
+
+  EXPECT_EQ(1u, command_runner_.commands_ran_.size());
```
bradking, line 3370: This verifies that only one command was executed, but not that it was the one we expected.  Please add additional checks.  Similarly in other cases below.

bradking (changes requested):

doc/manual.asciidoc
colincross: Yes, in BuildTest.ValidationDependsOnOutput.

I added BuildTest.ValidationCircular and BuildTest.ValidationWithCircularDependency, do those cover the cycles you were thinking of?

src/build.cc
colincross: Done.

src/build_test.cc
colincross: Done.

doc/manual.asciidoc
bradking: Yes, thanks.

colincross:
> > The edge that outputs the validation node can depend on the output of the edge that is being validated.
> 
> I might go the other way: the edge that outputs the node being validated could have a validation dependency on the validation output. That way running `ninja foo.c.o` would both compile and run validation, and consumers of the object files do not need to be aware of the validation step. That's purely a choice of the generator though.

I mentioned this only to highlight that unlike a regular dependency, a validation dependency does not form a cycle.  I added "if desired" to the end of this sentence to clarify that this isn't the only way to structure the dependencies.

src/build_test.cc
```diff
@@ -3083,6 +3083,40 @@ TEST_F(BuildTest, DyndepBuildDiscoverNewInput) {
   EXPECT_EQ("touch out", command_runner_.commands_ran_[2]);
 }
 
+TEST_F(BuildTest, DyndepBuildDiscoverNewInputWithValidation) {
+  // Verify that a dyndep file can be built and loaded to discover
+  // a new input to an edge.
```
bradking, line 3088: Did/do you intend to allow dyndep files to specify validation nodes?  I don't think they need to.

If not, please add a test case that verifies the dyndep parser rejects the syntax.

bradking (changes requested):

src/build_test.cc
colincross: Done.

bradking (commented):
The test cases I requested in previous reviews all LGTM.

src/build.cc
```diff
@@ -557,12 +557,12 @@ bool Builder::AddTarget(Node* target, string* err) {
     return false;
 
   if (Edge* in_edge = target->in_edge()) {
-    if (in_edge->outputs_ready())
-      return true;  // Nothing to do.
+    if (!in_edge->outputs_ready() &&
+        !plan_.AddTarget(target, err)) {
```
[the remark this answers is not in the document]
colincross: Yes, it causes a slight change in behavior if you attempt to build a leaf node that doesn't exist, printing `no work to do` instead of the previous `ninja: error: 'a' missing and no known rule to make it`.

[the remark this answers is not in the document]
colincross: Done.

IcedQuinn:
shouldn't these be mentioned in the manual also?

colincross:
> shouldn't these be mentioned in the manual also? i just checked the manual from the homepage and the changes aren't live.

They are mentioned in https://ninja-build.org/manual.html#ref_ninja_file and https://ninja-build.org/manual.html#validations, but not mentioned in https://ninja-build.org/manual.html#ref_dependencies and probably should be.

Creative-Commons-cntl (commented):
changes

PR #1805, opened 2020-06-24
Output exit code in FAILED message

### LINKED ISSUE #1507: Ninja hides the exit status of failing subprocesses
In build.cc, Ninja checks if the subprocess succeeded or failed

```
  // Print the command that is spewing before printing its output.
  if (!success) {
    string outputs;

    printer_.PrintOnNewLine("FAILED: " + outputs + "\n");
    printer_.PrintOnNewLine(edge->EvaluateCommand() + "\n");
  }
```

But by this point in the build, Ninja has already discarded the actual exit value from the subprocess.

There are a variety of tools that don't print anything to the console when they fail to run, relying on the user to instead check the exit status to see what the reason for failure was.

This is especially problematic for third party tools where the source code is not available.

It's, of course, possible to get the exit status printed to the console using a wrapper script, but as discussed at extreme length, wrapper scripts are desired by basically no one.

Ninja should print the exit status of failing subprocesses.

schultetwin1:
This fixes #1507 

Outputs the exit code after "FAILED" for any process that did not
complete successfully.

Output looks like "FAILED (\<exitcode\>): \<outputs\>"

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -68,7 +68,7 @@ bool DryRunCommandRunner::WaitForCommand(Result* result) {
    if (finished_.empty())
      return false;
 
-   result->status = ExitSuccess;
+   result->status.result = ExitSuccess;
    result->edge = finished_.front();
    finished_.pop();
    return true;
@@ -110,8 +110,7 @@ void BuildStatus::BuildEdgeStarted(const Edge* edge) {
 }
 
 void BuildStatus::BuildEdgeFinished(Edge* edge,
-                                    bool success,
-                                    const string& output,
+                                    const CommandRunner::Result* result,
                                     int* start_time,
                                     int* end_time) {
   int64_t now = GetTimeMillis();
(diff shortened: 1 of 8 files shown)


schultetwin1:
I'm not complete happy with the interface of this design (putting the exit code in parens). I'd love to hear other ideas if people have them.

Jonatino:
Any idea when or if this will be merged? It would help a lot when using ninja on CI to have proper sucess/errors. Right now I depend on checking the output to see if the build is there to determine if a ninja subprocess failed or not.

lazka:
I was just needing this to debug a failing command that didn't have any output and had to rebase this (some code has moved, but the basics still apply). Would indeed be nice to have this by default.

FrancoisBeaune:
Adding a vote to merge this PR, or another one that addresses the problem. It would have been very useful to us for debugging random command failures.

z2oh:
Just used a custom built ninja with this patch to debug a gnarly compiler crash. Adding my vote to have this feature merged!

mcprat:
it's a great idea, but I don't think this is the best way to write it

Felixoid:
Maybe the approach from #2540 is more straight forward?

Felixoid:
the #2540 is merged, there's no reason to keep it opened.

schultetwin1:
Looks great, thank you!

PR #1813, opened 2020-07-16
cmake: use PROJECT_{SOURCE,BINARY}_DIR instead of CMAKE_{SOURCE,BINAR…

ledocc:
…Y}_DIR

CMAKE_SOURCE_DIR refer to the full path to the top level of the current
CMake source tree
PROJECT_SOURCE_DIR refer to the source directory of the last call to the
project() command made in the current directory scope or one of its parents

when ninja is use as a subproject, the build fail because of this.


jhasse:
Thanks!

Issue #1571, opened 2019-05-12
Prevent ninja starting new job when physical memory become low


adelplanque:
I have a big project where
* 80% of jobs use a little amount of memory ==> jobs must be limited by CPU ressource
* 20% of jobs use large amount of memory ==> this jobs must be limited by memory ressouce

So like -l flags, it should be nice to have a -m flag :

    ninja -m 1G

==> don't start new job if physical available memory is below 1G

metti:
See #660 and #1354 for previous attempts.

kunaltyagi:
Any chance of the PR getting merged? Or is the addition of another cmdline flag a blocking change? It'd be great to not have `ninja || ninja -j1` as the build step 😄

jhasse:
> Any chance of the PR getting merged?

Not any time soon, sorry.

> Or is the addition of another cmdline flag a blocking change?

I would prefer something like `--keep-free-memory` instead of `-m`.

cmorty:
@jhasse : Any chance to move this forward with `--keep-free-memory`? I have some ugly template stuff that needs GB, while the rest needs far less than 100MB. Of course it would be perfect if a requirement could be bound to the job/jobpool, but not spawning new jobs below a certain limit should remove most of the pain.

jhasse:
@cmorty I'm not working on this.

Ideally I think this should be fixed by giving the kernel more information about what's happening: If the kernel knew that the processes that ninja spawns don't have to run in parallel, it could swap enough of them out if memory is low.

chen3feng:
Slow down the speed of job starting is also a considerable help.

greg7mdp:
This would be great to have! What's the holdup?

rustyx:
This is a must-have when doing LTO with clang on a modern CPU.
Running 32 `c++` processes in parallel takes <4 GB RAM, but 32 `ld` processes at the end take 100 GB RAM, causing my system to lock up :(
And reducing parallelism to `-j8` takes forever to compile.
At least need a separate argument to be able to specify compiler and linker parallelism separately (combinations can be computed relative to each other).

Marc-Pierre-Barbier:
If someone on macos can test https://github.com/ninja-build/ninja/pull/2605 it would be great

PR #1815, opened 2020-07-17
WIP: Allow running ninjas to share a job limit

davezarzycki:
Hello. I've been toying around with using POSIX semaphores to enable a process jobs limit across ninja invocations. This is only the first take and there is plenty of room for improvement (and robustness). That being said, it does work. Is this something that might be a candidate for upstreaming? High level feedback and style feedback is appreciated. The design of this patch will probably change considerably in order to recover from ninja crashing.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -544,6 +544,9 @@ else:
 if platform.is_aix() and not platform.is_os400_pase():
     libs.append('-lperfstat')
 
+if not platform.is_msvc():
+    libs.append('-lpthread')
+
 all_targets = []
 
 n.comment('Main executable is library plus main() function.')
--- a/src/build.cc
+++ b/src/build.cc
@@ -40,7 +40,57 @@
 #include "subprocess.h"
 #include "util.h"
 
+#ifndef _WIN32
+unsigned shared_ninja_jobs = 0;
+unsigned shared_ninja_local_jobs = 0;
+sem_t* shared_ninja_jobs_semaphore = nullptr;
+sem_t* shared_ninjas_semaphore = nullptr;
+char shared_ninja_jobs_semaphore_name[sizeof(SHARED_NINJA_JOBS_SEMAPHORE_PREFIX
+                                             "18446744073709551615")] =
+    "";  // UINT64_MAX
+char shared_ninjas_semaphore_name[sizeof(
+    SHARED_NINJAS_SEMAPHORE_PREFIX "18446744073709551615")] = "";  // UINT64_MAX
+#endif
+
 namespace {
+
+bool TryGlobalSpawnEnter() {
+#ifndef _WIN32
+  if (shared_ninja_jobs_semaphore == nullptr)
+    return true;
+
+  int current_ninjas;
+  if (sem_getvalue(shared_ninjas_semaphore, &current_ninjas) != 0)
+    Fatal("sem_getvalue: %s", strerror(errno));
+
+  // There is a technically race between getting the number of ninjas and
+  // waiting, so as a hack add one more phantom ninja just in case.
+  bool wait = shared_ninja_local_jobs <
+              (shared_ninja_jobs / unsigned(current_ninjas + 1));
+
+  if (wait) {
+    if (sem_wait(shared_ninja_jobs_semaphore) != 0)
+      Fatal("sem_wait: %s", strerror(errno));
+  } else if (sem_trywait(shared_ninja_jobs_semaphore) != 0) {
+    if (errno == EAGAIN)
+      return false;
+    Fatal("sem_trywait: %s", strerror(errno));
+  }
+  ++shared_ninja_local_jobs;
+#endif
+
+  return true;
+}
+
+void GlobalSpawnLeave() {
+#ifndef _WIN32
+  if (shared_ninja_jobs_semaphore == nullptr)
+    return;
+  --shared_ninja_local_jobs;
+  if (sem_post(shared_ninja_jobs_semaphore) != 0)
+    Fatal("sem_post: %s", strerror(errno));
+#endif
+}
 
 /// A CommandRunner that doesn't actually run the commands.
 struct DryRunCommandRunner : public CommandRunner {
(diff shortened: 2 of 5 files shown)


jhasse:
What's the use-case for this?

Is this something that could be fixed with a jobserver #1140?

davezarzycki:
This is similar to the GNU make jobserver solution with two key differences:
1) The GNU make jobserver implementation "only" supports recursion whereas this pull request allows both recursive and non-recursive invocations to share a limit. I say "only" in quotes because the GNU jobserver design could be extended to be more like like this patch but doing so would require a server process.
2) The GNU make jobserver protocol has become a "standard" that other programs now support.

The problem that motivated me is in fact, non-recursive unrelated ninjas trying to share a jobs limit. I often have a long running build or two running in the background while I actively work on something else, thus creating the potential for three builds to be running at once.

My approach would also help anybody recursively invoking ninja too but that problem isn't motivating this patch.

If #1140 limits jobserver protocol participation to recursive invocations (like GNU make), then no, #1140 does not solve my use case.

davezarzycki:
Also, I'd like to add that `ninja` is known for being fast and the jobserver protocol is not:

https://www.phoronix.com/scan.php?page=news_item&px=Linux-Pipe-Parallel-Job-Opt

Personally speaking, this kind of problem really ought to be solved with an atomic variable in shared memory (for the fast path) and only fall back to the kernel as needed.

jhasse:
> The problem that motivated me is in fact, non-recursive unrelated ninjas trying to share a jobs limit. I often have a long running build or two running in the background while I actively work on something else, thus creating the potential for three builds to be running at once.

Okay, couldn't this be solved by pausing any running ninja jobs when a new one starts (and continue them afterwards)?

davezarzycki:
> > The problem that motivated me is in fact, non-recursive unrelated ninjas trying to share a jobs limit. I often have a long running build or two running in the background while I actively work on something else, thus creating the potential for three builds to be running at once.
> 
> 
> 
> Okay, couldn't this be solved by pausing any running ninja jobs when a new one starts (and continue them afterwards)?

That really wouldn't work for me. This patch allows unrelated cron jobs (and sometimes interactively invoked ninjas) to share the machine without overwhelming it.

jhasse:
> That really wouldn't work for me.

Can you explain why? I don't mean that you would manually pause the other ninja jobs, but that it would happen automatically.

davezarzycki:
> > That really wouldn't work for me.
> 
> Can you explain why? I don't mean that you would manually pause the other ninja jobs, but that it would happen automatically.

Sorry, I updated the comment after I posted it. As the update said, I have cron jobs running, and I wouldn't want them pausing any interactive ninja. And some of these cron jobs auto-bisect with git, so they have the potential to run many ninja invocations that could pause interactive ones.

Ultimately, I plan on using this feature on a CI bot to let users dynamically share the machine without overwhelming it.

jhasse:
The cronjob's ninjas could work the other way around, i.e. they would wait for other running ninjas before starting themselves.

> Ultimately, I plan on using this feature on a CI bot to let users dynamically share the machine without overwhelming it.

What exactly do you mean by "overwhelming"? If you run x ninjas at the same time, each one will take x times longer to finish - there's no way to avoid that.

davezarzycki:
> The cronjob's ninjas could work the other way around, i.e. they would wait for other running ninjas before starting themselves.
> 
> > Ultimately, I plan on using this feature on a CI bot to let users dynamically share the machine without overwhelming it.
> 
> What exactly do you mean by "overwhelming"? If you run x ninjas at the same time, each one will take x times longer to finish - there's no way to avoid that.

If you run X ninjas without this patch, then they will use x times the amount of RAM (and incur needless context switching / cache thrashing). I really want to avoid my future CI bot running out of RAM/swap.

Also, the project I work on basically uses 96 cores for a few minutes to run ninja jobs until a critical section where only two cores are used (and then the test suite runs). So pausing/unpausing ninjas really is the wrong granularity when one ninja instance might only be using two of the 96 cores for a few minutes.

jhasse:
Okay I see. Would something like #1571 also help?

Regarding sharing job limit: I think having a ninja server process which could also be used to cache parsing of the build manifest would be the best approach.

davezarzycki:
My machines are not memory bound when running optimally (one compiler instance per CPU), so #1571 would not help.

I also agree that a ninja server process has many nice properties and would solve my goals. That being said, I'm trying to avoid getting tricked into major feature work. ;-)

If a serverless solution to shared job limit is undesirable, then I can simply close this PR and keep my patch local. I just wanted to 1) let you know my patch exists and 2) offer to work on making it merge ready if desired.

hzeller:
Reading through the motivation, it sounds like a highly specialized use-case.

You mention cron-jobs so I assume you run on Linux. Wouldn't cgroups help with this, which is a well-established OS feature ?
If all ninjas are launched in the same cgroup, you could use that to limit resources and it sounds like [concurrent processes](https://www.kernel.org/doc/Documentation/cgroup-v1/pids.txt) would be what you need here.
It would only require to makre sure that `EAGAIN` is properly handled in ninja, so if not, then this might still need some modification in ninja, and that would make things more robust.

This keeps the ninja code-base slim and managing the code-base between platforms more manageable (I see you had to add a few `#ifndef _WIN32` here and there).

(Note, I am not a Ninja maintainer, just reading through some pull requests, so these are merely my 2¢)

davezarzycki:
Hi @hzeller – Linux happens to be my primary development environment but it isn't my only development environment. Even if it were, cgroups only provides a static limit, whereas what I've implemented is a dynamic limit. Finally, the failure model for cgroups is awful: EGAIN from fork/clone? Most software doesn't handle that scenario at all, including ninja itself.

In any case, the code can and should be cleaned up. I think it can mostly if not completely contained into `subprocess-posix.cc` and that would clean up the `_WIN32` C preprocessor usage.

davezarzycki:
Everybody – please let me be honest. I have lots of criticisms about this pull request and the general problem space. I used to work at Apple for a long time at the POSIX layer, and I wrote launchd (like systemd but I loathe the comparison) and libdispatch (a thread pool and I/O library). I know a lot about how awful POSIX is and this exact PR is a great example. Serverless cross-process synchronization is notoriously difficult (to impossible) to get right in the POSIX world and I would totally respect you all if you don't want it for that reason.

That being said, lots of software is quite successful despite how awful the POSIX APIs are. In this case, this feature is opt-in via an environment variable and the POSIX failure modes can be documented. If people are okay with the POSIX failure scenarios, then great. If not, nobody is forcing them to use the feature.

jhasse:
> If a serverless solution to shared job limit is undesirable, then I can simply close this PR and keep my patch local.

Let's keep it open to see more feedback coming in :)

bonzini:
> Also, I'd like to add that ninja is known for being fast and the jobserver protocol is not:

The cost of the jobserver protocol is literally a few system calls per built file. That's nothing compared to the cost of the compiler.

I would prefer #1140 to go in since it offers more interoperability (not just with make but also with Cargo, gcc -flto, etc.&mdash;especially if ninja also included the server side), but this would also be a partial solution to the same problem.

davezarzycki:
For the record, I've been living with this pull request since it was created and it has a bug that I haven't fixed yet. Namely, the semaphore isn't reset properly when `ninja` is killed via control-C. Workaround: delete the named semaphores (wherever they exist for your OS).

Issue #420, opened 2012-09-17
-C option creates unwanted chdir message, add -q quiet option (patch attached)


rgeary1:
The -C option creates a "Changing to X directory" message which can't be suppressed. For our use case, ninja is part of a larger build process where this message isn't relevant. Since I can't use grep -v or sed  on the ninja output without losing the smart screen formatting, I've added a -q quiet option (patch below).
If someone else has another use for a quiet option later, it could be extended to take a number value as a quietness level.
This could be solved in other ways, this is just one solution.

From 1d11fc53cb4181199c065f138bc8513469ae1d7e Mon Sep 17 00:00:00 2001
From: Richard Geary richardg.work@gmail.com
Date: Fri, 14 Sep 2012 05:41:37 -0400
Subject: [PATCH 1/1] Add quiet -q option to suppress changedir message when using -C option
## Change-Id: I12f1bf232e29786e6699cbe9b0724425e7dcb0ff

 src/ninja.cc |   10 +++++++---
 1 files changed, 7 insertions(+), 3 deletions(-)

diff --git a/src/ninja.cc b/src/ninja.cc
index ad56f1c..9a7ac8b 100644
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -106,7 +106,7 @@ void Usage(const BuildConfig& config) {
 "options:\n"
 "  --version  print ninja version (\"%s\")\n"
 "\n"
-"  -C DIR   change to DIR before doing anything else\n"
+"  -C DIR   change to DIR before doing anything else. Use -q to supress directory change notification.\n"
 "  -f FILE  specify input build file [default=build.ninja]\n"
 "\n"
 "  -j N     run N jobs in parallel [default=%d]\n"
@@ -726,6 +726,7 @@ int NinjaMain(int argc, char*\* argv) {
   const char\* input_file = "build.ninja";
   const char\* working_dir = NULL;
   string tool_name;
-  bool printChdir = true;
  
  setvbuf(stdout, NULL, _IOLBF, BUFSIZ);

@@ -740,7 +741,7 @@ int NinjaMain(int argc, char*\* argv) {

   int opt;
   while (tool_name.empty() &&
-         (opt = getopt_long(argc, argv, "d:f:hj:k:l:nt:vC:V", kLongOptions,
-         (opt = getopt_long(argc, argv, "d:f:hj:k:l:nt:vC:Vq", kLongOptions,
                           NULL)) != -1) {
   switch (opt) {
     case 'd':
[...]

PR #1818, opened 2020-07-20
Add --quiet option that suppresses status updates.

hzeller:
Refined pull request after discussion in #1816

Signed-off-by: Henner Zeller <h.zeller@acm.org>

### DIFF
--- a/src/build.h
+++ b/src/build.h
@@ -162,8 +162,9 @@ struct BuildConfig {
                   failures_allowed(1), max_load_average(-0.0f) {}
 
   enum Verbosity {
-    NORMAL,
     QUIET,  // No output -- used when testing.
+    NO_STATUS_UPDATE,  // just regular output but suppress status update
+    NORMAL,  // regular output and status update
     VERBOSE
   };
   Verbosity verbosity;
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -207,6 +207,7 @@ void Usage(const BuildConfig& config) {
 "options:\n"
 "  --version      print ninja version (\"%s\")\n"
 "  -v, --verbose  show all command lines while building\n"
+"  --quiet        don't show progress updates.\n"
 "\n"
 "  -C DIR   change to DIR before doing anything else\n"
 "  -f FILE  specify input build file [default=build.ninja]\n"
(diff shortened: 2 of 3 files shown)


src/ninja.cc
```diff
@@ -207,6 +207,7 @@ void Usage(const BuildConfig& config) {
 "options:\n"
 "  --version      print ninja version (\"%s\")\n"
 "  -v, --verbose  show all command lines while building\n"
+"  --quiet        don't show progress updates.\n"
```
jhasse, line 210: remove . at the end

also not show if we should call it "progress" or "status". Maybe even "only show command output"?

src/build.h
```diff
@@ -162,8 +162,9 @@ struct BuildConfig {
                   failures_allowed(1), max_load_average(-0.0f) {}
 
   enum Verbosity {
-    NORMAL,
     QUIET,  // No output -- used when testing.
+    NO_STATUS_UPDATE,  // just regular output but suppress status update
+    NORMAL,  // regular output and status update
```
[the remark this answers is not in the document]
hzeller: I wanted to emphasize the 'level' of logging from very quiet to very verbose.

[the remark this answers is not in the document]
jhasse: Ah, good idea :)

src/ninja.cc
hzeller: How about `don't show progress status, just command output` ?

jhasse: Not sure, maybe someone else can chime in.

hzeller:
No message improvement suggestions so far; ready to merge ?

jhasse:
Yep. I'll need to finish the 1.10.1 release first though.

Leedehai:
Wondering if it's acceptable to suppress the `ninja: Entering directory ...` message using this option as well? @hzeller @jhasse

jhasse:
Yes, good catch.

Also this PR should add some testing to misc/output_test.py.

gjasny (commented):
Silencing ninja status will help us to drastically cut down log size.
But I'm wondering what will happen once we got rid of all compiler warnings and no output will appear for minutes.
Some CI systems might think that the build hangs.

From an architectural point of view: would it be possible to print progress at least once every n seconds?

Leedehai:
@gjasny I feel this looks like a mission creep. AFAIK, other build tools don't have a feature that prints progress periodically just to signal it's working - they either print logs, or don't. If the CI has problem with long silence, then IMO it's the CI's responsibility to accommodate, not the build tools'.

jhasse:
What do you think about
```diff
diff --git a/src/ninja.cc b/src/ninja.cc
index e085e62..5c6ac91 100644
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -1368,7 +1368,7 @@ NORETURN void real_main(int argc, char** argv) {
     // subsequent commands.
     // Don't print this if a tool is being used, so that tool output
     // can be piped into a file without this string showing up.
-    if (!options.tool)
+    if (!options.tool && config.verbosity != BuildConfig::NO_STATUS_UPDATE)
       printf("ninja: Entering directory `%s'\n", options.working_dir);
     if (chdir(options.working_dir) < 0) {
       Fatal("chdir to '%s' - %s", options.working_dir, strerror(errno));
```

to fix #420, too?

klimkin:
Would be nice to have this --quiet option or environment variable!

megabyde:
Any updates?

jhasse:
Closing in favor of #1968.

PR #1823, opened 2020-08-01
Use c++ instead of g++

serpent7776:
`configure.py` assumes `g++` is present on the system, which is not the case, e.g. for FreeBSD.
`c++` should be used insted, which should be a link to system c++ compiler. This will be `g++` for linux, but `clang++` for FreeBSD.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -269,7 +269,7 @@ if configure_env:
     n.variable('configure_env', config_str + '$ ')
 n.newline()
 
-CXX = configure_env.get('CXX', 'g++')
+CXX = configure_env.get('CXX', 'c++')
 objext = '.o'
 if platform.is_msvc():
     CXX = 'cl'


jhasse:
Are there systems where g++ exists but c++ doesn't?

Note that you can already change the compiler using the CXX environment variable.

serpent7776:
A few linux systems I've used linked `c++` to `g++`. FreeBSD links `c++` to `clang++`, but I'm not sure if it's like that for all systems. I don't know a system that doesn't provide `c++` (except for Windows).

I'm aware of the `CXX` variable, that's how FreeBSD ports build ninja.

I guess it can be dropped if there're any concerns that it can introduce an issue.

jhasse:
Let's see if someone complains. `c++` sounds like the better default to me.

PR #1834, opened 2020-08-26
Close BuildLog while running generators, fix #1724

### LINKED ISSUE #1724: Log files held open when re-generating build manifest

jhasse:
While #1780 delayed opening .ninja_log until the first write, this this didn't fully fix the issue on Windows: There might be build
statements which run before the generator resulting in an actual write to .ninja_log.

To fix this once and for all the BuildLog now gets closed before running the generator. `BuildLog::log_file_path_` won't be cleared so that it can be opened again after the generator finished.

See #1724. Not sure if the same needs to be done for .ninja_deps?

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -829,6 +829,10 @@ bool Builder::Build(string* err) {
     // See if we can start any more commands.
     if (failures_allowed && command_runner_->CanRunMore()) {
       if (Edge* edge = plan_.FindWork()) {
+        if (edge->GetBindingBool("generator")) {
+          scan_.build_log()->Close();
+        }
+
         if (!StartEdge(edge, err)) {
           Cleanup();
           status_->BuildFinished();
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -181,7 +181,7 @@ void BuildLog::Close() {
 }
 
 bool BuildLog::OpenForWriteIfNeeded() {
-  if (log_file_path_.empty()) {
+  if (log_file_ || log_file_path_.empty()) {
     return true;
   }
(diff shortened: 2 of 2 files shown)


mathstuf:
`.ninja_deps` probably needs it as well since `ninja -t recompact` is run during CMake's reconfigure.

jhasse:
Does `CONFIGURE_DEPENDS` (see https://gitlab.kitware.com/cmake/cmake/-/issues/20274#note_817221) need `.ninja_deps` though?

bradking:
I just built ninja with this change plus a local hack to make its version number 1.10.2.  The CMake `CONFIGURE_DEPENDS` test case now passes in the case where #1780  (and therefore 1.10.1) failed.

One might be able to produce a failure with `.ninja_deps` by constructing a test case in which a source file is compiled and its `.d` file integrated into the deps log, and *then* a `generator=1` rule runs.  I don't think CMake's test suite has such a case though.

jhasse:
Thx for testing!

> I don't think CMake's test suite has such a case though.

Does CMake allow such a thing? If not, I don't think that it's worth supporting.

bradking:
I don't think CMake has a way to encounter the `.ninja_deps` corner case, so I think we can skip handling that for now.

kayal28 (approved):

PR #1836, opened 2020-08-27
CMake: Add support for "browse" mode

### LINKED ISSUE #1822: CMake build doesn't support the browse tool
Enabling the browse requires additional logic to inline the `browse.py` into Ninja which is currently only supported by `configure.py` and not by the CMake build.

aharrison24:
Fixes #1822
Closes #1853

Adds support for `ninja -t browse` to CMake builds.

Browse mode requires a number of POSIX features to be available.
This commit adds configure-time checks that the 'unistd.h' header is
available and that the `inline.sh` script executes successfully. If the
checks pass then browse mode is enabled.

As discussed in #1853, when built via CMake the `ninja` executable
looks for a binary called `python` in the current path, in order to
launch the "browse" mode. The behaviour differs from that of the
configure.py script, which looks for a python executable that has the
*same name* as the python executable that invoked the configure script.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,6 +48,18 @@ else()
 endif()
 target_include_directories(libninja-re2c PRIVATE src)
 
+# --- Check for 'browse' mode support
+set(unsupported_browse_platforms
+	"Windows"
+	"SunOS"
+	"AIX"
+)
+if(${CMAKE_SYSTEM_NAME} IN_LIST unsupported_browse_platforms)
+	set(platform_supports_ninja_browse FALSE)
+else()
+	set(platform_supports_ninja_browse TRUE)
+endif()
+
 # Core source files all build into ninja library.
 add_library(libninja OBJECT
 	src/build_log.cc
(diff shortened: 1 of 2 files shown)


aharrison24:
Sorry, this obviously isn't ready to go. I'll look in to the CI failures.

jhasse:
Thx!

Instead of adding a list of unsupported operation systems, wouldn't a check for `unistd.h` work? I have no idea why SunOS and AIX aren't supported though.

alastair-harrison-oxb:
Yeah, I wasn't really sure what the reason for some platforms being unsupported was. So I just duplicated the logic from the configure script.

If what it comes down to is OS support for POSIX stuff like `execvp` and `fork` then we could instead use `try_compile` to determine whether those functions are available. Then only enable 'browse' mode if it looks like it will compile properly. I don't know if that would actually buy us much over just checking for `unistd.h`. Although I guess a check for `unistd.h` would also best be done in a `try_compile` command, since we can't be sure exactly where it might live.

aharrison24:
OK, I've taken a crack at doing some more robust platform checks before 'browse' mode is enabled. I did need to modify the `inline.sh` script so that it would return a non-zero exit code if any of its steps failed. I hope that's a reasonable change.

jhasse:
Looks good, thanks!

PR #1838, opened 2020-08-30
CMake: make test build optional

scivision:
For low-resource systems, it can be useful to greatly speed up the build
by not building the self-tests when they're not wanted. This also helps
avoid the Ninja build crashing since the self test build uses > 1GB RAM.


jhasse:
Simply use `cmake --build build-cmake --target ninja` :)

PR #1839, opened 2020-08-31
cmake: add BUILD_TESTING option (ON by default)

ledocc:
option provided by cmake's buildin Module CTest
enable_testing() is call by this Module

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -151,45 +151,45 @@ endif()
 
 include(CTest)
 if(BUILD_TESTING)
-	# Tests all build into ninja_test executable.
-	add_executable(ninja_test
-		src/build_log_test.cc
-		src/build_test.cc
-		src/clean_test.cc
-		src/clparser_test.cc
-		src/depfile_parser_test.cc
-		src/deps_log_test.cc
-		src/disk_interface_test.cc
-		src/dyndep_parser_test.cc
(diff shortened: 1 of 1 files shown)


scivision:
Thank you!  It is nice to use the canonical option as you have done vs. #1838.

My use case is building software suite natively on embedded systems, and one of the first steps is building Ninja, which uses too much RAM when ninja_test is built. And I don't want to built test suites for any of the suite to save time on a puny CPU. 
The users are engineers / scientists. The idea is to have a standalone system where the user doesn't need to setup a cross-compile toolchain when they might never or only once use it.

jhasse:
Thanks! While I went with tabs in CMakeLists.txt I think it makes more sense to follow the style of the rest of Ninja's source. Can you convert the tabs to 2 spaces in the lines you've touched?

ledocc:
For the "TAB or space" choice, I prefer space. But I follow the CMakeLists.txt style.
I can convert tabs to 2 spaces. But only for lines I have touch? This will mix tab an space in the same file.
Will be better to change all tabs to 2 space in a separate commit, isn't it ?

jhasse:
> But only for lines I have touch? 

Yes. I want to avoid a dedicated commit. I think there's already a mixture in the file.

ledocc:
done !

PR #1845, opened 2020-09-14
Add terse output mode

gjasny:
In contrast to the `--quiet` output mode, the `--terse` mode
periodically prints status information if none have been printed
during a certain time frame (right now 15 seconds).

Additionally the status is printed before any command output so
that the source of the output could be determined more easily.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -291,7 +291,8 @@ string BuildStatus::FormatProgressStatus(
 }
 
 void BuildStatus::PrintStatus(const Edge* edge, EdgeStatus status) {
-  if (config_.verbosity == BuildConfig::QUIET)
+  if (config_.verbosity == BuildConfig::QUIET
+      || config_.verbosity == BuildConfig::NO_STATUS_UPDATE)
     return;
 
   bool force_full_command = config_.verbosity == BuildConfig::VERBOSE;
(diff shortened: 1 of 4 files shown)


gjasny:
This PR sits on top of #1818 to avoid merge conflicts.

gjasny:
Sample output with timestamps:
```
NINJA_STATUS='[%e %f/%t] ' ~/src/ninja/cmake-build-debug/ninja --terse
[1.955 1/1694] Building CXX object modules/a.cpp.o
[17.944 114/1694] Generating bar.h
[33.023 168/1694] Generating foo.h
[37.162 207/1694] Linking CXX static library lib/baz.a
warning: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: archive library: lib/baz.a the table of contents is empty (no object file members in the library define global symbols)
[52.396 334/1694] Generating misc.cpp
```

As you could see the status lines for lines which do not produce output are more than 15s apart and skip many in-between-target outputs. The warning is printed with a status header and afterwards 15s elapse again, before printing another status.

gjasny:
I'm unsure about the `use_console()` condition and its use case. But as far as I understand it gets used if the process needs exclusive access to the console because it will perform I/O. In that case always printing the status line seems to be the correct way.

PR #1849, opened 2020-09-18
Generate directory for depfile

### LINKED ISSUE #1848: Ninja does not create depfile directories
Ninja creates the directories for all output files, but not the directory for a depfile. This is reasonable most of the time, as the depfile is commonly located in the same directory as the corresponding object file or the command being executed might create it. However, if neither of those conditions are true, then the command might fail on account of the directory not existing. Since a depfile is sort of an additional output file of a command, would it make sense for Ninja to also create the depfile directory?

rascani:
Ninja creates the directories for all output files, but not the
directory for a depfile. This works for most commands, as the depfile is
typically located in the same directory as the corresponding object file
or the command might generate the directory for it.  However, if it does
not, the command could fail due to directory not existing. This change
creates the directory for the depfile when creating the directories for
the output files.

Fixes #1848

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -908,6 +908,14 @@ bool Builder::StartEdge(Edge* edge, string* err) {
       return false;
   }
 
+  // Create the depfile directory
+  // XXX: this will block; do we care?
+  string depfile = edge->GetUnescapedDepfile();
+  if (!depfile.empty()) {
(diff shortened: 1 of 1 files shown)


philwo:
I think this PR can be closed, now that https://github.com/ninja-build/ninja/pull/2492 has been merged.

PR #1852, opened 2020-09-30
Add clang-analyze to build

jonesmz:


jhasse:
I think clang-tidy runs the analyzer, too, right? See #1768.

jonesmz:
Apparently it can, I didn't know that.

But it requires the specific clang-analyze functionality to be enabled.

Further, it looks like most clang-analyze checks aren't supported by clang-tidy:
https://clang-analyzer.llvm.org/available_checks.html
https://clang.llvm.org/extra/clang-tidy/checks/list.html

I count 11 non-OSX checks in clang-tidy, and over 25 non-OSX checks in clang-analyze.

jhasse:
Ah okay, than running the clang-analyze, too, seems a good idea. When does it run though? The output doesn't help me. Are you sure it would find something (and the GitHub Action check would fail)?

jonesmz:
The "scan-build" command inter-poses the clang-analyzer as a CC and CXX, runs the analysis, and then calls the "real" compiler.

In this case, it doesn't appear that the default-enabled checks are finding anything to complain about, but other projects I work with the default checks do complain.

Do you have any specific checks that you'd like enabled?

jhasse:
No, the default is fine :)

PR #1853, opened 2020-10-02
GitHub Actions: Upgrade packages before installing new ones

jhasse:


jhasse:
Uarg, apparently GitHub Actions isn't running a "normal" Ubuntu image where you can "install packages". See https://github.com/actions/virtual-environments/issues/46 and https://github.com/actions/virtual-environments/issues/357.

I need to set up running the tests in Docker ...

PR #1854, opened 2020-10-02
Fix clang-analyze build, need to also run during configure

jonesmz:
I apologize for not including this fix in the recently merged PR. I had the change in my local workspace, but apparently forgot to push it.

The scan-build command needs to run on both the configure and build step. My understanding is that somehow CMake is detecting that scan-build was used and doing something with it.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -27,7 +27,7 @@ jobs:
     - name: Build ninja
       shell: bash
       run: |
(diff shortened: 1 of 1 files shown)


jhasse:
Are you sure this doesn't affect the binary in any way?

jonesmz:
Closing in favor of: https://github.com/ninja-build/ninja/pull/1855

PR #1855, opened 2020-10-02
Add debug build with sanitizers support, fix scan-build

jonesmz:
This change adds a "debug" build where things that affect the binary for sake of analysis purposes can live.


jhasse:
I do not want to add code for this to CMakeLists.txt, sorry. This should rather be done via CXXFLAGS IMHO.

jonesmz:
Better?

jhasse:
Yes

PR #1862, opened 2020-10-20
Remove 'using namespace std', properly namespace all std symbols

jonesmz:
This PR changes nothing except removing 'using namespace std' from .cpp files, and adding std:: where needed. The one exception to this is adding #define NOMINMAX to src/includes_normalize-win32.cc to work around a windows problem where windows defines min and max as a macro.

If you want this PR broken up into individual files, let me know how many files you want per PR. I can make a new PR for each file if desired.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -323,7 +323,7 @@ if platform.is_msvc():
               # Disable size_t -> int truncation warning.
               # We never have strings or arrays larger than 2**31.
               '/wd4267',
-              '/DNOMINMAX', '/D_CRT_SECURE_NO_WARNINGS',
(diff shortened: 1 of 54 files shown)


jonesmz:
Github actions is failing with this error: curl: (7) Failed to connect to 2001:630:341:12::184: Cannot assign requested address

Looks to be a failure with github actions, not something related to the PR.

jhasse:
Not something we want to do any time soon, sorry. As there are already enough PRs in the backlog and this one would conflict very often, I'm closing it. Hope you understand.

PR #1872, opened 2020-11-04
Do not use /bin/sh for execution of built-in commands.

### LINKED ISSUE #1871: Use vanilla env to execute a subprocess
Ubuntu/Debian: 
```
kostja@hulk:~$ cat build.ninja 
rule test 
    command = echo -e 'foo\nbar'
    description = Test rule
build test: test 
kostja@hulk:~$ ninja test
[1/1] Test rule
-e foo
bar
kostja@hulk:~$ which echo
echo: shell built-in command
```
Fedora 32:
```
kostja@goliath~% cat ./build.ninja;
rule test                             
    command = echo 'foo\nbar'         
    description = Test rule           
build test: test                      
kostja@goliath~% ninja test         
[1/1] Test rule                       
foo\nbar                              
kostja@goliath~% cat ./build.ninja; 
rule test                             
    command = echo -e 'foo\nbar'      
    description = Test rule           
build test: test                      
kostja@goliath~% ninja test         
[1/1] Test rule                       
foo                                   
bar                           
```
In Subprocess, it's best to use /usr/bin/env instead of /bin/sh to ensure the path to built-ins does not depend on the derived environment, and thus is unstable in developer and pipleine environment.

kostja:
Portable shell programming is a challenge. /bin/sh defaults to bash
on most Linuxes, but not Debian and its descendants (where it points
to dash), which leads to different behaviour of shell builtins,
e.g. `echo`.

Autoconf's guide to portable shell programming calls this situation
a deadlock.

One way to ensure ninja behaviour is identical on all platforms is
require Bash (/bin/bash) in all cases..
Another is to avoid using the shell interpreter altogether.

This patch implements the second alternative.

Fixes #1871

### DIFF
--- a/src/subprocess-posix.cc
+++ b/src/subprocess-posix.cc
@@ -117,7 +117,11 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
   if (err != 0)
     Fatal("posix_spawnattr_setflags: %s", strerror(err));
 
-  const char* spawned_args[] = { "/bin/sh", "-c", command.c_str(), NULL };
+  // Avoid shell builtin expansion, such as $$ and echo, this is
+  // not portable.
+  string exec_command = "exec " + command;
+
+  const char* spawned_args[] = { "/bin/sh", "-c", exec_command.c_str(), NULL };
   err = posix_spawn(&pid_, "/bin/sh", &action, &attr,
         const_cast<char**>(spawned_args), environ);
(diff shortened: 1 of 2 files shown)


jhasse:
As you can see from the CI failures, this is a behaviour change many users of Ninja won't expect. Closing for now, needs more discussion in the issue.

PR #1876, opened 2020-11-11
Suppress MSVC warnings

orgads:
* D9025: overriding '/GR' with '/GR-'
* C4100: 'argc': unreferenced formal parameter
* C4244: '=': conversion from 'int' to 'char', possible loss of data (on tolower)
* C4267: 'initializing': conversion from 'size_t' to 'unsigned int', possible loss of data
* C4702: unreachable code (after Fatal)
* C4706: assignment within conditional expression
* C4996: 'strcpy': This function or variable may be unsafe


jonesmz:
Please fix the cause of the warnings. They exist for a reason

jhasse:
> D9025: overriding '/GR' with '/GR-'

This was fixed.

> C4100: 'argc': unreferenced formal parameter

Fix would be to move away from getopts (it's a C API and horrible). Not going to happen any time soon.

> C4244: '=': conversion from 'int' to 'char', possible loss of data (on tolower)

We could use `std::tolower(..., std::locale())` instead?

> C4267: 'initializing': conversion from 'size_t' to 'unsigned int', possible loss of data

I don't think it's worth fixing this one.

> C4702: unreachable code (after Fatal)

false-positive

> C4706: assignment within conditional expression

re2c generates this, nothing we can (easily) do about it.

> C4996: 'strcpy': This function or variable may be unsafe

Using C++ instead of the C str API would be the fix. Not that important though.

orgads:
> > C4100: 'argc': unreferenced formal parameter
> 
> Fix would be to move away from getopts (it's a C API and horrible). Not going to happen any time soon.

I only pasted the first occurrence, but there are tons of these...

> > C4244: '=': conversion from 'int' to 'char', possible loss of data (on tolower)
> 
> We could use `std::tolower(..., std::locale())` instead?

I doubt it will help.

I agree with the rest.

jhasse:
> I doubt it will help.

I've just tried it:

```diff
diff --git a/src/disk_interface.cc b/src/disk_interface.cc
index 594bc51..44d09d7 100644
--- a/src/disk_interface.cc
+++ b/src/disk_interface.cc
@@ -15,6 +15,7 @@
 #include "disk_interface.h"

 #include <algorithm>
+#include <locale>

 #include <errno.h>
 #include <stdio.h>
@@ -120,7 +121,8 @@ bool StatAllFilesInDir(const string& dir, map<string, TimeStamp>* stamps,
       // This is the case at least on NTFS under Windows 7.
       continue;
     }
-    transform(lowername.begin(), lowername.end(), lowername.begin(), ::tolower);
+    transform(lowername.begin(), lowername.end(), lowername.begin(),
+              [](char c) { return std::tolower(c, std::locale()); });
     stamps->insert(make_pair(lowername,
                              TimeStampFromFileTime(ffd.ftLastWriteTime)));
   } while (FindNextFileA(find_handle, &ffd));
@@ -178,8 +180,10 @@ TimeStamp RealDiskInterface::Stat(const string& path, string* err) const {
     dir = path;
   }

-  transform(dir.begin(), dir.end(), dir.begin(), ::tolower);
-  transform(base.begin(), base.end(), base.begin(), ::tolower);
+  transform(dir.begin(), dir.end(), dir.begin(),
+            [](char c) { return std::tolower(c, std::locale()); });
+  transform(base.begin(), base.end(), base.begin(),
+            [](char c) { return std::tolower(c, std::locale()); });

   Cache::iterator ci = cache_.find(dir);
   if (ci == cache_.end()) {
```

Fixes the warning indeed. I'm not sure if we should do it though, as this is code in the real disk interface not covered by unit tests.

jhasse:
As my patch is rather ugly, I think disabling the warning is okay. Thank you!

orgads:
Thank you :)

PR #1879, opened 2020-11-13
Fix building on AIX

### LINKED ISSUE #1844: Building ninja on AIX
Hi,
we encountered build issues when we tried to build ninja with IBM's Visual Age compiler 16.1 (using the xlclang++ frontend).

Here is what we had to change to make the  build go through:
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 65e42a4..c4aa588 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,10 @@
 project(ninja)
 
+cmake_policy(VERSION 3.15)
+message(WARNING "Using compiler ${CMAKE_CXX_COMPILER_ID}")
+
+
 # --- optional link-time optimization
 if(CMAKE_BUILD_TYPE MATCHES "Release")
        include(CheckIPOSupported)
@@ -87,6 +91,12 @@ else()
        target_sources(libninja PRIVATE src/subprocess-posix.cc)
 endif()
 
+if( ${CMAKE_SYSTEM_NAME} STREQUAL "AIX" )
+       target_compile_definitions(libninja PRIVATE  __STDC_FORMAT_MACROS)
+        target_sources(libninja PRIVATE src/getopt.c)
+       target_link_libraries(libninja PUBLIC -lperfstat)
+endif()
+
 #Fixes GetActiveProcessorCount on MinGW
 if(MINGW)
 target_compile_definitions(libninja PRIVATE _WIN32_WINNT=0x0601 __USE_MINGW_ANSI_STDIO=1)
@@ -95,7 +105,9 @@ endif()
 # Main executable is library plus main() function.
 add_executable(ninja src/ninja.cc)
 target_link_libraries(ninja PRIVATE libninja libninja-re2c)
-
+if( ${CMAKE_SYSTEM_NAME} STREQUAL "AIX" )
+       target_compile_definitions(ninja PRIVATE  __STDC_FORMAT_MACROS)
+endif()
 # Tests all build into ninja_test executable.
 add_executable(ninja_test
        src/build_log_test.cc
diff --git a/src/manifest_parser_perftest.cc b/src/manifest_parser_perftest.cc
index 67d11f9..a7a748c 100644
--- a/src/manifest_parser_perftest.cc
+++ b/src/manifest_parser_perftest.cc
@@ -25,6 +25,9 @@
 #ifdef _WIN32
 #include "getopt.h"
 #include <direct.h>
+#elif _AIX
+#include "getopt.h"
+#include <unistd.h>
 #else
 #include <getopt.h>
 #include <unistd.h>
```

kadler:
This PR synchronizes the configure.py and CMake builds as well as uses
the internal getopt implementation on AIX, since AIX getopt does not
support getopt_long. Previously, the IBM i configure.py build would
link with libutil for this, but as that's not a standard library, I've
switched to also use the internal getopt implementation.

Additionally, this fixes some crashes within the tests. A couple are
due to peculiarities of AIX's 32-bit memory model, while one is due to
a bug in the test causing an out of bounds read.

Fixes #1844

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,6 +112,9 @@ if(WIN32)
 	endif()
 else()
 	target_sources(libninja PRIVATE src/subprocess-posix.cc)
+	if(CMAKE_SYSTEM_NAME STREQUAL "OS400" OR CMAKE_SYSTEM_NAME STREQUAL "AIX")
+		target_sources(libninja PRIVATE src/getopt.c)
+	endif()
 endif()
 
 #Fixes GetActiveProcessorCount on MinGW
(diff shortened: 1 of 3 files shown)


src/hash_collision_bench.cc
```diff
@@ -27,9 +27,10 @@ int random(int low, int high) {
 
 void RandomCommand(char** s) {
   int len = random(5, 100);
-  *s = new char[len];
+  *s = new char[len+1];
```
jonesmz, line 30: Can you make a stand-alone PR for this?

kadler: Sure, no problem. I debated it originally.

kadler: Opened #1880

PR #1881, opened 2020-11-14
Vs quality of life

hdf89shfdfs:
First off this improves the visual studio appearance generated.

Before:
image

After:
image

Secondly add the /MP compiler flag to speed up builds for Visual Studio users.

Comment the MSVC compiler options, and use add_compile_options instead of modifying the CMAKE_CXX_FLAGS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,12 +33,6 @@ if(MSVC)
 	add_compile_options("/Zc:__cplusplus")
 	# Turn off RTTI
 	add_compile_options("$<$<COMPILE_LANGUAGE:CXX>:/GR->")
-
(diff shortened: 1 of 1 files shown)


jhasse:
Why isn't `USE_FOLDERS` on by default?

hdf89shfdfs:
@jhasse I'm not sure why. 

Most likely it would break compatibility with older cmake projects.

Good question to add to the cmake discourse: 
https://discourse.cmake.org/

jhasse:
Aren't that what CMake policies are for? I'm not sure.

I'm also not using VS, but I dislike adding subjective (someone might dislike folders; `/MP` breaks incremental compilation IIRC) options to `CMakeLists.txt`. Feel free to add some info to the wiki though.

PR #1884, opened 2020-11-23
disk_interface: Improve the stat cache handling for case sensitive folders on Windows

MartinBP:
disk_interface: Improve the stat cache handling for case sensitive folders on Windows

Do not convert paths to lower case before adding them to the stat cache or using them
as input to FindFirstFileExA. By not converting them to lower case, case sensitive folders
will now be handled correctly. Case insensitive folders still works as expected because casing doesn't matter.
Filenames are still converted to lower case, which works fine for both case sensitive and case insensitive folders.
The only thing not supported are two or more files with the exact same naming (except for casing) in the same case sensitive folder.


jhasse:
This will result in more cache misses in `RealDiskInterface::Stat` if `dir` differs in casing.

MartinBP:
That is correct. If the directories listed in build.ninja have inconsistent casing for the same folder then it will result in a cache miss.
If this is a worry, then it can be solved by creating a case insensitive comparator and use that to initialize the Cache.

```
#ifdef _WIN32
struct CaseInsensitiveCompare {
    bool operator()(const std::string &lhs, const std::string &rhs) const {
        return _stricmp(lhs.c_str(), rhs.c_str()) < 0;
    }
};
#endif

...

typedef std::map<std::string, DirCache, CaseInsensitiveCompare> Cache;
```

The downside is that two folders can't have the same name (except casing) within the same case sensitive folder, but this limitation also applies to files, so it might not be a bad idea.

jhasse:
Using `_stricmp` will have performance implications.

MartinBP:
I had a second take on this and because it was quite some time since last then I went ahead and did a rebase/merge.
These commits now show up in this pull request - is that a problem? Personally I think it looks a bit messy and it might be because I have done something in an incorrect order.

Back to the change.
The only thing that have changed compared to the original code is that the "dir/path" argument to FindFirstFileExA is no longer being converted to lower case.

jonesmz:
You conducted the rebase incorrectly. They shouldn't show up in a PR like this.

MartinBP:
> You conducted the rebase incorrectly. They shouldn't show up in a PR like this.

That's what I thought as well but I didn't know how to fix it in this pull request, so I deleted the branch and recreated it.
Is there anyway that I can add this new branch to this pull request or should I create a new?
The name of the branch is the same as the one I just deleted.

jonesmz:
You should be able to force-push to github to overwrite the branch on github with the one on your local disk.

MartinBP:
> You should be able to force-push to github to overwrite the branch on github with the one on your local disk.

I tried that but it will not let me reopen this pull request because "The windows_case_sensitive_folders branch was force-pushed or recreated". Sorry for mess...
I have created a new pull request.

PR #1887, opened 2020-11-28
Add FreeBSD support to GetProcessorCount

mjguzik:


src/util.cc
```diff
@@ -485,10 +489,17 @@ int GetProcessorCount() {
 #ifdef _WIN32
   return GetActiveProcessorCount(ALL_PROCESSOR_GROUPS);
 #else
-#ifdef CPU_COUNT
   // The number of exposed processors might not represent the actual number of
   // processors threads can run on. This happens when a CPU set limitation is
   // active, see https://github.com/ninja-build/ninja/issues/1278
+#ifdef __FreeBSD__
```
jhasse, line 495: `#if defined(__FreeBSD__)`

PR #1891, opened 2020-12-02
clang-tidy: add const to several global variables

neheb:
Found with cppcoreguidelines-avoid-non-const-global-variables

### DIFF
--- a/src/metrics.cc
+++ b/src/metrics.cc
@@ -76,9 +76,8 @@ int64_t TimerToMicros(int64_t dt) {
 
 }  // anonymous namespace
 
-
-ScopedMetric::ScopedMetric(Metric* metric) {
-  metric_ = metric;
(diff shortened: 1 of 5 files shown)


src/metrics.cc
```diff
@@ -76,9 +76,8 @@ int64_t TimerToMicros(int64_t dt) {
 
 }  // anonymous namespace
 
-
-ScopedMetric::ScopedMetric(Metric* metric) {
-  metric_ = metric;
+ScopedMetric::ScopedMetric(const Metric* metric) {
+  metric_ = const_cast<Metric*>(metric);
```
jhasse, line 80: I don't think `const_cast` is a good idea here.

jhasse (changes requested):
I'm not sure this is worth it.

PR #1893, opened 2020-12-08
Added fuzzer with build script and seed

AdamKorcz:
This PR adds a fuzzer with a sample build file to use as seed and a build script to integrate Ninja into OSS-fuzz.

I will shortly be setting up the integration on the OSS-fuzz side. Once the files are merged here, Ninja is ready to be integrated into OSS-fuzz which will allow the fuzzer in this fuzzer and future fuzzers to be run continuously on the OSS-fuzz platform. If bugs are found from the fuzzers, maintainers will be notified with an email containing a link to a detailed bug report with stacktrace and reproducible test case.

The fuzzer does expose a memory leak. To me it looks like everything that should be freed in the fuzzer gets freed. In case I am wrong I appreciate feedback.


jhasse:
Did you know about the [existing fuzzing instructions](https://github.com/ninja-build/ninja/wiki#using-afl-fuzz)?

AdamKorcz:
Yep and they are great! The benefit of OSS-Fuzz is that ninja will be fuzzed with 3 different fuzzers (libFuzzer, AFL and Hongfuzz). The fuzzing will also be performed in a continuous manner so the fuzzing corpus is accumulated a lot over time, meaning the fuzzer will have much more opportunity to get deeper into the code. I can also set it up such that the fuzzing runs as integrated into the CI, which is very useful in terms of catching bugs in new code.

jhasse:
I see.

build.sh and sample_ninja_build should include "fuzz" or something similar in their name. Or maybe moved into misc/oss-fuzz?

Issue #1682, opened 2019-11-14
Support quite rules


kostja:
Add an option to execute a rule quietly please.
```
rule mode_list
    command = echo debug release dev
    description =
build mode_list: mode_list
```
```
 % ninja mode_list
[1/1] echo debug release dev
debug release dev
```

chen3feng:
I also think this feature is useful, is a PR welcome?

Issue #1537, opened 2019-03-12
Is it possible to leave stderr messages from the compiler on stderr and not redirect them to stdout?


tpimh:
I understand that default ninja behavior is to only print to stdout, but I wonder if it's possible to implement this as an option like done in this branch: https://github.com/dendy/ninja/tree/stderr

jhasse:
What do you need it for?

tpimh:
I would like to separate build output into too files: `build.log` for informational output and `build.err` for warnings and errors. Redirecting stdout to one file and stderr to another works for GNU Make, but not for Ninja.

jhasse:
You won't know which warning or error belongs to which build output line. What's the reason behind the two files?

tpimh:
Mostly just warning detection. If there are warnings, they should be either fixed or disabled. The warning usually contains source filename and line number, so there is no need to find corresponding lines of ninja's own output.

jhasse:
I see. Would #480 be enough for your use-case?

The reason I'm asking is: We're trying to keep the number of options as low as possible.

tpimh:
I'm not sure how could possibly be the solution for my problem. Probably there may be a very hacky solution with NINJA_STATUS variable, but I can't think of one right now.
On the other hand if split stderr is implemented, `--quiet` that disables all ninja's own output and only prints compiler output would be as easy as `ninja --split-sderr >/dev/null`.

jhasse:
> I'm not sure how that could possibly be the solution for my problem.

You would run `ninja --quiet` and could detect all the warnings. (I'm aware that `--quiet` isn't implemented yet)

tpimh:
But in this case I will not get any ninja's own output. I would like to keep both, but split them.

jhasse:
Why?

tpimh:
In current build script I print a dot for every line on stdout and an x for every line on stderr. Saving both stdout and stderr into separate files. This keeps the CI from stopping because of the lack of output, gives a pretty visual output that helps to detect new warnings and errors and provides separate readable build logs with warning descriptions. This works for other build systems, but not for ninja as there is no way to tell if a line came from stderr or stdout. This is how I did it: https://github.com/tpimh/nenuzhnix-build/commit/722ece30d2b3801b7871ae091223e261c38e671d (still to be rewritten properly)

jhasse:
[...]

PR #1898, opened 2020-12-18
Add --quiet and -q flags

xyproto:
* Add a `--quiet` and a `-q` flag.
* Also respect the `QUIET` environment variable, if it's set to `1`.
* Makes ninja not print status updates, `Entering directory` or `no work to do`.

See also: #420, #480, #1537, #1818 and #1682


megabyde:
Any update on this? We have big builds and the CI logs are monstrous with the default mode.

jhasse:
Why did you create a new PR when there are already so many existing ones?

We didn't want `-q` as that means `--question` in Make and might confuse some people. We also don't want to introduce a new envirnoment variable for this, the flag is enough.

xyproto:
@jhasse

The two previous pull requests both have unresolved issues and/or are abandoned. #420 is from September 2012.

`--ask` is often used to avoid naming the flag `--question`. I think `--question` should rather be `--ask`.

`--quiet` in combination with `-q` are very common.

jhasse:
> I think --question should rather be --ask.

I agree, but we can't change that now.

> --quiet in combination with -q are very common.

Yes, but not with Make. And many Ninja users are coming from Make, that's why I think we shouldn't add the short option.

xyproto:
I see that `--question` is used by make, but I don't see why an unconventional use of a flag by `make` should be relevant, as long as `ninja` is not a drop-in replacement for `make`.

Additionally, many developers use `ninja` exclusively via `cmake`.

While many Ninja users may be used to `make`, I like to think they are able to accept and understand that different utilities take different flags.

megabyde:
What about `-s, --silent, --quiet` from `make`?

xyproto:
I believe `--quiet` is much, **much** more common than `--silent`. Here's a data-point from my current system (Arch Linux, various packages installed):

    grep -a -c -l "\-\-silent" /usr/bin/* 2>/dev/null | wc -l
    91
    
    grep -a -c -l "\-\-quiet" /usr/bin/* 2>/dev/null | wc -l
    469
    
While this is not an exact science, I would expect you to get similar results for your systems.

I really believe `make` is the odd one out here, in terms of flag usage, and not a program to emulate.

jhasse:
> I don't see why an unconventional use of a flag by make should be relevant, as long as ninja is not a drop-in replacement for make.

Even if Ninja is not a drop-in replacement for Make, Make is still a lot more important to Ninja than other command line tools.

In general I'm not a fan of one letter options - better be explicit.

megabyde:
What about using `--silent`?

jhasse:
What's wrong with `--quiet` only?

xyproto:
* Both `-q` and `--quiet` are very conventional.
* `ninja` is not `make` and already does not behave exactly like `make` (and why should it?)

I think this patch would be beneficial to Ninja like it is.

xyproto:
Is there any chance that this pull request will be merged, or should I just close it?

jhasse:
Closing in favor of #1968.

PR #1899, opened 2020-12-18
Status changes to support frontends

colincross:
This pull request takes the less controversial status refactoring changes from #1210 as an incremental step towards supporting frontends.  It moves the status printing code into a separate file and calls it through an interface to allow multiple implementations.


src/status.cc
```diff
@@ -0,0 +1,266 @@
...
+
+  if (edge->use_console() || printer_.is_smart_terminal())
+    PrintStatus(edge, start_time_millis);
+
+  if (edge->use_console())
+    printer_.SetConsoleLocked(true);
+}
+
+void StatusPrinter::BuildEdgeFinished(Edge* edge, int64_t end_time_millis,
+                                      bool success, const string& output) {
```
jhasse, line 62: In your original implementation you've used `const CommandRunner::Result*` here. I guess you've changed it since you only need `success` and `output`? What about exitcodes other than 0 (success == true) or 1 (success == false)?

src/ninja.cc
```diff
@@ -1366,9 +1373,10 @@ NORETURN void real_main(int argc, char** argv) {
     // Don't print this if a tool is being used, so that tool output
     // can be piped into a file without this string showing up.
     if (!options.tool)
-      printf("ninja: Entering directory `%s'\n", options.working_dir);
+      Info("Entering directory `%s'", options.working_dir);
     if (chdir(options.working_dir) < 0) {
       Fatal("chdir to '%s' - %s", options.working_dir, strerror(errno));
+      exit(1);
```
jhasse, line 1379: `Fatal` already calls `exit(1)`.

src/ninja.cc
```diff
@@ -1366,9 +1373,10 @@ NORETURN void real_main(int argc, char** argv) {
     // Don't print this if a tool is being used, so that tool output
     // can be piped into a file without this string showing up.
     if (!options.tool)
-      printf("ninja: Entering directory `%s'\n", options.working_dir);
+      Info("Entering directory `%s'", options.working_dir);
```
jhasse, line 1376: `status->Info`?

src/status.cc
colincross: That was part of a later patch in the series, this is a pure move of the code from build.cc to status.cc.

src/ninja.cc
colincross: Fixed.

src/ninja.cc
colincross: Fixed.

src/status.cc
jhasse: Ah my bad, thanks.

jdrouhard:
FYI, this does break pipelines that previously relied on all output being printed on stdout. This changes error messages when a subcommand fails to print to stderr now, which could cause issues with processes that read both stdout and stderr and previously relied on errors being serialized with normal output.

Actual subcommands' stderrs are still printed to stdout. It's only ninja's own error messages that are now stderr.

Was this intentional? @colincross

jhasse:
Thanks for bringing this up! I restored the old behavior and added a test so that we don't regress.

Issue #1882, opened 2020-11-19
Build fails when using case sensitive folders on Windows


MartinBP:
Ninja is unable to find files in folders using upper case letters when case sensitive folders are enabled.

Issue can be reproduced by setting up the following folder structure and using the build.ninja file below.
[test_ninja.zip](https://github.com/ninja-build/ninja/files/5569895/test_ninja.zip)

```
`-- test_ninja
     |-- bar
     |    `-- hello.c
     |-- Foo
     |    `-- hello.c
```

test_ninja must be a case sensitive folder.
`fsutil.exe file setCaseSensitiveInfo .\test_ninja\ enable`

```
# build.ninja
cc     = gcc

rule compile
  command = $cc -c $in -o $out

rule link
  command = $cc $in -o $out

#Example 1: Using incorrect casing for 'foo'. Ninja correctly reports an error
#build hello.o: compile foo/hello.c
#Output => ninja: error: 'foo/hello.c', needed by 'hello.o', missing and no known rule to make it

#Example 2: Using correct casing for 'Foo'. Ninja reports the same error, but it should be able to find it.
#build hello.o: compile Foo/hello.c
#Output => ninja: error: 'Foo/hello.c', needed by 'hello.o', missing and no known rule to make it

#Example 3: Using incorrect casing for 'Bar'. Ninja reports the expected error "No such file or directory"
#build hello.o: compile Bar/hello.c
#[1/2] gcc -c Bar/hello.c -o hello.o
#FAILED: hello.o
#gcc -c Bar/hello.c -o hello.o
#gcc: error: Bar/hello.c: No such file or directory
#gcc: fatal error: no input files
#compilation terminated.
#ninja: build stopped: subcommand failed.

#Example 4: Using correct casing for 'bar'. Everything is lower case and so it works
#build hello.o: compile bar/hello.c

build hello: link hello.o

default hello
```

I would expect Example 2 above to work.
Changing the following function RealDiskInterface::Stat in disk_interface.cc to not perform a transform to lower case makes it work as expected.

```
TimeStamp RealDiskInterface::Stat(const string& path, string* err) const {
....

**Removing the two lines below fixes the issue**
  //transform(dir.begin(), dir.end(), dir.begin(), ::tolower);
  //transform(base.begin(), base.end(), base.begin(), ::tolower);

....
}
```

jonesmz:
Does it work if you apply this patch?
https://github.com/ninja-build/ninja/pull/1671

[...]

PR #1901, opened 2021-01-03
disk_interface: Improve the stat cache handling for case sensitive folders on Windows

MartinBP:
disk_interface: Improve the stat cache handling for case sensitive folders on Windows

The path used as argument to FindFirstFileExA is no longer being converted to lowercase.
By not converting the path to lower case, case sensitive folders will now be handled correctly.
Case insensitive folders still works as expected because casing doesn't matter on those.
All entries in the stat cache remains lowercase to avoid potential cache misses.

### DIFF
--- a/src/disk_interface.cc
+++ b/src/disk_interface.cc
@@ -180,12 +180,13 @@ TimeStamp RealDiskInterface::Stat(const string& path, string* err) const {
     dir = path;
   }
 
-  transform(dir.begin(), dir.end(), dir.begin(), ::tolower);
+  string dir_lowercase = dir;
+  transform(dir.begin(), dir.end(), dir_lowercase.begin(), ::tolower);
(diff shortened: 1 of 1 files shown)


Bagira80:
Is this PR the fix for #1882?

Is there something missing which prevents it to be merged and released? (I am hitting the same problem as explained in #1882.)

MartinBP:
> Is this PR the fix for #1882?
> 
> Is there something missing which prevents it to be merged and released? (I am hitting the same problem as explained in #1882.)

Yes this is a fix for #1882. 

I haven't worked on this since i did the PR and I'm not aware of anything missing except for the thing I mentioned in https://github.com/ninja-build/ninja/issues/1882#issuecomment-732379254
This is however a very special case that I can live with and I do not believe it will affect that many (if any at all).

PR #1905, opened 2021-01-25
Add Haiku support.

korli:
Add include of sys/select.h to get FD_* definitions.

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -625,6 +625,10 @@ double GetLoadAverage() {
     return -0.0f;
   return 1.0 / (1 << SI_LOAD_SHIFT) * si.loads[0];
 }
+#elif defined(__HAIKU__)
+double GetLoadAverage() {
+    return -0.0f;
+}
 #else
 double GetLoadAverage() {
   double loadavg[3] = { 0.0f, 0.0f, 0.0f };


jonesmz:
Is the modification to configure.py useful or necessary?

jonesmz:
No cmake changes needed?

korli:
the modification in configure.py is needed for --platform: https://github.com/ninja-build/ninja/blob/a30ccaef7314f5aa858ba3167af08be4b518d960/configure.py#L208

korli:
> No cmake changes needed?

AFAIK we don't build ninja with cmake on Haiku. Should we?

jonesmz:
cmake is preferred over configure.py. No one should be using configure.py for anything.

korli:
> cmake is preferred over configure.py. No one should be using configure.py for anything.

Well, I saw ArchLinux using configure.py :)

I tested with cmake and it builds out of the box, nice. Should I clean up the pull request, removing configure.py changes?

jonesmz:
Personally i think you should, but I am not a Ninja maintainer, so do what you think is best.

src/subprocess-posix.cc
```diff
@@ -21,6 +21,7 @@
 #include <unistd.h>
 #include <stdio.h>
 #include <string.h>
+#include <sys/select.h>
```
jhasse, line 24: why?

korli: to get FD_* definitions from sys/select.h

https://github.com/ninja-build/ninja/blob/master/src/subprocess-posix.cc#L334
https://pubs.opengroup.org/onlinepubs/007904975/basedefs/sys/select.h.html

korli: I checked again, and it seems to work fine without it. Must be a restover. It's now removed.

PR #1907, opened 2021-01-29
Clarify purpose for implicit dependencies

aDotInTheVoid:
This confused me for a bit, so I've tried to clarify it.

In some cases the script would want to be in `$in`, eg
```ninja
rule runscript
  command = sh $in
```

However if the script reads a file, and which file it reads is encoded in the script itself, it's clear the file will never show up in the arguments. 

Another example would be something like
```ninja
build ninja: link $builddir/ninja.o | $builddir/libninja.a
  libs = -lninja
```
but that requires knowing how linker arguments work, which is probably more confusing.

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -949,8 +949,9 @@ source file of a compile command.
 +
 This is for expressing dependencies that don't show up on the
 command line of the command; for example, for a rule that runs a
-script, the script itself should be an implicit dependency, as
-changes to the script should cause the output to rebuild.
+script that reads a hardcoded file, the hardcoded file should 
+be an implicit dependency, as changes to the file should cause 
+the output to rebuild, even though it doesn't show up in the arguments.
 +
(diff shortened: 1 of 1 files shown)


jhasse:
the script itself should still be an implicit dependy, shouldn't it?

aDotInTheVoid:
This is my understanding is that depending on how you set up the rule to run the script, it should either be implicit or explicit.

Eg with
```ninja
rule runscript
  command = sh script.sh
```
you dont need `script.sh` in `$in`, so you should use an implicit dependency to invoke this rule


But in
```ninja
rule runscript
  command = sh $in
```
You do want the script to be in `$in` so it should be explicit dependencys.

This (for my) is what made the docs confuzing, and it is possible that my understanding is wrong.

jhasse:
I think you're correct.

jhasse:
Your example is better :) Thanks for improving the docs!

PR #1910, opened 2021-02-09
CMake: add feature_summary()

eszlari:


jhasse:
Thanks, but see https://github.com/ninja-build/ninja/pull/1770#issuecomment-713011149.

eszlari:
What about the other commit?

jhasse:
I'm not a CMake expert, but it looks cleaner to me. On the other hand: If it ain't broken, don't fix it ;)

If you create a new PR with those changes, I would wait for feedback from some others who know more about CMake than me.

I wouldn't touch the AIX code though, as that's completely untested by CI.

PR #1911, opened 2021-02-09
cmake: add_compile_options / PROJECT_SOURCE_DIR

eszlari:


scivision:
these are all good modernizations of the CMake syntax and per recommended practices

Issue #1886, opened 2020-11-25
ninja -t cleandead should consider not failing on failure to remove


mathstuf:
The cleandead tool is an assistance, and as such should not trigger failures for "unimportant" problems. One of which is failing to delete one of these files. This was reported to [CMake's Discourse](https://discourse.cmake.org/t/automatic-call-of-cleandead-ninja-tool/2180) but belongs here.

jonesmz:
If the tools whole purpose for existence is to delete file(sl, and the file cannot be deleted, then returning non-success is the correct action.

mathstuf:
The purpose (AFAIK) is to help remove files which used to be part of the build, but are no longer in it (e.g., you've removed a source file, so its object file is now no longer needed). It's a utility tool. If you clean the vacuum the deps log and then do cleandead, it will leave files behind since ninja has removed its knowledge of them. It's not a perfect tool, but an assistant. If it gets in the way like this (which is obviously "something weird going on"), just ignoring the issue (or noting it in output with `-v`) should be sufficient.

jonesmz:
Ok, then I suggest you edit the first comment on this issue to more clearly describe the problem.

A link to a third party resource isn't reliable, as it might be removed or altered.

trmagma:
I am the poster of the problem at cmake-discourse.
One source of the problem are version controlled files, which are copied from source to build directory. In some SCMs ( e.g. like clearcase or perforce) all files are read only (in source directory) until explicit checked out. A simple copy will just keep the permissions. I assume, even if the read only state was explicit set during the copy ninja will fail (not tested).

For this simple case it would be sufficient if ninja tries harder to delete the file by trying to remove the read only permission (similar like the -f option of the  rm shell command).  If this also failed, this might be an error.  
 
CMake seems to call ninja -t cleandead  automatically after generating the ninja build scripts (which would do there job perfectly) but fails, cause the ninja call failed. Think this call of cmake is perfectly right,  cause it must be done after the new build scripts are created, but before the next build.

jhasse:
> If you clean the vacuum the deps log [...]

What do you mean by that?

> For this simple case it would be sufficient if ninja tries harder to delete the file by trying to remove the read only permission (similar like the -f option of the rm shell command). If this also failed, this might be an error.

Hm ... not sure if this should be the default behavior though.

Why not remove the read-only permission when copying the files to the build directory?

mathstuf:
[...]

PR #1913, opened 2021-02-12
Restore toleration of missing to-be-deleted files on Windows

bradking:
Since #1892 we use `DeleteFile` without checking if it failed due to a path missing.  Fix the `GetFileAttributes` logic to check for multiple variants of "file not found" as we do in `StatSingleFile`.  ~~Then fall back to the original `remove()` logic that already checks for `ENOENT` on failure.~~  Add the check after `DeleteFile` too.

Issue: #1886

### DIFF
--- a/src/disk_interface.cc
+++ b/src/disk_interface.cc
@@ -272,9 +272,9 @@ int RealDiskInterface::RemoveFile(const string& path) {
     return 1;
   }
   if (attributes & FILE_ATTRIBUTE_READONLY) {
-    // On non-Windows systems remove will happily delete read-only files. On
-    // Windows Ninja should behave the same. See
-    // https://github.com/ninja-build/ninja/issues/1886
+    // On non-Windows systems, remove() will happily delete read-only files.
+    // On Windows Ninja should behave the same:
+    //   https://github.com/ninja-build/ninja/issues/1886
     SetFileAttributes(path.c_str(), attributes & ~FILE_ATTRIBUTE_READONLY);
   }
   if (!DeleteFile(path.c_str())) {
--- a/src/disk_interface.cc
+++ b/src/disk_interface.cc
@@ -268,16 +268,23 @@ int RealDiskInterface::RemoveFile(const string& path) {
 #ifdef _WIN32
   DWORD attributes = GetFileAttributes(path.c_str());
   if (attributes == INVALID_FILE_ATTRIBUTES) {
-    if (GetLastError() == ERROR_FILE_NOT_FOUND) {
+    DWORD win_err = GetLastError();
+    if (win_err == ERROR_FILE_NOT_FOUND || win_err == ERROR_PATH_NOT_FOUND) {
       return 1;
     }
   } else if (attributes & FILE_ATTRIBUTE_READONLY) {
     // On non-Windows systems, remove() will happily delete read-only files.
     // On Windows Ninja should behave the same:
     //   https://github.com/ninja-build/ninja/issues/1886
+    // Skip error checking.  If this fails, accept whatever happens below.
     SetFileAttributes(path.c_str(), attributes & ~FILE_ATTRIBUTE_READONLY);
   }
   if (!DeleteFile(path.c_str())) {
+    DWORD win_err = GetLastError();
+    if (win_err == ERROR_FILE_NOT_FOUND || win_err == ERROR_PATH_NOT_FOUND) {
+      return 1;
+    }
+    // Report as remove(), not DeleteFile(), for cross-platform consistency.
     Error("remove(%s): %s", path.c_str(), GetLastErrorString().c_str());
     return -1;
   }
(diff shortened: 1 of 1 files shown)


bradking:
I run nightly testing of CMake with a nightly build of `ninja`.  After #1892, I started noticing failures with output from ninja:

```
Cleaning... ninja: error: remove(...): The system cannot find the path specified.
```

bradking:
Cc: @jhasse

jhasse:
Thx!

You've also added the A suffix to the winapi functions, any reason for that?

bradking:
The `A` suffix on winapi functions is already used elsewhere in the same source file.  I added it for consistency.  If we add unicode support in the future, they should all be converted to `W` suffixes together.

src/disk_interface.cc
```diff
@@ -266,22 +266,19 @@ FileReader::Status RealDiskInterface::ReadFile(const string& path,
...
+  if (attributes == INVALID_FILE_ATTRIBUTES) {
+    DWORD win_err = GetLastError();
+    if (win_err == ERROR_FILE_NOT_FOUND || win_err == ERROR_PATH_NOT_FOUND) {
+      return 1;
+    }
+  } else if (attributes & FILE_ATTRIBUTE_READONLY) {
+    // On non-Windows systems remove() will happily delete read-only files.
+    // On Windows Ninja should behave the same.
+    // See https://github.com/ninja-build/ninja/issues/1886
+    SetFileAttributesA(path.c_str(), attributes & ~FILE_ATTRIBUTE_READONLY);
```
mathstuf, line 279: If this fails, are we just deferring the error to the actual `DeleteFile` call? Might want to mention the explicit lack of error checking if that's the case.

jhasse: Yes. Note that this was the case before, nothing really to do with this commit (that's why I dislike the added A, because it creates a bigger diff than necessary).

jhasse:
> Then fall back to the original remove() logic that already checks for ENOENT on failure.

Is this really needed? Couldn't this patch be broken down to just adding ` || win_err == ERROR_PATH_NOT_FOUND`?

bradking:
I was trying to minimize the net change from before #1892 rather than a current net change, since the older code has been deployed and working for years.   Try

```
git diff 2d7f7e55~1 HEAD -- src/disk_interface.cc
```

If you want to switch to `DeleteFileA`  then I can re-work the change.

src/disk_interface.cc
bradking: I updated the branch to make the changes as very granular commits.

bradking: I added an explicit comment about the lack of error checking on `SetFileAttributesA`.

src/disk_interface.cc
```diff
@@ -266,22 +266,16 @@ FileReader::Status RealDiskInterface::ReadFile(const string& path,
...
-  }
-  if (attributes & FILE_ATTRIBUTE_READONLY) {
-    // On non-Windows systems remove will happily delete read-only files. On
-    // Windows Ninja should behave the same. See
-    // https://github.com/ninja-build/ninja/issues/1886
-    SetFileAttributes(path.c_str(), attributes & ~FILE_ATTRIBUTE_READONLY);
-  }
-  if (!DeleteFile(path.c_str())) {
-    Error("remove(%s): %s", path.c_str(), GetLastErrorString().c_str());
-    return -1;
```
bradking, line 282: The `DeleteFile` code path is removed in favor of `remove()` by the last commit of this PR.  The idea is to minimize the net change from the code prior to #1892.  If we want to keep `DeleteFile` on Windows, we can just drop that commit.  I don't think `DeleteFile` is semantically much different than `remove()` on Windows.  It's not synchronous anyway.

bradking:
@jhasse something is wrong with the macOS builder in CI.

jhasse:
I'm not a fan of the A/W suffix. When we switch to UNICODE this will just create an unnecessary huge diff. `std::wstring` has a `c_str()` function, too, so that line could eventually be untouched. That's why I didn't add it in #1892.

The reason to catch an acceptable error directly after `GetFileAttributes` was to avoid the subsequent calls.

I also deliberately used `DeleteFile` instead of `remove`, because it has better Unicode support. Otherwise we would have to `#ifdef`on Windows anyway and use `_wremove` and `_wcserror` instead.

> @jhasse something is wrong with the macOS builder in CI.

Seems to be working now.

jhasse (changes requested):
I kept the `remove(%s)` wording as people might have grep'ed for it.

Can you keep the DeleteFile code path?

jhasse (approved):

bradking:
Thanks for explaining your reasoning.  I added a corresponding comment about `remove()`.

jhasse:
Thank you for the PR!

jhasse:
I've just discovered that Windows now supports UTF-8 and we might not need `UNICODE` after all ...
#1915

If that really works, using `remove` sounds like the better idea. Sorry for the bad timing. If you find another issue with `DeleteFile`, feel free to to create a PR with your idea of using `remove` on Windows, too.

PR #1915, opened 2021-02-17
Use UTF-8 on Windows 10 Version 1903, fix #1195

### LINKED ISSUE #1195: Unicode support on Windows
Currently Ninja uses ANSI codepage and ANSI WinAPI functions on Windows. This isn't good because that doesn't allow to use filenames and environment variables outside of ANSI codepage.

Ninja should use Wide WinAPI functions everywhere and then decode/encode them to/from UTF-8 internally.

Also currently `ninja` rule files are expected to be in ANSI codepage which isn't portable between different Windows machines since they can use different ANSI codepages. Solution would be to use UTF-8 encoding for ninja files. I guess to be backward compatible there will be needed some flag to enable UTF-8 files.

Incomplete example patch

``` diff
diff --git a/src/subprocess-win32.cc b/src/subprocess-win32.cc
index 4bab719..82ec616 100644
--- a/src/subprocess-win32.cc
+++ b/src/subprocess-win32.cc
@@ -73,8 +73,12 @@ HANDLE Subprocess::SetupPipe(HANDLE ioport) {
 }

 bool Subprocess::Start(SubprocessSet* set, const string& command) {
-  HANDLE child_pipe = SetupPipe(set->ioport_);
+  wstring wideCommand;
+  if (!UTF8ToWide(command.c_str(), wideCommand)) {
+    Fatal("Subprocess::Start: Failed to encode command string to Wide string");
+  }

+  HANDLE child_pipe = SetupPipe(set->ioport_);
   SECURITY_ATTRIBUTES security_attributes;
   memset(&security_attributes, 0, sizeof(SECURITY_ATTRIBUTES));
   security_attributes.nLength = sizeof(SECURITY_ATTRIBUTES);
@@ -86,7 +90,7 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
   if (nul == INVALID_HANDLE_VALUE)
     Fatal("couldn't open nul");

-  STARTUPINFOA startup_info;
+  STARTUPINFOW startup_info;
   memset(&startup_info, 0, sizeof(startup_info));
   startup_info.cb = sizeof(STARTUPINFO);
   if (!use_console_) {
@@ -106,7 +110,7 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {

   // Do not prepend 'cmd /c' on Windows, this breaks command
   // lines greater than 8,191 chars.
-  if (!CreateProcessA(NULL, (char*)command.c_str(), NULL, NULL,
+  if (!CreateProcessW(NULL, (LPWSTR)wideCommand.c_str(), NULL, NULL,
                       /* inherit handles */ TRUE, process_flags,
                       NULL, NULL,
                       &startup_info, &process_info)) {
diff --git a/src/util.cc b/src/util.cc
index e31fd1f..f40b0ab 100644
--- a/src/util.cc
+++ b/src/util.cc
@@ -90,6 +90,21 @@ void Error(const char* msg, ...) {
   fprintf(stderr, "\n");
 }

+#ifdef _WIN32
+bool UTF8ToWide(string utf8, wstring &wide) {
+  bool success = false;
+  const int wlength = MultiByteToWideChar(CP_UTF8, 0, utf8.c_str(), int(utf8.size()), NULL, 0);
+  wchar_t* wdata = new wchar_t[wlength];
+  int r = MultiByteToWideChar(CP_UTF8, 0, utf8.c_str(), int(utf8.size()), wdata, wlength);
+  if (r > 0) {
+    wide = wstring(wdata, wlength);
+    success = true;
+  }
+  delete[] wdata;
+  return success;
+}
+#endif
+
 bool CanonicalizePath(string* path, unsigned int* slash_bits, string* err) {
   METRIC_RECORD("canonicalize str");
   size_t len = path->size();
diff --git a/src/util.h b/src/util.h
index cbdc1a6..da6bbcd 100644
--- a/src/util.h
+++ b/src/util.h
@@ -99,6 +99,9 @@ bool Truncate(const string& path, size_t size, string* err);
 #endif

 #ifdef _WIN32
+/// Encode UTF8 string to Wide string
+bool UTF8ToWide(string utf8, wstring &wide);
+
 /// Convert the value returned by GetLastError() into a string.
 string GetLastErrorString();
```

jhasse:
Allows Ninja to use descriptions, filenames and environment variables with characters outside of the ANSI codepage on Windows. Build manifests are now UTF-8 by default (this change needs to be emphasized in the release notes).

WriteConsoleOutput doesn't support UTF-8, but it's deprecated on newer Windows 10 versions anyway (or as Microsoft likes to put it: "no longer a part of our ecosystem roadmap"). We'll use the VT100 sequence just as we do on Linux and macOS.

https://docs.microsoft.com/en-us/windows/uwp/design/globalizing/use-utf8-code-page
https://docs.microsoft.com/en-us/windows/console/writeconsoleoutput
https://docs.microsoft.com/de-de/windows/console/console-virtual-terminal-sequences

Fixes #1195 (I hope).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -137,6 +137,10 @@ endif()
 add_executable(ninja src/ninja.cc)
 target_link_libraries(ninja PRIVATE libninja libninja-re2c)
 
+if(WIN32)
+  target_sources(ninja PRIVATE windows/ninja.manifest)
+endif()
+
 # Adds browse mode into the ninja binary if it's supported by the host platform.
 if(platform_supports_ninja_browse)
 	# Inlines src/browse.py into the browse_py.h header, so that it can be included
--- a/src/line_printer.cc
+++ b/src/line_printer.cc
@@ -87,22 +87,27 @@ void LinePrinter::Print(string to_print, LineType type) {
     GetConsoleScreenBufferInfo(console_, &csbi);
 
     to_print = ElideMiddle(to_print, static_cast<size_t>(csbi.dwSize.X));
-    // We don't want to have the cursor spamming back and forth, so instead of
-    // printf use WriteConsoleOutput which updates the contents of the buffer,
-    // but doesn't move the cursor position.
-    COORD buf_size = { csbi.dwSize.X, 1 };
-    COORD zero_zero = { 0, 0 };
-    SMALL_RECT target = {
-      csbi.dwCursorPosition.X, csbi.dwCursorPosition.Y,
-      static_cast<SHORT>(csbi.dwCursorPosition.X + csbi.dwSize.X - 1),
-      csbi.dwCursorPosition.Y
-    };
-    vector<CHAR_INFO> char_data(csbi.dwSize.X);
-    for (size_t i = 0; i < static_cast<size_t>(csbi.dwSize.X); ++i) {
-      char_data[i].Char.AsciiChar = i < to_print.size() ? to_print[i] : ' ';
-      char_data[i].Attributes = csbi.wAttributes;
+    if (supports_color_) {  // this means ENABLE_VIRTUAL_TERMINAL_PROCESSING
+                            // succeeded
+      printf("%s\x1B[K", to_print.c_str());  // Clear to end of line.
+      fflush(stdout);
+    } else {
+      // We don't want to have the cursor spamming back and forth, so instead of
+      // printf use WriteConsoleOutput which updates the contents of the buffer,
+      // but doesn't move the cursor position.
+      COORD buf_size = { csbi.dwSize.X, 1 };
+      COORD zero_zero = { 0, 0 };
+      SMALL_RECT target = { csbi.dwCursorPosition.X, csbi.dwCursorPosition.Y,
+                            static_cast<SHORT>(csbi.dwCursorPosition.X +
+                                               csbi.dwSize.X - 1),
+                            csbi.dwCursorPosition.Y };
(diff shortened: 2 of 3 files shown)


MrTrillian (approved):

MrTrillian:
@jhasse I'm happy to see Ninja supporting Unicode, but I want to point out consequences to this change that should be taken into account:

This is a breaking change for ninja build files. Since Ninja feeds the raw bytes from those files into -A Win32 functions, this means that Ninja build files will now have to be encoded as utf-8, not ANSI (which is good, but a breaking change). In particular, cmake will have to be updated to not generate Ninja build files as ANSI.

There might also be repercussions with the "include prefix" feature. Ninja does binary comparison of the string coming from its build file and that coming from the `cl.exe` process output. Previously the first of these strings was ANSI and now it will be UTF-8, so this might cause mismatches depending on the encoding that `cl.exe` uses.

jhasse:
Indeed, thanks for pointing that out.

Note that this will only result in issues when there are non-ASCII characters in the build manifest. Something that would have resulted in problems on Windows anyway so I doubt that many people relied on that. It will still be the first point in the release notes ;)

bradking:
I've opened [CMake Issue 21866](https://gitlab.kitware.com/cmake/cmake/-/issues/21866) for this, thanks.

bradking:
#1918 proposes an additional tool to help generators determine the correct encoding for `build.ninja` files.

rllane17 (commented):
approve

Issue #1897, opened 2020-12-16
1.10.2 fails to build on mingw


ilg-ul:
I'm maintaining the [xPack Ninja Build](https://github.com/xpack-dev-tools/ninja-build-xpack/releases), a cross-platform binary distribution of Ninja, and when I tried to build 1.10.2, the build failed while compiling `browse.cc`, which uses `pipe()` and `fork()`, not available on mingw-w64.

As a quick workaround I patched `CMakeLists.txt` to disable browser support:

```
--- CMakeLists-orig.txt	2020-12-16 17:08:37.000000000 +0200
+++ CMakeLists.txt	2020-12-16 17:24:52.000000000 +0200
@@ -71,7 +71,11 @@
 	set(${RESULT} "${PLATFORM_HAS_UNISTD_HEADER}" PARENT_SCOPE)
 endfunction()
 
-check_platform_supports_browse_mode(platform_supports_ninja_browse)
+if(WIN32)
+  set(platform_supports_ninja_browse FALSE)
+else()
+  check_platform_supports_browse_mode(platform_supports_ninja_browse)
+endif()
 
 # Core source files all build into ninja library.
 add_library(libninja OBJECT
```

I'm not very familiar with the Ninja internals, and I don't know if this is an acceptable solution or not, so, on short term, I would appreciate an advice if I can proceed with my 1.10.2 release, and, on long term, an official solution.

For the records, the issue was introduced with 1.10.2, since both 1.10.0 and 1.10.1 compiled just fine with mingw.

ilg-ul:
Is it ok to disable building `browse.cc`? I would like to know if I can proceed with my 1.10.2 release using this workaround.

Thank you.

scivision:
Are you building from CMake directly e.g. `cmake -B build; cmake --build build`?  For me this works from MinGW64 via MSYS2 on both Ninja `master` and `v1.10.2` using GCC 10.2.0

jonesmz:
Have you considered adding a job to the github ci, in `.github/workflows/windows.yml` to let compiling under MinGW64 be tested regularly?

jonesmz:
If you wanted to add mingw support, it would be easy to add to this file,

https://github.com/ninja-build/ninja/blob/634682ef5d18a21837b68a152e41b6573d0e2e83/.github/workflows/windows.yml

 From this pull request: https://github.com/ninja-build/ninja/pull/2109

[...]

PR #1916, opened 2021-02-18
cmake: add browse check status

scivision:
This adds traditional CMake configure-time print outs to raise awareness if browse will be built or not. Re: #1897 

```
-- Checking if browse capable
-- Checking if browse capable - no
```


CMakeLists.txt
```diff
@@ -65,12 +67,19 @@ function(check_platform_supports_browse_mode RESULT)
 	if(NOT inline_result EQUAL "0")
 		# The inline script failed, so browse mode is not supported.
 		set(${RESULT} "0" PARENT_SCOPE)
+		message(STATUS "Checking if browse capable - no")
```
jhasse, line 70: I would just print a warning message here (maybe not for Windows) and drop all the other output.

jhasse (changes requested):

Issue #1251, opened 2017-02-24
mixing depfile absolute paths with manifest relative paths breaks rebuilds


bradking:
The ninja manual warns against such mixture, but it has become a serious limitation.

The problem is that people want their build systems to invoke the compiler with an absolute path to the source file.  This produces better output for IDEs to match error messages and refer back to the original files.  It also may affect the paths recorded in debug information.  However, this conflicts with ninja's preference for relative paths in the build manifest.

Here is a script demonstrating the issue: [ninja-bug.bash.txt](https://github.com/ninja-build/ninja/files/799752/ninja-bug.bash.txt)

bradking:
Cc: @nico @chaoren

Over in CMake, there is relevant discussion in [issue 16675](https://gitlab.kitware.com/cmake/cmake/issues/16675), [issue 13894](https://gitlab.kitware.com/cmake/cmake/issues/13894), and [MR 520](https://gitlab.kitware.com/cmake/cmake/merge_requests/520).

zlolik:
As workaround you can set object output directory explicitly, even "." or "..", but absolute path works also.
[ninjatst.bash.txt](https://github.com/ninja-build/ninja/files/822532/ninjatst.bash.txt)

bradking:
@zlolik IIUC you suggest placing absolute paths in the build manifest (`build.ninja`) graph nodes.  Certainly that works (and is what CMake does in out-of-source builds), but my point here is that Ninja discourages use of absolute paths and yet IDEs require them in output.

It is also hypothetically possible that a compiler given `-c foo.c` could generate an absolute path in the depfile anyway.  If that happens then Ninja will not recognize the dependency due to the same problem reported here.

Fixing this will require Ninja to be able to recognize when a relative path and absolute path are the same file and have only one graph node internally instead of two.  Doing this only for depfile processing may be sufficient, as it is reasonable to ask the `build.ninja` generators to name paths consistently within the build manifest.

zlolik:
@bradking First of all, I am agree with you that we have an issue with ninja and I only suggest workaround.
I think there are 3 linked issues:
1. Logic with input absolute/relative path, more details in #1152 .
2. Logic to force output absolute/relative path. The problem with compilers - the only way to said to compiler that you want to have absolute path in output is use absolute path in command line. Not ninja issue but ours.
3. Logic to parse depfile absolute/relative path. This issue.

I use 2 workarounds:
[...]

PR #1917, opened 2021-02-25
ninja_workdir: Add top-level binding to set working directory

bradking:
Add a binding that the build manifest generator can use to tell ninja the working directory to use for the build.  This allows a command like

```console
$ ninja -f /path/to/some/build.ninja
```

to work, regardless of the caller's working directory, if `build.ninja` has the binding.

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -828,6 +828,11 @@ Two variables are significant when declared in the outermost file scope.
   discussion of the build log>>.  (You can also store other build output
   in this directory.)
 
+`ninja_workdir`:: _(Available since Ninja 1.11.)_ If present, sets the
+  working directory in which the build will take place.  Otherwise, the
+  build will take place in the working directory in which `ninja` is
+  invoked, or that specified by `ninja -C`.
+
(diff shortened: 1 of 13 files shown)


bradking:
@jhasse while working on #1251 I realized that the proposed `ninja_workdir` setting is useful on its own.  This PR adds it.

The `Linux / test` job is failing in CI, but I don't see what this PR's changes have to do with it.

jhasse:
Search the log for `error:`.

jhasse:
Thanks for the PR btw! I'm still not 100% sure if this is the best approach.

bradking:
@jhasse thanks.  I updated the PR to fix the clang-tidy diagnostic.  CI is now clean.

jhasse:
The use-case in the initial comment isn't strong enough on its own IMHO.

bradking:
> The use-case in the initial comment isn't strong enough on its own IMHO.

Fair enough.  I'll continue discussion over in #1251.

bradking:
Closing for now.  As discussed above, this may not be worthwhile on its own.

PR #1920, opened 2021-03-01
doc: clarify purpose of 'wincodepage' tool output

bradking:
Document the intended purpose of each line of output.

Avoid implying support for arbitrary code pages in `build.ninja`.


bradking:
Cc: @jhasse @MrTrillian

MrTrillian:
Thanks for Cc'ing, @bradking . See my last comment on #1918

bradking:
#1921 is an alternative approach that minimizes the information in the output.  I'll let @jhasse decide which approach to merge, and close the other one.

bradking:
Based on the discussion in #1918 I think #1921 is the preferred solution, so I'll close this alternative.

PR #1921, opened 2021-03-01
wincodepage: minimize to indicate UTF-8 or not

bradking:
The ANSI code page identifier is more information than generator
programs actually need.  The encoding of `build.ninja` is always
either UTF-8 or the system-wide ANSI code page.  Reduce the output
to provide no more information than the generator programs need.

The Console code page can be obtained in other ways, so drop it.

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -308,17 +308,21 @@ file. _Available since Ninja 1.10._
 if they have one).  It can be used to know which rule name to pass to
 +ninja -t targets rule _name_+ or +ninja -t compdb+.
 
-`wincodepage`:: available on Windows hosts.  Prints the ANSI code page
-used by `ninja`, whose encoding is expected in `build.ninja`.  Also prints
-the Console code page for reference.  The output has the form:
+`wincodepage`:: Available on Windows hosts (_since Ninja 1.11_).
+Prints the Windows code page whose encoding is expected in the build file.
+The output has the form:
 +
 ----
-ANSI code page: %u
-Console code page: %u
+Build file encoding: <codepage>
 ----
 +
-where each `%u` is an integer code page identifier, expressed in decimal.
-_Available since Ninja 1.11._
+Additional lines may be added in future versions of Ninja.
++
+The `<codepage>` is one of:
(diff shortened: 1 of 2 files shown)


bradking:
Cc: @jhasse @MrTrillian

doc/manual.asciidoc
```diff
@@ -308,17 +308,21 @@ file. _Available since Ninja 1.10._
...
 +
-where each `%u` is an integer code page identifier, expressed in decimal.
-_Available since Ninja 1.11._
+Additional lines may be added in future versions of Ninja.
++
+The `<codepage>` is one of:
+
+`UTF-8`::: Encode as UTF-8.
+
+`ANSI`::: Encode to the system-wide ANSI code page.
```
bradking, line 325: @MrTrillian is there a name besides `ANSI` that would be more accurate?

jhasse:
What about renaming the tool to `winutf8` and then just exit with 0 (success, utf-8 enabled) or 1 (something else)?

This would make it fairly easy for CMake to check if UTF-8 can be used, wouldn't it? Even for older ninja versions a non existing tool will return non-0, too.

MrTrillian (approved):

doc/manual.asciidoc
MrTrillian: ANSI is the established term for this 👍

bradking:
> What about renaming the tool to winutf8 and then just exit with 0 (success, utf-8 enabled) or 1 (something else)?

I actually implemented that locally first and decided on this instead.  With exit-code only, human usage is very confusing.  There is no output.  It is also more error prone because running `ninja -t winutf8 extra-arg` will report an error and return non-zero even if the encoding is UTF-8.

KyleFromKitware:
I agree, the output should be human-readable.

bradking:
@jhasse this PR should be good to go, and updates the already-merged feature from #1918.  We'd like to get it merged so we can update the corresponding [CMake change](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5860).

bradking:
Thanks!

PR #1924, opened 2021-03-05
ninja_workdir: reconcile relative and absolute paths in depfile,dyndep files with build manifest

bradking:
A depfile (or dyndep file) may contain relative and/or absolute paths based on compiler
behavior and the include directory flags given.  It is not always
possible to make these exactly match corresponding paths in the build
manifest.  If Ninja does not find an exact match for a depfile path, try
converting it to the opposite kind of path (absolute or relative) and
look again.  If neither matches, add a node with the path as-is.

Fixes: #1251

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -589,6 +589,14 @@ build edges such that it is not an error if the listed dependency is
 missing.  This allows you to delete a header file and rebuild without
 the build aborting due to a missing input.
 
+Paths emitted by the compiler may be either absolute or relative paths,
+depending on compiler behavior and the include directory flags given to it.
+Paths loaded from a depfile are matched to paths in build statements
+if they match exactly.  _Since Ninja 1.11_, paths are also matched if
+one of them is relative and matches the other when prefixed by the
+working directory (determined as documented by the
+<<ref_toplevel, top-level variable>> `ninja_workdir`).
+
 depfile
 ^^^^^^^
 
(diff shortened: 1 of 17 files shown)


bradking:
Based on #1917.

doc/manual.asciidoc
```diff
@@ -589,6 +589,14 @@ build edges such that it is not an error if the listed dependency is
 missing.  This allows you to delete a header file and rebuild without
 the build aborting due to a missing input.
 
+Paths emitted by the compiler may be either absolute or relative paths,
+depending on compiler behavior and the include direcory flags given to it.
```
lukka, line 593: typo: `directory`

bradking:
Closing for now.  I have some ideas on a more general solution.

jhasse:
Could `builddir` be reused for this?

bradking:
Looking the history and current functionality of `builddir`, it appears to be intended for intermediate files such as Ninja's build log, and is unrelated to the location of the build manifest or the current working directory.  Furthermore, it is allowed to be a relative path.  Therefore I don't think it's suitable for this PR's purpose.

jhasse:
With CMake it isn't set and implicitly set to the path of the build manifest. CMake could set it to the absolute path and not have a compatibility problem with older ninja versions. Then we use builddir (if absolute) to reconcile paths in depfiles (i.e. your patch).

bradking:
The condition for using `builddir` in this way could be "is absolute and `realpath(builddir) == getcwd()`".

jhasse:
ah yes good point, we don't want it to mess up dependencies when it is NOT set to the directory of the build manifest

PR #1925, opened 2021-03-09
depfile,dyndep: reconcile relative and absolute paths with build manifest

bradking:
A depfile (or dyndep file) may contain relative and/or absolute paths based on compiler
behavior and the include directory flags given. It is not always
possible to make these exactly match corresponding paths in the build
manifest. If Ninja does not find an exact match for a depfile path, try
converting it to the opposite kind of path (absolute or relative) and
look again. If neither matches, add a node with the path as-is.

Fixes: #1251

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -589,6 +589,13 @@ build edges such that it is not an error if the listed dependency is
 missing.  This allows you to delete a header file and rebuild without
 the build aborting due to a missing input.
 
+Paths emitted by the compiler may be either absolute or relative paths,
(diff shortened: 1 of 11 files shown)


bradking:
This version differs from #1924 in that it does not add the `ninja_workdir` binding and relies only on `getcwd` to recover the prefix for relative paths in `build.ninja`.  This is a bit less powerful for generators that want to support certain combinations involving symbolic links, but solves #1251 without introducing any new interfaces.

bradking:
Closing for now.  I have some ideas on a more general solution.

PR #1933, opened 2021-03-22
Follow up fixes to #1753

### LINKED ISSUE #1932: build.ninja generator not re-running when it should
#1753 broke CMake's test suite.  I must have neglected to run it before my approval in that PR.

I tracked it down to this example:

```bash
rm -f .ninja_deps .ninja_log
cat >build.ninja <<EOF
rule generator
  command = touch build.ninja
  generator = 1
build build.ninja: generator | implicit_generator_input
EOF
sleep 1
touch implicit_generator_input
ninja
ls -tr build.ninja implicit_generator_input
```

Prior to the change, the output is:

```
[1/1] touch build.ninja
ninja: no work to do.
implicit_generator_input
build.ninja
```

After the change, the output is:

```
ninja: no work to do.
build.ninja
implicit_generator_input
```

jdrouhard:
These are followup fixes for #1753 and fixes #1932.

- Check the output file's mtime against the most recent input time before checking the build log
- For outputs created by generator rules, record the restat'd output's mtime in the build log
- Always include depfile mtime in build log mtime if special deps binding isn't specified
- Fix performance regression (change `Stat()` to `StatIfNecessary()` in `StartEdge()`)

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -750,7 +750,7 @@ bool Builder::FinishCommand(CommandRunner::Result* result, string* err) {
     return plan_.EdgeFinished(edge, Plan::kEdgeFailed, err);
   }
 
-  TimeStamp most_recent_input_mtime = 0;
+  TimeStamp record_mtime = 0;
   if (!config_.dry_run) {
     // Restat the edge outputs
     bool node_cleaned = false;
@@ -761,6 +761,8 @@ bool Builder::FinishCommand(CommandRunner::Result* result, string* err) {
       TimeStamp new_mtime = (*o)->mtime();
       if (new_mtime == -1)
         return false;
+      if (new_mtime > record_mtime)
+        record_mtime = new_mtime;
       if (old_mtime == new_mtime && edge->GetBindingBool("restat")) {
         // The rule command did not change the output.  Propagate the clean
         // state through the build graph.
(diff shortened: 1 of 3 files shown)


jdrouhard:
cc: @bradking @jhasse

bradking:
I've confirmed this fixes #1932, but some of CMake's test suite still fails.  I'll track down the next case.

bradking:
See additional case posted in https://github.com/ninja-build/ninja/issues/1932#issuecomment-804275428.

bradking:
@jdrouhard as of commit `9fb5867362` all but one of CMake's tests pass.  I'm still tracking down the remaining failure.

bradking:
See additional case posted in https://github.com/ninja-build/ninja/issues/1932#issuecomment-804399236.

src/build.cc
```diff
@@ -692,7 +692,7 @@ bool Builder::StartEdge(Edge* edge, string* err) {
   Node* most_recent_input = NULL;
   for (vector<Node*>::iterator i = edge->inputs_.begin();
        i != edge->inputs_.end() - edge->order_only_deps_; ++i) {
-    if (!(*i)->Stat(disk_interface_, err))
+    if (!(*i)->StatIfNecessary(disk_interface_, err))
```
jdrouhard, line 695: This was discussed in #1753 but I think the performance tradeoff should lean toward speed over correctness. If we err on the side of speed, we just might have unnecessary rebuilds on future runs (as opposed to the alternative which is thinking outputs are up to date when they are actually dirty).

bradking:
As of commit `f2205713` all of CMake's tests pass.  However, commit `53f96fed` breaks one again.

src/build.cc
bradking: A build system should not re-run commands when not needed.  We don't know how many downstream dependencies might need to rebuild unnecessarily.

jdrouhard: We discussed this on the original PR, I think you actually wondered why this wasn't `StatIfNecessary()` to begin with.

This shouldn't cause ninja to rerun commands if the inputs don't change, which is identical behavior to what was here before #1753.

jdrouhard:
> As of commit `f2205713` all of CMake's tests pass.  However, commit `53f96fed` breaks one again.
> 
> 

What test? We really should port more of CMake's ninja tests to native ninja tests.

bradking:
>>  commit 53f96fed breaks one again.
> What test? 

It's the `Qt5Autogen.RerunMocBasic` test.  I won't have time until later to try to narrow it down to a ninja-only example.

src/build.cc
bradking: Here is an example that change affects:

```bash
rm -f .ninja_deps .ninja_log out1 out2 in1 in2imp

>in1
echo >build.ninja '
rule touch_out_discovered_during_build
  command = touch $out $discovered

rule touch_dep_discovered_during_build
  command = touch $out; echo "$out: $discovered" > $depfile

build out1: touch_out_discovered_during_build in1
  discovered = in2imp

build out2: touch_dep_discovered_during_build in1 || out1
  discovered = in2imp
  depfile = out2.d
  deps = gcc

default out2'

ninja >/dev/null 2>&1

sleep 1
>in1
ninja
ninja -d explain
```

With `Stat`, or prior to #1753, the output is:

```
[1/2] touch out1 in2imp
[2/2] touch out2; echo "out2: in2imp" > out2.d
ninja: no work to do.
```

With `StatIfNecessary` after #1753, the output is:

```
[1/2] touch out1 in2imp
[2/2] touch out2; echo "out2: in2imp" > out2.d
ninja explain: recorded mtime of out2 older than most recent input in2imp (... vs ...)
[1/1] touch out2; echo "out2: in2imp" > out2.d
```

jdrouhard: Yeah, this test is unfortunate. Using `StatIfNecessary()` could lead to infinite rebuild loops (target would never be clean) with this specific relationship of rules and implicit dependencies being updated by the rules. `out1` has no specific knowledge that it updated `in2imp` so there's nothing we can do when finishing the command for `out1` to update the cached mtime for `in2imp`.

If we want to keep the better build correctness that #1753 provides, it looks like there's no way to avoid `Stat()` when edges are started which does have a performance penalty.

bradking: IIUC #1753 provides robustness against edits made to the source tree *during the build* by processes *outside the build system*.  We need to decide if supporting that is worth making everyone pay for the extra `Stat` on every build just in case someone does that.

bradking: This investigation led me to find that [CMake Issue 20025](https://gitlab.kitware.com/cmake/cmake/-/issues/20025#note_925598) has the same root cause, even with Ninja 1.10 (which does not have #1753).  Cases like this might be resolvable using a `dyndep` binding to declare the dynamically-discovered implicit output so that Ninja knows about it.  I still need to try that out to be sure.

However, CMake's cases are just examples of existing cases that may exist in the wild.  Many projects would have had no choice but to express things as in my above example because `dyndep` support is recent.

bradking: I suggest dropping commit `53f96fed` for now so we can get everything else fixed.  Then open an issue to discuss this optimization.

jdrouhard: Yeah, that sounds good. I'm going to go ahead and add regression tests for this case anyhow.

jhasse:
Hi! Thanks for your work. I haven't followed everything yet.

I think a good idea would be to actually revert the merge commit of #1753, then add the test cases (except those for #1753) and then draft a new version of #1753 with the fixups. What do you think?

We could also merge this PR, but then the actual change (and the fixes for it) would be scattered.

jdrouhard:
I can do that. I'll open another PR with the revert + new tests, and a second one that re-lands the change along with the fixes.

src/build.cc
bradking: > might be resolvable using a dyndep binding... I still need to try that out to be sure.

Unfortunately that won't work without additional changes to Ninja.  See the [explanation](https://gitlab.kitware.com/cmake/cmake/-/issues/20025#note_926850) I posted in the CMake issue.

PR #1936, opened 2021-03-24
Change build log to always log the most recent input mtime

jdrouhard:
This is the same PR as #1753 but includes the fixes from #1933. Some additional discussion about the slight performance regression this introduces should be done. The tradeoff this PR introduces is improved robustness against external processes modifying build input files during the build at the cost of a `stat()` on each input just before an edge's command is run.

Commit message follows:

---------

If an edge's output files' mtimes are compared to the most recent
input's mtime, edges might be calculated as clean even if they are
actually dirty. While an edge is running its command to produce its
outputs and an input to the edge is updated before the outputs are
written to disk, then subsequent runs will think that the outputs are
newer than the inputs, even though the inputs have actually been updated
and may be different than what were used to produce those outputs.

Ninja will now restat all inputs just prior to running an edge's command
and remember the most recent input mtime. When the command completes, it
will stat any discovered dependencies from dep files (if necessary),
recalculate the most recent input mtime, and log it to the build log
file. On subsequent runs, ninja will use this value to compare to the
edge's most recent input's mtime to determine whether the outputs are
dirty.

This extends the methodology used by restat rules to work in all cases.
Restat rules are still unique in that they will clean the edge's output
nodes recursively if the edge's command did not change the output, but
in all cases, the mtime recorded in the log file is now the most recent
input mtime. See the new tests for more clarification.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -270,19 +270,11 @@ bool Plan::CleanNode(DependencyScan* scan, Node* node, string* err) {
 #define MEM_FN mem_fn  // mem_fun was removed in C++17.
 #endif
     if (find_if(begin, end, MEM_FN(&Node::dirty)) == end) {
-      // Recompute most_recent_input.
-      Node* most_recent_input = NULL;
-      for (vector<Node*>::iterator i = begin; i != end; ++i) {
-        if (!most_recent_input || (*i)->mtime() > most_recent_input->mtime())
-          most_recent_input = *i;
-      }
-
       // Now, this edge is dirty if any of the outputs are dirty.
       // If the edge isn't dirty, clean the outputs and mark the edge as not
       // wanted.
       bool outputs_dirty = false;
-      if (!scan->RecomputeOutputsDirty(*oe, most_recent_input,
-                                       &outputs_dirty, err)) {
+      if (!scan->RecomputeOutputsDirty(*oe, &outputs_dirty, err)) {
         return false;
       }
       if (!outputs_dirty) {
@@ -696,6 +688,20 @@ bool Builder::StartEdge(Edge* edge, string* err) {
       return false;
   }
 
+  // Find the most recent mtime of any (existing) non-order-only input
+  Node* most_recent_input = NULL;
+  for (vector<Node*>::iterator i = edge->inputs_.begin();
+       i != edge->inputs_.end() - edge->order_only_deps_; ++i) {
+    if (!(*i)->Stat(disk_interface_, err))
+      return false;
+    if (!most_recent_input || (*i)->mtime() > most_recent_input->mtime())
+      most_recent_input = *i;
+  }
+
+  edge->most_recent_input_ = most_recent_input;
+  if (most_recent_input)
+    edge->most_recent_input_mtime_ = most_recent_input->mtime();
+
   // start command computing and run it
   if (!command_runner_->StartCommand(edge)) {
     err->assign("command '" + edge->EvaluateCommand() + "' failed.");
@@ -744,20 +750,20 @@ bool Builder::FinishCommand(CommandRunner::Result* result, string* err) {
     return plan_.EdgeFinished(edge, Plan::kEdgeFailed, err);
   }
 
-  // Restat the edge outputs
-  TimeStamp output_mtime = 0;
-  bool restat = edge->GetBindingBool("restat");
+  TimeStamp record_mtime = 0;
   if (!config_.dry_run) {
+    // Restat the edge outputs
     bool node_cleaned = false;
-
     for (vector<Node*>::iterator o = edge->outputs_.begin();
          o != edge->outputs_.end(); ++o) {
-      TimeStamp new_mtime = disk_interface_->Stat((*o)->path(), err);
+      TimeStamp old_mtime = (*o)->mtime();
+      (*o)->Stat(disk_interface_, err);
+      TimeStamp new_mtime = (*o)->mtime();
       if (new_mtime == -1)
         return false;
-      if (new_mtime > output_mtime)
-        output_mtime = new_mtime;
-      if ((*o)->mtime() == new_mtime && restat) {
+      if (new_mtime > record_mtime)
+        record_mtime = new_mtime;
+      if (old_mtime == new_mtime && edge->GetBindingBool("restat")) {
         // The rule command did not change the output.  Propagate the clean
         // state through the build graph.
         // Note that this also applies to nonexistent outputs (mtime == 0).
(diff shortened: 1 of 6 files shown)


jdrouhard:
cc: @bradking @jhasse

src/build.cc
```diff
@@ -768,32 +774,40 @@ bool Builder::FinishCommand(CommandRunner::Result* result, string* err) {
...
+      for (vector<Node*>::iterator i = deps_nodes.begin();
+          i != deps_nodes.end(); ++i) {
+        (*i)->StatIfNecessary(disk_interface_, err);
+        if ((*i)->mtime() > record_mtime)
+          record_mtime = (*i)->mtime();
+      }
+    }
+
+    // Take into account the mtime of the depfile. Special depfiles (denoted with a "deps"
+    // binding) are handled above
```
bradking, line 803: The "Special depfiles" are just depfiles that are loaded into the DepsLog. Perhaps:

```c++
// If we have 'depfile' but not 'deps' then the depfile is not
// loaded into the deps log.  Take its mtime into account.
```

src/build.cc
```diff
@@ -768,32 +774,40 @@ bool Builder::FinishCommand(CommandRunner::Result* result, string* err) {
...
+    string depfile = edge->GetUnescapedDepfile();
+    if (deps_type.empty() && !depfile.empty()) {
+      TimeStamp depfile_mtime = disk_interface_->Stat(depfile, err);
+      if (depfile_mtime == -1)
+        return false;
+      if (depfile_mtime > record_mtime)
+        record_mtime = depfile_mtime;
```
bradking, line 810: What if the `depfile` names an implicit dependency that is newer than itself or any other known input?  In that case we won't log the proper mtime for the edge.  With the most-recent-input-mtime approach, we do need to load each `depfile` immediately after it is written even when not using the deps log.  It is no longer sufficient to defer loading it to future builds.

src/build.cc
```diff
@@ -768,32 +774,40 @@ bool Builder::FinishCommand(CommandRunner::Result* result, string* err) {
...
+    // Generator edges should just use the edge's latest output mtime since it's possible that
+    // the generator rule could create/update implicit inputs as part of the command, and
+    // we don't want to get into an infinite loop.
+    if (!edge->GetBindingBool("generator")) {
+      record_mtime = edge->most_recent_input_mtime_;
+
+      // If there were any added deps, compute the most recent input mtime
+      for (vector<Node*>::iterator i = deps_nodes.begin();
+          i != deps_nodes.end(); ++i) {
+        (*i)->StatIfNecessary(disk_interface_, err);
```
bradking, line 796: What if the newly-known input was modified externally between the time that the edge's command actually used it, and the time that we are now stat-ing it?

src/build.cc
```diff
@@ -696,6 +688,20 @@ bool Builder::StartEdge(Edge* edge, string* err) {
       return false;
   }
 
+  // Find the most recent mtime of any (existing) non-order-only input
+  Node* most_recent_input = NULL;
+  for (vector<Node*>::iterator i = edge->inputs_.begin();
+       i != edge->inputs_.end() - edge->order_only_deps_; ++i) {
+    if (!(*i)->Stat(disk_interface_, err))
```
bradking, line 695: For reference, this being `Stat` rather than `StatIfNecessary` is the crux of the performance impact of this approach.  It is needed to account for cases in which an input is modified after the build plan was created but before an edge runs.

bradking (changes requested):
I've confirmed that CMake's test suite is happy with this PR as of commit `b0f83c32`.

The goal of the change in this PR is to improve robustness against changes to sources that are made *during the build* by processes that are *not part of the build*, such as a human with an external editor.  The performance cost is an extra `stat` for every input of every edge that executes during each build.

src/build.cc
jdrouhard: Since we didn't know about it when we started, the best we can do is get the time now. It's why the PR isn't 100% robust around external input modification during the build. The only way to get 100% correctness is to use dyndep to discover all inputs before the actual edge command is started. To be fair, doing it this way is no worse than what currently exists in ninja (other than the possibility of an unneeded rebuild on a subsequent run - if this input wasn't known to this edge specifically but was known during the initial build plan generation).

We can't make this `Stat()` as it defeats the purpose of the PR. Since `deps_nodes` contains all dependencies from the "special" dep file (not only those that weren't known at edge command start time), using `Stat()` here would mean that changes to the input while the edge command is run would be lost on future builds.

A better way to do this would be to somehow track which deps were discovered that are "new" and were not counted for during the initial edge start input dep list. We could `Stat()` only those, and simply bypass doing anything with the ones we know we counted for when edge was started.

Unfortunately, the `RecordDeps()` function used below this short circuits actually figuring out _which_ deps are new in favor of simply replacing the list with the new one if they are detected as different. I could take a crack at improving this dependency discovery/updating to be a bit more intelligent without sacrificing any performance, but I'm not sure where that will take me.

src/build.cc
jdrouhard: True, parsing the depfile and discovering new inputs during `FinishCommand()` is likely necessary, now. I can work on that in the next week.

src/build.cc
jdrouhard: @bradking I had another thought this morning. What if, instead of trying to be intelligent about tracking the most recent input mtime for an output, we simply _touch_ the output file just prior to starting the edge command, and remember the mtime of the file at that point (and record it when processing the successful command later)? If the output doesn't exist yet, we can simply delete it immediately after grabbing the mtime. We'd probably only have to do this for the first output of the edge (instead of attempting to do it for all outputs).

If the command fails, we won't record any new output mtime in the log file, so subsequent runs would still use the last-successful recorded mtime to compare against current inputs.

I think this solves the performance issue (touch/stat of one file per edge vs a stat of all input files per edge) as well as the hole this PR currently has related to inputs not known to the edge when the command is started.

bradking: Interesting idea.  Do we even need to touch the actual output?  We could touch another file, or even try to just compute the current time ourselves.

jdrouhard: I started with computing the time ourselves, but determined it's bad to compare filesystem mtimes to system times. This at least keeps ninja comparing filesystem times to each other.

We could touch another file, but I was concerned that a destination file on one filesystem might have different time stamping granularity (or times, in the case of NFS) than whatever temporary file we decided to use.

bradking: It feels risky to touch the actual output in case the build is interrupted.

Another approach is to touch a `.ninja_tmp` file in the same directory as the first output, get its mtime, and then remove it.

jdrouhard: Yep, that's exactly what I'm trying in a separate branch right now.

jdrouhard: This is working. I'm pretty excited about this alternative, it's much less intrusive than this change, is more performant, and I *think* is correct even more of the time.

bradking: Great!  That approach should probably be in a new PR.

jdrouhard:
Closed in favor of #1943.

PR #1937, opened 2021-03-24
dyndep: reconcile dyndep-specified outputs with depfile-specified inputs

bradking:
When a path loaded from a depfile does not have a node, we create a new node with a phony edge producing it.  If we later load a dyndep file that specifies the same node as an output of a known edge, we previously failed with a "multiple rules generate ..." error.  Instead, since the conflicting edge was internally generated, replace the node's input edge with the now-known real edge that produces it.

While at it, add a dedicated test for dyndep-discovered implicit dependencies.  Previously this was covered only as part of more complex tests.


bradking:
@jhasse this resolves a dyndep+depfile corner case I found while investigating #1932 and #1933.

jhasse:
Thanks! Will this fix https://gitlab.kitware.com/cmake/cmake/-/issues/20025 ?

bradking:
[CMake Issue 20025](https://gitlab.kitware.com/cmake/cmake/-/issues/20025) needs this change, but will also need changes to CMake to use this combination of features.  The case in that issue has build-time-discovered outputs of the `cmake_autogen` command that are also build-time-discovered dependencies reported by the compiler in a depfile.

Issue #1940, opened 2021-03-26
restat needlessly rebuilding files


dmoody256:
Using this ninja file:

```
rule INSTALL
    command = rm -f $out; ln $in $out
    restat = 1
build install/output1.txt: INSTALL src/input1.txt
build install/output2.txt: INSTALL install/output1.txt | install/output3.txt
build install/output3.txt: INSTALL src/input2.txt
```

And creating fake inputs in this order:

```
mkdir src
touch src/input1.txt
touch src/input2.txt
```

Ninja will rebuild files that have already been built on consecutive runs when nothing has changed:

```
> ninja --version
1.10.0.git.kitware.jobserver-1
> ninja -d explain | cat
ninja explain: output install/output1.txt doesn't exist
ninja explain: install/output1.txt is dirty
ninja explain: output install/output3.txt doesn't exist
ninja explain: install/output3.txt is dirty
[1/3] rm -f install/output1.txt; ln src/input1.txt install/output1.txt
[2/3] rm -f install/output3.txt; ln src/input2.txt install/output3.txt
[3/3] rm -f install/output2.txt; ln install/output1.txt install/output2.txt
> ninja -d explain | cat
ninja explain: restat of output install/output2.txt older than most recent input install/output3.txt (1616728218815774370 vs 1616728221187789429)
[1/1] rm -f install/output2.txt; ln install/output1.txt install/output2.txt
> ninja -d explain | cat
ninja: no work to do.
```

jdrouhard:
Can you try #1943 and see if that fixes this issue?

dmoody256:
> Can you try #1943 and see if that fixes this issue?

@jdrouhard I confirmed that branch fixes this issue.

PR #1941, opened 2021-03-26
Changed restat mtime to be set the first time restat edges are built.

dmoody256:
Fixes https://github.com/ninja-build/ninja/issues/1940


jhasse:
Can you add a unit test for #1940?

jdrouhard:
> Can you add a unit test for #1940?

A unit test to show this would be great, and @dmoody256, could you try the branch for #1943 as a fix for #1940 first? It changes what is recorded as the output mtime in the build log.

dmoody256:
@jdrouhard @jhasse unit test added

Issue #1733, opened 2020-02-06
Add a performance test script / "performance regression tests"


jhasse:
Currently this is done manually by building Chromium before and after a commit. I think it would be good to have a script, which you can pass two commit hashes and a path to a Chromium source, which will then compare the `-d stats` result.

Maybe add the functionality to `misc/measure.py`?

See https://github.com/ninja-build/ninja/pull/1577#issuecomment-517797439

nightlark:
How long does it take to build chromium?

jonesmz:
On my 24 core server. About 2 hours.

Its pretty horrible

ClausKlein:
use ccache, that helps ...

scivision:
Would it be useful to have smaller but also ubiquitous packages? Maybe in Fortran too, like Lapack that is a couple thousand files mostly in Fortran. 
Typically for benchmarking one tries to do several repeated runs to mitigate unpredictable externally-driven slowdowns.
Maybe look to the packages Phoronix frequently uses for build benchmarking?

PR #1943, opened 2021-03-26
Provide resiliency against inputs changing during the build

### LINKED ISSUE #1162: Long running compile commands can cause incorrect dirty status computation

jdrouhard:
This is a better, faster approach than #1753 for providing robustness against inputs changing during the build. It fixes #1162  and #1940.

When recording the outputs' mtimes when the command finishes, use the
remembered temporary file's mtime from when the edge started for
recording in the build log. Subsequent runs will use that as the mtime
for the output.

This provides robustness against inputs changing while the command
itself is running. If an input is changed, the subsequent run will
detect the output as dirty since its recorded mtime reflects when the
build command began, not when the output was actually written to disk.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -518,6 +518,10 @@ Builder::Builder(State* state, const BuildConfig& config,
       start_time_millis_(start_time_millis), disk_interface_(disk_interface),
       scan_(state, build_log, deps_log, disk_interface,
             &config_.depfile_parser_options) {
+  lock_file_path_ = ".ninja_lock";
+  string build_dir = state_->bindings_.LookupVariable("builddir");
+  if (!build_dir.empty())
+    lock_file_path_ = build_dir + "/" + lock_file_path_;
 }
 
 Builder::~Builder() {
@@ -552,6 +556,10 @@ void Builder::Cleanup() {
         disk_interface_->RemoveFile(depfile);
     }
   }
+
+  string err;
+  if (disk_interface_->Stat(lock_file_path_, &err) > 0)
+    disk_interface_->RemoveFile(lock_file_path_);
 }
 
 Node* Builder::AddTarget(const string& name, string* err) {
@@ -704,13 +712,24 @@ bool Builder::StartEdge(Edge* edge, string* err) {
 
   status_->BuildEdgeStarted(edge, start_time_millis);
 
-  // Create directories necessary for outputs.
+  TimeStamp build_start = -1;
+
+  // Create directories necessary for outputs and remember the current
+  // filesystem mtime to record later
   // XXX: this will block; do we care?
   for (vector<Node*>::iterator o = edge->outputs_.begin();
        o != edge->outputs_.end(); ++o) {
     if (!disk_interface_->MakeDirs((*o)->path()))
       return false;
+    if (build_start == -1) {
+      disk_interface_->WriteFile(lock_file_path_, "");
+      build_start = disk_interface_->Stat(lock_file_path_, err);
+      if (build_start == -1)
+        build_start = 0;
+    }
   }
+
+  edge->command_start_time_ = build_start;
 
   // Create response file, if needed
   // XXX: this may also block; do we care?
@@ -770,55 +789,42 @@ bool Builder::FinishCommand(CommandRunner::Result* result, string* err) {
   }
 
   // Restat the edge outputs
-  TimeStamp output_mtime = 0;
-  bool restat = edge->GetBindingBool("restat");
+  TimeStamp record_mtime = 0;
   if (!config_.dry_run) {
+    const bool restat = edge->GetBindingBool("restat");
+    const bool generator = edge->GetBindingBool("generator");
     bool node_cleaned = false;
+    record_mtime = edge->command_start_time_;
 
-    for (vector<Node*>::iterator o = edge->outputs_.begin();
-         o != edge->outputs_.end(); ++o) {
-      TimeStamp new_mtime = disk_interface_->Stat((*o)->path(), err);
-      if (new_mtime == -1)
-        return false;
-      if (new_mtime > output_mtime)
-        output_mtime = new_mtime;
-      if ((*o)->mtime() == new_mtime && restat) {
-        // The rule command did not change the output.  Propagate the clean
-        // state through the build graph.
-        // Note that this also applies to nonexistent outputs (mtime == 0).
-        if (!plan_.CleanNode(&scan_, *o, err))
+    // restat and generator rules must restat the outputs after the build
+    // has finished. if record_mtime == 0, then there was an error while
+    // attempting to touch/stat the temp file when the edge started and
+    // we should fall back to recording the outputs' current mtime in the
+    // log.
+    if (record_mtime == 0 || restat || generator) {
+      for (vector<Node*>::iterator o = edge->outputs_.begin();
+           o != edge->outputs_.end(); ++o) {
+        TimeStamp new_mtime = disk_interface_->Stat((*o)->path(), err);
+        if (new_mtime == -1)
           return false;
-        node_cleaned = true;
+        if (new_mtime > record_mtime)
+          record_mtime = new_mtime;
+        if ((*o)->mtime() == new_mtime && restat) {
+          // The rule command did not change the output.  Propagate the clean
+          // state through the build graph.
+          // Note that this also applies to nonexistent outputs (mtime == 0).
+          if (!plan_.CleanNode(&scan_, *o, err))
+            return false;
+          node_cleaned = true;
+        }
       }
     }
-
     if (node_cleaned) {
-      TimeStamp restat_mtime = 0;
-      // If any output was cleaned, find the most recent mtime of any
-      // (existing) non-order-only input or the depfile.
-      for (vector<Node*>::iterator i = edge->inputs_.begin();
-           i != edge->inputs_.end() - edge->order_only_deps_; ++i) {
-        TimeStamp input_mtime = disk_interface_->Stat((*i)->path(), err);
-        if (input_mtime == -1)
-          return false;
-        if (input_mtime > restat_mtime)
-          restat_mtime = input_mtime;
-      }
-
-      string depfile = edge->GetUnescapedDepfile();
-      if (restat_mtime != 0 && deps_type.empty() && !depfile.empty()) {
-        TimeStamp depfile_mtime = disk_interface_->Stat(depfile, err);
-        if (depfile_mtime == -1)
-          return false;
-        if (depfile_mtime > restat_mtime)
-          restat_mtime = depfile_mtime;
-      }
+      record_mtime = edge->command_start_time_;
 
       // The total number of edges in the plan may have changed as a result
       // of a restat.
       status_->PlanHasTotalEdges(plan_.command_edge_count());
-
-      output_mtime = restat_mtime;
     }
   }
 
@@ -832,7 +838,7 @@ bool Builder::FinishCommand(CommandRunner::Result* result, string* err) {
 
   if (scan_.build_log()) {
     if (!scan_.build_log()->RecordCommand(edge, start_time_millis,
-                                          end_time_millis, output_mtime)) {
+                                          end_time_millis, record_mtime)) {
       *err = string("Error writing to build log: ") + strerror(errno);
       return false;
     }
--- a/src/build_test.cc
+++ b/src/build_test.cc
@@ -611,6 +611,7 @@ bool FakeCommandRunner::StartCommand(Edge* edge) {
       fs_->WriteFile(edge->outputs_[0]->path(), content);
   } else if (edge->rule().name() == "touch-implicit-dep-out") {
     string dep = edge->GetBinding("test_dependency");
+    fs_->Tick();
     fs_->Create(dep, "");
     fs_->Tick();
     for (vector<Node*>::iterator out = edge->outputs_.begin();
@@ -627,7 +628,12 @@ bool FakeCommandRunner::StartCommand(Edge* edge) {
     fs_->Create(dep, "");
   } else if (edge->rule().name() == "generate-depfile") {
     string dep = edge->GetBinding("test_dependency");
+    bool touch_dep = edge->GetBindingBool("touch_dependency");
     string depfile = edge->GetUnescapedDepfile();
+    if (touch_dep) {
+      fs_->Tick();
+      fs_->Create(dep, "");
+    }
     string contents;
     for (vector<Node*>::iterator out = edge->outputs_.begin();
          out != edge->outputs_.end(); ++out) {
@@ -635,6 +641,20 @@ bool FakeCommandRunner::StartCommand(Edge* edge) {
       fs_->Create((*out)->path(), "");
     }
     fs_->Create(depfile, contents);
+  } else if (edge->rule().name() == "long-cc") {
+    string dep = edge->GetBinding("test_dependency");
+    string depfile = edge->GetUnescapedDepfile();
+    string contents;
+    for (vector<Node*>::iterator out = edge->outputs_.begin();
+        out != edge->outputs_.end(); ++out) {
+      fs_->Tick();
+      fs_->Tick();
+      fs_->Tick();
+      fs_->Create((*out)->path(), "");
+      contents += (*out)->path() + ": " + dep + "\n";
+    }
+    if (!dep.empty() && !depfile.empty())
+      fs_->Create(depfile, contents);
   } else {
     printf("unknown command\n");
     return false;
@@ -689,6 +709,18 @@ bool FakeCommandRunner::WaitForCommand(Result* result) {
     result->status = ExitFailure;
   else
     result->status = ExitSuccess;
+
+  // This rule simulates an external process modifying files while the build command runs.
+  // See TestInputMtimeRaceCondition and TestInputMtimeRaceConditionWithDepFile.
+  // Note: only the first and third time the rule is run per test is the file modified, so
+  // the test can verify that subsequent runs without the race have no work to do.
+  if (edge->rule().name() == "long-cc") {
+    string dep = edge->GetBinding("test_dependency");
+    if (fs_->now_ == 4)
+      fs_->files_[dep].mtime = 3;
+    if (fs_->now_ == 10)
+      fs_->files_[dep].mtime = 9;
+  }
 
   // Provide a way for test cases to verify when an edge finishes that
   // some other edge is still active.  This is useful for test cases
@@ -1471,7 +1503,7 @@ TEST_F(BuildWithLogTest, ImplicitGeneratedOutOfDate) {
 TEST_F(BuildWithLogTest, ImplicitGeneratedOutOfDate2) {
   ASSERT_NO_FATAL_FAILURE(AssertParse(&state_,
 "rule touch-implicit-dep-out\n"
-"  command = touch $test_dependency ; sleep 1 ; touch $out\n"
+"  command = sleep 1 ; touch $test_dependency ; sleep 1 ; touch $out\n"
 "  generator = 1\n"
 "build out.imp: touch-implicit-dep-out | inimp inimp2\n"
 "  test_dependency = inimp\n"));
@@ -1482,6 +1514,29 @@ TEST_F(BuildWithLogTest, ImplicitGeneratedOutOfDate2) {
   fs_.Tick();
 
   string err;
+
+  EXPECT_TRUE(builder_.AddTarget("out.imp", &err));
+  EXPECT_FALSE(builder_.AlreadyUpToDate());
+
+  EXPECT_TRUE(builder_.Build(&err));
+  EXPECT_TRUE(builder_.AlreadyUpToDate());
+
+  command_runner_.commands_ran_.clear();
+  state_.Reset();
+  builder_.Cleanup();
+  builder_.plan_.Reset();
+
+  EXPECT_TRUE(builder_.AddTarget("out.imp", &err));
+  EXPECT_TRUE(builder_.AlreadyUpToDate());
+  EXPECT_FALSE(GetNode("out.imp")->dirty());
+
+  command_runner_.commands_ran_.clear();
+  state_.Reset();
+  builder_.Cleanup();
+  builder_.plan_.Reset();
+
+  fs_.Tick();
+  fs_.Create("inimp", "");
 
   EXPECT_TRUE(builder_.AddTarget("out.imp", &err));
   EXPECT_FALSE(builder_.AlreadyUpToDate());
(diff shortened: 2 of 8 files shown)


jdrouhard:
cc: @bradking @jhasse 

If we are more comfortable with this approach, we can just close #1936. I apologize for all the back and forth on the previous attempt to provide the input file robustness. This seems so much cleaner, I wish I had arrived here to begin with.

jdrouhard:
@jhasse this is ready to merge and is far superior to the original approach in #1753, let me know if you have any questions about this. I will close #1936 since this supersedes it and is much improved over it.

@bradking having a positive CMake test run with this branch would be a good sanity check.

bradking:
The CMake test suite fails several tests with this change:

* `RunCMake.Ninja`
* `RunCMake.NinjaMultiConfig`
* `RunCMake.file`
* `RunCMake.try_compile`

@jdrouhard I'll contact you externally about reproducing the failures.

jdrouhard:
All CMake Tests pass with this branch, now. The failing tests were due to a slight hole in a regression test I added earlier in #1935. Generator rules should continue to record outputs' mtimes as they are on disk when a command finishes instead of using the time when the command is started. This branch now makes generator rules use the existing behavior instead of the new behavior.

@bradking feel free to double check this. I ran the entire CMake test suite, including all Qt5 tests.
@jhasse this is ready for review and merging.

jdrouhard:
> @jdrouhard the `RunCMake.Ninja` test still fails. Please run it with Qt 5.15 available.

@bradking I ran with Qt 5.14 and I cannot reproduce. If there was a specific change in Qt 5.15 that breaks this unrelated test, I'd be surprised. What's the specific condition that is failing?

src/build.cc
```diff
@@ -679,14 +679,33 @@ bool Builder::StartEdge(Edge* edge, string* err) {
...
+        err->assign("could not create temporary file to determine command "
+                    "start time of '" + edge->EvaluateCommand() + "'");
+        return false;
+      }
+
+      build_start = disk_interface_->Stat(tmp_path, err);
+      disk_interface_->RemoveFile(tmp_path);
+
+      if (build_start == -1)
+        return false;
```
jhasse, line 703: ninja shouldn't fail if it can't create the file, but rather behave as before.

src/build_test.cc
```diff
@@ -1750,15 +1800,17 @@ TEST_F(BuildTest, RspFileSuccess)
   EXPECT_TRUE(builder_.Build(&err));
   ASSERT_EQ(3u, command_runner_.commands_ran_.size());
 
-  // The RSP files were created
-  ASSERT_EQ(files_created + 2, fs_.files_created_.size());
+  // The RSP files and temp file to acquire output mtimes were created
+  ASSERT_EQ(files_created + 3, fs_.files_created_.size());
   ASSERT_EQ(1u, fs_.files_created_.count("out 2.rsp"));
   ASSERT_EQ(1u, fs_.files_created_.count("out 3.rsp"));
+  ASSERT_EQ(1u, fs_.files_created_.count(".ninja_tmp"));
```
jhasse, line 1807: IMHO this is an implementation detail and shouldn't be part of the test.

src/build_test.cc
```diff
@@ -1750,15 +1800,17 @@ TEST_F(BuildTest, RspFileSuccess)
...
 
-  // The RSP files were removed
-  ASSERT_EQ(files_removed + 2, fs_.files_removed_.size());
+  // The RSP files and temp file to acquire output mtimes were removed
+  ASSERT_EQ(files_removed + 3, fs_.files_removed_.size());
   ASSERT_EQ(1u, fs_.files_removed_.count("out 2.rsp"));
   ASSERT_EQ(1u, fs_.files_removed_.count("out 3.rsp"));
+  ASSERT_EQ(1u, fs_.files_removed_.count(".ninja_tmp"));
```
jhasse, line 1813: This should definitely be part of the test though. Maybe fs_ should have a method to count files and we should check that everything was cleaned up.

src/disk_interface_test.cc
```diff
@@ -198,7 +198,7 @@ TEST_F(DiskInterfaceTest, MakeDirs) {
   EXPECT_EQ(0, fclose(f));
 #ifdef _WIN32
   string path2 = "another\\with\\back\\\\slashes\\";
-  EXPECT_TRUE(disk_.MakeDirs(path2.c_str()));
+  EXPECT_TRUE(disk_.MakeDirs(path2));
```
jhasse, line 201: Good catch! Too bad our CI doesn't run clang-tidy on Windows (yet).

jhasse (commented):
I'm wondering what the performance implications of touching a file are? I'm also scared about edge-cased like NFS mounted filesystems.

Shouldn't one file be enough? I know that this will result in unnecessary rebuilds if an input is changed while its edge hasn't started, but better safe than sorry (and it would bring clarity to the performance implication of this PR).

I think one file might be a good idea. First of all we can create it next to .ninja_deps and .ninja_log (a directory we definitely have write access to). Second we might re-use it as a locking mechanism: Currently it's possible to run ninja multiple times in a directory resulting in all sorts of bugs. What do you think?

---

Another thought: What about the hash approach [Evan suggested](https://github.com/ninja-build/ninja/pull/1165#issuecomment-594233368)? While this would be a HUGE change, it might be something we want to do in the future anyway, see https://github.com/ninja-build/ninja/issues/1459

jdrouhard:
> I'm wondering what the performance implications of touching a file are? I'm also scared about edge-cased like NFS mounted filesystems.

For any edge, it only touches a single file. I don't believe there will be much performance impact here since the edge command will be writing its outputs to the disk when it runs anyway.

> Shouldn't one file be enough? I know that this will result in unnecessary rebuilds if an input is changed while its edge hasn't started, but better safe than sorry (and it would bring clarity to the performance implication of this PR).
>
> I think one file might be a good idea. First of all we can create it next to .ninja_deps and .ninja_log (a directory we definitely have write access to). Second we might re-use it as a locking mechanism: Currently it's possible to run ninja multiple times in a directory resulting in all sorts of bugs. What do you think?

As we've discovered in the previous PRs #1753 and #1932 we really do need to track the beginning of each edge separately. However, I can change it so the temp file we use lives next to `.ninja_deps` and `.ninja_log` but each edge will still touch/stat it to get the current time. I could clean up a bit of the refactoring I did around `DirName(...)` that way.

Using it as a locking mechanism could also come in handy if we decide to clean the file up at the end of the run. I'm not going to tackle that in this PR, though.

> Another thought: What about the hash approach [Evan suggested](https://github.com/ninja-build/ninja/pull/1165#issuecomment-594233368)? While this would be a HUGE change, it might be something we want to do in the future anyway, see #1459

This would be a massive change that would likely break many many things. There are many things to consider with a hash approach, and in general, using a hash instead of mtime has its own set of problems that are just different than mtimes. At least the approach in this PR makes it pretty resilient and is a definite improvement over current functionality.

bradking:
> I ran with Qt 5.14 and I cannot reproduce. If there was a specific change in Qt 5.15 that breaks this unrelated test, I'd be surprised. What's the specific condition that is failing?

@jdrouhard there is a section of the `RunCMake.Ninja` test that only runs with Qt 5.15, and not 5.14.

src/build.cc
jdrouhard: done

src/build_test.cc
jdrouhard: the file is created by the virtual file system during the test; putting the assert here makes it clear why there were 3 files created.

jdrouhard:
@bradking tested with Qt 5.15. restat rules must also restat the outputs when the command finishes (along with generator rules).
@jhasse I've changed it so the `.ninja_tmp` file is created once per build next to `.ninja_log`/`.ninja_deps` and if it can't be touched/stat'd when an edge starts, then it falls back to old behavior.

Let me know what else you'd like to see.

dmoody256:
@jdrouhard could you try out my unit test from #1941, wether this goes in or that PR, I would like the unit test to be added to protect the hardlinking scenario that is causing me trouble currently in my builds.

bradking:
> tested with Qt 5.15. restat rules must also restat the outputs 

Thanks.  With that fix, this PR as of commit `8d191a9550` passes CMake's entire test suite.

jdrouhard:
> @jdrouhard could you try out my unit test from #1941, wether this goes in or that PR, I would like the unit test to be added to protect the hardlinking scenario that is causing me trouble currently in my builds.

I can try your test on my branch. I'd like to see this get merged as is, and then we can change your PR to just be a new regression test, if that works for you?

dmoody256:
@jdrouhard sure thats fine, i'll try to follow up in my PR

jdrouhard:
@jhasse this should be good to go. If you need more time to look it over, maybe you could at least add a label for the next ninja release?

jdrouhard:
@jhasse ping

jhasse:
> I'm not sure what the alternative is, record the same mtime for every single output whose command was run during the course of the build?

I thought the latest input's mtime?

jdrouhard:
> > I'm not sure what the alternative is, record the same mtime for every single output whose command was run during the course of the build?
> 
> 
> 
> I thought the latest input's mtime?

That was the attempt that doesn't work 100% of the time because ninja doesn't necessarily know all of the inputs when an edge command runs.

Can we merge this PR so it can bake in master for awhile and get more eyes and testing on it? It fully passes the CMake test suite and I've added a number of regression tests in an earlier PR, as well as more here. Performance impact should be minimal, and any of the minimal impact is amortized over the course of the build (no performance impact to build graph generation).

jhasse:
> That was the attempt that doesn't work 100% of the time because ninja doesn't necessarily know all of the inputs when an edge command runs.

But on the next run ninja would then know if the input was known previously, right?

jdrouhard:
> > That was the attempt that doesn't work 100% of the time because ninja doesn't necessarily know all of the inputs when an edge command runs.
> 
> 
> 
> But on the next run ninja would then know if the input was known previously, right?

No, it doesn't work with 100% accuracy because ninja needs to record some value in the log when the command finishes. If it doesn't know the inputs until after the command, then the whole reason for the PR is circumvented. There wouldn't be a way for ninja to know if one of those inputs changed while the command was running.

The only way to completely solve it is for ninja to remember the time the command starts, and that time needs to be directly comparable to other input mtimes. That way, ninja can compare the time the command started to the inputs' mtimes on subsequent builds to see if anything changed since the command ran that produced the output. That is exactly what this PR does.

jhasse:
> If it doesn't know the inputs until after the command, [...]

It doesn't know *all* inputs. For the inputs it knows, it would work, wouldn't it?

Since the major use-case for this is source files being changed in an editor while running ninja and the unknown inputs are generated files in the build directory most of the time (correct me if I'm wrong here), I think that would be good enough while not being 100% accurate.

jdrouhard:
> > If it doesn't know the inputs until after the command, [...]
> 
> 
> 
> It doesn't know *all* inputs. For the inputs it knows, it would work, wouldn't it?
> 
> 
> 
> Since the major use-case for this is source files being changed in an editor while running ninja and the unknown inputs are generated files in the build directory most of the time (correct me if I'm wrong here), I think that would be good enough while not being 100% accurate.


It wouldn't work even with inputs it does know unless we stat every input right before every command. This was exactly what #1753 did but was rolled back due to performance concerns and CMake test suite failures. #1936 was the follow up that did the same thing with the CMake tests fixed, but it still had the performance issue of way too many stat calls.

This PR _does_ work 100% of the time with _less_ performance impact since it doesn't require stat'ing every input before every command.

jhasse:
> It wouldn't work even with inputs it does know unless we stat every input right before every command.

Why not?

src/build.cc
```diff
@@ -745,55 +764,42 @@ bool Builder::FinishCommand(CommandRunner::Result* result, string* err) {
   }
 
   // Restat the edge outputs
-  TimeStamp output_mtime = 0;
-  bool restat = edge->GetBindingBool("restat");
+  TimeStamp record_mtime = 0;
   if (!config_.dry_run) {
+    bool const restat = edge->GetBindingBool("restat");
+    bool const generator = edge->GetBindingBool("generator");
```
jhasse, line 770: please use west-const

jdrouhard:
> > It wouldn't work even with inputs it does know unless we stat every input right before every command.
> 
> 
> 
> Why not?

This has been discussed at length in those other PRs I mentioned. I'm pretty sure you were part of those discussions. 

In any case, the build log needs to remember an mtime it can use to compare against on subsequent runs. Where do we source that mtime? It has to come from something at the time the build command begins or we lose the "anchor" time for knowing whether something was changed since the inputs were read by whatever command produced the output.

If we get it from the most recent mtime, then we have to know the most recent input's mtime when the build command starts. That's #1753 and #1936. To know the most recent input's mtime is a performance bottleneck since we have to stat every known input in order to find that mtime. And if we don't know all the inputs, it won't be an accurate time anyway.

Currently ninja doesn't track anything like that at all, it simply records the output's mtime after the command finishes.

Like I said, this attempt (most recent input mtime) was tried in those PRs. You were willing to give them a shot, I'm asking you give this PR the same as it solves the problem better. Let's just let it hit master do it gets more exposure for awhile.

jhasse:
> In any case, the build log needs to remember an mtime it can use to compare against on subsequent runs.

Yes, but only because we always compare against that mtime. Couldn't ninja use the output mtime instead if it was comparing against an (at the time of the previous run) unknown input?

src/build.cc
```diff
@@ -679,14 +687,25 @@ bool Builder::StartEdge(Edge* edge, string* err) {
...
+
+  // Create directories necessary for outputs and remember the current
+  // filesystem mtime to record later
   // XXX: this will block; do we care?
   for (vector<Node*>::iterator o = edge->outputs_.begin();
        o != edge->outputs_.end(); ++o) {
     if (!disk_interface_->MakeDirs((*o)->path()))
       return false;
+    if (build_start == -1) {
+      disk_interface_->WriteFile(tmp_mtime_file_path_, "");
```
jhasse, line 700: Which command updates the mtime in `RealDiskInterface::WriteFile`? Couldn't `fwrite` just do nothing if it was passed an empty buffer?

src/graph.cc
```diff
@@ -240,37 +240,34 @@ bool DependencyScan::RecomputeOutputDirty(const Edge* edge,
...
+  // output file's actual mtime and simply check the recorded mtime from
+  // the log against the most recent input's mtime (see below)
+  bool used_restat = false;
+  if (edge->GetBindingBool("restat") && build_log() &&
+      (entry = build_log()->LookupByOutput(output->path()))) {
+    used_restat = true;
+  }
+
+  // Dirty if the output is older than the input.
+  if (!used_restat && most_recent_input && output->mtime() < most_recent_input->mtime()) {
```
jhasse, line 264: `output->mtime()` is the `command_start_time_` of the previous run, right?

src/build_test.cc
jhasse: Still, I would rather remove the explicit check for `ASSERT_EQ(files_created + 2, fs_.files_created_.size());`. *This* unit test shouldn't care.

src/build_test.cc
```diff
@@ -1646,10 +1700,56 @@ TEST_F(BuildWithLogTest, RestatMissingInput) {
...
+  ASSERT_EQ("", err);
+  EXPECT_EQ(1u, command_runner_.commands_ran_.size());
+  EXPECT_EQ(1u, builder_.plan_.command_edge_count());
+}
+
 TEST_F(BuildWithLogTest, GeneratedPlainDepfileMtime) {
   ASSERT_NO_FATAL_FAILURE(AssertParse(&state_,
 "rule generate-depfile\n"
-"  command = touch $out ; echo \"$out: $test_dependency\" > $depfile\n"
+"  command = sleep 1 ; touch $out ; echo \"$out: $test_dependency\" > $depfile\n"
```
jhasse, line 1752: Why this change?

jdrouhard:
> > In any case, the build log needs to remember an mtime it can use to compare against on subsequent runs.
> 
> 
> 
> Yes, but only because we always compare against that mtime. Couldn't ninja use the output mtime instead if it was comparing against an (at the time of the previous run) unknown input?

You are now suggesting a build log version bump where we record both the output's actual mtime and a maybe-incorrect most recent input mtime in the build log? We would _also_ have to be able to identify which of those two we should use based on context we don't have. Somehow we'd have to track whether the known input set changed for a particular output since the last run, and that gets ambiguous (not possible) with restat and generator rules which create inputs for other outputs, including order only dependencies that create inputs for other edges when they run.

Doing it as you suggest is actually a more intrusive and scary change for a solution that wouldn't even provide the same safety as this PR.

I'm not sure why we are drilling down on this again when we've already discussed this at length. Is there an issue with this PR or are you just trying to understand it better?

src/build_test.cc
```diff
@@ -627,14 +628,32 @@ bool FakeCommandRunner::StartCommand(Edge* edge) {
     fs_->Create(dep, "");
   } else if (edge->rule().name() == "generate-depfile") {
     string dep = edge->GetBinding("test_dependency");
+    bool touch_dep = edge->GetBindingBool("touch_dependency");
     string depfile = edge->GetUnescapedDepfile();
+    fs_->Tick();
```
jhasse, line 633: Move this into `if (touch_dep)`?

src/build.cc
jdrouhard: fwrite will update the file to be empty if it is passed an empty buffer, it can't simply decide to do nothing.

src/graph.cc
jdrouhard: Here, `output->mtime()` is the actual mtime of the file on disk when this build was started (it's calculated the first time the output is considered when generating the build graph).

src/build_test.cc
jdrouhard: Ok, I'll remove it.

src/build_test.cc
jdrouhard: These command lines aren't used by the test runner, they are "documentation" for what's happening in the fake command runner, so I was just updating this to match what this rule will actually do in the test.

jhasse:
I'm reluctant to merge this PR because I think it's an invasive change to how ninja works - after this we won't be comparing output mtimes anymore. As you notice I haven't fully understood all the edge cases, too, and don't want to do a revert again.

There's also the problem that we don't have any good performance testing in place so the implications are *probably* small. Hopefully. For every use-case?

Another thing that would need to change, but it's a bigger issue of ninja: The unit tests only use the `VirtualFileSystem` and not the real thing. That's a problem especially for stuff that relies on filesystem behavior like this PR.

Last but not least I always thought about adding a locking mechanism to ninja, so that two runs don't interfere with each other. Maybe we could re-use `.ninja_tmp` for that (and probably rename it to `ninja.lock` or something like that).

Another idea to think about would be to re-use `.ninja_log` to get the mtime. Not sure if that is feasible.

> Somehow we'd have to track whether the known input set changed for a particular output since the last run, and that gets ambiguous (not possible) with restat and generator rules which create inputs for other outputs, including order only dependencies that create inputs for other edges when they run.

Why wouldn't it be possible?

src/build.cc
jhasse: what if it is already empty?

jdrouhard: It will still write the empty string to the file and update the mtime.

jhasse: Is this documented specified somewhere?

Wouldn't matter if we had real-world tests we could run on all systems :(

jdrouhard:
> I'm reluctant to merge this PR because I think it's an invasive change to how ninja works - after this we won't be comparing output mtimes anymore. As you notice I haven't fully understood all the edge cases, too, and don't want to do a revert again.

I understand. I know it's a big change, but with any good improvements like this, there's a bit of risk involved. It solves a number of issues that have been opened related to this problem and definitely makes ninja more resilient to real word usages.

In this case, it will be still be comparing output mtimes, but what it compares against is sometimes the time the command that produced it began, and other times is exactly what it is now (mtime of the output when it finished which is the case for restat and generator rules).


> There's also the problem that we don't have any good performance testing in place so the implications are *probably* small. Hopefully. For every use-case?

This is true. The only way to gauge the true performance impact is to let it be used in as many real world situations as it can before a release is made with the change in it. If you have suggestions for how we can increase the confidence level in the performance impact, I'm all ears. My anecdotal evidence is from a Linux perspective where I daily run builds with 4000 edges that each have hundreds of inputs. There's no discernible performance impact vs current master.

> Another thing that would need to change, but it's a bigger issue of ninja: The unit tests only use the `VirtualFileSystem` and not the real thing. That's a problem especially for stuff that relies on filesystem behavior like this PR.

Agreed. We can unit test all day long but if a real file system doesn't behave exactly the way a test author assumes, the unit tests won't capture real world behavior.

System tests like these are typically more brittle as they will rely on factors outside the control of the test, but having some would be a good step forward in any case. 

> Last but not least I always thought about adding a locking mechanism to ninja, so that two runs don't interfere with each other. Maybe we could re-use `.ninja_tmp` for that (and probably rename it to `ninja.lock` or something like that).

That seems like a good change. Maybe a follow up to this PR after it's merged?

> > Somehow we'd have to track whether the known input set changed for a particular output since the last run, and that gets ambiguous (not possible) with restat and generator rules which create inputs for other outputs, including order only dependencies that create inputs for other edges when they run.
> 
> 
> Why wouldn't it be possible?

Without actually trying this change, I can't say for certain, but after my experience with the various attempts to solve this problem, I know that it's not quite so simple as tracking the inputs for an edge. There are some unique rules that can exist that don't really mesh with simply checking whether inputs have changed.

It might be possible. I do think l, however, that this change would be even more invasive than what this PR does, and would only _sometimes_ help the resiliency.

jhasse:
Thank you for being so patient with me :) Sorry it all takes so long!

> If you have suggestions for how we can increase the confidence level in the performance impact, I'm all ears.

Fix #1733 and let a script run on CI (although I'm not how sure how consistent performance is during a run on GitHub Actions - between runs it probably is comparable).

src/build.cc
jdrouhard: https://en.cppreference.com/w/c/io/fwrite

While it's ultimately up to the filesystem implementation to update the mtime if a file is written to, fwrite() does guarantee that the file is written to. We specifically take advantage of using the filesystem for acquiring an mtime because we compare this acquired time to the input mtimes on subsequent runs, which are presumably residing on the same filesystem (or one with times that are closely in sync).

src/build.cc
jdrouhard: done.

src/build_test.cc
jdrouhard: changed back since this rule isn't touching the dependency and i moved the sleep into the if block as suggested.

src/build_test.cc
jdrouhard: done.

jdrouhard:
> Thank you for being so patient with me :) Sorry it all takes so long!

No worries, thanks for seeing this through. I've been trying to get this problem fixed for about 5 years so I'm glad there's finally traction for it.

> > If you have suggestions for how we can increase the confidence level in the performance impact, I'm all ears.
> 
> Fix #1733 and let a script run on CI (although I'm not how sure how consistent performance is during a run on GitHub Actions - between runs it probably is comparable).

Are you making #1733 a prerequisite for getting this PR merged, or can we merge this, assuming we can get to #1733 before the next release?

src/build.cc
jhasse: > If size or count is zero, fwrite returns zero and performs no other action.

Sounds like it doesn't write to the file.

jdrouhard: Apologies, it's not `fwrite()` that causes the mtime to be updated, it's the `fopen(path, "w")` call that does so. The fact that we're writing 0 bytes is actually an optimization that means the `fwrite()` call itself is a no-op (like you say).

`fopen()` creates the file if it does not exist, or deletes (overwrites) the file if it already exists. Either way, mtime is updated.

jdrouhard:
ping @jhasse -- any other comments before you get this merged in?

jhasse:
> Are you making #1733 a prerequisite for getting this PR merged [...]

No, but I think I will release 1.11 first, then merge this PR into master to get some more testing. Maybe someone will fix #1733 along the way, but I doubt it :(

jdrouhard:
Ping @jhasse @nico 

This has been thoroughly tested by the full cmake test suite and we've been using this patch at my place of work for nearly a year with zero issues. 

I know you wanted to release a new version before merging this so it could have maximal time in HEAD for testing, but it doesn't seem like a release is coming soon. 

Can we get this merged so we can get more exposure? I don't think this PR is going to have any discernible performance impact.

jdrouhard:
> > Are you making #1733 a prerequisite for getting this PR merged [...]
> 
> No, but I think I will release 1.11 first, then merge this PR into master to get some more testing. Maybe someone will fix #1733 along the way, but I doubt it :(

Looks like 1.11.0 was released recently. Congrats on the release! Looks like a great one. I rebased this PR on current master. Could we get it merged soon to let this bake as long as possible before the next release? Thanks @jhasse!

jdrouhard:
I don’t have a preference whatsoever on the lock file name. Which do you prefer?

jhasse:
Let's use `.ninja_lock` which fits in with the other files.

In a future PR we would then write the PID to this file to check if it's a zombie file, right? I don't have any experience in how to do directory locking :D

jdrouhard:
> Let's use `.ninja_lock` which fits in with the other files.

Done.

> In a future PR we would then write the PID to this file to check if it's a zombie file, right? I don't have any experience in how to do directory locking :D

On Linux at least, this is pretty easy. We can make ninja acquire a lock on this file before running, which would block other ninja processes from running concurrently. I'm not as familiar with the windows side of things but I'm sure something similar can be achieved.

craigscott-crascit:
>We can make ninja acquire a lock on this file before running, which would block other ninja processes from running concurrently.

Some projects implement logic that invokes the build within a build (nested invocation). I'm not saying that's a good pattern, but I've seen that in real world projects more than once. It's often used to run some kind of test or utility task during the build. For such logic, if ninja blocks another ninja process from running in a build directory, then such child ninja invocations on the same build will never get a chance to run and will appear to hang.

jhasse:
Good point. We could provide a flag to disable the check for such setups (and for speed).

PR #1949, opened 2021-04-08
Consider the remaining job capacity in main loop

Flowdalic:
This changes CanRunMore() to return an int instead of a bool. The
return value is the remaining capacity of new jobs that can be spawned
without saturating a potential enabled load limitation (if ninja's -l
option is used). We assume that every started edge increases the load
by one, hence the available "load capacity" is the maximum load minus
the current load.

So, instead of spawning new jobs only limited by the remaining job
capacity ("-j") if the load average limitation is not yet reached, we
only spawn so many new jobs that would saturate the load limitation.

This helps to speed up multiple parallel builds on the same host, as
the following benchmark shows. We compare the total build times of 8
parallel builds of LLVM on a 256-core system using "ninja -j 258".

ninja-master:        1351 seconds
ninja-load-capacity:  920 seconds

That is, with this commit, the whole process becomes 1.46× faster.

The used benchmark script created and prepared 8 build directories,
records the start time, spawns 8 subshells invoking "ninja -j 258",
awaits the termination of those subshells, and records the end
time. Besides the total running time, it also outputs /proc/loadavg,
provides an indication of where the performance is gained:

ninja-master:         3.90  93.94 146.38 1/1936 209125
ninja-load-capacity: 92.46 210.50 199.90 1/1936 36917

So with this change, ninja uses the available hardware cores better in
the presence of competing ninja processes, while avoiding to overload
the system.

The used benchmark script for reference:
```
#!/usr/bin/env bash
set -euo pipefail

VANILLA_NINJA=~/code/ninja-master/build/ninja
LOAD_CAPACITY_AWARE_NINJA=~/code/ninja-load-capacity/build/ninja

CMAKE_NINJA_PROJECT_SOURCE=~/code/llvm-project/llvm

declare -ir PARALLEL_BUILDS=8
readonly TMP_DIR=$(mktemp --directory --tmpdir=/var/tmp)

cleanup() {
	rm -rf "${TMP_DIR}"
}
trap cleanup EXIT

BUILD_DIRS=()
for i in $(seq 1 ${PARALLEL_BUILDS}); do
	BUILD_DIR="${TMP_DIR}/${i}"
	mkdir "${BUILD_DIR}"
	(
		cd "${BUILD_DIR}"
		cmake -G Ninja "${CMAKE_NINJA_PROJECT_SOURCE}" &> "${BUILD_DIR}/build.log"
	)&
	BUILD_DIRS+=("${BUILD_DIR}")
done
wait

NPROC=$(nproc)
MAX_LOAD=$(( NPROC + 2 ))
SLEEP_SECONDS=300

for NINJA_BIN in "${VANILLA_NINJA}" "${LOAD_CAPACITY_AWARE_NINJA}"; do
	for BUILD_DIR in "${BUILD_DIRS[@]}"; do
		(
			"${NINJA_BIN}" -C "${BUILD_DIR}" clean &> "${BUILD_DIR}/build.log"
		)&
	done
	wait

	echo "Starting build with ${NINJA_BIN} using -j ${MAX_LOAD}"
	START=$(date +%s)
	for BUILD_DIR in "${BUILD_DIRS[@]}"; do
		(
			"${NINJA_BIN}" -C "${BUILD_DIR}" -l "${MAX_LOAD}" &> "${BUILD_DIR}/build.log"
		)&
	done
	wait
	STOP=$(date +%s)

	DELTA_SECONDS=$((STOP - START))
	echo "Using ${NINJA_BIN} to perform ${PARALLEL_BUILDS} of ${CMAKE_NINJA_PROJECT_SOURCE}"
	echo "took ${DELTA_SECONDS} seconds on this ${NPROC} core system using -j ${MAX_LOAD}"
	echo "/proc/loadavg:"
	cat /proc/loadavg
	echo "ninja --version:"
	"${NINJA_BIN}" --version

	echo "Sleeping "${SLEEP_SECONDS}" seconds to bring system into quiescent state"
	sleep ${SLEEP_SECONDS}
done
```

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -18,6 +18,8 @@
 #include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <climits>
+#include <stdint.h>
 #include <functional>
 
 #if defined(__SVR4) && defined(__sun)
@@ -46,7 +48,7 @@ struct DryRunCommandRunner : public CommandRunner {
   virtual ~DryRunCommandRunner() {}
 
   // Overridden from CommandRunner:
-  virtual bool CanRunMore() const;
+  virtual size_t CanRunMore() const;
   virtual bool StartCommand(Edge* edge);
   virtual bool WaitForCommand(Result* result);
 
@@ -54,8 +56,8 @@ struct DryRunCommandRunner : public CommandRunner {
   queue<Edge*> finished_;
 };
 
-bool DryRunCommandRunner::CanRunMore() const {
-  return true;
+size_t DryRunCommandRunner::CanRunMore() const {
+  return SIZE_MAX;
 }
 
 bool DryRunCommandRunner::StartCommand(Edge* edge) {
@@ -455,7 +457,7 @@ void Plan::Dump() const {
 struct RealCommandRunner : public CommandRunner {
   explicit RealCommandRunner(const BuildConfig& config) : config_(config) {}
   virtual ~RealCommandRunner() {}
-  virtual bool CanRunMore() const;
+  virtual size_t CanRunMore() const;
   virtual bool StartCommand(Edge* edge);
   virtual bool WaitForCommand(Result* result);
   virtual vector<Edge*> GetActiveEdges();
@@ -478,12 +480,26 @@ void RealCommandRunner::Abort() {
   subprocs_.Clear();
 }
 
-bool RealCommandRunner::CanRunMore() const {
+size_t RealCommandRunner::CanRunMore() const {
   size_t subproc_number =
       subprocs_.running_.size() + subprocs_.finished_.size();
-  return (int)subproc_number < config_.parallelism
-    && ((subprocs_.running_.empty() || config_.max_load_average <= 0.0f)
-        || GetLoadAverage() < config_.max_load_average);
+
+  int64_t capacity = config_.parallelism - subproc_number;
+
+  if (config_.max_load_average > 0.0f) {
+    int load_capacity = config_.max_load_average - GetLoadAverage();
+    if (load_capacity < capacity)
+      capacity = load_capacity;
+  }
+
+  if (capacity < 0)
+    capacity = 0;
+
+  if (capacity == 0 && subprocs_.running_.empty())
+    // Ensure that we make progress.
+    capacity = 1;
+
+  return capacity;
 }
 
 bool RealCommandRunner::StartCommand(Edge* edge) {
@@ -634,8 +650,13 @@ bool Builder::Build(string* err) {
   // Second, we attempt to wait for / reap the next finished command.
   while (plan_.more_to_do()) {
     // See if we can start any more commands.
-    if (failures_allowed && command_runner_->CanRunMore()) {
-      if (Edge* edge = plan_.FindWork()) {
+    if (failures_allowed) {
+      size_t capacity = command_runner_->CanRunMore();
+      while (capacity > 0) {
+        Edge* edge = plan_.FindWork();
+        if (!edge)
+          break;
+
         if (edge->GetBindingBool("generator")) {
           scan_.build_log()->Close();
         }
@@ -654,11 +675,19 @@ bool Builder::Build(string* err) {
           }
         } else {
           ++pending_commands;
-        }
 
-        // We made some progress; go back to the main loop.
-        continue;
+          --capacity;
+
+          // Re-evaluate capacity.
+          size_t current_capacity = command_runner_->CanRunMore();
+          if (current_capacity < capacity)
+            capacity = current_capacity;
+        }
       }
+
+       // We are finished with all work items and have no pending
+       // commands. Therefore, break out of the main loop.
+       if (pending_commands == 0 && !plan_.more_to_do()) break;
     }
 
     // See if we can reap any finished commands.
--- a/src/build_test.cc
+++ b/src/build_test.cc
@@ -15,6 +15,8 @@
 #include "build.h"
 
 #include <assert.h>
+#include <climits>
+#include <stdint.h>
 
 #include "build_log.h"
 #include "deps_log.h"
@@ -473,7 +475,7 @@ struct FakeCommandRunner : public CommandRunner {
       max_active_edges_(1), fs_(fs) {}
 
   // CommandRunner impl
-  virtual bool CanRunMore() const;
+  virtual size_t CanRunMore() const;
   virtual bool StartCommand(Edge* edge);
   virtual bool WaitForCommand(Result* result);
   virtual vector<Edge*> GetActiveEdges();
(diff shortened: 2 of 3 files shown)


Flowdalic:
I am not sure if the Windows build failure is caused by my changes or something else. Since I am not a windows person, any help is appreciated. Thanks. :)

jhasse:
Why is this faster?

Flowdalic:
> Why is this faster?

Due the reduction of the total number of involved context switches.

I tried to explain the situation better in the new commit message, aver realizing that it is maybe not so obvious and that I failed to do so in the initial commit message. Please have a look at the new commit message and let me know if you have any further questions. :)

jhasse:
Thanks! The new commit message is great.

I'm not a user of the load limit parameter, maybe someone else can comment / review.

The Windows build failure is a general problem (see master branch) which I don't understand either :(

Flowdalic:
Friendly reminder :)

Flowdalic:
@jhasse friendly reminder. Anything I can do to get this merged?

src/build_test.cc
```diff
@@ -473,7 +474,7 @@ struct FakeCommandRunner : public CommandRunner {
       max_active_edges_(1), fs_(fs) {}
 
   // CommandRunner impl
-  virtual bool CanRunMore() const;
+  virtual int CanRunMore() const;
```
jonesmz, line 477: I think this should be returning a size_t.

You can't have a negative capacity, only 0 or more.

int also has platform specific limitations. it's 32bit, even on a 64bit machine.

Flowdalic:
Looks like someone needs to approve the CI workflows. Thanks.

jonesmz:
@Flowdalic it's not that the CI workflow needs approval.

It's that you have compiler errors that the CI workflow caught.

```
 In file included from /usr/include/c++/4.8.2/cstdint:35:0,
                 from /__w/ninja/ninja/src/build.cc:22:
/usr/include/c++/4.8.2/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support is currently experimental, and must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
 #error This file requires compiler and library support for the \
```

Personally, i don't think NInja should be requiring C++03 or earlier for Unix build targets, but it does.

So either re-work your code to not trigger these errors, or put your foot down and only offer these changes with C++11 support.

Flowdalic:
> @Flowdalic it's not that the CI workflow needs approval.

I am not sure about. I see here 

> 3 workflows awaiting approval
> First-time contributors need a maintainer to approve running workflows. Learn more.
> 1 successful check

```
/usr/include/c++/4.8.2/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support is currently experimental, and must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
 #error This file requires compiler and library support for the \
```

fixed by replacing `cstdint` with `stdint.h`.

Flowdalic:
Squashed everything into a single commit. Someone may want to approve the workflows. Anything else to get this merged?

Flowdalic:
Friendly Ping. Anything I can do to get this merged?

Flowdalic:
Another friendly Ping. Anything I can do to get this merged?

src/build.cc
```diff
@@ -18,6 +18,8 @@
 #include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <climits>
+#include <stdint.h>
```
jonesmz, line 22: cstdint

src/build.cc
```diff
@@ -46,16 +48,16 @@ struct DryRunCommandRunner : public CommandRunner {
...
   virtual bool WaitForCommand(Result* result);
 
  private:
   queue<Edge*> finished_;
 };
 
-bool DryRunCommandRunner::CanRunMore() const {
-  return true;
+size_t DryRunCommandRunner::CanRunMore() const {
+  return SIZE_MAX;
```
jonesmz, line 60: std::numeric_limits<size_t>::max()

src/build.cc
```diff
@@ -460,12 +462,26 @@ void RealCommandRunner::Abort() {
...
 
-bool RealCommandRunner::CanRunMore() const {
+size_t RealCommandRunner::CanRunMore() const {
   size_t subproc_number =
       subprocs_.running_.size() + subprocs_.finished_.size();
-  return (int)subproc_number < config_.parallelism
-    && ((subprocs_.running_.empty() || config_.max_load_average <= 0.0f)
-        || GetLoadAverage() < config_.max_load_average);
+
+  long capacity = config_.parallelism - subproc_number;
```
jonesmz, line 469: No code should ever use `long`. It is evil.
It changes size based on the operating system used, and not the underlying processor architecture. This leads to lots of surprises.

Use either `size_t` or `ssize_t`

src/build_test.cc
```diff
@@ -15,6 +15,8 @@
 #include "build.h"
 
 #include <assert.h>
+#include <climits>
+#include <stdint.h>
```
jonesmz, line 19: cstdint

src/build.cc
Flowdalic: The logic requires `capacity` to become negative, so `size_t` is not an option. And since `ssize_t` only guarantees to be able to hold one negative value, i.e.,  `-1`, it is also appears to be no viable option here.

Flowdalic:
Noe that I had to revert the suggested changes do ninja not support C++11 yet.

src/build.cc
```diff
@@ -473,12 +475,26 @@ void RealCommandRunner::Abort() {
...
+  if (config_.max_load_average > 0.0f) {
+    int load_capacity = config_.max_load_average - GetLoadAverage();
+    if (load_capacity < capacity)
+      capacity = load_capacity;
+  }
+
+  if (capacity < 0)
+    capacity = 0;
+
+  if (!capacity && subprocs_.running_.empty())
```
jhasse, line 493: can you change this to `capacity == 0` to make it a bit more clear?

src/build.cc
jhasse: can you use int64_t instead?

src/build.cc
```diff
@@ -649,11 +670,17 @@ bool Builder::Build(string* err) {
...
+          --capacity;
+
+          // Re-evaluate capacity.
+          size_t current_capacity = command_runner_->CanRunMore();
+          if (current_capacity < capacity)
+            capacity = current_capacity;
+        }
       }
+
+      if (!plan_.more_to_do()) break;
```
jhasse, line 683: isn't this redundant?

Flowdalic:
@jhasse thanks for your review. I've addressed your comments.

src/build.cc
Flowdalic: No, otherwise you may end up in the `ninja: build stopped: stuck [this is a bug].` case at the end of a build. However, I wonder if this should be

`if (pending_commands == 0 && !plan_more_to_do()) break;`

Flowdalic: Yep, otherwise we may no reap the last few finished commands, fixed that.

Issue #1946, opened 2021-03-30
Build outputs modified after the build outside of ninja are not rebuilt


Dragnalith:
In the case you manually modified a build output, the output will not be rebuilt to its correct state because ninja does not consider it dirty. I think the behavior should be considered as bug, because _having a build node is in a state different from the one it would have if the rule command of its edge was executed again_ is the definition of _being dirty_.

I have a simple fix the following PR #1945.

PR #1951, opened 2021-04-14
Make outputs modified outside of the build system considered dirty

Dragnalith:
This pull request is a fix for #1946.

Today if an output of the build is modified from outside of the build system (manually or by another process) ninja will not consider it to be dirty.

I think if an output state is different from the state it would have if the rule command was executed again, then it should be considered dirty. (Edit: Since in some scenario like when you consciously modified outputs for quick testing you may not want ninja to override your change, my implementation makes it optional. For backward compatibility default ninja keep the same behavior, but if run with -m flag, it will considered modified outputs as dirty).

Situation where it is a problem: 
- You have an action registered as order-only dependency which generate some C++ header. 
- You include one of generated file for the time to one of your .cpp, so ninja does not consider it yet as implicit dependency. 
- You open the generated file with your favorite text editor, and by mistyping on your keyboard you modify the generated header to an invalid C++ state.
- As consequence, ninja will not try to re-generate the header, because it is not yet a dependency of the build, and ninja will fail the build because the header is not valid C++.

My modification considered an output is dirty if its mtime is more recent than the one in the build log. The mtime of the output can be older than the one on the build log in the case its build edge is mark with 'restat = 1'. In that case if the rule command did not modified the output, its build log entry will inherit the mtime of its most recent inputs or depfile.


bradking:
Cc: @jdrouhard

jdrouhard:
Few thoughts:

1. This is incompatible with #1943 as it stands. That PR has been reviewed and is just waiting on being merged. If I understand what this PR is trying to solve, I think #1943 does it as well. It changes how the mtime stored in the log is calculated and used, and should be resilient to dependencies (even undiscovered ones) being changed while running a build. It doesn't fix changing leaf node outputs after a run has completed and expecting them to rebuild, however.
2. If we want to make ninja more generally consider outputs that have changed since the previous run as dirty, I think that merits broader discussion. Build systems generally only care about whether the inputs that produce the outputs have changed, not whether outputs themselves might be different. I'm not directly opposed to the idea, but I think we'd need a build log version bump and we'd need to restat every output when commands have finished (not just generator and restat rules) and store that mtime separately (and use it only to compare the output's mtime on subsequent builds).
3. I don't think adding a command line option is appropriate here. ninja should either always do this or not.

Dragnalith:
@jdrouhard 

I have looked at #1943 and I understand what it solves and how it is incompatible with my PR.

Actually it is not an implementation incompatibility, it is a ninja specification incompatibility: _is the purpose of a ninja build to give the same result as a full rebuild?_

If the answer is yes, the problem addressed by #1943 (i.e what happen if inputs change after starting executing a rule command but before the output has been written) should lead to a ninja build failure. The current solution of #1943 is to succeed the build, but consider the edge dirty, so it will re-execute next time ninja is called.

I understand it may exist reason that in practice we do not want ninja to tend to a correct incremental build.

Dragnalith:
@jdrouhard I think the only to make both approaches compatible will be to add a new column the build log to store the actual mtime of the output. There would be two mtimes: `record_mtime` and `actual_mtime`. `record_mtime` would be used as it is today, to know if an output should be rebuilt, and `actual_mtime` would let ninja know if the output has been modified since the last time it has been logged.

Issue #1954, opened 2021-04-16
Fully suppress ninja status when NINJA_STATUS is the empty string


EliRibble:
Currently if a user sets the environment variable `NINJA_STATUS` to the empty string then `ninja` will show status lines with no header but include each edge's description.

Instead `ninja` should detect when the environment variable is empty and fully suppress status output.

This is useful in situations where ninja is writing output to a file and the user wants the file to be very short and only show failure information, like capturing output in a build system.

jhasse:
What if the user wants to only show edge description? ;)

It's too bad that the description wasn't part of `NINJA_STATUS` (see #1045) from the start. Changing the behavior now is prone to cause users too complain after updating.

Also see #480.

EliRibble:
Okay, so would you accept a PR that implements #480 ? Because I could write that instead...

jhasse:
Check out #1818. It's basically done and just needs to be rebased against master and have the small change applied that I've suggested in my last comment.

PR #1955, opened 2021-04-16
Suppress progress output when NINJA_STATUS is empty.

EliRibble:
If log verbosity is set to fully verbose it overrides the NINJA_STATUS
environment variable. Otherwise, this allows users to suppress most
build output on very large builds.

Issue: https://github.com/ninja-build/ninja/issues/1954


jhasse:
Closing in favor of #1968.

PR #1963, opened 2021-04-26
Reliable ETA and progress percentage.

### LINKED ISSUE #115: Estimate time left on build

LebedevRI:
Reliable ETA and progress percentage.

This has been bugging me for *years*. :)
Count of finished edges isn't a great statistic, it isn't
really obvious if LLVM will take 8 minues to build, or 10 minutes.

But, it's actually pretty straight-forward to get some
more useful information. We already know how much time each edge
has taken, so we could just do the dumb thing, and assume that
every edge in the plan takes the same amount of time.

Or, we can do better. `.ninja_log` already contains
the historical data on how long each edge took to produce it's outs,
so we simply need to ensure that we populate edges with that info,
and then we can greatly improve our predictions.
The math is pretty simple i think.

This is largely a port of a similar change i did to LLVM LIT:
https://reviews.llvm.org/D99073

With this, i get something quite lovely:
```
llvm-project/build-Clang12$ NINJA_STATUS="[%f/%t %p %P][%w + %W] " /repositories/ninja/build-Clang-debug/ninja opt
[288/2527  11%   4%][00:27 + 08:52] Building CXX object lib/DebugInfo/CodeView/CMakeFiles/LLVMDebugInfoCodeView.dir/AppendingTypeTableBuilder.cpp.o
```

I hope people will find this useful, and it could be merged.
CC @nico 

Please let me know which kinds of test coverage this needs?

Fix #115.

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -204,7 +204,11 @@ Several placeholders are available:
 `%o`:: Overall rate of finished edges per second
 `%c`:: Current rate of finished edges per second (average over builds
 specified by `-j` or its default)
-`%e`:: Elapsed time in seconds.  _(Available since Ninja 1.2.)_
+`%e`:: Elapsed time in seconds. _(Available since Ninja 1.2.)_
+`%E`:: Remaining time (ETA) in seconds. _(Available since Ninja 1.12.)_
+`%w`:: Elapsed time in [h:]mm:ss format. _(Available since Ninja 1.12.)_
+`%W`:: Remaining time (ETA) in [h:]mm:ss format. _(Available since Ninja 1.12.)_
+`%P`:: The percentage (in ppp% format) of time elapsed out of predicted total runtime. _(Available since Ninja 1.12.)_
 `%%`:: A plain `%` character.
 
 The default progress status is `"[%f/%t] "` (note the trailing space
--- a/src/graph.h
+++ b/src/graph.h
@@ -250,6 +250,10 @@ struct Edge {
   bool is_phony() const;
   bool use_console() const;
   bool maybe_phonycycle_diagnostic() const;
+
+  // Historical info: how long did this edge take last time,
+  // as per .ninja_log, if known? Defaults to -1 if unknown.
+  int64_t prev_elapsed_time_millis = -1;
 };
 
 struct EdgeCmp {
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -156,6 +156,10 @@ struct NinjaMain : public BuildLogUser {
   /// @return true if the manifest was rebuilt.
   bool RebuildManifest(const char* input_file, string* err, Status* status);
 
+  /// For each edge, lookup in build log how long it took last time,
+  /// and record that in the edge itself. It will be used for ETA predicton.
+  void ParsePreviousElapsedTimes();
+
   /// Build the targets listed on the command line.
   /// @return an exit code.
   int RunBuild(int argc, char** argv, Status* status);
@@ -287,6 +291,19 @@ bool NinjaMain::RebuildManifest(const char* input_file, string* err,
   }
 
   return true;
+}
+
+void NinjaMain::ParsePreviousElapsedTimes() {
+  for (Edge* edge : state_.edges_) {
+    for (Node* out : edge->outputs_) {
+      BuildLog::LogEntry* log_entry = build_log_.LookupByOutput(out->path());
+      if (!log_entry)
+        continue;  // Maybe we'll have log entry for next output of this edge?
+      edge->prev_elapsed_time_millis =
+          log_entry->end_time - log_entry->start_time;
+      break;  // Onto next edge.
+    }
+  }
 }
 
 Node* NinjaMain::CollectTarget(const char* cpath, string* err) {
@@ -1597,6 +1614,8 @@ NORETURN void real_main(int argc, char** argv) {
       status->Error("rebuilding '%s': %s", options.input_file, err.c_str());
       exit(1);
     }
+
+    ninja.ParsePreviousElapsedTimes();
 
     int result = ninja.RunBuild(argc, argv, status);
     if (g_metrics)
--- a/src/status.cc
+++ b/src/status.cc
@@ -14,6 +14,15 @@
 
 #include "status.h"
 
+#ifdef _WIN32
+#include "win32port.h"
+#else
+#ifndef __STDC_FORMAT_MACROS
+#define __STDC_FORMAT_MACROS
+#endif
+#include <cinttypes>
+#endif
+
 #include <stdarg.h>
 #include <stdlib.h>
 
@@ -27,11 +36,9 @@
 using namespace std;
 
 StatusPrinter::StatusPrinter(const BuildConfig& config)
-    : config_(config),
-      started_edges_(0), finished_edges_(0), total_edges_(0), running_edges_(0),
-      time_millis_(0), progress_status_format_(NULL),
+    : config_(config), started_edges_(0), finished_edges_(0), total_edges_(0),
+      running_edges_(0), progress_status_format_(NULL),
       current_rate_(config.parallelism) {
-
   // Don't do anything fancy in verbose mode.
   if (config_.verbosity != BuildConfig::NORMAL)
     printer_.set_smart_terminal(false);
@@ -41,8 +48,32 @@ StatusPrinter::StatusPrinter(const BuildConfig& config)
     progress_status_format_ = "[%f/%t] ";
 }
 
-void StatusPrinter::PlanHasTotalEdges(int total) {
-  total_edges_ = total;
+void StatusPrinter::EdgeAddedToPlan(const Edge* edge) {
+  ++total_edges_;
+
+  // Do we know how long did this edge take last time?
+  if (edge->prev_elapsed_time_millis != -1) {
+    ++eta_predictable_edges_total_;
+    ++eta_predictable_edges_remaining_;
+    eta_predictable_cpu_time_total_millis_ += edge->prev_elapsed_time_millis;
+    eta_predictable_cpu_time_remaining_millis_ +=
+        edge->prev_elapsed_time_millis;
+  } else
+    ++eta_unpredictable_edges_remaining_;
+}
+
+void StatusPrinter::EdgeRemovedFromPlan(const Edge* edge) {
+  --total_edges_;
+
+  // Do we know how long did this edge take last time?
+  if (edge->prev_elapsed_time_millis != -1) {
+    --eta_predictable_edges_total_;
+    --eta_predictable_edges_remaining_;
+    eta_predictable_cpu_time_total_millis_ -= edge->prev_elapsed_time_millis;
+    eta_predictable_cpu_time_remaining_millis_ -=
+        edge->prev_elapsed_time_millis;
+  } else
+    --eta_unpredictable_edges_remaining_;
 }
 
 void StatusPrinter::BuildEdgeStarted(const Edge* edge,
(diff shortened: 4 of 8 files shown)


LebedevRI:
So far, i'm aware of two caveats:
1. All times are in milliseconds. That may or may not affect the quality of prediction. Honestly, i'm not sure why they are in milliseconds. Even if they were in nanoseconds, `int64_t` would be enough for [~3 centuries](https://www.wolframalpha.com/input/?i=%282%5E63%29-1+nanoseconds). `.ninja_log` file size conservation?
2. The prediction becomes bogus if the last rebuild was recovered from `ccache`.
   I guess we need to stop using previous times if the actual rate is very different from the predicted one.

jhasse:
Can you add "Fix #115." to the commit message?

LebedevRI:
> So far, i'm aware of two caveats:
> 
> 1. All times are in milliseconds. That may or may not affect the quality of prediction. Honestly, i'm not sure why they are in milliseconds. Even if they were in nanoseconds, `int64_t` would be enough for [~3 centuries](https://www.wolframalpha.com/input/?i=%282%5E63%29-1+nanoseconds). `.ninja_log` file size conservation?
> 2. The prediction becomes bogus if the last rebuild was recovered from `ccache`.
>    I guess we need to stop using previous times if the actual rate is very different from the predicted one.

FWIW i've fixed second point (by hardcoding some magic numbers).
First one i'm not so sure.

Will look into failing build.

mathstuf (commented):
Overall the code looks OK to me. One question I have is how this interacts with `restat = 1` removing edges and dyndeps adding and removing edges from the plan during the build. Can tests be added?

LebedevRI:
> Overall the code looks OK to me. One question I have is how this interacts with `restat = 1` removing edges and dyndeps adding and removing edges from the plan during the build.

I have no idea what `restat = 1` is, but this does gracefully handles adding/removing edges
(that haven't started yet i presume?) during the build. That was indeed important to handle.
See `builder_->status_->EdgeAddedToPlan(edge);` in `Plan::EdgeWanted()` e.g.

> Can tests be added?

I haven't looked into that yet.

LebedevRI:
Did some analysis. With clean build:
image
With rebuild from ccache:
image

For ccache the largest dt is 0.2s at 100% completion,
while for clean build the minimal dt is always larger than ~2.0s after ~5% completion.
So looks like we can say that we can reuse the previous timings if the curr/prev averages differ by less than ~10x.

[ninja_log.clean-sheet.csv](https://github.com/ninja-build/ninja/files/6409649/ninja_log.clean-sheet.csv)
[ninja_log.ccache.csv](https://github.com/ninja-build/ninja/files/6409650/ninja_log.ccache.csv)

LebedevRI:
(please don't merge. this is still broken when e.g. ccache is used)

LebedevRI:
Hm, and i just realized this is still here.

Essentially, this functionality requires that the last execution time, and the time it will take this time,
are similar. If they are *significantly* different (e.g. if on the second run it's being restored from ccache,
or it was restored from ccache last time, but this time it didn't hit), prediction will just be wrong.

As far as i can tell, the best thing we can do, is to heuristically stop using last execution time
when the new time is significantly different. It kinda works, but not really.

Can codeowners here please comment if the functionality this PR provides is acceptable given that drawback?

jonesmz:
>  If they are significantly different (e.g. if on the second run it's being restored from ccache,
or it was restored from ccache last time, but this time it didn't hit), prediction will just be wrong.

I don't think there's any meaningful way to do anything about this. It's probably best to assume multiple runs of the same target will probably take about the same amount of time. There's no other method to do the estimation without trying to add a lot of rules like "If ccache, all bets are off. If distcc, all bests are off, etc".

LebedevRI:
So i take, in other words this would be fine.
Let me rebase this...

LebedevRI:
Okay, rebased.
I've tested it somewhat, and it does appear to be still working after the rebase.
What is needed to get this through the finish line this time? :)

jonesmz:
> What is needed to get this through the finish line this time? :)

I honestly couldn't say. I'm not a maintainer, i don't have commit / merge rights.

You could create some test cases for the unit beyond the two you added.

Some interesting situations could be

1. A test where the build times do stay the same, but the order of execution changes, and the estimates should reflect that.
2. A test where the build times change a lot vs. the first run, and the estimates basically don't change for the targets that haven't yet exceeded their budget.

If you wanted to make it even more sophisticated, you could try to add some additional output like, if build times are taking half as long start adjusting expected completion down slightly for every target that took 1/2 or less as long as expected.

But I'm not sure it's worth adding additional logic like that right now.

src/graph.h
```diff
@@ -250,6 +250,10 @@ struct Edge {
   bool is_phony() const;
   bool use_console() const;
   bool maybe_phonycycle_diagnostic() const;
+
+  // Historical info: how long did this edge take last time,
+  // as per .ninja_log, if known? Defaults to -1 if unknown.
+  int64_t prev_elapsed_time;
```
jonesmz, line 256: Consider instead using std::chrono so that you can represent actual time values, and not just only, well, whatever unit this is in. Seconds?

jonesmz: Personally I'd *really* like to see std::chrono used, but it's not my call for ninja.

If this PR gets merged, i'll replace your time keeping type with std::chrono in my fork/patch series though (splinter-build, on github), because I'd like to have this change, but would rather have std::chrono :-)

src/ninja.cc
```diff
@@ -289,6 +293,22 @@ bool NinjaMain::RebuildManifest(const char* input_file, string* err,
   return true;
 }
 
+void NinjaMain::ParsePreviousElapsedTimes() {
+  for (vector<Edge*>::iterator edge = state_.edges_.begin(),
```
jonesmz, line 297: Ninja uses C++11 now.

I strongly recommend using a range-based forloop so that this nested forloop pair isn't eye-bleed to read.

src/status.cc
```diff
@@ -14,6 +14,15 @@
 
 #include "status.h"
 
+#ifdef _WIN32
+#include "win32port.h"
```
jonesmz, line 18: Is this actually necessary? As far as I know, MSVC with C++11 supports inttypes.h properly?

src/graph.h
LebedevRI: Yep, naked unitless data is a problem, i agree. I would think it would make sense to change
all relevant occurrences at once, so i'm not sure an one-off change is a win?

src/status.cc
```diff
@@ -14,6 +14,15 @@
...
+#else
+#ifndef __STDC_FORMAT_MACROS
+#define __STDC_FORMAT_MACROS
+#endif
+#include <inttypes.h>
```
jonesmz, line 23:
```
#include <cinttypes>
```

src/status.cc
```diff
@@ -27,11 +36,13 @@
...
 StatusPrinter::StatusPrinter(const BuildConfig& config)
-    : config_(config),
-      started_edges_(0), finished_edges_(0), total_edges_(0), running_edges_(0),
-      time_millis_(0), progress_status_format_(NULL),
+    : config_(config), started_edges_(0), finished_edges_(0), total_edges_(0),
+      running_edges_(0), time_millis_(0), cpu_time_millis_(0),
+      time_predicted_percentage_(0.0), eta_predictable_edges_total_(0.0),
+      eta_predictable_cpu_time_total_(0.0), eta_predictable_edges_remaining_(0),
+      eta_predictable_cpu_time_remaining_(0),
+      eta_unpredictable_edges_remaining_(0), progress_status_format_(NULL),
```
jonesmz, line 44: Now that we have C++11: For newly added member variables, lets move their initialization and default value to the same place in the class body.

Continuing to have to maintain this list of initializes is a major eye-sore and prone to mistakes.

src/status.cc
LebedevRI: Right, i agree that it should be `#include <cinttypes>`, but there are zero `#include <cinttypes>`,
and 5 `#include <inttypes.h>`...

src/status.cc
```diff
@@ -58,11 +91,96 @@ void StatusPrinter::BuildEdgeStarted(const Edge* edge,
     printer_.SetConsoleLocked(true);
 }
 
-void StatusPrinter::BuildEdgeFinished(Edge* edge, int64_t end_time_millis,
-                                      bool success, const string& output) {
+void StatusPrinter::RecalculateProgressPrediction() {
+  time_predicted_percentage_ = 0.0;
```
jonesmz, line 95: Weak recommendation:

I think this kind of routine should be separated out into a function that can be given to the main loop as a parameter

Here you use the strategy of "If so many edges match what we saw last time with X ration", but other strategies might be better suited for a given environment. E.g. My environment would be better suited with a rolling-average calculation  that I'd parameterize in the ninjafile and/or cli args. and future estimates adjusted based on the current ratio instead of clamping

src/status.cc
```diff
@@ -58,11 +91,96 @@ void StatusPrinter::BuildEdgeStarted(const Edge* edge,
...
+                            actual_average_cpu_time_millis);
+
+    // Let's say that the average times should differ by less than 10x
+    use_previous_times = ratio < 10;
+  }
+
+  int edges_with_known_runtime = finished_edges_;
+  if (use_previous_times)
+    edges_with_known_runtime += eta_predictable_edges_remaining_;
+  if (edges_with_known_runtime == 0)
```
jonesmz, line 133: Weak recommendation:

I personally find code without {}'s to be hard to reason about and understand. So for that, i'd say add curly braces. But that's not necessary or really even implied by the current codebase.

src/status.cc
```diff
@@ -58,11 +91,96 @@ void StatusPrinter::BuildEdgeStarted(const Edge* edge,
...
+  time_predicted_percentage_ = cpu_time_millis_ / total_cpu_time;
+}
+
+void StatusPrinter::BuildEdgeFinished(Edge* edge, int64_t start_time_millis,
+                                      int64_t end_time_millis, bool success,
+                                      const string& output) {
   time_millis_ = end_time_millis;
   ++finished_edges_;
 
+  int64_t elapsed = end_time_millis - start_time_millis;
```
jonesmz, line 174: elapsed is never used except for the next line. Maybe eschew the variable entirely.

src/status.h
```diff
@@ -43,10 +46,15 @@ struct Status {
 /// human-readable strings to stdout
 struct StatusPrinter : Status {
   explicit StatusPrinter(const BuildConfig& config);
-  virtual void PlanHasTotalEdges(int total);
+
+  /// Callbacks for the Plan to notify us about adding/removing Edge's.
+  virtual void EdgeAddedToPlan(const Edge* edge);
```
jonesmz, line 51: These functions override the function in the base class.

Use the "override" keyword.

If they're the most-derived function in the codebase for this class hierarchy, also use the "final" keyword to help the compiler perform function-devirtualization (important for performance)

src/status.cc
jonesmz: Ninja's current maintainer doesn't like changes of the nature "do a specific thing to the whole codebase", and prefers incremental "while we're here" changes.

src/ninja.cc
```diff
@@ -289,6 +293,18 @@ bool NinjaMain::RebuildManifest(const char* input_file, string* err,
   return true;
 }
 
+void NinjaMain::ParsePreviousElapsedTimes() {
+  for (Edge* edge : state_.edges_) {
+    for (Node* out : edge->outputs_) {
+      BuildLog::LogEntry* log_entry = build_log_.LookupByOutput(out->path());
```
jonesmz, line 299: Wow that's an *enormous* improvement in readability. Nice!

src/graph.h
jonesmz: Well this is a new feature, and i don't see any documentation in the current code for what `prev_elapsed_time` represents in terms of units.

So for this new code, it should probably be std::chrono, and if you need to adapt from something that's already present, just adapt in the few places where that needs to happen.

Other raw-integer time variables can be adjusted later by someone else.

src/status.cc
```diff
@@ -27,11 +36,9 @@
 using namespace std;
 
 StatusPrinter::StatusPrinter(const BuildConfig& config)
-    : config_(config),
-      started_edges_(0), finished_edges_(0), total_edges_(0), running_edges_(0),
-      time_millis_(0), progress_status_format_(NULL),
+    : config_(config), started_edges_(0), finished_edges_(0), total_edges_(0),
```
jonesmz, line 39: you're now initializing `started_edges_(0)` where previously this was not done.

I have no objection, just pointing it out to you.

src/status.h
LebedevRI: If i change these, the rest of the `virtual`-that-should-be-`override` get diagnosed.
I think that cleanup should be done separately from this change.

src/status.h
```diff
@@ -71,7 +79,30 @@ struct StatusPrinter : Status {
...
+  /// How much cpu clock elapsed so far?
+  int64_t cpu_time_millis_ = 0;
+
+  /// What percentage of predicted total time have elapsed already?
+  double time_predicted_percentage_ = 0.0;
+
+  /// Out of all the edges, for how many do we know previous time?
+  int eta_predictable_edges_total_ = 0;
+  /// And how much time did they all take?
+  int64_t eta_predictable_cpu_time_total_ = 0;
```
jonesmz, line 95: Either switch to std::chrono, or document the unit(s) these variables are in terms of.

src/status.h
jonesmz: How are they not already getting diagnosed? The compiler being overly helpful?

I think you should just do your new code correctly for C++11, and not worry about pre-existing code having new warnings. The codebase already has warnings anyway.

But like I said, i have no control over committing / merging, so do whatever gets your change merged the quickest.

src/graph.h
LebedevRI: My main problem with that is that it (and some other changes) goes directly against the current code style,
and having uniform code is generally better, and can be uplifted all at once,
as opposed to having a weird mismatch of different coding styles.

But if the maintainers (@jhasse ? @nico ?) can chime in and review the change as a whole,
i can sure make those changes if they are deemed better by maintainers.

src/status.cc
LebedevRI: Well, that seems surprising to me, but i don't follow this codebase to know either way, so okay then.

src/graph.h
jonesmz: There's precedent that "update a specific thing for the whole codebase" will be rejected, and only "while we're already modifying / adding, do it 'correctly'" will be accepted.

But whatever you want to do. Not my call.

src/status.cc
jonesmz: Plenty of discussion that I've had on the subject over the last few years: Most recent is here, there were other PRs where the topic is discussed, but I'm not able to find them within just a minute.
https://github.com/ninja-build/ninja/pull/2154

LebedevRI:
@jonesmz thank you for taking a look!
I fear the rest of the review notes are rather contentious, so i'd rather want to hear from maintainers first..

jonesmz:
Yep, whatever gets the feature merged is fine.

LebedevRI:
@jhasse (?) - gentle ping.

Thank you.

LebedevRI:
Re build failure: 
i wonder if instead of `curl -L -O https://github.com/Kitware/CMake/releases/download/v3.16.4/cmake-3.16.4-Linux-x86_64.sh`
it would be more stable to use https://apt.kitware.com/

doc/manual.asciidoc
```diff
@@ -204,7 +204,9 @@ Several placeholders are available:
 `%o`:: Overall rate of finished edges per second
 `%c`:: Current rate of finished edges per second (average over builds
 specified by `-j` or its default)
-`%e`:: Elapsed time in seconds.  _(Available since Ninja 1.2.)_
+`%e`:: Elapsed time in hh:mm:ss format.  _(Available since Ninja 1.2.)_
```
mathstuf, line 207: I don't think changing this is nice. Can a new specifier be used instead? Also, it seems the test suite actually gets `h:mm:ss` for less-than-10-hour estimates?

doc/manual.asciidoc
```diff
@@ -204,7 +204,9 @@ Several placeholders are available:
...
+`%E`:: Remaining time (ETA) in hh:mm:ss format.  _(Available since Ninja 1.12.0.)_
+`%P`:: The percentage of time elapsed out of predicted total runtime.  _(Available since Ninja 1.12.0.)_
```
mathstuf, line 209: A note that this includes the `%` would be handy. The padded-to-width-3 may also be relevant.

doc/manual.asciidoc
LebedevRI: Thanks for taking a look!
I've squatted on `%e` because printing it in seconds seems rather pointless to me
yet the abbrev seems appropriate.
Are there stability guarantees for these placeholders?

mathstuf: I don't know about stability guarantees, but there are definitely tools that hurk and die if a non-default `NINJA_STATUS` is used because they try to parse it for their own progress meters in the IDE UI or whatever (Qt Creator is/was one? Maybe they've gone to forcing their own `NINJA_STATUS` instead these days).

mathstuf: And I do have a concern about always including the hours is that it takes up space when ~most builds are probably well under an hour (especially incremental which is where I'd expect a human to be waiting for things and watching the `%e` anyways).

LebedevRI: I do believe both of these concerns are now addressed :)

src/status.cc
```diff
@@ -201,16 +323,74 @@ string StatusPrinter::FormatProgressStatus(const char* progress_status_format,
...
+          snprintf(buf, sizeof(buf), "?");
+        else {
+          switch (*s) {
+          case 'e':  // elapsed, seconds
+          case 'E':  // ETA, seconds
+            snprintf(buf, sizeof(buf), "%.3f", sec);
+            break;
+          case 'w':  // elapsed, human-readable
+          case 'W':  // ETA, human-readable
+            if (sec >= 60 * 60)
```
LebedevRI, line 379: Hm, but we should probably print hours for both predicates if *either* of them requires them?
It's a bit asymmetrical otherwise...

doc/manual.asciidoc
```diff
@@ -204,7 +204,11 @@ Several placeholders are available:
 `%o`:: Overall rate of finished edges per second
 `%c`:: Current rate of finished edges per second (average over builds
 specified by `-j` or its default)
-`%e`:: Elapsed time in seconds.  _(Available since Ninja 1.2.)_
+`%e`:: Elapsed time in seconds. _(Available since Ninja 1.2.)_
+`%E`:: Remaining time (ETA) in seconds. _(Available since Ninja 1.12.0.)_
+`%w`:: Elapsed time in [h:]mm:ss format. _(Available since Ninja 1.12.)_
+`%W`:: Remaining time (ETA) in [h:]mm:ss format. _(Available since Ninja 1.12.0.)_
+`%P`:: The percentage (in ppp% format) of time elapsed out of predicted total runtime. _(Available since Ninja 1.12.0.)_
```
mathstuf, line 211: Following other entries, just `1.12` seems to be more warranted to me.

src/status.cc
mathstuf: Behavior changing based on whether `%w` and `%W` are both used in the same format would be weird to me…

It could also waste horizontal space with long descriptions and end up eliding out anyways.

LebedevRI: > based on whether `%w` and `%W` are both used in the same format

That's not what i said. I wouldn't consider whether or not they are **both** present,
but merely the fact that one of them would be printed with hours,
regardless of whether or not said predicate is actually in the format string.

mathstuf: That's fine; the calculations would need to be performed ahead of time instead of as-needed when the format specifier is found then.

LebedevRI: Yay, now that is something i can live with :)

mathstuf (approved):

LebedevRI:
ping

LebedevRI:
> ping

LebedevRI:
> > ping

LebedevRI:
> > > ping

LebedevRI:
> > > > ping

@jhasse

LebedevRI:
> > > > > ping

LebedevRI:
> > > > > > ping

LebedevRI:
> > > > > > > ping

LebedevRI:
> > > > > > > > ping

mathstuf:
Please do not repeatedly "poke" issues/PRs with content-less comments such as this. Note that there has been activity lately (mostly merging bug fixes and some CI updates). Hopefully there will be more time in the future for such feature PRs to get the attention the need and deserve, but "pings" really do not help.

However, given the CI changes lately, a rebase of this PR to adapt to the new configuration is likely useful.

LebedevRI:
ping.

jhasse:
Thank you for your hard work and your patience! A great feature :)

The pings resulted in me unsubscribing and not interested in me having a look at this PR. Please refrain from doing that next time.

LebedevRI:
Wow thank you @jhasse, i had almost given up on this thing :)

> The pings resulted in me unsubscribing and not being interested in having a look at this PR. Please refrain from doing that next time.

Yeah, believe me, i very  *VERY* ***VERY*** much know the feeling :(
Sorry about that, i didn't quite know what the preferred etiquette was here, and thought i was helping.

PR #1964, opened 2021-04-29
Set output mtime of phony edges to the latest inputs

rascani:
This commit fixes issue #478.

Observed:
Real edges depending on a phony edge will not be marked as dirty or
rebuilt if the phony's (real) inputs are updated.

Expected:
An edge should always be rebuilt if its inputs or transitive inputs are
newer than the output's mtime.

Change:
Node::mtime_ was overloaded, 0 represented "does not exist". This change
disambiguates it by adding Node::exists_. Then to fix the observed
behaviour, Node::UpdatePhonyMtime was added to update the mtime if the
node does not exist.

Add tests BuildTest.PhonyUseCase# GraphTest.PhonyDepsMtimes.
Unit tests will also test for always-dirty behaviour if a phony rule has
no inputs.

This patch is a rebase of #1021 onto the master branch of 2021-04-29 with some additional changes addressing the previous review commentary.

### DIFF
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -31,7 +31,19 @@
 using namespace std;
 
 bool Node::Stat(DiskInterface* disk_interface, string* err) {
-  return (mtime_ = disk_interface->Stat(path_, err)) != -1;
+  METRIC_RECORD("node stat");
+  mtime_ = disk_interface->Stat(path_, err);
+  if (mtime_ == -1) {
+    return false;
+  }
+  exists_ = (mtime_ != 0) ? ExistenceStatusExists : ExistenceStatusMissing;
+  return true;
+}
+
+void Node::UpdatePhonyMtime(TimeStamp mtime) {
+  if (!exists()) {
+    mtime_ = std::max(mtime_, mtime);
+  }
 }
 
 bool DependencyScan::RecomputeDirty(Node* node, string* err) {
@@ -237,6 +249,14 @@ bool DependencyScan::RecomputeOutputDirty(const Edge* edge,
               output->path().c_str());
       return true;
     }
+
+    // Update the mtime with the newest input. Dependents can thus call mtime()
+    // on the fake node and get the latest mtime of the dependencies
+    if (most_recent_input) {
+      output->UpdatePhonyMtime(most_recent_input->mtime());
+    }
+
+    // Phony edges are clean, nothing to do
     return false;
   }
 
@@ -487,7 +507,7 @@ string Node::PathDecanonicalized(const string& path, uint64_t slash_bits) {
 void Node::Dump(const char* prefix) const {
   printf("%s <%s 0x%p> mtime: %" PRId64 "%s, (:%s), ",
          prefix, path().c_str(), this,
-         mtime(), mtime() ? "" : " (:missing)",
+         mtime(), exists() ? "" : " (:missing)",
          dirty() ? " dirty" : " clean");
   if (in_edge()) {
     in_edge()->Dump("in-edge: ");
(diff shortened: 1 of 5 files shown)


src/graph.h
```diff
@@ -113,9 +122,19 @@ struct Node {
   /// Possible values of mtime_:
   ///   -1: file hasn't been examined
   ///   0:  we looked, and file doesn't exist
-  ///   >0: actual file's mtime
+  ///   >0: actual file's mtime, or the latest mtime of its dependencies if it doesn't exist
```
rascani, line 125: On the previous review, there was [comment](https://github.com/ninja-build/ninja/pull/1021#discussion_r307730374) expressing concern about changing the semantics of mtime(). Here are the instances where it is called in non-test cases outside of DependencyScan:

* [Builder::Cleanup](https://github.com/ninja-build/ninja/blob/master/src/build.cc#L535) - mtime() is used to determine whether we need to remove an output file when the build is interrupted. This change would produce a false positive here, but the ramifications would be that we call RemoveFile on an output that doesn't exist. 
* [Builder::FinishCommand](https://github.com/ninja-build/ninja/blob/master/src/build.cc#L760) - As far as I can tell, this function does not get called for phony edges, as there is no associated command.
* [Plan::CleanNode](https://github.com/ninja-build/ninja/blob/master/src/build.cc#L276) - My understanding is that this method is working its way back up the graph, marking nodes as clean after a command has been completed without changes. The mtime() call is being used to calculate the most recent input, which I believe would be fine.
* [ImplicitDepLoader::LoadDepsFromLog](https://github.com/ninja-build/ninja/blob/master/src/graph.cc#L627) - I don't believe this code path would be reached for phony edges.

rascani:
Are there any remaining concerns with this PR?

src/graph.cc
```diff
@@ -31,7 +31,16 @@
 using namespace std;
 
 bool Node::Stat(DiskInterface* disk_interface, string* err) {
-  return (mtime_ = disk_interface->Stat(path_, err)) != -1;
+  METRIC_RECORD("node stat");
+  mtime_ = disk_interface->Stat(path_, err);
+  exists_ = (mtime_ > 0) ? ExistenceStatusExists : ExistenceStatusMissing;
```
johanneslerch, line 36: I think checking for mtime_ > 0 is not consistent with previous behavior. In graph.h the exists() function was implemented by checking mtime_ != 0 and not > 0. This should be fine as long as mtime_ can not have negative values, like -1000. Unfortunately, we observed that on windows for a file dated 31.12.1979 23:00 the mtime return by disk_interface->Stat is a very large negative number leading to ninja complaining about missing files that actually exist.

We will try with the following fix:
```suggestion
  exists_ = (mtime_ != 0) ? ExistenceStatusExists : ExistenceStatusMissing;
```

Nevertheless, thanks for this PR, it saved us a lot of time! We are using it in a build system with hundreds of users.

rascani: Great catch! Looking at the RealDiskInterface::Stat code, it looks like it adjusts the Windows epoch to the year 2000, which would explain the negative value. Unfortunately, Stat() also returns -1 on error. In that case, we should probably continue to treat the existence status as unknown.

IMO, it's probably less than ideal that mtimes could be negative values, but I believe adjusting the epoch would break existing build logs. That's probably outside the scope of this PR, so I'll match existing behavior for now.

xxthunder (approved):
Great, thank you for fixing that.

rascani:
@jhasse - Do you have any remaining concerns about this PR? I'd love to land it

nico:
I think this looks fine, but it also looks a bit scary -- I wouldn't be surprised if it breaks some corner case around generator edges or similar. But I can't think of a better way to find problems than to land and see if we run into problems in practice. So let's merge this, and maybe you can send an email to the mailing list asking people to test. Worst case, we can back it out again.

rascani:
Thank you!

PR #1977, opened 2021-06-02
cmake: use `check_symbol_exists` for browse support

mathstuf:
Just because `unistd.h` exists does not mean it has the APIs that are
needed. This occurs when cross-compiling using Fedora's packaged MinGW
toolchain.


jhasse:
Adding a Fedora MinGW build to GitHub Actions would be nice.

PR #1979, opened 2021-06-03
util: Remove unnecessary CanonicalizePath error handling

bradking:
Since #1314, the only failure case in the `CanonicalizePath` implementation is the "empty path" error.  Update call sites that might have empty paths to explicitly check for them before calling CanonicalizePath.  Note that the depfile parser ensures all parsed outs and deps are non-empty.  Remove error handling from the signature to simplify call sites.


src/clean.cc
```diff
@@ -189,21 +189,21 @@ int Cleaner::CleanTargets(int target_count, char* targets[]) {
   LoadDyndeps();
   for (int i = 0; i < target_count; ++i) {
     string target_name = targets[i];
-    uint64_t slash_bits;
-    string err;
-    if (!CanonicalizePath(&target_name, &slash_bits, &err)) {
-      Error("failed to canonicalize '%s': %s", target_name.c_str(), err.c_str());
+    if (target_name.empty()) {
+      Error("empty path");
```
jhasse, line 193: Keep `Error("failed to canonicalize '%s': empty path", target_name.c_str());` so that the error stays the same?

bradking: I actually preserved it originally, but then decided to update it since no other code paths use the `failed to canonicalize` wording.  I've restored it for now so that this PR remains refactoring-only.

PR #1982, opened 2021-06-09
ninja_syntax.py: add pathlib support

aDotInTheVoid:
Note that this currently casts all not `str`s to `str`, which would also effect how `int`s are handleled, I'm not sure if you want this. Instead, I could change it to only cast on `pathlib` types.

Also, I may need to change it, depending on which python versions you want to suport (pathlib was added in 3.4)

### DIFF
--- a/misc/ninja_syntax.py
+++ b/misc/ninja_syntax.py
@@ -25,6 +25,8 @@ import re
 import textwrap
 
 def escape_path(word):
+    if not isinstance(word, str):
+        word = str(word)
     return word.replace('$ ', '$$ ').replace(' ', '$ ').replace(':', '$:')
 
 class Writer(object):
--- a/misc/ninja_syntax_test.py
+++ b/misc/ninja_syntax_test.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 import unittest
+import pathlib
 
 try:
     from StringIO import StringIO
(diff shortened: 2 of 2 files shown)


scivision:
Last I knew, the project is looking to keep Python 2.7 compatibility, perhaps given the fundamental nature of Ninja.  https://github.com/ninja-build/ninja/pull/1767#discussion_r535037338

jhasse:
I don't understand. What is the motivation for this change in the first place?

(feel free to reopen)

remexre:
The motivation is to support the `pathlib.Path` objects introduced in Python 3; these provide a less error-prone API for paths by making it impossible to accidentally do e.g. `"some/dir" + "some/file"` or `if ("some/dir/" + "/" + "some/file") == "some/dir/some/file":`.

However, since these objects aren't strings anymore, they need to either be stringified by ninja _syntax.py or by the caller (or port ninja_syntax.py to use `Path`s internally instead of strings); the former would be nicer as a user.

remexre:
(doesn't look like I can reopen, but I'd like this feature to exist, so I would if I could...)

jhasse:
Ah I see. I still think we need to support Python 2 for the time being though.

@aDotInTheVoid Out of interest: Are you able to re-open the PR?

remexre:
Unless I misunderstand, this PR only requires Python 3 for the newly added tests; would it be mergable if that test were gated behind a version check?

jhasse:
Yes :)

PR #1986, opened 2021-06-19
Fix memory leak in manifest_parser.

AmbientLighter:


rillian (approved):
lgtm, fwiw.

How often can this leak happen? If it weren't for the `scoped_ptr` comment I would have expected the intent was to leak until shutdown on error, like with the globals, but I guess there can be an arbitrary number of invalid rules?

src/manifest_parser.cc
```diff
@@ -141,31 +141,37 @@ bool ManifestParser::ParseRule(string* err) {
   if (env_->LookupRuleCurrentScope(name) != NULL)
     return lexer_.Error("duplicate rule '" + name + "'", err);
 
-  Rule* rule = new Rule(name);  // XXX scoped_ptr
+  Rule* rule = new Rule(name);
```
jonesmz, line 144: use raii

```cpp
#ifdef cpp11 
    struct Unleaker
    {
        Unleaker(Rule* rule) : m_rule(rule){}
        ~Unleaker(){ delete m_rule; }
        Rule* get(){ return m_rule; }
        Rule*m_rule;
    };
#else
    using Unleaker = std::unique_ptr<Rule>;
#endif
    Unleaker rule(new Rule(name));
```
    then reference as "rule.get()" instead of rule.

jhasse:
I think this leak is intentional.

PR #1994, opened 2021-07-20
compdb: escape control characters in JSON strings

### LINKED ISSUE #1938: compdb should escape control characters
[RFC7159](https://tools.ietf.org/html/rfc7159.html#section-7): All Unicode characters may be placed within the quotation marks, except for the characters that must be escaped: quotation mark, reverse solidus, and the **control characters** (U+0000 through U+001F).

https://github.com/ninja-build/ninja/blob/8cd25aa78c2ff2ca8b3712f8c3630d1a095856d9/src/ninja.cc#L769-L776

Example:

```
$ echo -e 'rule cc\n  command = gcc $in\t-o $out\nbuild foo: cc foo.c' | tee build.ninja
rule cc
  command = gcc $in	-o $out
build foo: cc foo.c
$ ninja -t compdb
[
  {
    "directory": "/tmp/tmp.fj51AyZzjC",
    "command": "gcc foo.c	-o foo",
    "file": "foo.c",
    "output": "foo"
  }
]
```
should be `"command": "gcc foo.c\t-o foo"`

afq984:
Fixes #1938


src/json.h
```diff
@@ -0,0 +1,45 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef NINJA_JSON_H_
+#define NINJA_JSON_H_
+
+#include <stdio.h>
+
+void EncodeJSONString(FILE* stream, const char* str) {
```
jhasse, line 20: Maybe `std::ostream&` instead of `FILE*` so that we can use a `std::stringstream` instead of `tmpfile()`?

afq984: In that case, shall we limit the use of `std::cout` to `printCompdb` only, or propagate it to ToolCompilationDatabase to be consistent? std::ios_base::sync_with_stdio by default is on and tends to slow down stuff a lot.

Or make it `std::string EncodeJSONString(const std::string& str)`? We're already calling c_str() in `printCompdb` anyway

(I [originally](https://github.com/ninja-build/ninja/commit/be0daf4870635ec5076b2caff500237a3b2519e8#diff-6d51fafa9848c43417816f023be8ee23ceabf5963620c740aadb0c8d1930580bR33) used `open_memstream` to avoid `tmpfile` but  `open_memstream` isn't available on Windows)

afq984: I just did some `time ninja -t compdb` experiments with build.ninja generated from chromium (ran only once):
* [master](https://github.com/ninja-build/ninja/commit/c8f8f2a9e3016ab7a9ecb2e8b084cf441f3ae88e) 7.50 sec
* `void EncodeJSONString(FILE* stream, const char* str)` 7.30 sec; [code](https://github.com/ninja-build/ninja/commit/450f977cd0a0c240fde3fa6d58a10b5ac0342801)
* `std::string EncodeJSONString(const std::string& str)` with fwrite 6.86 sec; [code](https://github.com/afq984/ninja/commit/82a27729ede4184cabfcf8fd3465659a3c64b0d2)
* `EncodeJSONString(std::ostream& os, const std::string& in)` with std::stringstream::str().c_str() and fwrite 11.39sec; [code](https://github.com/afq984/ninja/commit/9e4bf10300b61b5562e75e297a6c8f9de3d4c6bc)
* `void EncodeJSONString(std::ostream& os, const std::string& in) with std::cout` 13.94 sec; [code](https://github.com/afq984/ninja/commit/d1d5be20e0e685bdc9cad92f3c03d10dcd95e786)

jhasse: That's a bit sad :( Thanks for testing.

afq984: @jhasse I updated the PR to use `std::string EncodeJSONString(const std::string& str)`, please take a look!

src/json.h
```diff
@@ -0,0 +1,56 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef NINJA_JSON_H_
+#define NINJA_JSON_H_
+
+#include <cstdio>
+#include <string>
+
+std::string EncodeJSONString(const std::string& in) {
```
jhasse, line 21: Move definition to its own cc file.

src/json.h
```diff
@@ -0,0 +1,56 @@
...
+      out += "\\\\";
+    else if (c == '\"')
+      out += "\\\"";
+    else
+      out += c;
+  }
+  return out;
+}
+
+void PrintJSONString(const std::string& in) {
```
jhasse, line 51: for this function only move the definition.

jhasse (approved):

Issue #1998, opened 2021-07-23
-d stats overhead is a bit too high, due to canonicalize str and friends


randomascii:
On a typical Chrome NOP build the ninja -d stats output looks something like this:

metric                  count   avg (us)        total (ms)
.ninja parse            6580    3823.8          25160.6
canonicalize str        6240450 0.0             36.9
canonicalize path       6251390 0.0             25.7
lookup node             6339406 0.0             29.3
.ninja_log load         1       119672.0        119.7
.ninja_deps load        1       564744.0        564.7
node stat               169000  9.9             1679.1
depfile load            327     331.5           108.4

While looking at an ETW trace of this NOP build I noticed that RtlQueryPerformanceCounter accounted for 707 of 13,494 samples, or more than 5%. Testing without -d stats showed that this entry went away entirely.

It seems that when -d stats is specified that ninja probably calls QueryPerformanceCounter twice for each event. ~19 million of the events - roughly 99% - are for canonicalize str, canonicalize path, and lookup node. The total cost of all of these events is roughly 90 ms. That means that the overhead of measuring the cost of those three event types is roughly seven times higher than the cost of those three event types.

In order to ensure that -d stats is not a burden we should remove those three measurements. In addition to improving performance (easily measured by timing ninja with and without) it will simplify the output and make the non-zero averages more visible.

randomascii:
Another issue with the -d stats flag is that TimerToMicros is called for each sample. This is not as expensive as RtlQueryPerformanceCounter but it is easily avoidable - the conversion only needs to be done a dozen times, when printing the results, instead of 20 million times. The total cost on a profiling run with ninja symbols was ~310 ms, according to the ETW sampling profiler, which adds another 30-35% to the cost of the profiling.

randomascii:
Now that canonicalize str, canonicalize path, and lookup node are not being measured by -d stats the cost of -d stats is dramatically lower, so optimization is not needed. Greater precision and portability of the -d stats code would still be nice to have.

PR #2000, opened 2021-07-26
Optimize ninja -d stats

randomascii:
-d stats enables instrumented profiling of key functions in ninja.
However, some of those functions are invoked 6+ million times in a NOP
build of Chromium and the cost of measuring those functions dwarfs the
cost of the functions. Here is typical -d stats output for a Chromium
build:

.ninja parse            6580    4197.5          27619.5
canonicalize str        6240450 0.0             47.3
canonicalize path       6251390 0.0             33.5
lookup node             6339402 0.0             37.2
.ninja_log load         1       176226.0        176.2
.ninja_deps load        1       465407.0        465.4
node stat               168997  8.8             1482.9
depfile load            327     352.7           115.3

99% of the measurements are in three functions. The total measurement
cost (per ETW sampled profiling) is 700-1200 ms, which is many times
greater than the costs of the functions.

With this change typical output looks like this:

.ninja parse            6580    3526.3          23203.2
.ninja_log load         1       227305.0        227.3
.ninja_deps load        1       485693.0        485.7
node stat               168997  9.6             1615.0
depfile load            327     383.1           125.3

This resolves issue #1998.


jonesmz:
I think you would be better off making these calls to METRIC_RECORD a compile-time option, instead of removing them entirely.

randomascii:
> I think you would be better off making these calls to METRIC_RECORD a compile-time option, instead of removing them entirely.

That sounds reasonable. What is the recommended way to do that in ninja? In the context of Chromium I would use a gn arg to drive a /D compiler switch to drive the conditionality - what is the equivalent in this context?

jonesmz:
1. Find the definition of the macro METRIC_RECORD, add an #ifdef around that definition that can make it a no-op if some appropriate preprocessor definition is/isn't present and/or set to an appropriate value.
2. Modify the CMake code to have an OPTION() function so that cmake will keep track of some compile-time option for whether to enable or disable the METRIC_RECORD stuff.
3. Modify the CMake code for the ninja-executable so that if the OPTION you created is set, it adds the preprocessor definition you decided on in step 1.

jonesmz:
Another approach would be to make this a runtime option for the Ninja binary. Where some CLI option passed to Ninja enabled / disabled the METRIC_RECORD macro's guts.

randomascii:
There's already a command-line option to enable/disable the METRIC_RECORD macro's guts, and I assume that the overhead of the macro is negligible when it is disabled. The problem is that when METRIC_RECORD is enabled, through -d stats, the overhead is significant due to these three high-volume/low-value uses.

So, if I add an #ifdef it wouldn't be around METRIC_RECORD (no need) but around these three specific instances. Does that still sound worthwhile? I haven't hacked on CMake in years but I could take a look.

I'm not sure anybody would use this flag, however. For instance, when profiling EvalString::Evaluate I just added a METRIC_RECORD macro to do profiling, then deleted it when I was done. I think the same idea makes sense for these functions. If somebody wants to work on them they can add and then remove the macros. It's just a line of code.

randomascii:
I'm closing this one and replacing with a cleaner PR, found here: https://github.com/ninja-build/ninja/pull/2002

Issue #1999, opened 2021-07-23
ninja allocation overhead is high


randomascii:
A NOP build of chrome currently takes about 12.5 seconds on my Z840. Half of this time is in ninja.exe itself, with most of the rest (~42%) in ntdll.dll and ntoskrnl.exe, and most of that time is related to allocating, freeing, zering, and faulting in memory. If ninja did fewer allocations it might be measurably faster.

I recorded my profile without ninja symbols so I don't yet know where in ninja the allocations are coming from but I wanted to file this placekeeper issue now in case I don't get to it soon.

randomascii:
EvalString::Evaluate concatenates numerous strings and might benefit from abseil's support for this. The mixture of RAW and environment variable strings makes this slightly complicated, but not unsolvable. 99% of the ~6.45 million calls in a Chromium build use only raw strings. So, we could do an initial pass to see if they are all raw strings and then use abseil's concatenation functions if they are. Or do the equivalent and just reserve the right amount of space.

jonesmz:
Ninja does not depend on third party code (Or didn't, last I checked).

Ninja could gain 90% of the benefits that Abseil's string code provides by using a function like this:

https://github.com/splinter-build/splinter/blob/10-string_concat/src/string_concat.h

However, as that function's implementation hard-depends on C++17, it's going to be a loooong wait for Ninja to use it.

randomascii:
It's just ~4 lines of code to reserve the right size. The code is simple enough, but unfortunately the gain is fairly small so it's still debatable. ~50 ms savings?

https://github.com/ninja-build/ninja/pull/2001

Most of the time no concatenation happens because the list only has a single entry. Special casing that breaks the return optimization, but that is probably why avoiding reallocs doesn't make much of a difference.

PR #2001, opened 2021-07-26
Avoid reallocs in EvalString::Evaluate

randomascii:
Concatenating strings often requires repeated reallocations as the
string grows, plus the cost of copying. These costs are not huge but
they are easily avoidable when, as in this case, we usually know the
final size.

This saves about 50 ms on a NOP-build of Chromium (not as much as I
had hoped).

This helps a little bit with issue #1999


jonesmz:
The changes to METRIC_RECORD aren't related to eval_env.cc.

I recommend you remove them from this PR, since you also modify those entries in the other PR you created

randomascii:
Uggh. Yep, that was a mistake. I messed up my master branch and it's now polluting other branches. I'm going to close both PRs and reopen clean ones.

randomascii:
I'm going for the classic git workflow of deleting my repo and starting over.

PR #2002, opened 2021-07-26
Optimize ninja -d stats

randomascii:
-d stats enables instrumented profiling of key functions in ninja.
However, some of those functions are invoked 6+ million times in a NOP
build of Chromium and the cost of measuring those functions dwarfs the
cost of the functions. Here is typical -d stats output for a Chromium
build:

.ninja parse            6580    4197.5          27619.5
canonicalize str        6240450 0.0             47.3
canonicalize path       6251390 0.0             33.5
lookup node             6339402 0.0             37.2
.ninja_log load         1       176226.0        176.2
.ninja_deps load        1       465407.0        465.4
node stat               168997  8.8             1482.9
depfile load            327     352.7           115.3

99% of the measurements are in three functions. The total measurement
cost (per ETW sampled profiling) is 700-1200 ms, which is many times
greater than the costs of the functions.

With this change typical output looks like this:

.ninja parse            6580    3526.3          23203.2
.ninja_log load         1       227305.0        227.3
.ninja_deps load        1       485693.0        485.7
node stat               168997  9.6             1615.0
depfile load            327     383.1           125.3

This resolves issue #1998.

### DIFF
--- a/src/state.cc
+++ b/src/state.cc
@@ -19,7 +19,6 @@
 
 #include "edit_distance.h"
 #include "graph.h"
-#include "metrics.h"
 #include "util.h"
 
 using namespace std;
@@ -104,7 +103,6 @@ Node* State::GetNode(StringPiece path, uint64_t slash_bits) {
 }
 
 Node* State::LookupNode(StringPiece path) const {
-  METRIC_RECORD("lookup node");
   Paths::const_iterator i = paths_.find(path);
   if (i != paths_.end())
     return i->second;
--- a/src/util.cc
+++ b/src/util.cc
@@ -56,7 +56,6 @@
 #endif
 
 #include "edit_distance.h"
-#include "metrics.h"
 
 using namespace std;
 
@@ -118,7 +117,6 @@ void Info(const char* msg, ...) {
 }
 
 void CanonicalizePath(string* path, uint64_t* slash_bits) {
-  METRIC_RECORD("canonicalize str");
   size_t len = path->size();
   char* str = 0;
   if (len > 0)
@@ -138,7 +136,6 @@ static bool IsPathSeparator(char c) {
 void CanonicalizePath(char* path, size_t* len, uint64_t* slash_bits) {
   // WARNING: this function is performance-critical; please benchmark
   // any changes you make to it.
-  METRIC_RECORD("canonicalize path");
   if (*len == 0) {
     return;
   }


randomascii:
Apologies for the chaos. This is a repeat of the previous pull request but without the extraneous merge commits and other nonsense. I also removed two #includes that are no longer needed.

On the previous PR it was suggested that maybe this should be a compile-time option instead of deleting the lines entirely. In my opinion it is better to delete the lines. If somebody is working on these functions and wants to profile them then they can add back the necessary lines (and the #include). I just did that when working on EvalString::Evaluate and it worked very nicely. There is nothing magical about these three functions to justify leaving #ifdefed profiling commands in them but not elsewhere.

The other METRIC_RECORD entries in the ninja code remain and are still producing statistics. I am removing these three because have a specifically poor cost-to-benefit ratio. I hope that this makes sense.

randomascii:
The original PR, now closed, is here: https://github.com/ninja-build/ninja/pull/2000

randomascii:
Any thoughts on this? It seems to me that we either have to have an extremely efficient HighResTimer() function (__rdtsc() is the only thing fast enough) or we have to remove these three uses of METRIC_RECORD. Spending 1+ seconds to record times that are not very accurate (and have, historically, been off by a factor of 50) doesn't seem useful. I say this as somebody who always runs ninja with the -d stats flag.

jhasse:
> I'm not sure anybody would use this flag, however. For instance, when profiling `EvalString::Evaluate` I just added a METRIC_RECORD macro to do profiling, then deleted it when I was done. I think the same idea makes sense for these functions. If somebody wants to work on them they can add and then remove the macros. It's just a line of code.

I agree.

Let's see if someone else has some feedback, otherwise I'll merge this for the next release.

emartinfigma:
I think "lookup node" is effectively measuring hash time, and "canonicalize path" is measuring path canonicalization, which I know @nico spent some time on, so I CC him just in case.  I agree that someone interested in profiling these can either reinstrument them locally or use a profiler instead.

evmar:
I think "lookup node" is effectively measuring hash time, and "canonicalize path" is measuring path canonicalization, which I know @nico spent some time on, so I CC him just in case.  I agree that someone interested in profiling these can either reinstrument them locally or use a profiler instead.

randomascii:
Is this waiting on anything? I run with -d stats a lot and I'd like to skip the extra second of latency for these stats.

jonesmz:
Can this change be done with a compile-time option, instead of removing the associated code? That would let someone who still wants to measure these two cases do so if they need to without having to dig through the code to find the right functions to instrument.

They can be turned off by default, of course.

randomascii:
A compile-time option could be used, but there would be as much code controlling the statistics as in the statistics, and I'm not sure it would make them much more discoverable than being in github history. The profiling of these events hasn't worked for years (it still doesn't, although I'm working on improving that) and nobody noticed, so I'm not sure that preserving them makes sense.

The last comment from @jhasse suggested that just deleting them was best, FWIW.

jhasse:
> I agree that someone interested in profiling these can either reinstrument them locally or use a profiler instead.

this

PR #2003, opened 2021-07-26
Avoid reallocs in EvalString::Evaluate

randomascii:
Concatenating strings often requires repeated reallocations as the
string grows, plus the cost of copying. These costs are not huge but
they are easily avoidable when, as in this case, we usually know the
final size.

This saves about 50 ms on a NOP-build of Chromium (not as much as I
had hoped).

### DIFF
--- a/src/eval_env.cc
+++ b/src/eval_env.cc
@@ -99,7 +99,15 @@ string BindingEnv::LookupWithFallback(const string& var,
 }
 
 string EvalString::Evaluate(Env* env) const {
+  // Estimate how much space will be needed for the concatenation. This will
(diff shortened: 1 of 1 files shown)


randomascii:
This is a recreation of a previous pull requests. Now it is properly isolated on its own branch with no other changes.

I noticed after I created the PR that the commit message had two of its lines truncated. I can always recreate the PR if that needs fixing.

The gains for this change are not large, so maybe it's not worth it. I was really hoping for more, but the main cost of this function is creating ~6 million strings. Unless we can std::move from the parsed_ list this function probably can't be made much faster.

jhasse:
> 50 ms on a NOP-build of Chromium

How long is the whole NOP-build?

Issue #2004, opened 2021-07-27
Stopwatch::Now is too expensive, too imprecise


randomascii:
While investigating https://github.com/ninja-build/ninja/issues/1998 I noticed that Stopwatch::Now() calls the HighResTimer and then converts to microseconds. This means that Stopwatch::Now can only reliably measure the duration of events that take at least ~5 us or so - below that the rounding errors start to become significant.

"canonicalize path", as one concrete example, is measured 6,251,390 times during a NOP build of Chromium, for a total duration of 25.7 ms. That means that the average duration of this function is 4 ns.

That explains why the TimerToMicros time is greater than the "canonicalize path" time.

I propose that StopwatchNow be changed to return raw HighResTimer data. One many modern Windows machines this is 10 MHz which is a bit more precise (although still not great for measuring 4 ns intervals) and will be much more efficient. The conversion to ms/micro-seconds/ns can be done when printing.

The Stopwatch class is used in few enough places to make this a simple change.

jonesmz:
Out of curiosity, would using this be sufficient resolution and speed? https://en.cppreference.com/w/cpp/chrono/high_resolution_clock

randomascii:
The VC++ version of chrono wraps QueryPerformanceCounter, so it would have the same 10 MHz frequency (on most machines) that we have now. Unfortunately it adds more overhead, such as querying for the frequency on every sample, and it converts to a standard frequency at each sample point, which means more multiplications and divisions. It's standard frequency is GHz so it doesn't lose any precision.

That is, it would be more expensive than the current implementation, but more precise. It would be more expensive and equally precise as my proposed fix. It would be portable, sort of, although the documentation recommends against using it because the implementations are so inconsistent.

It would probably be fine as long as we weren't trying to measure things like "canonicalize str" that happen 6+ million times and have extremely short durations.

I'm tempted to swap from QueryPerformanceCounter to __rdtsc. That would be less expensive and more precise. It's not guaranteed for general purpose timing, but seems fine for this task. I could make it conditional on CPU type so the QueryPerformanceCounter code would still be available. It's just two functions that need to change.

jonesmz:
Have you raised a bug report on Microsoft's std::chrono implementation?

The whole bloody point of the high_resolution_clock is to get the best resolution without a big perf impact.

randomascii:
Somebody must have complained because more recent versions of VC++ 2019 cache the frequency, so that's progress.

[...]

PR #2005, opened 2021-07-27
Make Stopwatch and ScopedMetric more precise

randomascii:
Stopwatch and ScopedMetric both used to store their times in
microseconds. The high-frequency timer source on Windows is typically
10 MHz, so 0.1 microsecond precision. Truncating to microseconds throws
away precision, and is also (in the ScopedMetric case) quite expensive.

In the case of Chromium there were some measurements that were done
millions of times with an average duration of ~0.003 to 0.004
microseconds. These measurements could only be statistically measured
by a 1 MHz clock, and their total times were typically reported as
~15-24 ms:

canonicalize str        6240450 0.0             23.6
canonicalize path       6251390 0.0             15.7
lookup node             6339402 0.0             19.5
.ninja_log load         1       166196.0        166.2

With this change a (presumed more accurate) picture appears:

canonicalize str        6240450 0.2             1492.4
canonicalize path       6251390 0.2             1165.0
lookup node             6339402 0.1             934.9
.ninja_log load         1       152829.0        152.8

It turns out these functions are about 50 times more expensive than
previously thought (but still cheap).

We would have to switch to __rdtsc() as our clock source to see for sure
whether this new data is accurate, but it is likely that it is _more_
accurate, and it is certainly cheaper. The old code was expensive enough
that QueryPerformanceCounter showed up clearly on profiles. It will now
be called six orders of magnitude less frequently.


randomascii:
I'm tempted to go farther and switch this to use __rdtsc(). This will give much greater precision (~300 times more on typical machines). If we were trying to do long-term stable timing or run on old CPUs then this might not be a great choice but for "ninja -d stats" it seems appropriate.

jonesmz:
Can c++ 11's std::chrono be used for this, instead of resorting to platform specific APIs?

https://en.cppreference.com/w/cpp/chrono/high_resolution_clock

randomascii:
They could, but it's not clear that they should. More detailed answer here: https://github.com/ninja-build/ninja/issues/2004#issuecomment-887819134

randomascii:
I cleaned up the frequency initialization code to fix the duplication that I had caused, so that's better.

I also created a proof-of-concept RDTSC version.

https://github.com/randomascii/ninja/commit/62ab72a3e90041f986bd2ccb0e79c1c670c05e4e

On my workstation this records time at 2594 MHz. It seems to show longer times for .ninja_log load which I need to investigate. If we want to measure things like canonicalize str and canonicalize path and lookup node which take 0.1 to 0.2 microseconds then we should consider using this more precise code. If not then QPC, or even std::chrono::steady_clock could be used.

jonesmz:
Well, as a person who is not a ninja project member, so I don't really get to have the final say:

I don't think Ninja should be increasing the amount of platform specific code.

Even if std::chrono isn't literally the fastest possible implementation, it's my opinion that it should be used unless there is an overwhelming reason not to.

randomascii:
I've created three different Stopwatch/ScopedMetric changes.

This one is the simplest change, just improving the existing timer code to make it faster and more accurate.

Another one uses __rdtsc() for maximum precision and performance. It could actually be used to measure sub-microsecond events. Using __rdtsc has some caveats and is non-portable, so there are definite tradeoffs. https://github.com/randomascii/ninja/tree/faster_now

The third option uses std::chrono::steady_clock::now(), which makes it more accurate than the current implementation, but a bit slower. It's main advantage is portability. It needs a bit of cleanup before being made into a PR. Its performance on VC++ will improve when Microsoft releases some optimizations I suggested: https://github.com/randomascii/ninja/tree/steady_clock_now

Each one has some advantages. They are all better than the current code. The second one is the most radical so I'd probably recommend either the first or third. Thoughts?

randomascii:
The steady_clock_now option requires C++ 11 which ninja currently doesn't use, so it is out of the running unless the requirements for building ninja are increased. That is, right now I get this error message on Linux builds:

'''
This file requires compiler and library support for the ISO C++ 2011 standard. This support is currently experimental, and must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
'''

So, __rdtsc or the simple cleanup PR, which is to say, this one. I recommend this one.

jonesmz:
Note that C++11 is required on Windows for Ninja.

I think it's perfectly reasonable to require C++11 on other platforms in order to use fully platform agnostic code that won't require tweaking and/or #ifdefs for every random platform.

PR #2009, opened 2021-07-29
Make Stopwatch and ScopedMetric more precise

randomascii:
Stopwatch and ScopedMetric both used to store their times in
microseconds. The high-frequency timer source on Windows is typically
10 MHz, so 0.1 microsecond precision. Truncating to microseconds throws
away precision, and is also (in the ScopedMetric case) quite expensive.

In the case of Chromium there were some measurements that were done
millions of times with an average duration of ~0.003 to 0.004
microseconds. These measurements could only be statistically measured
by a 1 MHz clock, and their total times were typically reported as
~15-24 ms:

canonicalize str        6240450 0.0             23.6
canonicalize path       6251390 0.0             15.7
lookup node             6339402 0.0             19.5
.ninja_log load         1       166196.0        166.2

With this change a (presumed more accurate) picture appears:

canonicalize str        6240450 0.3             1989.9
canonicalize path       6251390 0.2             1300.9
lookup node             6339402 0.2             1172.4
.ninja_log load         1       168528.0        168.5

It turns out these functions are about 50 times more expensive than
previously thought (but still cheap).

This change experimentally switches to chrono::steady_clock as our clock
source. This is similar in cost to the previous implementation, but not
as fast as QPC without conversion would be. This code is significantly
more accurate when measuring short time periods, and this code is
portable.

In short, this data is more precise, more accurate, similar cost, and
portable. This code is still expensive enough that
QueryPerformanceCounter will show up on profiles unless we get rid of
the three very-high-frequency profiling events.


randomascii:
This is a proof-of-concept, using portable code to measure time. It's still too expensive to use with "canonicalize str" and friends, and I need to remove the test code. Typical test code on my Windows machine look like this:

459.3 ms for 10000000 calls to HighResTimer().
240.2 ms for 10000000 calls to HighResTimerQPC().

So, it is slower than my other optimized versions, but as fast as the existing code, roughly.

randomascii:
This might be correct but I'm not sure that the checks on Linux are actually running. PTAL.

PR #2017, opened 2021-08-19
status: print roughly estimated remaining time

dmantipov:
Add NINJA_STATUS environment variable format specifier
to print the roughly estimated remaining time.

Signed-off-by: Dmitry Antipov <dantipov@cloudlinux.com>

### DIFF
--- a/src/status.cc
+++ b/src/status.cc
@@ -215,6 +215,15 @@ string StatusPrinter::FormatProgressStatus(const char* progress_status_format,
         break;
       }
 
+        // Roughly estimated remaining time in seconds.
+      case 'R': {
+        snprintf(buf, sizeof(buf), "%.1f",
+                 ((time_millis_ / 1e3) / finished_edges_) *
+                 (total_edges_ - finished_edges_));
+        out += buf;
+        break;
+      }
+
       default:
         Fatal("unknown placeholder '%%%c' in $NINJA_STATUS", *s);
         return "";


src/status.cc
```diff
@@ -215,6 +215,15 @@ string StatusPrinter::FormatProgressStatus(const char* progress_status_format,
         break;
       }
 
+        // Roughly estimated remaining time in seconds.
+      case 'R': {
+        snprintf(buf, sizeof(buf), "%.1f",
+                 ((time_millis_ / 1e3) / finished_edges_) *
```
jonesmz, line 221: what is 1e3? is that a valid C++ identifier?

src/status.cc
```diff
@@ -215,6 +215,15 @@ string StatusPrinter::FormatProgressStatus(const char* progress_status_format,
...
+                 (total_edges_ - finished_edges_));
```
jonesmz, line 222: Given that ninja tasks can have any arbitrary completion time, this is only useful for the first run of the ninjafile, where no information about how long a given task takes is available at all.

Considering that most of the time, ninja is used to incrementally build an existing set of tasks a second or more time, i personally consider this type of estimation to be useless.

A proper time remaining feature needs to track how long a particular task has taken in the past over either a rolling average, or true average, and use that per-task information to estimate completion time.

ilyapopov:
See also #1963 with a more comprehensive approach.

jhasse:
Closing in favor of #1963.

PR #2019, opened 2021-08-25
Add critical path scheduler to improve build times

### LINKED ISSUE #376: Use better logic for picking which node to build first

peterbell10:
This is based on #1333 and fixes #376.

I'm interesting in seeing #1333 merged, but unfortunately it seems to have languished. When I tried it, it was completely broken. This maintains the same principal, but reworks it to: address review comments, fix bugs, and improve time complexity of operations.

For each edge depended on by the targets being built, we calculate its priority as the longest path through the weighted directed graph of dependencies. Where weights come from the build log's last execution time. Ready build tasks are then always executed highest priority first. 

#1333 maintained a `std::list` of all edges in priority order. This PR instead uses a `std::priority_queue` of only the ready edges (wrapped in the `EdgeQueue` class). This method has log time `push`, and `pop` so  avoids the `O(n)` traversal we had with the list.


src/build.cc
```diff
@@ -75,6 +75,16 @@ bool DryRunCommandRunner::WaitForCommand(Result* result) {
 
 }  // namespace
 
+
+bool EdgeQueue::EdgePriorityCompare::operator()(const Edge* e1, const Edge* e2) const {
```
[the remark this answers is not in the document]
BenFrantzDale: You could template it. (And if you *really* mean for it to be just `Edge*`, you can static_assert that the template *is* a forward-declared type. As in
```
class Edge; // Fwd declare
...
    template <typename EdgeType>
    [[nodiscard]] bool operator()(const EdgeType* e1, const EdgeType* e2) const {
        static_assert(std::is_same_v<EdgeType, Edge>); // Avoiding including `graph.h`.
        ...
```

src/build.cc
```diff
@@ -424,6 +436,109 @@ void Plan::UnmarkDependents(const Node* node, set<Node*>* dependents) {
...
+    }
+  }
+
+  while (!edgesQ.empty()) {
+    Edge* e = edgesQ.front();
+    edgesQ.pop();
+    active_edges.erase(e);
+
+    for (std::vector<Node*>::iterator it = e->inputs_.begin(),
+                                      end = e->inputs_.end();
```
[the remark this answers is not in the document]
BenFrantzDale: Also means no `std::chrono::duration<int64_t, std::milli>` :-(

Lezcano:
Is there any progress in merging this PR? It would be such a massive time saver when building large projects.

jhasse:
I'll have a look at this immediately after releasing 1.11.0.

Lezcano:
Would it be possible to try to get this in 1.11? Otherwise it won't be in an official release for quite a long time.
Thank you in advance!

jhasse:
The same goes for all the other PRs - some of which are even older. 1.11.0 is late already and I don't have the time to review and test all of them.

src/build.h
```diff
@@ -35,6 +35,42 @@ struct Node;
...
+
+  Edge* pop() {
+    Edge* ret = queue_.top();
+    queue_.pop();
+    return ret;
+  }
+
+  void clear() {
+    queue_ = std::priority_queue<Edge*, std::vector<Edge*>, EdgePriorityCompare>();
+  }
```
digit-google, line 67: Note that a simpler version to get the same result is to subclass std::priority_queue which exposes the container as a protected member named `c` as in:

```
class EdgeQueue : public std::priority_queue<Edge*, std:;vector<Edge*>, EdgePriorityCompare> {
 public:
  void clear() { c.clear(); }
}
```

src/build.h
```diff
@@ -35,6 +35,42 @@ struct Node;
 struct State;
 struct Status;
 
+
+// Set of ready edges, sorted by priority
+class EdgeQueue {
+  struct EdgePriorityCompare {
```
digit-google, line 41: Can you add this to the Edge class directly (e.g. Edge::PriorityCompare) and move EdgeQueue definition to `src/graph.h` where it belongs. Also a simple unit-test would be welcome.

src/graph.h
```diff
@@ -171,6 +172,11 @@ struct Edge {
 
   void Dump(const char* prefix="") const;
 
+  int64_t critical_time() const { return critical_time_; }
```
digit-google, line 175: Can you add a comment explaining what this value is and how it is supposed to be used. Which units too?

digit-google: Also consider explaining why this is default-initialized tp -1, this is subtle and better explained in a comment.

src/build.h
peterbell10: Moved to `src/graph.h` but also had to move `EdgePriorityCompare` outside the class for sub-classing to work.

src/graph.h
```diff
@@ -343,4 +348,32 @@ struct DependencyScan {
   DyndepLoader dyndep_loader_;
 };
 
+// Prioritize edges by largest critical time first
```
digit-google, line 379: nit: This comment is confusing, because if this struct is used with std::sort(), it will actually place the edges with the smallest (not largest) critical time first, and otherwise smallest ids. A better comment would just state that instead.

The code works because std::priority_queue<> implements a maximal-heap, so pop() will always returns the "largest" item for the comparator being used, in this case, it will mean the one with the largest critical time, and otherwise the largest id. If all critical times are equal, e.g. for a clean build, this reverts the order currently used by the implementation, this is subtle, and may be even undesirable (does it make clean builds slower?).

src/graph.h
```diff
@@ -343,4 +348,32 @@ struct DependencyScan {
...
+    const int64_t ct1 = e1->critical_time();
+    const int64_t ct2 = e2->critical_time();
+    if (ct1 != ct2) {
+      return ct1 < ct2;
+    }
+    return e1->id_ < e2->id_;
+  }
+};
+
+// Set of ready edges, sorted by priority
```
digit-google, line 391: nit: a priority queue is not a set (i.e. it can contain duplicates, which can have surprising consequences). Also please clarify "sorted by priority" because this is very subtle. For example, what about:

```
A priority queue holding non-owning Edge pointers, its pop() method returns the edge with the largest ciritical_time() in the queue, and largest id() otherwise.
```

Also please call this EdgePriorityQueue or EdgeCriticalTimePriorityQueue for clarity.

src/build.h
digit-google: I meant move it to the Edge class itself, but that ok. Thanks for doing this.

src/build.cc
```diff
@@ -437,6 +442,150 @@ void Plan::UnmarkDependents(const Node* node, set<Node*>* dependents) {
...
+  if (!build_log)
+    return;
+
+  METRIC_RECORD("ComputePriorityList");
+  // Remove duplicate targets
+  {
+    std::set<const Node*> seen;
+    SeenBefore<Node> seen_before(&seen);
+    targets_.erase(std::remove_if(targets_.begin(), targets_.end(), seen_before),
+                   targets_.end());
```
digit-google, line 533:
```
// Remove duplicates from targets_.
std::set<const Node*> unique_targets(targets_.begin(), targets_.end());
targets_.assign(unique_targets.begin(), unique_targets.end());
```

src/build.cc
```diff
@@ -172,10 +173,12 @@ void Plan::ScheduleWork(map<Edge*, Want>::iterator want_e) {
   Pool* pool = edge->pool();
   if (pool->ShouldDelayEdge()) {
     pool->DelayEdge(edge);
-    pool->RetrieveReadyEdges(&ready_);
+    EdgeSet new_edges;
+    pool->RetrieveReadyEdges(&new_edges);
+    ready_.push_multiple(new_edges.begin(), new_edges.end());
```
digit-google, line 178: nit: this seems inefficient, can we consider Pool::RetrieveReadyEdges() pushing directly into |ready_| instead? Since the EdgePriorityQueue is in `src/graph.h` now, that shouldn't be a problem.

src/build.cc
```diff
@@ -187,7 +190,9 @@ bool Plan::EdgeFinished(Edge* edge, EdgeResult result, string* err) {
   // See if this job frees up any delayed jobs.
   if (directly_wanted)
     edge->pool()->EdgeFinished(*edge);
-  edge->pool()->RetrieveReadyEdges(&ready_);
+  EdgeSet new_edges;
+  edge->pool()->RetrieveReadyEdges(&new_edges);
+  ready_.push_multiple(new_edges.begin(), new_edges.end());
```
digit-google, line 195: ditto

src/build.cc
```diff
@@ -437,6 +442,150 @@ void Plan::UnmarkDependents(const Node* node, set<Node*>* dependents) {
...
+
+  // total time if building all edges in serial. This value is big
+  // enough to ensure higher priority target's initial critical time
+  // is always bigger than lower ones
+  int64_t total_time = AssignEdgeRuntime(build_log, want_);
+
+  // Use backflow algorithm to compute critical times for all nodes, starting
+  // from the destination nodes.
+  // XXX: ignores pools
+  std::queue<Edge*> breadthFirstEdges;  // Queue, for breadth-first traversal
```
digit-google, line 544: nit: keep variable names consistent, i.e. use `breadth_first_edges`, or simply `work_queue` because it doesn't really matter here.

src/graph_test.cc
```diff
@@ -944,3 +944,56 @@ TEST_F(GraphTest, PhonyDepsMtimes) {
...
+
+  queue.push(edges[1]);
+  queue.push(edges[2]);
+  queue.push(edges[0]);
+
+  for (int i = 0; i < n_edges; ++i) {
+    EXPECT_EQ(queue.top(), edges[i]);
+    queue.pop();
+  }
+  EXPECT_TRUE(queue.empty());
```
digit-google, line 996: very nice. Thanks.

src/graph.h
```diff
@@ -172,6 +173,10 @@ struct Edge {
 
   void Dump(const char* prefix="") const;
 
+  // Critical time is the estimated exection time in ms of the edges
+  // forming the longest time-weighted path to the target output.
+  // This quantity is used as a priority during build scheduling.
+  // NOTE: Defaults to -1 as a marker smaller than any valid time
```
digit-google, line 203: nit: s/exection/execution time :-)

Also, it may be more useful to document the use case than the implementation (which is not totally clear here), for example something like:

```
// Critical time is an estimated execution time, in milliseconds, for all commands that lead to the generation of
// the target output. It is used by the build scheduler to optimize total build time. Should be default-initialized to -1
// as a marker smaller than any valid duration.
```

src/graph.h
```diff
@@ -204,6 +215,8 @@ struct Edge {
   BindingEnv* env_;
   VisitMark mark_;
   size_t id_;
+  int64_t run_time_ms_;
```
digit-google, line 218: Why not use accessor methods like `run_time_ms()` and `set_run_time_ms()` for this member? build.cc sets the value directly, which seems wrong. Can you add a comment e.g. "Run time in milliseconds for the edge's command, taken from the build log when present, estimated otherwise."

src/build.cc
```diff
@@ -437,6 +442,150 @@ void Plan::UnmarkDependents(const Node* node, set<Node*>* dependents) {
...
+    std::vector<int64_t>::iterator p75_it = durations.begin() + p75_idx;
+    std::nth_element(durations.begin(), p75_it, durations.end());
+    p75_time = *p75_it;
+  }
+
+  for (std::map<Edge*, Plan::Want>::const_iterator it = want.begin(),
+                                                   end = want.end();
+       it != end; ++it) {
+    Edge* edge = it->first;
+    if (edge->run_time_ms_ >= 0) {
```
digit-google, line 511: I had to check twice that this code does the rigth thing for phony edges. It does because edge->run_time_ms_ is default-initialized to 0, and while phony edges are ignored in the loop above, they will be ignored here due to the default value. This is not obvious though. Ca you make it clearer, i.e. either with a comment, or use a named constant, e.g.:

```
  const int64_t kUnknownRunTime = -1;  // special value used as a marker between the two loops below.
  ...
   if (!entry) {
     missed_durations = true;
     edge->run_time_ms = kUnknownRunTime;
   }
   ...
     if (edge->run_time_ms != kUnknownRunTime) {
       continue;
     }
```

src/build.cc
```diff
@@ -437,6 +442,150 @@ void Plan::UnmarkDependents(const Node* node, set<Node*>* dependents) {
...
+  std::set<const Edge*> active_edges;   // Set of in breadthFirstEdges
+  SeenBefore<Edge> seen_edge(
+      &active_edges);  // Test for uniqueness in breadthFirstEdges
+
+  for (std::vector<const Node*>::reverse_iterator it = targets_.rbegin(),
+                                                  end = targets_.rend();
+       it != end; ++it) {
+    if (Edge* in = (*it)->in_edge()) {
+      // Use initial critical time: total_time * N. This means higher
+      // priority targets always get a higher critical time value
```
digit-google, line 554: I am confused by this comment, by "higher priority targets", do you mean "targets that appear first on the ninja command-line will always have a larger critical value"? Is this why you are using a reverse iterator here? It may be simpler to just iterate in the right order and adjust the priority_weight computation, e.g.:

```
for (size_t n = 0; n < targets_.size(); ++n) {
  Node* target = targets_[n];
  if (Edge* in = target->in_edge()) {
    // Add a bias to ensure that targets that appear first in |targets_| have a larger critical time than
    // those that follow them. E.g. for 3 targets: [2*total_time, total_time, 0].
    int64_t priority_weight = (targets_.size() - n - 1) * total_time;
    ...
  }
}
```

src/build.cc
```diff
@@ -437,6 +442,150 @@ void Plan::UnmarkDependents(const Node* node, set<Node*>* dependents) {
...
+      if (!seen_edge(in)) {
+        breadthFirstEdges.push(in);
+      }
+    }
+  }
+
+  while (!breadthFirstEdges.empty()) {
+    Edge* e = breadthFirstEdges.front();
+    breadthFirstEdges.pop();
+    active_edges.erase(e);
```
digit-google, line 564: Any reason why you remove the edge from this set, since the `seen_edge()` call below willl insert new ones in it? This doesn't seem necessary (but please correct me if I'm wrong), and is slightly confusing (i.e. if this was used in an unrestricted graph, this would probably infinite-loop in cycles due to the removal).

src/build.cc
peterbell10: I see your point, but it does somewhat matter. A depth first traversal would be more likely to get a node wrong and have to re-do work.

src/graph.h
peterbell10: I think that phrasing is more confusing. "all commands that lead to the generation of the target output." sounds like the sum of the entire sub-graph, not just the longest path.

src/build.cc
peterbell10: This is strictly necessary for the algorithm. If one of the inputs to this edge gets its critical time updated, then this edge must be put back into the work queue.

digit-google (commented):
Sorry about all the nitpicks, this is still a very exciting PR. I'm worried about all the heuristics being used and how to measure if they do not regress things (e.g. clean builds), or what impact to these algorithms would have in the future.

There must be a way to make this completely testable and simulable, i.e. have a pre-determined build plans with recorded run times for all commands, and compute (without running any real commands) how well that would work, compared to the naive implementation, with a clean build log, a build log that is 80% full only (to simulate real incremental builds), and a full one. What do you think?

src/build.cc
digit-google: Subtle, I see, can you add a comment to clarify that then?

src/graph.h
digit-google: On second thought I think you are correct about the phrasing, I stand corrected :)

peterbell10:
>Sorry about all the nitpicks, this is still a very exciting PR. I'm worried about all the heuristics being used and how to measure if they do not regress things (e.g. clean builds), or what impact to these algorithms would have in the future.

Something worth noting about clean builds is it should still be a win. Even if there is no build log, the runtime of each edge will be set to 1 and the critical time becomes the depth of the node in the graph. This is also a sensible scheduling algorithm because a higher depth job is likely to open up more opportunities for parallelism. So for example, it would always prioritize running a code-generator before compiling plain files linked into the same library which is a win because more files can be compiled in parallel.

>There must be a way to make this completely testable and simulable, i.e. have a pre-determined build plans with recorded run times for all commands, and compute (without running any real commands) how well that would work, compared to the naive implementation, with a clean build log, a build log that is 80% full only (to simulate real incremental builds), and a full one. What do you think?

I'll see what I can do. I think it should be possible to write a build log then check the computed critical times. Something like that anyway. 

For a general simulation I'm sure I can work up a python script to simulate a build, although generating random build graphs and timings that are actually realistic may be tricky.

kinke:
> Something worth noting about clean builds is it should still be a win.

In my case, the initial build took the same time as with ninja master (I've rebased this PR). What I very much like is that I can manually tweak the ninja cmdline and add known long-running targets there; adding such an object file reduced the initial build time from 33 seconds to 26 seconds. Subsequent builds with `.ninja_log` are done in 24 seconds. So I absolutely love it, thanks dude!

FWIW, some info about my very imbalanced build (few long-running processes):
```
    Longest build steps:
           0.4 weighted s to build .reggae/objs/sil.objs/__dub__/core/source/kaleidic/sil/std/core/c... (8.4 s elapsed time)
           0.4 weighted s to build .reggae/objs/sil.objs/__dub__/core-datetime/source/kaleidic/sil/s... (9.3 s elapsed time)
           0.4 weighted s to build .reggae/objs/sil.objs/__dub__/extra/source/kaleidic/sil/std/extra... (10.2 s elapsed time)
           0.4 weighted s to build .reggae/objs/sil.objs/__dub__/extra-xlsx/source/kaleidic/sil/std/... (10.7 s elapsed time)
           0.5 weighted s to build .reggae/objs/sil.objs/__dub__/extra/source/kaleidic/sil/std/extra... (11.4 s elapsed time)
           0.5 weighted s to build .reggae/objs/sil.objs/__dub__/extra-technical/source/kaleidic/sil... (13.2 s elapsed time)
           0.6 weighted s to build .reggae/objs/sil.objs/__dub__/extra-kaleidic/source/kaleidic/sil/... (13.8 s elapsed time)
           0.8 weighted s to build .reggae/objs/sil.objs/__dub__/symmetry-imap/source/imap_auth.o (18.7 s elapsed time)
           0.8 weighted s to build .reggae/objs/sil.objs/__dub__/core/source/kaleidic/sil/std/core_a... (18.9 s elapsed time)
           4.1 weighted s to build sil (4.1 s elapsed time)
    Time by build-step type:
           4.1 s weighted time to generate 1 (no extension found) files (4.1 s elapsed time sum)
          19.5 s weighted time to generate 269 .o files (466.2 s elapsed time sum)
    23.5 s weighted time (470.3 s elapsed time sum, 20.0x parallelism)
    270 build steps completed, average of 11.48/s
```

digit-google:
Just to say that the current patch doesn't change the build time for Fuchsia builds (it may even slow it down by a few percent but it' s hard to be certain), however we rely heavily on pools, which may be limiting the benefit of the new algorithm.

travisdowns:
I can't wait for this change. In any cases it would cut our build times by a third, because we have one file (with no input task dependencies) which takes 2x as long as most files: this file should be built immediately at the start but ninja choose to build it near the end, adding about 1 minute to a 3 minute build. Almost any old scheduling algorithm examining `.ninja_log` would solve this immediately.

hadrielk:
@peterbell10 where did you get such a beautiful chart?

I've been staring at tables, waterfall charts and flame graphs, but I'd much prefer to see build-times your way.

def-:
That's ninjatracing: https://github.com/nico/ninjatracing

jhasse:
Closing in favor of #2177.

kinke:
FWIW, I've published a drop-in ninja release including this PR as main change, making the historical-build-times usage opt-in via `--cp`: https://github.com/symmetryinvestments/ninja/releases/tag/v1.11.1-sym1 (and https://github.com/symmetryinvestments/gha-setup-ninja along with it)

travisdowns:
@kinke - awesome, I'm going to try this soon.

Just to clarify (since you can't create issues on forks), without `--cp` this release would work the same as upstream ninja 1.11.1?

kinke:
> Just to clarify (since you can't create issues on forks), without --cp this release would work the same as upstream ninja 1.11.1?

Nope; there's still the default prioritization by explicit cmdline order / node depth. `--cp` really only affects whether the previous `.ninja_log` timings are to be used.

PR #2020, opened 2021-08-27
Make ScopedMetric portable, accurate, and efficient

randomascii:
ScopedMetric used to use separate code paths for Win32 and other
platforms. std::chrono makes it practical to use the same code
everywhere. This requires C++ 11 so that is specified in the
configuration file (C++ 11 was already used on Win32 builds).

This change also makes ScopedMetric more accurate because it
postpones converting the raw deltas until it is time to report them.
Previously they would be converted at each measurement site, potentially
losing almost a microsecond each time.

Postponing the conversion also reduces the cost of ScopedMetric.

The performance of std::chrono is currently harmed a bit by some
design and code-gen issues in VC++, but bugs have been filed and these
should be fixed. The overhead is low enough to be inconsequential for
the O(15,000) ScopedMetric objects created in a build of Chrome.

The net result is that the ScopedMetric code is now portable, more
accurate, and as fast or faster than it used to be.


randomascii:
I think this change is a nice improvement to ninja (std::chrono was suggested on the issue #2004 discussion), but it requires C++ 11, so that needs approval from a project maintainer, and I need to know how to enable it. I thought that adding set(CMAKE_CXX_STANDARD 11) to CMakeLists.txt would do the trick but the failures on ubuntu:16.04 suggest I am doing something wrong, or else C++ 11 is not supported there at all, or else we are using a version of CMake prior to 3.1.

I'm happy to polish this off if this change is wanted, and some help in enabling C++11 would be appreciated.

https://stackoverflow.com/questions/10851247/how-do-i-activate-c-11-in-cmake

CONTRIBUTING.md
```diff
@@ -15,7 +15,7 @@ Generally it's the
 a few additions:
 
 * Any code merged into the Ninja codebase which will be part of the main
-  executable must compile as C++03. You may use C++11 features in a test or an
+  executable must compile as C++11. You may use C++14 features in a test or an
   unimportant tool if you guard your code with `#if __cplusplus >= 201103L`.
```
jonesmz, line 19: change the 201103L identifier to match the C++14 one.

src/metrics.cc
```diff
@@ -112,18 +95,23 @@ void Metrics::Report() {
...
     printf("%-*s\t%-6d\t%-8.1f\t%.1f\n", width, metric->name.c_str(),
            metric->count, avg, total);
   }
 }
 
-uint64_t Stopwatch::Now() const {
-  return TimerToMicros(HighResTimer());
+double Stopwatch::Elapsed() const {
+  // Convert to micros after converting to double to minimize error.
+  return 1e-6 * TimerToMicros(static_cast<double>(NowRaw() - started_));
```
jonesmz, line 108: 1e-6 is a real thing? I've never seen this syntax.

CONTRIBUTING.md
randomascii: Good catch. I'll do that.

src/metrics.cc
randomascii: Yep. 1 times 10 to the -6 power. It's standard exponential notation. I guess you can never tell what will be recognizable or not to other software developers.

Any thoughts on how to enable C++11 on Ubuntu 2016.04?

CONTRIBUTING.md
randomascii: Done.

CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ else()
 endif()
 
 # --- compiler flags
+set(CMAKE_CXX_STANDARD 11)
```
Sarcasm, line 19: It is my understanding that `CMAKE_CXX_STANDARD`
is more targeted to toolchain files than for the projects directly:

- https://gitlab.kitware.com/cmake/cmake/issues/17146#note_299405

Whereas `target_compile_features(libninja PUBLIC cxx_std_11)`,
just specifies a minimum required standard,
but allow a developer or package to upgrade the standard seamlessly,
e.g. a packager may want to upgrade to C++20 for ABI-compatibility reasons.
This can be done using `CMAKE_CXX_STANDARD` from a toolchain file,
or from the command line:

    cmake -DCMAKE_CXX_STANDARD=20 ...

CMake will also respect the `-std=<standard>` flag,
if specified specified from the environment variable `CXXFLAGS`:

    env CXXFLAGS=-std=c++20 cmake ...

or from the command line `CMAKE_CXX_FLAGS`:

    cmake -DCMAKE_CXX_FLAGS=-std=c++20 ...

One thing to be aware with this method,
is that if one wants to make sure, e.g. in CI,
that the minimum C++ standard is actually tested,
it should do so explicitly, using one of the aforementioned methods.

randomascii:
Okay, it looks like CMakeLists.txt is used on OSX and some Linux build tests, but not everywhere. So we need to specify C++11 in some way in CMakeLists.txt and configure.py, which I am now doing. And it seems to work!

This PR is a mess (too many commits) so I'm going to squash it and make a new branch, and remove the non-C++11 version of the code (it's not used anywhere) and then land this. I'd like to add comments to CMakeLists.txt and configure.py pointing at the other one so that it is more likely that they will be kept in sync in the future.

Thoughts?

randomascii:
Let me know how this looks. If the changes look good then I'll create a new branch with these changes collapsed and we can land that.

jhasse:
I think it would be best to have the switch to C++11 in its own PR. There are a few `#ifdef`s we can drop then, too.

Also please keep your expectations about having this land any time soon low. (sorry in advance)

src/metrics.cc
jonesmz: Sorry, I didn't see this before.

> Any thoughts on how to enable C++11 on Ubuntu 2016.04?

You're referring to the CI build instructions? You'd have to modify .github/workflows/* and CMakeLists.txt

But I think you've already answered your question, give you created https://github.com/ninja-build/ninja/pull/2089

src/metrics.cc
```diff
@@ -34,49 +29,35 @@ Metrics* g_metrics = NULL;
 
 namespace {
 
-#ifndef _WIN32
 /// Compute a platform-specific high-res timer value that fits into an int64.
 int64_t HighResTimer() {
```
jonesmz, line 33: With the assumption that the C++11 PR gets merged, is there a reason not to have HighResTimer return the result of chrono::steady_clock::now() directly? E.g. why convert to int at all?

jonesmz: For that matter, why bother with this function anyway? E.g. just replace everywhere it's called with std::chrono::steady_clock::now()

src/metrics.cc
```diff
@@ -34,49 +29,35 @@ Metrics* g_metrics = NULL;
...
-  timeval tv;
-  if (gettimeofday(&tv, NULL) < 0)
-    Fatal("gettimeofday: %s", strerror(errno));
-  return (int64_t)tv.tv_sec * 1000*1000 + tv.tv_usec;
+  auto now = chrono::steady_clock::now();
```
jonesmz, line 34: Please use std:: namespace on new or modified code.

jhasse:
Can you rebase this on master now that C++11 is enabled?

randomascii:
Rebase done, and the results look good. The branch is a huge mess so let me know if you'd rather I create a new PR with everything squashed.

atetubou (commented):
Could you revert change for configure.py?

randomascii:
> Could you revert change for configure.py?

Good catch. Done.

PR #2030, opened 2021-09-23
Persistent mode, take 3

tmandry:
This is a resurrection of #1438 with some improvements. In particular, the persistent build server supports building arbitrary targets and invoking arbitrary tools. Some changes in options will still cause it to decline a request and exit, but this is somewhat conservative and could be relaxed in the future.

This is a draft because Windows support still needs these improvements applied, and it would be nice to write some tests for the IPC implementation.


src/manifest_parser.h
```diff
@@ -15,10 +15,16 @@
 #ifndef NINJA_MANIFEST_PARSER_H_
 #define NINJA_MANIFEST_PARSER_H_
 
+#include <string>
+#include <vector>
+using namespace std;
```
jonesmz, line 20: Please do not add "using namespace std;" to any files, and for new or modified code, add "std::" where appropriate.

tmandry:
@jhasse, my team was discussing the need for this again today. Are you interested in reviewing? I can get a cleaned up version ready if so.

PR #2034, opened 2021-10-07
Fixed not-linking pthreads that has broken compilation with MSYS2 MinGW.

KOLANICH:


KOLANICH:
It seems to be either a defect in MSYS2 MinGW-w64 toolchain with posix runtime, or in CMake. Anyway, the build is broken and we have to do something about it.

PR #2040, opened 2021-11-03
minidump-win32 is needed by native GNU-like Clang on Windows

### LINKED ISSUE #2006: Unable to build form source with clang
```
lld-link: error: undefined symbol: void __cdecl CreateWin32MiniDump(struct _EXCEPTION_POINTERS *)
>>> referenced by C:/Users/a7bd50f1/Downloads/ninja-1.10.2/src/ninja.cc
>>>               CMakeFiles/ninja.dir/src/ninja.cc.obj
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

The source file containing the implementation of the missing function is conditionally included into compilation inside an `if(MSVC)`. It seems that the flag is not set for clang with GNU-like command line (clang.exe), but it is for MSVC style (clang-cl.exe), even though in both cases MSVC runtime libraries and headers are used, in both cases the target is `x86_64-pc-windows-msvc`. Removing the check makes the ninja build just fine.

The problematic part of CMakeLists.txt:
```
if(MSVC)
	target_sources(libninja PRIVATE src/minidump-win32.cc)
endif()
```
`cmake` was invoked with `CC=clang CXX=clang++ cmake .. -GNinja`
A snippet from the CMake compiler discovery log:
```
-- The C compiler identification is Clang 12.0.1 with GNU-like command-line
-- The CXX compiler identification is Clang 12.0.1 with GNU-like command-line
-- Check for working C compiler: C:/Program Files/LLVM/bin/clang.exe - skipped
-- Check for working CXX compiler: C:/Program Files/LLVM/bin/clang++.exe - skipped

-- Configuring done
-- Generating done
```

scivision:
The entire minidump-win32.cc is guarded by ifdef _MSC_VER, so
MSYS2/MinGW Clang will just ignore minidump-win32.cc contents.

fixes #2006

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,10 +122,8 @@ if(WIN32)
 		src/msvc_helper-win32.cc
 		src/msvc_helper_main-win32.cc
 		src/getopt.c
+		src/minidump-win32.cc
 	)
-	if(MSVC)
-		target_sources(libninja PRIVATE src/minidump-win32.cc)
-	endif()
 else()
 	target_sources(libninja PRIVATE src/subprocess-posix.cc)
 	if(CMAKE_SYSTEM_NAME STREQUAL "OS400" OR CMAKE_SYSTEM_NAME STREQUAL "AIX")


jhasse:
The fix looks fine, but the CI job is for MSVC-like Clang, so I don't see the need to add it (in this PR at least).

PR #2042, opened 2021-11-03
cmake: use generator expression to tersify

scivision:
This PR refactors CMakeLists.txt to use CMake generator expressions to compact the logic. This can help reduce typos and make the script less verbose.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,7 @@ else()
 endif()
 
 # --- compiler flags
+
 if(MSVC)
 	set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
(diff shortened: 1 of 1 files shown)


jonesmz:
If you're making changes to these, please also convert to adding the compile definitions to only the targets, and not to the global compile definitions list, so that some day Ninja can be consumed as a library without the project using it needing to jump through a bunch of hoops.

jhasse:
Am I the only one who finds

```cmake
  if(WIN32)
    target_sources(ninja_test PRIVATE src/includes_normalize_test.cc src/msvc_helper_test.cc)
  endif()
```

a lot more readable than

```cmake
   "$<$<BOOL:${WIN32}>:src/includes_normalize_test.cc;src/msvc_helper_test.cc>"
```

?

PR #2050, opened 2021-11-15
CPack: create .msi installer for winget package

### LINKED ISSUE #2022: Support WinGet package manager for Windows 10+
It would be really great to have ability to just install ninja like that out of the box:

`windget install --id=NinjaBuild.NinjaBuild`

Here is some instructions how it can be done: [Microsoft Docs](https://docs.microsoft.com/en-us/windows/package-manager/package/manifest?tabs=minschema%2Cversion-example)

scivision:
(will) fixes #2022 

Uses CPack + WiX to create a Ninja .msi installer, necessary for winget package creation. 

TODO: need .yaml for winget


foobar442130:
You may be slightly interested in my PR for reasons related to the project call:
https://github.com/ninja-build/ninja/pull/2053

.github/workflows/linux.yml
```diff
@@ -2,7 +2,9 @@ name: Linux
 
 on:
   pull_request:
+    paths: "!**.yml"
```
jonesmz, line 5: This doesn't seem like the best rule, since it would prevent github actions from executing when someone only changed linux.yml, for example.

CMakeLists.txt
```diff
@@ -2,8 +2,16 @@ cmake_minimum_required(VERSION 3.15)
...
 
-project(ninja)
+project(ninja
+	LANGUAGES C CXX
+	DESCRIPTION "Ninja is a small build system with a focus on speed."
+	HOMEPAGE_URL "https://ninja-build.org/"
+	VERSION 1.10.2
+)
+
+cmake_dependent_option(create_msi "Create .msi installer" OFF WIN32 OFF)
```
jonesmz, line 14: An alternative to making a new option would be to always run the cpack code.

Generating the cpack output doesn't create the package, something still has to call cpack.

You would still have to gate the cpack code this PR adds on the WIX generator being available though.

PR #2053, opened 2021-11-20
Win32: Add VERSIONINFO resource to

foobar442130:
This will let people see the API version in the explorer property box.
This will also let users see the version via powershell.

Basically this works by creating a `resource compiler` file. And replacing and using the @ONLY feature to string replace with CMake variables now specified by the project call.

Credit to glfw for figuring out how this works:
https://github.com/glfw/glfw/commit/2551829f75d293d7b4e910b4957798caff01b847#diff-7053816704745f00f0a83c382f0399eb9b4142c6b4b6087934aa3efd4e1b63ce


foobar442130:
Here is how it looks on my machine:

image

image

CMakeLists.txt
```diff
@@ -3,7 +3,10 @@ cmake_minimum_required(VERSION 3.15)
 include(CheckSymbolExists)
 include(CheckIPOSupported)
 
-project(ninja)
+project(NINJA
```
foobar442130, line 6: I renamed the project to not have to deal with casing issues.

ninja_VERSION -> NINJA_VERSION

windows/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+target_sources(ninja PRIVATE 
+    ninja.manifest
+    ninja.rc.in
+)
+
+configure_file(ninja.rc.in ninja.rc @ONLY)
```
foobar442130, line 6: The @ONLY will replace variables with the @NINJA_VERSION_MAJOR@ with the equivalent CMake variable

foobar442130:
Looking at this project further it seems the real build system is the python script and not the cmake. So I'm not sure my change will be accepted. But I hope my point about the resource file gets into the project in some form.

I'm more than willing to change things to get this into ninja.

foobar442130:
I couldn't figure out how the python script worked. So I just made a solution that relied less on CMake.

I ran CMake verbosely and this is the command that needs to be run for MSVC in the ninja python script.

```
ClCompile:
  C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.30.30705\bin\HostX64\x64\CL.exe /c /Zi /nologo /W4 /WX- /diagnostics:column /Od /Ob0 /D _MBCS /D WIN32 /D _WINDOWS /D _CRT_SECURE_NO_WARNINGS /D "
  CMAKE_INTDIR=\"Debug\"" /Gm- /EHsc /RTC1 /MTd /GS /fp:precise /Zc:wchar_t /Zc:forScope /Zc:inline /GR- /Fo"ninja.dir\Debug\\" /Fd"ninja.dir\Debug\vc143.pdb" /external:W4 /Gd /TP /wd4100 /wd4267 /wd4706 /wd4702 /wd4244 /err 
  orReport:queue  /Zc:__cplusplus D:\github\ninja\src\ninja.cc
  ninja.cc
# This command needs to come before the linker step for ninja.exe
ResourceCompile:
  C:\Program Files (x86)\Windows Kits\10\bin\10.0.19041.0\x64\rc.exe /D WIN32 /D _DEBUG /D _WINDOWS /D _CRT_SECURE_NO_WARNINGS /D "CMAKE_INTDIR=\\\"Debug\\\"" /l"0x0409" /nologo /fo"ninja.dir\Debug\ninja.res" D:\github\ninja
  \windows\ninja.rc
Link:
  C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.30.30705\bin\HostX64\x64\link.exe /ERRORREPORT:QUEUE /OUT:"D:\github\ninja\build\vs\Debug\ninja.exe" /INCREMENTAL /ILK:"ninja.dir\Debug\ninja.ilk"
   /NOLOGO kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib /MANIFEST /MANIFESTUAC:"level='asInvoker' uiAccess='false'" /manifest:embed /manifestinput:D:\gi 
  thub\ninja\windows\ninja.manifest /DEBUG /PDB:"D:/github/ninja/build/vs/Debug/ninja.pdb" /SUBSYSTEM:CONSOLE /TLBID:1 /DYNAMICBASE /NXCOMPAT /IMPLIB:"D:/github/ninja/build/vs/Debug/ninja.lib" /MACHINE:X64  /machine:x64 ninj 
  a.dir\Debug\ninja.res
```

foobar442130:
@jhasse in order to get this PR accepted would it be fair to say I'll need to implement the necessary resource compiler functionality in the python script?

jhasse:
No, the official binaries are built using CMake.

foobar442130:
> No, the official binaries are built using CMake.

Oh in that case my first commit was much more on point. I'll revise it.

foobar442130:
I've update the PR to parse the version information from the C++ source code. That way there are less release steps 😊

I also added information for the copyright years 2013-2021, and the later is done via the build system.

foobar442130:
I opted for parsing the source code instead of just using pure CMake, to minimize any disruptions to anyone's workflow on the project. Because while the official binaries are built using CMake I don't want to impair anyone's workflow that uses the configure.py script.

windows/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+    ninja.rc.in
+)
+
+string(TIMESTAMP NINJA_CURRENT_YEAR "%Y")
+
+configure_file(ninja.rc.in ninja.rc @ONLY)
+
+target_sources(ninja PRIVATE
+    ${CMAKE_CURRENT_BINARY_DIR}/ninja.rc
+)
```
jonesmz, line 12: For what it's worth, I don't think it makes sense to create a subdirectory just for windows. This is only 12 lines of cmake code, it should just go into the main cmakelists.txt

jhasse:
Sorry for the late response.

I agree with @jonesmz. Please fix the merge conflict and try to simplify the change. After that reopen the PR.

Issue #1900, opened 2020-12-21
Should support filename longer than 260 characters


6ziv:
Well, I can see that it is a limitation from Microsoft Windows.
But MSDN suggests that this limitation may be removed.
[Enable Long Paths in Windows 10, Version 1607, and Later](https://docs.microsoft.com/zh-cn/windows/win32/fileio/maximum-file-path-limitation#enable-long-paths-in-windows-10-version-1607-and-later)
And there are cases we have to use long paths (when handling some really complicated 3rd-party projects).
Maybe ninja could only raise a warning, but do not stop the build, like cmake does?

jhasse:
Duplicate of #1514.

jhasse:
#1514 was removed by GitHub because the user who created was banned or something. No idea why they won't simply remove his posts only ...

Anyway, I will reopen this one then. #1514 can still be accessed via the GitHub API: https://api.github.com/repos/ninja-build/ninja/issues/1514/comments

jhasse:
https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation

LDong-Arm:
> https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation

Hi @jhasse,
 
From the page you linked, the user needs to edit the registry to enable long path support, and also the application needs to declare it in its own manifest (pasted from that page):
```
<application xmlns="urn:schemas-microsoft-com:asm.v3">
    <windowsSettings xmlns:ws2="http://schemas.microsoft.com/SMI/2016/WindowsSettings">
        <ws2:longPathAware>true</ws2:longPathAware>
    </windowsSettings>
</application>
```
Presumably adding this to [ninja.manifest](https://github.com/ninja-build/ninja/blob/master/windows/ninja.manifest) should work? Or how would this impact compatibility with old Windows?

Thanks.

LDong-Arm:
Oh https://github.com/ninja-build/ninja/pull/1939 already does that actually.

emmenlau:
So, what are the next steps for this issue?

parbo:
`RealDiskInterFace` uses posix file functions like fopen, and also the *A versions of win32 calls, while the [long paths page](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry) says only *W versions are affected. Another option might be to prepend `\\?\`, but this only works for absolute paths.

parbo:
[...]

PR #1939, opened 2021-03-25
Add long path support for Windows

majaeger:
Currently, ninja only works for paths less than MAX_PATH (255) on Windows. This change makes ninja long-path aware by manifesting the application as "longPathAware", and using the unicode version of windows file APIs

See the below docs for more info:
https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats#skip-normalization
https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation#enable-long-paths-in-windows-10-version-1607-and-later

PR #2056, opened 2021-11-23
Add long path support for Windows [Approach #2]

majaeger:
Currently, ninja only works for paths less than MAX_PATH (255) on Windows. This change makes ninja long-path aware by manifesting the application as "longPathAware", and using the unicode version of windows file APIs. 

See the below docs for more info:
https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats#skip-normalization
https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation#enable-long-paths-in-windows-10-version-1607-and-later

This is a second attempt based off of https://github.com/ninja-build/ninja/pull/1939 , which attempts to address some of the shortcomings / difficulties with the previous approach.

Difference:
The main difference between this and the previous PR, all windows APIs were moved to the UNICODE version. In this one, both UNICODE and non-UNICODE are supported based on the BUILD_UNICODE option set in CMakeLists.txt. This has a couple of advantages:

* For non-unicode version, there are no extra copies incurred to convert paths to wstrings
* Compiling for the non-UNICODE version executes the same code paths (mostly) as for non-windows platforms. One of the big difficulties I hit previously was getting the correct behavior between windows and non-windows platforms
* No core logic changes. Given how deep the concept of paths as std::string is ingrained in the ninja code-base, trying to convert everything to use wstring (and still support std::string for non-windows) is a massive undertaking (And I found that I missed a ton of instances in the previous PR, especially around C file functions). By converting to wstring on-demand, there are no core logic changes with this PR. Just conversions to/from std::string/std::wstring when calling platform APIs.

Based on these performance results, I've enabled the BUILD_UNICODE option by default. If this is not desirable, I can switch it to OFF instead.

| Test Name      | UNICODE     | ANSI     |
| :------------- | :----------: | -----------: |
|  build_log_perftest | min 13ms  max 14ms  avg **13.2ms** | min 14ms  max 14ms  avg 14.0ms |
| canon_perftest | min 189ms  max 204ms  avg 196.8ms | min 187ms  max 207ms  avg **192.2ms** |
| clparser_perftest | Parse 16384 times in 3019ms avg **184.3us** | Parse 8192 times in 2367ms avg 288.9us   |
| depfile_parser_perftest | (Unknown what args to use) |  |
| hash_collision_bench | 0 collisions after 20000000 runs | 0 collisions after 20000000 runs |
| manifest_parser_perftest  | (Failed to run python script) |  |


jonesmz:
Could you provide a comparison / contrast against https://github.com/ninja-build/ninja/pull/1671 ?

majaeger:
> Could you provide a comparison / contrast against #1671 ?

Sure thing @jonesmz . I would say that the fundamental concept between the two is similar, but they have different goals.

The linked PR aims to fully support Unicode within ninja. This means it cares about string contents as well as string type, which results in changes that ripple through a large portion of the ninja code base. It also means that many code paths are different between windows and linux builds. The fact that long paths are supported with this approach is just a side-effect of using the Unicode version of windows C / Win32 APIs. It also fully removes the ability to build for windows with no UNICODE support.

In contrast, this PR enables UNICODE for the sole purpose of enabling long path support. That means that std:;string is still the main representation of a path within ninja, and std::string are converted to std::wstring on-demand only when being passed into win32 / CRT functions which take a file path. This means that no core-logic is actually changed between the unicode/non-unicode builds, at the cost of some minor overhead when copying a std::string to a std::wstring. It also maintains the ability to build without UNICODE defined, which greatly simplifies debugging issues between the windows build and linux build. The downside is that, while UNICODE is defined, this change doesn't actually make any attempt to ensure unicode string contents are handled differently than ascii (Same behavior as today)

Milerius:
when this will be merged and a new release will be available?

CMakeSettings.json
```diff
@@ -0,0 +1,26 @@
+﻿{
```
cristianadam, line 1: The `CMakeSettings.json` is not needed for this PR.

src/file_path.h
```diff
@@ -0,0 +1,123 @@
...
+#endif /* !_TCHAR_DEFINED */
+
+#ifdef UNICODE
+typedef std::wstring file_string_t;
+#else
+typedef std::string file_string_t;
+#endif
+
+inline std::string NarrowPath(const std::wstring& path) {
+  return std::string(path.begin(), path.end());
```
digit-google, line 45: Are you sure that this is what you want here? losing information silently doesn't sound like a great idea.

digit-google:
Both this PR and #1671 seem to move in the right direction, but I would suggest creating a dedicated FilePath class that holds a file path in the "native" format, irrespective of "UNICODE", with methods to wrap fopen()/unlink()/stat()/whatever, plus some helpers like FilePath::AppendUTF8(). Of course only use them when actually accessing the file system, Ninja can keep using UTF-8 strings for its internal representation of paths in the build graph, for example.

This will make the code easier to maintain, and avoids introducing yet another build configuration setting / variant for the Win32 build.

And to be clear, I'd happily introduce a PR to do that after any one of these two is merged.

stephanosio:
Just adding a reference to the issue that this PR fixes.

Closes #1900.

src/file_path.h
majaeger: The goal of this PR is solely to add long path support by using the unicode version of all windows APIs. It does not actually add support for Unicode, which is a much broader change. So the behavior of what contents a path may contain are the same in this PR as in the existing ninja releases

majaeger:
@digit-google , I 100% agree. That's the path I originally started down, however it ended up being a much broader change than I felt comfortable introducing. Especially since it impacted the linux build as well, and I do not have as much experience there.

src/util.h
```diff
@@ -113,13 +116,50 @@ bool Truncate(const std::string& path, size_t size, std::string* err);
...
+#define getcwd _wgetcwd
+#define mkdir _wmkdir
+
+#define t_snprintf _snwprintf
+#define t_splitpath _wsplitpath
+#define t_stricmp _wcsicmp
+#define t_strlen wcslen
+
+#define PATH_MAX 2048
+#else
```
digit-google, line 134: the multiple `#else` and `#endif` directives are a little hard to follow. Can you use comment to clarify them a little, e.g.:

```
#ifdef _MSC_VER
....
#ifdef UNICODE
...
#else  // !UNICODE
...
#endif // !UNICODE
#else // !_MSC_VER
...
#endif // !_MSC_VER
```

src/test.cc
```diff
@@ -69,7 +69,7 @@ char* mkdtemp(char* name_template) {
 string GetSystemTempDir() {
 #ifdef _WIN32
   char buf[1024];
-  if (!GetTempPath(sizeof(buf), buf))
+  if (!GetTempPathA(sizeof(buf), buf))
```
digit-google, line 72: nit: since this calls GetTempPathA() directory, consider using `buf[MAX_PATH + 1]` here (see https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppatha)

src/file_path.h
digit-google: This should at a minimum call WideCharToMultiByte() to ensure that the result matches either UTF-8 or the current codepage, instead of truncating whatever UTF-16 values in the input path to 8-bits. For example, there is one example of calling FindFirstFileExW() explicitly, whose result is passed to this NarrowPath() function, and this is not correct.

https://docs.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-widechartomultibyte

majaeger: How would you recommend handling errors from WideCharToMultiByte?

digit-google: It depends on the call site (e.g. for FindFirstFileExW() you would just ignore the file). It's probably better not to have a generic NarrowPath() function at all, and handle the wide-to-byte conversion errors explicitly at the small number of call sites where they could happen.

MehdiChinoune:
Does it work on MinGW?
If this is restricted to MSVC, then the title should be changed.

digit-google:
It looks like the code change should work on Mingw, only the CMakeList changes needs an update. I wonder if this needs an update in `configure.py` too? The latter is not used to generate release binaries for Windows, but I assume some people would wnat a Unicode-aware Windows binary generated with it too.

jhasse:
I would prefer it if Microsoft lifts the 260 character limit for *A functions instead.

jonesmz:
That's unlikely for new versions of windows, and never going to happen for existing versions.

Is ninja planning to drop support for e.g. windows 7, 8, 8.1 and 10?

jhasse:
> [...] and never going to happen for existing versions.

Microsoft has added UTF-8 support to Windows 10 via Windows Update so I don't see why this is never going to happen.

> Is ninja planning to drop support for e.g. windows 7, 8, 8.1 and 10?

There is no "ninja" entity which "plans" anything.

jonesmz:
> Microsoft has added UTF-8 support to Windows 10 via Windows Update so I don't see why this is never going to happen.

You want to simultaneously support versions of Linux so old that they don't have C++11 compilers, while also drop support for versions of windows older than windows 10? That's inconsistent. Are we supporting older systems, or not?

> There is no "ninja" entity which "plans" anything.

The group of people who are entitled to make decisions about what code gets merged into the codebase is what i meant by "ninja" here.

That means, effectively, you, since you're the only one out of the current list of maintainers on the Ninja-build organization on github who's done anything of substance in years.

So are you planning to drop support for older versions of windows, or not? Personally, I'm all for it, and would be happy to hear that's the plan.

majaeger:
I agree it is unlikely that the *A functions are updated to change behavior due to back-compat requirements. While it would be more convenient if we could just use the A functions in this code base, under the hood the A functions also widen the string and forward the calls to the W version of the APIs as well (After performing additional legacy logic related to DOS paths), so there's not actually a big difference.

From an outside contributor's perspective, I think the most helpful thing for me would have been if c++17 could be used easily throughout ninja, which would enable use of std::filesystem and std::string_view APIs . I think that would get the code closer to the kind of change @digit-google described above.

jonesmz:
@majaeger I haven't rebased on latest ninja codebase in quite a while, life and all that.

But you're welcome to take a look at my fork of ninja, where I did add a patch to use std::filesystem, and another patch to use std::string_view.

I'd be happy to collaborate on things like this with you.

https://github.com/splinter-build/splinter/tree/16-std-filesystem

Note that that repo is organized as a series of patches, so a branch starting with N is intended to also include the commits from a branch starting with N-1, and so on.

This, of course, can make it confusing to work with, but the intention of that project has always been to get the patches applied to ninja, and as such i thought it would be easier to work in terms of patches, instead of doing a hard-fork.

jonesmz:
well, i spent the evening rebasing all of my patches on the latest changes. Couldn't get the filesystem branch to build (see 16-filesystem-broken), but the other ones should all build, or be pretty close. Looks like there are some CI failures due to environmental issues that i haven't had time to mess with.

Regardless, the std::filesystem changes aren't terribly far from working again, so like i said, happy to collaborate if you have any interest.

jonesmz:
> While it would be more convenient if we could just use the A functions in this code base, under the hood the A functions also widen the string and forward the calls to the W version of the APIs as well (After performing additional legacy logic related to DOS paths), so there's not actually a big difference.

What's the runtime performance of this?

If the utf-8 strings are being widened every time one of these functions is called, how is that any better, in terms of runtime perf? It seems to me like, if the strings have to be widened at every call, we'd be much better off with this implementation: https://github.com/ninja-build/ninja/pull/1671

majaeger:
@jonesmz , the performance numbers for this change are in the initial PR description. It's basically a win in some tests and a slight loss in others. This is because the win32 API *A() functions themselves also widen the strings passed in, but do additional processing regarding legacy DOS paths. By doing the widening in the ninja code and using the *W() functions, we skip that additional processing, and the path gets passed directly to the filesystem driver.

I haven't looked through all of the changes you linked, but in general I expect performance to be about the same unless ninja uses  wide strings everywhere, thus eliminating the need for either the ninja code or the win32 A* functions to widen the strings.

I did actually see your fork before I started work on this. While I personally like the direction you've taken a lot of the code, I'm currently restricted in my corporate build environment to the main project. If my free time opens up though, I'll keep an eye out for issues in your repo that I might be able to assist with :)

jhasse:
If a new feature only works on newer Windows versions, that's not "dropping support for older versions". See UTF-8 or ANSI color output for example.

If a PR would result in ninja.exe not running anymore (or a regression) on Windows 7, I wouldn't merge it.

emmenlau:
@majaeger I was trying to pull your changes into Ninja 1.10.2, but its not so easy. I think it may help if you _rebased_ your work on Ninja master, rather than pulling Ninja master into your branch. Or maybe my git-foo is not sufficient to cherry-pick your commits.

majaeger:
@emmenlau , I pushed a new version which is rebased on upstream/master

emmenlau:
Awesome, thanks! I can confirm that your changes work successfully in our build system to compile hundreds of common open source projects. The changes also resolve the following [issue where Qt can not be compiled](https://bugreports.qt.io/browse/QTBUG-101181).

digit-google:
Can you rework this into a smaller and cleaner list of commits? The erroneous additions / deletions / last-minute compilation fixes are annoying when inspecting the history, and will make bisecting difficult in the future when we need this. Ideally each commit should compile and work properly.

Unfortunately, I am not maintainer and cannot approve running workflows to check that everything works as expected. Will try locally.

.gitignore
```diff
@@ -43,3 +43,5 @@
 /.clangd/
 /compile_commands.json
 /.cache/
+/.vs
+/out/
```
Neumann-A, line 47:
```suggestion
```

CMakeLists.txt
```diff
@@ -152,7 +161,7 @@ add_executable(ninja src/ninja.cc)
 target_link_libraries(ninja PRIVATE libninja libninja-re2c)
 
 if(WIN32)
-  target_sources(ninja PRIVATE windows/ninja.manifest)
+	target_sources(ninja PRIVATE windows/ninja.manifest)
```
Neumann-A, line 164: remove ws change.

CMakeLists.txt
```diff
@@ -183,53 +192,53 @@ endif()
...
+		target_link_libraries(${perftest} PRIVATE libninja libninja-re2c)
+	endforeach()
+
+	if(CMAKE_SYSTEM_NAME STREQUAL "AIX" AND CMAKE_SIZEOF_VOID_P EQUAL 4)
+		# These tests require more memory than will fit in the standard AIX shared stack/heap (256M)
+		target_link_options(hash_collision_bench PRIVATE "-Wl,-bmaxdata:0x80000000")
+		target_link_options(manifest_parser_perftest PRIVATE "-Wl,-bmaxdata:0x80000000")
+	endif()
+
+	add_test(NAME NinjaTest COMMAND ninja_test)
```
Neumann-A, line 241: remove ws change

CMakeSettings.json
```diff
@@ -0,0 +1,26 @@
...
+      "configurationType": "Release",
+      "inheritEnvironments": [ "msvc_x64_x64" ],
+      "buildRoot": "${projectDir}\\out\\build\\${name}",
+      "installRoot": "${projectDir}\\out\\install\\${name}",
+      "cmakeCommandArgs": "",
+      "buildCommandArgs": "",
+      "ctestCommandArgs": ""
+    }
+  ]
+}
```
Neumann-A, line 26:
```suggestion
```

Neumann-A: Consider using `CMakePresetsUser.json` for your usage.

BillyONeal:
> I would prefer it if Microsoft lifts the 260 character limit for *A functions instead.
> > [...] and never going to happen for existing versions.
> 
> Microsoft has added UTF-8 support to Windows 10 via Windows Update so I don't see why this is never going to happen.

To my limited understanding the path limit remains on the A functions for backcompat reasons (e.g. apps that pass around fixed MAX_PATH sized buffers before calling these).

src/file_path.cc
```diff
@@ -0,0 +1,37 @@
...
+#include <vector>
+
+#ifdef UNICODE
+
+#define WIN32_LEAN_AND_MEAN
+#include <Windows.h>
+
+bool NarrowPath(const std::wstring& path, std::string* narrowPath, std::string* err) {
+  std::vector<char> pathChars(PATH_MAX, '\0');
+  if (!WideCharToMultiByte(CP_UTF8, WC_ERR_INVALID_CHARS, path.c_str(), -1, pathChars.data(), pathChars.size(), nullptr, nullptr)) {
```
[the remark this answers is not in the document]
jhasse: Since #1915 Ninja expects UTF-8.

jhasse:
@BillyONeal The change is behind a manifest setting so it wouldn't break backwards compatibility.

emmenlau:
@jhasse, do you think this PR is something that can be considered for integration? The change helps us quite a bit building Qt on MSVC.

jhasse:
@emmenlau IMHO this PR is a workaround for something Microsoft needs to fix. Have you contacted them about it?

emmenlau:
@jhasse I agree that Microsoft should fix this in their code.

But from the top of my head, I would know at least ten more things that are significantly broken in Microsoft code. When I was younger, I sometimes even went to the effort to contact them. But not a single time has this lead to anything, not even feedback on my report. So I personally would not be willing to invest into Microsoft fixing their code, I'd much rather have something working. And looking at this thread, I see a number of other responses that seem to hint at a similar feeling or stance.

If you are saying you prefer keeping ninja free of workarounds for other peoples bugs, I can understand. But it would be just nice to know if this PR here has a potential future in ninja, because currently it helps me a lot to get my work done and I'm a bit concerned that as a patch, it may not stay applicable and maintained forever.

jhasse:
You're probably right. I haven't decided on what to do regarding this issue. Not something for 1.11.0 though, as there's the UTF-8 change already.

Neumann-A:
>  IMHO this PR is a workaround for something Microsoft needs to fix. Have you contacted them about it?

@jhasse: The A apis are **DOCUMENTED** to be limited by **MAX_PATH** (260 CHARS). (see e.g. here: [GetLongPathNameA](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getlongpathnamea))

So instead of requesting something unreasonable from MS, fix the usage of the wrong API in ninja. 
From my view this is definitely a bug in ninja not reading the MS docs well enough.

jhasse:
@Neumann-A I know, I want Microsoft to add a way to opt-in to lift this limitation just as it was added for the W APIs.

emmenlau:
Thanks a lot for the thumbs up, @jhasse ! I can very much relate that finding the right balance what to put into ninja and what not can be a challenge. But this PR seems to have potential to make the lives of a lot of people better, so I'm happy if it can be considered 😀.

What I can say so far is that we've been using this patch-set successfully in ninja master c5e50948 since about a month on Windows 11, Windows 10, Ubuntu 20.04 and macOS 11.5 and it resolved the MSVC Qt build problems.

Let me know if there is anything else I can do to move it forward. @majaeger I guess it would be great if it can be kept up to date against master, and maybe the commits could be squashed into a smaller set?

@jhasse, does that even make sense at this point in time, or should this PR wait for a thorough review from you or other core developers first?

Neumann-A:
To add to this:
https://github.com/microsoft/vcpkg/pull/23201
also uses this patch + a newer version of ninja. No problems observed in any ports depending on `zlib` or `meson` which are the entry points for the ninja-dependency in that PR.

jhasse:
> @jhasse, does that even make sense at this point in time, or should this PR wait for a thorough review from you or other core developers first?

I don't have time to review this right now and as I said, I haven't decided what I think would be the best way forward. I'm very sorry :(

> Depending on utf-8 support is nice if you don't care about past Windows, but otherwise you're leaving out many customers who have unicode in their paths.

I don't understand. UTF-8 *is* Unicode.

BillyONeal:
> > Depending on utf-8 support is nice if you don't care about past Windows, but otherwise you're leaving out many customers who have unicode in their paths.
> 
> I don't understand. UTF-8 _is_ Unicode.

Past Windows, as in already shipped releases. Even if the policy gets changed to what you want it would only affect whatever the most current Win10 and Win11 releases are.

src/file_path.h
```diff
@@ -0,0 +1,124 @@
...
+
+#endif
+
+inline bool NarrowPath(const std::string& path, std::string* narrowPath, std::string* err) {
+  *narrowPath = path;
+  return true;
+}
+
+inline std::wstring WidenPath(const std::string& path) {
+  return std::wstring(path.begin(), path.end());
```
Osyotr, line 56: Shouldn't this also call `MultiByteToWideChar`?

src/file_path.cc
```diff
@@ -0,0 +1,37 @@
...
+#include <string>
+#include <vector>
+
+#ifdef UNICODE
+
+#define WIN32_LEAN_AND_MEAN
+#include <Windows.h>
+
+bool NarrowPath(const std::wstring& path, std::string* narrowPath, std::string* err) {
+  std::vector<char> pathChars(PATH_MAX, '\0');
```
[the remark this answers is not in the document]
jonesmz: Ninja builds as pre-C++11.
std::array<> is not available until C++17.
see: https://github.com/ninja-build/ninja/pull/2089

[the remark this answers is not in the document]
Osyotr: `char[]` then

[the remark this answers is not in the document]
jonesmz: Well, yes, I agree with avoiding a std::vector that's just going to be thrown away. 

either char[], or std::string with "resize" called on it.

strega-nil-ms:
> I don't understand. UTF-8 _is_ Unicode.

UTF-8 is not Unicode - UTF-8 is a standard representation of unicode text, but there are many other representations of unicode text, including (but not limited to):

* The three standard representations:
  - UTF-8
  - UTF-16
  - UTF-32
* Many non-standard representations, including the most important - GB18030, the official encoding of the PRC

Windows and environments like it (Java, Javascript, SMS) were based originally on UCS-2 since they came before Unicode had decided to break its 65,536 character limit, and so they are now WTF-16 -- UTF-16 but allowing unpaired surrogates (although to be fair, not supporting unpaired surrogates is not a thing anyone cares about unless you're a file system implementation).

This fundamentally means that, to talk to these systems, you _need_ to convert to UTF-16 on the boundaries.

jhasse:
I misunderstood your sentence about leaving out past Windows, of course I know UTF-8 is not Unicode itself but a Unicode encoding.

> This fundamentally means that, to talk to these systems, you *need* to convert to UTF-16 on the boundaries.

You don't *need* to do that, otherwise Ninja wouldn't work on Windows before this PR. You do need to convert to UTF-16 to gain long path support.

src/file_path.cc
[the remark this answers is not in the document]
abique: `std::string` will call malloc as well as `std::vector`.

abique:
Hi,

I don't understand the resistance here to add long path support.

It is thousand of hours if not more that people loose every single month because of the long path issue on Windows. And when the build fails it is very hard to detect and extremely disappointing to find that in 2022 we're stuck because of long path issue on Windows. This leads to bad practices, like building in `C:\x\` to have the shortest path possible and hope that it will work.

I believe it should be a top priority. We're lucky to have people from Microsoft helping out to find the best solution in regard of Windows history and backward compatibility which you seem to be concerned about.

I hope this gets solved soon.

Regards,
Alex.

lygstate:
ping for merge? anything stop this?

lygstate:
For older OS, we can always using UNC path to handle long path.
I am not sure if this PR are fixing the issue properly, also we can using utf-8 everywhere, there is no double about

awarwick:
> I don't have time to review this right now and as I said, I haven't decided what I think would be the best way forward. I'm very sorry :(

@jhasse do you have time to review this? If not, can you pass this off to someone else to review? This is blocking our company as well. Our paths aren't even that long -- it's just generated intermediary files which can be nearly 2x the length of the actual paths checked into source which are causing problems, and it's a bit unfortunate for something as robust as ninja to be stuck on dinosaur OS APIs that enable these problems.

jonesmz:
My workplace just burnt 3 engineers for > 1 month working around this same issue. 

Ultimately we played whack-a-mole with the various scripts that could generate long paths, shortening folder names whereever we could, to get under the limit.

old-pigeon:
I compiled the latest Ninja git master with this patch in hope to resolve this issue: https://github.com/microsoft/vcpkg/issues/23714

Sadly, it didn't help. Only this fixed the problem: https://github.com/microsoft/vcpkg/issues/23714#issuecomment-1075002693

So it seems this patch is either not complete (catching all cases?) or is buggy.

Neumann-A:
> So it seems this patch is either not complete (catching all cases?) or is buggy.

This is because there are other tools being a road block.. In the qt case there are moc/uic/rcc which don't have long paths enabled (although they easily could) but the ultimativ roadblock will be the MSVC tools which heavily use MAX_PATH and are thus limited. So unless you use a full llvm toolchain which includes stuff like rc/mt you won't get full long path support but only a few extra characters.

Ju1He1:
> > So it seems this patch is either not complete (catching all cases?) or is buggy.
> 
> This is because there are other tools being a road block.. In the qt case there are moc/uic/rcc which don't have long paths enabled (although they easily could) but the ultimativ roadblock will be the MSVC tools which heavily use MAX_PATH and are thus limited. So unless you use a full llvm toolchain which includes stuff like rc/mt you won't get full long path support but only a few extra characters.

so there is no chance ninja + msvc will receive a patch to resolve this nasty issue?

tntljc:
> > So it seems this patch is either not complete (catching all cases?) or is buggy.
> 
> This is because there are other tools being a road block.. In the qt case there are moc/uic/rcc which don't have long paths enabled (although they easily could) but the ultimativ roadblock will be the MSVC tools which heavily use MAX_PATH and are thus limited. So unless you use a full llvm toolchain which includes stuff like rc/mt you won't get full long path support but only a few extra characters.

Ninja supporting the long path is still valuable at least for my project. The long path won't be involved in any msvc compilation but it blocked CMake + ninja.

emmenlau:
> Ninja supporting the long path is still valuable at least for my project. The long path won't be involved in any msvc compilation but it blocked CMake + ninja.

Well, if you can build ninja from source, this PR here worked nicely for us for about a year now. If you can not build from source let me know, maybe I can send a pre-built Windows executable.

old-pigeon:
When it worked for over a year, can this "just" be merged?

abique:
This branch has conflict, so the PR should be updated.

emmenlau:
> When it worked for over a year, can this "just" be merged?

Well, as was discussed [before](https://github.com/ninja-build/ninja/pull/2056#issuecomment-1076791642), there is still no consensus inside the ninja team if this PR is the right way forward. Kindly pinging @jhasse for suggestions where to go from here...

abique:
Could the long path feature be turned off using a command line switch?

emmenlau:
> Could the long path feature be turned off using a command line switch?

Why would you want to turn it off?

abique:
> > Could the long path feature be turned off using a command line switch?
> 
> Why would you want to turn it off?

It seems that the ninja team was against having long path for some reasons, to not break legacy build systems etc...
So having a switch to turn it on/off solves the issue with builds that requires long path disabled.

emmenlau:
> It seems that the ninja team was against having long path for some reasons, to not break legacy build systems etc...
> So having a switch to turn it on/off solves the issue with builds that requires long path disabled.

I do not think that this is the main concern. I think the main concern is that this PR adds a significant change to the codebase, and there are other ways how these changes could be implemented. Depending on the implementation, there are different pro's and con's considering the complexity of the change-set and the effort of future maintenance.

This is just my understanding. But I would be quite surprised if anybody would explicitly disable long path support when its merged. The downsides I can see are in the code complexity and maintenance.

Osyotr:
The amount of changes in this PR can be reduced by removing `#if _UNICODE` logic and always using W-versions of  WinAPI.
There is no point in supporting A-versions.

jhasse:
> I think the main concern is that this PR adds a significant change to the codebase, and there are other ways how these changes could be implemented.

Well kind of. If you mean by "other changes" changes by Microsoft and not in Ninja's code: Support for long paths in combination with UTF-8.

cristianadam:
At https://github.com/ninja-build/ninja/pull/2225 I have a very small PR that just adds the "longPathAware" manifest, which allows usage of long paths if the proper registry key is set.

It has 0 code changes and for the users that need long path support they can have it.

majaeger:
@jhasse , I am very skeptical that the change in #2225 resolved all of the long path issues that this PR handles. I also can't continue to keep up this branch updated forever if it is not going to be considered though. If I perform testing on #2225 for additional cases and find it is still not sufficient, will this PR actually be considered?

jhasse:
IMHO all additional cases not solved by #2225 should be reported to Microsoft. This really shouldn't be app developers responsibility.

saschasc:
I tested this PR and it solved the long path issues of ninja on Windows. First I tried #2225 and it did not solve anything. I don't understand your reasoning why Microsoft should solve this?

jhasse:
ninja isn't the only Windows program. There are thousands of others. It's less work if Microsoft puts in the (probably huge? No idea, my bet would be that there code sucks BIG TIME. Still:) work to fix this in Windows so that all the thousands apps only need #2225 and not #2056, than every program doing #2056.

cristianadam:
> In the qt case there are moc/uic/rcc which don't have long paths enabled (although they easily could) but the ultimativ roadblock will be the MSVC tools which heavily use MAX_PATH and are thus limited. So unless you use a full llvm toolchain which includes stuff like rc/mt you won't get full long path support but only a few extra characters.

I took care of moc/uic/rcc in Qt 6.5 with the `longPathAware` manifest at https://codereview.qt-project.org/c/qt/qtbase/+/447113, similar to https://github.com/ninja-build/ninja/pull/2225 of ninja.

Soon you can on Windows use: clang-cl, Qt, and ninja to build applications for Windows using the MSVC ABI.

levicki:
@jhasse 
> ninja isn't the only Windows program. There are thousands of others.

But it is the one used to build say [xformers](https://github.com/facebookresearch/xformers) from source code, and now that is failing thanks to ninja still not supporting long paths after 4 years of deliberation on how to proceed on the path issue(s) reported in the repo.

I don't see what is stopping the team from fixing ths?

 - Using *W APIs should have happened long ago
 - Why not use [std::filesystem::path](https://en.cppreference.com/w/cpp/filesystem/path) instead of rolling your own filepath class as suggested elsewhere? It supports all relevant conversions.
 - Manifest will be ignored on older Windows which don't support it.
 - ninja can check the registry key on startup and print a warning for the user to enable it if they have issues with long paths.
 - on older Windows versions ninja can prepend `\\?\` to the path and override `MAX_PATH` limitation like pretty much all apps dealing with paths on Windows were doing since like forever.

It's 2024 for God sake, issues with long paths (and/or spaces in paths) should be a thing of the past.

jhasse:
It was reported that long path support works with current master, please give it a try and comment there if it doesn't work for you: https://github.com/ninja-build/ninja/issues/1900#issuecomment-1817532728

emmenlau:
> It was reported that long path support works with current master, please give it a try and comment there if it doesn't work for you

The solution in master does not work for me. I'm sorry, I thought this was known? A number of people keep on having problems with the solution in master, me for one.

I've enabled the registry setting for long paths in Windows 11, then built ninja from latest master, and it *does* work in many cases, where it failed without the manifest fix/registry hack before.

However, there are still some cases where the build fails. I could not really figure out why, but for example, Qt fails to build for me (fully reproducible). In contrast, I've switched to the PR https://github.com/ninja-build/ninja/pull/2056 from Matt Jaeger successfully since a rather long time, and subsequently never had any long path issues ever since.

levicki:
> It was reported that long path support works with current master, please give it a try and comment there if it doesn't work for you: [#1900 (comment)](https://github.com/ninja-build/ninja/issues/1900#issuecomment-1817532728)

I can do that but I don't know whether putting the `ninja.exe` built from master into a folder which is in the `PATH` is enough for `pip` to use that version when building a wheel provided that no `ninja` package is installed in current `venv` or machine wide?

See, that's why we need a release. Install the package and Bob's your uncle instead of having to figure out a way to override each and every build system which is now using `ninja`.

PR #2063, opened 2021-12-29
Use python3 instead of python

maruel:
Some systems do not have python available. Explicitly using python3
makes it clearer and more consistent.

misc/output_test.py was already using "#!/usr/bin/env python3" so there
was some inconsistency already.


maruel:
For the record, .github/workflows/linux.yml explicitly uses "python3".

There is user-visible change, e.g. ninja -t browse will run python3 instead of python.

jhasse:
Some systems do not have python3 available (see CI failure for example). Those scripts are compatible with both python 2 and 3 and therefore I think using just `python` is more clear.

maruel:
Fair enough. I realized after making the PR that I could pass `--with-python=python3` to configure. Apologies for not realizing before making the PR.

maruel:
Actually there's one file to fix to make this work 100% with "python" is not in PATH. I will make a separate PR.

PR #2064, opened 2021-12-31
Make manifest_parser_perftest respect NINJA_PYTHON

maruel:
The NINJA_PYTHON define is set by passing --with-python to configure.py.
The default value is set to the python executable running configure.py.

The problem is that one perf test was hardcoding "python" instead of
using NINJA_PYTHON.

To reproduce, run with and without this patch the following commands:

  git clean -f -d -x
  python3 configure.py --bootstrap
  ./ninja manifest_parser_perftest
  ./manifest_parser_perftest

Without this change, output is:

  Creating manifest data...sh: 1: python: not found


maruel:
Checks failure is odd, I swear I tested locally.

maruel:
Oh I see now that CMakeLists.txt only set NINJA_PYTHON (hardcoded to python BTW) for browse.cc and nothing else. I never edited a CMake file before so I'll try my best.

maruel:
I'm doing it wrong; looks like it needs extra escaping on Windows?

https://ci.appveyor.com/project/nico/ninja/builds/42049759/job/y9whh9imjah27dml

```
FAILED: build/manifest_parser_perftest.obj 
cl /showIncludes /nologo /Zi /W4 /WX /wd4530 /wd4100 /wd4706 /wd4244 /wd4512 /wd4800 /wd4702 /wd4819 /wd4127 /wd4355 /wd4091 /GR- /wd4267 /DNOMINMAX /D_CRT_SECURE_NO_WARNINGS /D_HAS_EXCEPTIONS=0 /DNINJA_PYTHON="python.exe" /FS /Ox /DNDEBUG /GL -I. -c src\manifest_parser_perftest.cc /Fobuild\manifest_parser_perftest.obj /Fdbuild\manifest_parser_perftest.pdb
src\manifest_parser_perftest.cc(51): error C2065: 'python': undeclared identifier
src\manifest_parser_perftest.cc(51): error C2143: syntax error: missing ';' before 'string'
```

maruel:
Oh well, I give up at this point as I don't have access to Visual Studio installed on a Windows machine.

PR #2065, opened 2021-12-31
Disable re2c from embedding the version number in the source file

maruel:
The checked in sources (depfile_parser.cc and lexer.cc) have the re2c version
embedded in it. The output didn't change since re2c 1.3, except for the
different version number.

A reproducible build would use a pinned version of re2c anyway, so there
is no need to hardcode its value in the checked in generated files.


maruel:
I just realized that I forgot to update [misc/ci.py](https://github.com/ninja-build/ninja/blob/master/misc/ci.py).

@jhasse do you want me to send another PR to trim the two entries in this file?

jhasse:
What do you mean exactly?

Issue #760, opened 2014-05-06
Ninja scans dependencies before order-only dependencies are run


mathstuf:
There is a [CMake bug](http://www.cmake.org/Bug/view.php?id=14747) about ninja not running this properly:

``` ninja
rule R
    command = $COMMAND
build always: R
    COMMAND = echo 1 >> out
build out: phony out
build out-copy: R out || always
    COMMAND = cp out out-copy
build all: phony out-copy always
default all
```

Basically, ninja is scanning the dependencies of `out-copy` before `always` is run and therefore skips the fact that `out` has changed. Is it intended that dependency scanning is only guaranteed to be done after direct dependencies, not order-only dependencies or a bug? If the former, documentation that this is the case would be nice.

ahartmetz:
Is anybody still interested in implementing Brad's suggestion?
I use CMake with Ninja, I find the byproducts thing to be a bit of a bummer because it's hard to explain to people who just want to build something without becoming an expert in build system matters, and I enjoy working on graph algorithms. If nobody else picks this up, I'll add it to the pile of stuff I want to do in "my copious free time" :blush:.

arjanhouben:
I think this bug should be given more attention, as builds can very easily be incorrect without the user knowing it.
I use a lot of the following constructs.
```
add_custom_target(
   generate_version
    COMMAND
    <generate version.tmp>
    copy_if_different version.tmp version.h
)
add_executable( my_exe main.cpp version.h )
target_link_libraries( my_exe some_lib )
add_dependencies( my_exe generate_version )
```
This is an oversimplified example, leaving out the GENERATED property for the **version.h** file for example.
The problem here is that while **my_exe** will probably build without issues, if the only thing to change is the contents of **version.h**, **my_exe** will not be updated.
This _may_ be obvious enough since the build step will be pretty much instant. 
However, as in my case often is, if **some_lib** is updated which should lead to a different **version.h**, **my_exe** is NOT updated, since the only thing to happen is to re-link the executable.
Another problem with this, is that it is non-trivial to check afterwards, as the timestamps of **version.h**, **some_lib** and **my_exe** _might_ actually be exactly as you expect, while still not having included the latest content of **version.h** in **my_exe**!

mathstuf:
FYI, the better way to do that is `add_custom_command(OUTPUT version.h)` followed by `add_custom_target(generate_version DEPENDS version.h)`. The dependencies actually work that way.

arjanhouben:
[...]

PR #2067, opened 2022-01-06
Print "explain" debug just before each command is run

drothlis:
The explanations are printed just before each command is run. Before, it would potentially print pages & pages of explanations at the beginning, and it was hard to tell which explanation caused a particular command to run.

Edges that are pruned due to "restat" used to print all their explanations at the beginning, cluttering up the output with explanations that didn't actually cause any rebuilding. Now, those explanations aren't printed if those edges aren't run.

### DIFF
--- a/src/debug_flags.cc
+++ b/src/debug_flags.cc
@@ -12,6 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <stdio.h>
+#include <map>
+#include <vector>
+#include <string>
+
+#include "graph.h"
+
 bool g_explaining = false;
 
 bool g_keep_depfile = false;
@@ -19,3 +26,21 @@ bool g_keep_depfile = false;
 bool g_keep_rsp = false;
 
 bool g_experimental_statcache = true;
+
+// Reasons each Node needs rebuilding, for "-d explain".
+typedef std::map<const Node*, std::vector<std::string> > Explanations;
+static Explanations explanations_;
+
+void record_explanation(const Node* node, std::string explanation) {
+  explanations_[node].push_back(explanation);
+}
+
+void print_explanations(FILE *stream, const Edge* edge) {
+  for (std::vector<Node*>::const_iterator o = edge->outputs_.begin();
+       o != edge->outputs_.end(); ++o) {
+    for (std::vector<std::string>::iterator s = explanations_[*o].begin();
+         s != explanations_[*o].end(); ++s) {
+      fprintf(stream, "ninja explain: %s\n", (*s).c_str());
+    }
+  }
+}
--- a/src/debug_flags.h
+++ b/src/debug_flags.h
@@ -17,12 +17,20 @@
 
 #include <stdio.h>
 
-#define EXPLAIN(fmt, ...) {                                             \
-  if (g_explaining)                                                     \
-    fprintf(stderr, "ninja explain: " fmt "\n", __VA_ARGS__);           \
+#define EXPLAIN(node, fmt, ...) {                                       \
+  if (g_explaining) {                                                   \
+    char buf[1024];                                                     \
+    snprintf(buf, 1024, fmt, __VA_ARGS__);                              \
+    record_explanation(node, buf);                                      \
+  }                                                                     \
 }
 
+struct Edge;
+struct Node;
+
 extern bool g_explaining;
+void record_explanation(const Node* node, std::string reason);
+void print_explanations(FILE *stream, const Edge* node);
 
 extern bool g_keep_depfile;
 
--- a/misc/output_test.py
+++ b/misc/output_test.py
@@ -19,35 +19,56 @@ default_env.pop('CLICOLOR_FORCE', None)
 default_env['TERM'] = ''
 NINJA_PATH = os.path.abspath('./ninja')
 
+class BuildDir:
+    def __init__(self, build_ninja: str):
+        self.build_ninja = build_ninja
+        self.d = None
+
+    def __enter__(self):
+        self.d = tempfile.TemporaryDirectory()
+        with open(os.path.join(self.d.name, 'build.ninja'), 'w') as f:
+            f.write(self.build_ninja)
+            f.flush()
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.d.cleanup()
+
+    def run(
+        self,
+        flags: str = '',
+        pipe: bool = False,
+        env: Dict[str, str] = default_env,
+    ) -> str:
+        ninja_cmd = '{} {}'.format(NINJA_PATH, flags)
+        try:
+            if pipe:
+                output = subprocess.check_output(
+                    [ninja_cmd], shell=True, cwd=self.d.name, env=env)
+            elif platform.system() == 'Darwin':
+                output = subprocess.check_output(['script', '-q', '/dev/null', 'bash', '-c', ninja_cmd],
+                                                 cwd=self.d.name, env=env)
+            else:
+                output = subprocess.check_output(['script', '-qfec', ninja_cmd, '/dev/null'],
+                                                 cwd=self.d.name, env=env)
+        except subprocess.CalledProcessError as err:
+            sys.stdout.buffer.write(err.output)
+            raise err
+        final_output = ''
+        for line in output.decode('utf-8').splitlines(True):
+            if len(line) > 0 and line[-1] == '\r':
+                continue
+            final_output += line.replace('\r', '')
+        return final_output
+
 def run(
     build_ninja: str,
     flags: str = '',
     pipe: bool = False,
     env: Dict[str, str] = default_env,
 ) -> str:
-    with tempfile.TemporaryDirectory() as d:
-        with open(os.path.join(d, 'build.ninja'), 'w') as f:
-            f.write(build_ninja)
-            f.flush()
-        ninja_cmd = '{} {}'.format(NINJA_PATH, flags)
-        try:
-            if pipe:
-                output = subprocess.check_output([ninja_cmd], shell=True, cwd=d, env=env)
-            elif platform.system() == 'Darwin':
-                output = subprocess.check_output(['script', '-q', '/dev/null', 'bash', '-c', ninja_cmd],
-                                                 cwd=d, env=env)
-            else:
-                output = subprocess.check_output(['script', '-qfec', ninja_cmd, '/dev/null'],
-                                                 cwd=d, env=env)
-        except subprocess.CalledProcessError as err:
-            sys.stdout.buffer.write(err.output)
-            raise err
-    final_output = ''
-    for line in output.decode('utf-8').splitlines(True):
-        if len(line) > 0 and line[-1] == '\r':
-            continue
-        final_output += line.replace('\r', '')
-    return final_output
+    with BuildDir(build_ninja) as b:
+        return b.run(flags, pipe, env)
 
 @unittest.skipIf(platform.system() == 'Windows', 'These test methods do not work on Windows')
 class Output(unittest.TestCase):
--- a/misc/output_test.py
+++ b/misc/output_test.py
@@ -11,6 +11,7 @@ import subprocess
 import sys
 import tempfile
 import unittest
+from textwrap import dedent
 from typing import Dict
 
 default_env = dict(os.environ)
@@ -21,7 +22,7 @@ NINJA_PATH = os.path.abspath('./ninja')
 
 class BuildDir:
     def __init__(self, build_ninja: str):
-        self.build_ninja = build_ninja
+        self.build_ninja = dedent(build_ninja)
         self.d = None
 
     def __enter__(self):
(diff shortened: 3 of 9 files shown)


drothlis:
Example output:

<table>
<tr>
<th>Command</th>
<th>this branch</th>
<th>master</th>
</tr>
<tr>
<td>ninja -f misc/explain.ninja -d explain -v</td>
<td>

```
ninja explain: .FORCE is dirty
[1/5] [ -e input1 ] || touch input1
ninja explain: .FORCE is dirty
[2/5] [ -e input2 ] || touch input2
ninja explain: input1 is dirty
[3/5] echo "Writing!" > mid1
ninja explain: mid1 is dirty
[4/5] echo "Writing!" > mid2
ninja explain: mid2 is dirty
ninja explain: input2 is dirty
[5/5] echo "Writing!" > output
```

</td>
<td>

```
ninja explain: output .FORCE of phony edge with no inputs doesn't exist
ninja explain: .FORCE is dirty
ninja explain: input1 is dirty
ninja explain: mid1 is dirty
ninja explain: mid2 is dirty
ninja explain: .FORCE is dirty
ninja explain: input2 is dirty
[1/5] [ -e input1 ] || touch input1
[2/5] [ -e input2 ] || touch input2
[3/5] echo "Writing!" > mid1
[4/5] echo "Writing!" > mid2
[5/5] echo "Writing!" > output
```

</td>
</tr>
<tr>
<td>ninja -f misc/explain.ninja -d explain -v</td>
<td>

```
ninja explain: .FORCE is dirty
[1/5] [ -e input1 ] || touch input1
ninja explain: .FORCE is dirty
[2/3] [ -e input2 ] || touch input2
```

</td>
<td>

```
ninja explain: output .FORCE of phony edge with no inputs doesn't exist
ninja explain: .FORCE is dirty
ninja explain: input1 is dirty
ninja explain: mid1 is dirty
ninja explain: mid2 is dirty
ninja explain: .FORCE is dirty
ninja explain: input2 is dirty
[1/5] [ -e input1 ] || touch input1
[2/3] [ -e input2 ] || touch input2
```

</td>
</tr>
<tr>
<td>touch input1</td>
<td></td>
<td></td>
</tr>
<tr>
<td>ninja -f misc/explain.ninja -d explain -v</td>
<td>

```
ninja explain: .FORCE is dirty
[1/5] [ -e input1 ] || touch input1
ninja explain: .FORCE is dirty
[2/5] [ -e input2 ] || touch input2
ninja explain: input1 is dirty
ninja explain: output mid1 older than most recent input input1 (1641477997421878016 vs 1641478036759380371)
[3/5] echo "Writing!" > mid1
ninja explain: mid1 is dirty
[4/5] echo "Writing!" > mid2
ninja explain: mid2 is dirty
ninja explain: input2 is dirty
[5/5] echo "Writing!" > output
```

</td>
<td>

```
ninja explain: output .FORCE of phony edge with no inputs doesn't exist
ninja explain: .FORCE is dirty
ninja explain: input1 is dirty
ninja explain: mid1 is dirty
ninja explain: mid2 is dirty
ninja explain: .FORCE is dirty
ninja explain: input2 is dirty
[1/5] [ -e input1 ] || touch input1
ninja explain: output mid1 older than most recent input input1 (1641478283252786871 vs 1641478287856962471)
[2/5] [ -e input2 ] || touch input2
[3/5] echo "Writing!" > mid1
[4/5] echo "Writing!" > mid2
[5/5] echo "Writing!" > output
```

</td>
</tr>
<tr>
<td>touch input2</td>
<td></td>
<td></td>
</tr>
<tr>
<td>ninja -f misc/explain.ninja -d explain -v</td>
<td>

```
ninja explain: .FORCE is dirty
[1/5] [ -e input1 ] || touch input1
ninja explain: .FORCE is dirty
[2/3] [ -e input2 ] || touch input2
ninja explain: mid2 is dirty
ninja explain: input2 is dirty
ninja explain: output output older than most recent input input2 (1641478037915424518 vs 1641478085829253903)
[3/3] echo "Writing!" > output
```

</td>
<td>

```
ninja explain: output .FORCE of phony edge with no inputs doesn't exist
ninja explain: .FORCE is dirty
ninja explain: input1 is dirty
ninja explain: mid1 is dirty
ninja explain: mid2 is dirty
ninja explain: .FORCE is dirty
ninja explain: input2 is dirty
[1/5] [ -e input1 ] || touch input1
[2/3] [ -e input2 ] || touch input2
ninja explain: output output older than most recent input input2 (1641478288869001069 vs 1641478292477138679)
[3/3] echo "Writing!" > output
```

</td>
</tr>
</table>

drothlis:
You can see that inputs that were dirty, but then marked clean because of `restat`, **are** still printed if the rule was still run for another reason. You can see this in the line "ninja explain: input2 is dirty" after `touch input1`.

But at least it isn't showing inputs that aren't involved in any rules being re-built.

drothlis:
I squashed & force-pushed a fix for C++98 compatibility to fix [this build failure](https://github.com/ninja-build/ninja/runs/4729530208).

drothlis:
For a real-world comparison, at my day job a no-op build with "-d explain" prints 1,800 lines (before ninja starts building anything); with my patches the total output is 93 lines. (As you might imagine we use *restat* a lot.) You can see an example of this build output at https://david.rothlis.net/ninja-explain/

src/debug_flags.cc
```diff
@@ -12,10 +12,35 @@
...
 
 bool g_keep_depfile = false;
 
 bool g_keep_rsp = false;
 
 bool g_experimental_statcache = true;
+
+// Reasons each Node needs rebuilding, for "-d explain".
+typedef std::map<const Node*, std::vector<std::string> > Explanations;
+static Explanations explanations_;
```
digit-google, line 32: A few remarks: this is probably not the right place to put this, consider moving this to its own header + source files (e.g. explanations.h + explanations.cc), and declare a single Explanations class which Record() and Print() methods instead. Also an explanations_test.cc would be welcome to check that this works correctly even if the implementation changes in the future.

Also please avoid introducing new global variables. An Explanations* pointer should be passed to all functions that need to call the EXPLAIN() macro instead (I know this sucks, probably introduce this in its own commit). The Explanations object should probably be owned by the Builder class.

src/debug_flags.cc
```diff
@@ -12,10 +12,35 @@
...
+
+void record_explanation(const Node* node, std::string explanation) {
```
digit-google, line 34: nit: use a `StringPiece` type instead of `std::string` here to avoid an un-necessary string copy, e.g.:

```
void record_explanation(const Node* node, StringPiece explanation) {
  explanations_[node].emplace_back(explanation.str_, explanation.len_);
}
```

src/debug_flags.h
```diff
@@ -17,12 +17,20 @@
...
 #include <stdio.h>
 
-#define EXPLAIN(fmt, ...) {                                             \
-  if (g_explaining)                                                     \
-    fprintf(stderr, "ninja explain: " fmt "\n", __VA_ARGS__);           \
+#define EXPLAIN(node, fmt, ...) {                                       \
+  if (g_explaining) {                                                   \
+    char buf[1024];                                                     \
+    snprintf(buf, 1024, fmt, __VA_ARGS__);                              \
+    record_explanation(node, buf);                                      \
```
digit-google, line 24: Consider doing the temporary buffer allocation directly in a `record_explanation`, as in:

```
void record_explanation(const Node*, const char* fmt, ...);
```

digit-google: Actually, with a dedicated class and pointed object, you can drop the macro entirely and replace all instances of `EXPLAIN(.....)` with `explanations->Record(node, ....)` which will be clearer.

src/debug_flags.cc
digit-google: And the pointer could be used for checks instead of the `g_explaining` global flag.

misc/explain.ninja
```diff
@@ -0,0 +1,23 @@
+# Use like:
```
digit-google, line 1: It is not clear how to use this file by reading this comment. I know that all the details are in the conversation for this PR, but this is not enough. Please add a Python3 regression test under `misc/explanation_test.py` that actually invokes all the steps described in the conversation for this PR, and checks that everything works correctly. See `misc/output.py` for inspiration, or add something to it instead.

In all cases, every new feature should have proper unit and regression tests, otherwise maintenance of the code base is complicated (how do we know if a change actually breaks stuff people rely on), and slows down future reviews (there is only so much that human review can catch).

digit-google:
I've added quite a few comments here, but I'd like to clarify that this is a good PR.

One important remark though is that this changes Ninja's output. This _may_ be a touchy topic because some people actually try to parse it automatically in various ways, so any change here can risk breaking someone's workflow / scripts. However given that all lines still begin with the same `ninja explain: ` prefix, this risk is probably limited. I wonder what @jhasse thinks about this topic. A simple work-around would be to add one more configuration flag to change the behavior, but the less there are the better for everyone :)

src/debug_flags.cc
drothlis: Which global variable do you mean? If you're talking about `explanations_`, doesn't `static` make it local to the compilation unit? (It's quite possible I'm wrong -- it's been a long time since I've done C++).

Thanks for the review, by the way!

digit-google: Yes, I am talking about `explanations` which is global variable (in the sense of mutable global state). These are best avoided because they tend to hide dependencies from the source code, they have their own life-cycle which can introduce unexpected coupling issues, they make unit-testing more difficult, and they also complicate future refactors (e.g. there are plans to turn Ninja into a library, but it could be anything that tries to change the graph data structures); also they are generally not thread-safe. So let's not introduce more of them if possible :-)

drothlis:
I have pushed a unit test. I'll work on @digit-google's other suggestions, I think they make sense.

> See misc/output.py for inspiration, or add something to it instead.

Thanks for pointing me at that, it was very helpful.

> The Explanations object should probably be owned by the Builder class.

Thanks for this guidance too.

drothlis:
Note that this pull request interacts with ninja's default non-verbose output, because the "explain" output prints a newline, so when ninja updates the status output from `[1/3]` to `[2/3]` it will print anew instead of overwriting the previous line, and it will print the "explain" output for that build line again:

<table>
<tr><th>ninja -d explain</th><th>ninja -d explain --verbose</th></tr>
<tr><td>

```
ninja explain: .FORCE is dirty
[0/3] [ -e input ] || touch input
ninja explain: .FORCE is dirty
[1/3] [ -e input ] || touch input
ninja explain: input is dirty
[1/3] cp input mid
ninja explain: input is dirty
[2/3] cp input mid
ninja explain: mid is dirty
[2/3] cp mid output
ninja explain: mid is dirty
[3/3] cp mid output
```

</td><td>

```
ninja explain: .FORCE is dirty
[1/3] [ -e input ] || touch input
ninja explain: input is dirty
[2/3] cp input mid
ninja explain: mid is dirty
[3/3] cp mid output
```

</td></tr>
</table>

I don't think it's a problem in practice, because it doesn't really make sense not to use `--verbose` if you're asking for *extra* debugging with `-d explain`. Maybe `-d explain` should imply `--verbose`? Either way I don't think it's a blocker.

enerjazzer:
Is there any update on this? It's an important feature....

mathstuf:
This is somewhat related to #760. If that were implemented (very unlikely as it is a major rework of the internals), the "do we need to run this" wouldn't even be done until the node is considered to run and buffering wouldn't have to be done.

enerjazzer:
> This is somewhat related to #760. If that were implemented (very unlikely as it is a major rework of the internals), the "do we need to run this" wouldn't even be done until the node is considered to run and buffering wouldn't have to be done.

What I'm more interested in is seeing "why this target is considered dirty" (i.e. "which dependency is dirty"). Currently ninja just prints it's dirty and that's it, which is not helpful when you try to pin down the dirty dependency when a target has a lot of dependencies.

anthonyalayo:
I landed here for the same problems as others. Can we have an update on this?

Spongman:
what happened to this? what needs to happen to get this out of the bikeshed?

drothlis:
I don't intend to do any further work on this, so anyone's welcome to take over. I do still use this in my daily workflow — it works well enough for me as is, and I simply don't have the time.

Spongman:
yeah, i can imagine it's quite demotivating, seeing a problem, coming up with a solution, offering it to the community, and having it die on the vine in code-review hell.
well, thanks for trying anyway.

jhasse:
I agree that this is a good change! Sorry I didn't have enough time two years ago to merge this right away.

> what needs to happen to get this out of the bikeshed?

The conflicts would need to be resolved.

drothlis:
Excellent! I have rebased on master. The patch to `graph.cc`, `status.cc` (now `status_printer.cc`), and `output_test.py` didn't apply cleanly so I re-did those manually; everything else was fairly straight-forward.

drothlis:
@jhasse I have rebased / fixed the conflicts, and all the CI checks have passed, so it's ready for you to have a look when you have the time. :-)

jhasse:
I agree with all of @digit-google's remaining comments. Still I merged it as is, as you said you don't indent to work more on this. We can do the adjustments now with other PRs / directly on master.
The output test is also nice, thanks!

emosy:
shout-out to everyone who got this over the finish line after 2 years of abandonment!

digit-google:
FYI, I uploaded PR #2451 to cleanup the code base from the global variable (no change in behavior, just saner sources to better help understand what's going on under the hood).

PR #2071, opened 2022-01-19
Allow linking Ninja binary with extra library.

digit-google:
This patch adds a NINJA_EXTRA_LINK_LIBRARY option to
`CMakeLists.txt`, and a --link-with option to
`configure.py`.

These allow building a version of the Ninja binary that
is linked against a better allocation library.

For example linking against rpmalloc [1] on Linux yields
a notable speedup (10.4 -> 9.1 seconds for a no-op Fuchsia
build with warm filesystem caches).

Example usage:

  mkdir build-cmake && cd build-cmake && rm -rf *
  export LDFLAGS="-lpthread -ldl"
  cmake .. -DCMAKE_BUILD_TYPE=Release -GNinja ... \
           -DNINJA_EXTRA_LINK_LIBRARY=/path/to/librpmalloc_wrapper.a

Or:

  export LDFLAGS="-lpthread -ldl"
  ./configure.py --bootstrap --link-with=/path/to/librpmalloc_wrapper.a

Where the extra linker flags are required by rpmalloc in this case.

[1] https://github.com/mjansson/rpmalloc

Change-Id: I5a50f818f43516ccd82047ce39b6ada156b5232a

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -151,6 +151,15 @@ endif()
 add_executable(ninja src/ninja.cc)
 target_link_libraries(ninja PRIVATE libninja libninja-re2c)
 
+# Define NINJA_EXTRA_LINK_LIBRARY to add new link-time libraries
+# to the final ninja executable. This can be used to link a custom
+# allocator to speed up the tool.
+if(NINJA_EXTRA_LINK_LIBRARY)
+  add_library(extra_link STATIC IMPORTED)
+  set_property(TARGET extra_link PROPERTY IMPORTED_LOCATION "${NINJA_EXTRA_LINK_LIBRARY}")
+  target_link_libraries(ninja PRIVATE extra_link)
+endif()
+
 if(WIN32)
   target_sources(ninja PRIVATE windows/ninja.manifest)
 endif()
--- a/configure.py
+++ b/configure.py
@@ -223,6 +223,8 @@ parser.add_option('--with-python', metavar='EXE',
 parser.add_option('--force-pselect', action='store_true',
                   help='ppoll() is used by default where available, '
                        'but some platforms may need to use pselect instead',)
+parser.add_option('--link-with', metavar='LIB',
+                  help='Link Ninja executable with additional library LIB.')
 (options, args) = parser.parse_args()
 if args:
     print('ERROR: extra unparsed command-line arguments:', args)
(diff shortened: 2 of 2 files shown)


jonesmz:
This is a really odd way to approach this.

The idiomatic cmake way to link a target with an extra library is to add the cmakelist as a subdirectory of a new top-level cmake list and call "target_link_libraries(ninja PRIVATE mynewlibrary)"

Or to modify the LDFLAGS environment variable.

Or to set CMAKE_EXE_LINKER_FLAGS

Or any of the other environment variables that can influence this.

Specifically adding a new option to the cmake lists file is unnecessary and weird.

jonesmz:
Further, if you *really* want to approach the problem like this, you need to ensure (and provide test cases) for both static *and* dynamic libraries. Currently it looks like you're only providing support for static libraries.

digit-google:
Unfortunately, using LDFLAGS does not work, because CMake will put the flags at the start of the final link command, while the allocator library needs to appear at the end, after other static libraries used by the executable (libninja here). Otherwise, its content will be ignored due to the way Unix linking works.

You can verify this with the following (in an umodified ninja checkout):

```
touch /tmp/libfoo.a
mkdir build
cd build
LDFLAGS=/tmp/libfoo.a cmake ..
cat CMakeFiles/ninja.dir/link.txt 
```

CMAKE_EXE_LINKER_FLAGS suffers from the same problem, except that it also requires the use of a toolchain file (setting it on the command-line doesn't seem to work and actually prevents the generation of valid Makefiles  for reasons I didn't have the time to debug).

This is also configure.py had to be modified to guarantee the same thing.

If you are aware of any environment or CMake variable that could guarantee this behaviour, I would be happy to use it instead.

Using Ninja as a CMake sub-project is an interesting idea though, I'll try it and report quickly if it works correctly.

Finally, can you clarify why you think supporting dynamic libraries here would be useful? On Unix it is already possible to use LD_PRELOAD or DYLD_INSERT_LIBRARIES to use an allocator shared library. On Windows, you would need to link the binary against the import library for the DLL, which is a static library anyway.

Also, the point of using a static library is to generate a single executable that can be used anywhere and distributed easily.

digit-google:
And the Ninja subproject idea actually works! For the record, here's the CMakeLists.txt that I'm using:

```
project(ninja-rpmalloc)
add_subdirectory("${NINJA_SOURCE_DIR}" ninja)
target_link_libraries(ninja PRIVATE "${RPMALLOC_LIB}" -lpthread -ldl)
```

Where I use `cmake .. -DNINJA_SOURCE_DIR=/path/to/ninja-src -DRPMALLOC_LIB=/path/to/rpmalloc-lib && make ninja` to generate a `ninja` binary that is properly linked against rpmalloc (I verified this manually and running some benchmarks).

It would be nice to be able to do something similar with `configure.py`, but this really  isn't a blocker for me, thus I'm closing this PR. Thanks for the suggestion.

PR #2073, opened 2022-01-25
Unit tests path for Windows-compiled

Gesugao-san:
Compiling Ninja on Windows creates the file in a different path, and the path syntax in Windows is different from Linux.


jhasse:
This depends on the generator used by CMake. Something like `ctest` could be used to have a Generator-independent way to run the unit tests.

PR #2075, opened 2022-02-03
Make the output of `ninja -t inputs` deterministic

digit-google:
This sorts the output of `ninja -t inputs` to make it
deterministic and remove duplicates.

### DIFF
--- a/src/graph_test.cc
+++ b/src/graph_test.cc
@@ -215,6 +215,39 @@ TEST_F(GraphTest, RootNodes) {
   }
 }
 
+TEST_F(GraphTest, CollectInputs) {
+  ASSERT_NO_FATAL_FAILURE(AssertParse(
+      &state_,
+      "build out$ 1: cat in1 in2 in$ with$ space | implicit || order_only\n"));
+
+  std::vector<std::string> inputs;
+  Edge* edge = GetNode("out 1")->in_edge();
+
+  // Test without shell escaping.
+  inputs.clear();
+  edge->CollectInputs(false, &inputs);
+  EXPECT_EQ(5u, inputs.size());
+  EXPECT_EQ("in1", inputs[0]);
+  EXPECT_EQ("in2", inputs[1]);
+  EXPECT_EQ("in with space", inputs[2]);
+  EXPECT_EQ("implicit", inputs[3]);
+  EXPECT_EQ("order_only", inputs[4]);
+
+  // Test with shell escaping.
+  inputs.clear();
+  edge->CollectInputs(true, &inputs);
+  EXPECT_EQ(5u, inputs.size());
+  EXPECT_EQ("in1", inputs[0]);
+  EXPECT_EQ("in2", inputs[1]);
+#ifdef _WIN32
+  EXPECT_EQ("\"in with space\"", inputs[2]);
+#else
+  EXPECT_EQ("'in with space'", inputs[2]);
+#endif
+  EXPECT_EQ("implicit", inputs[3]);
+  EXPECT_EQ("order_only", inputs[4]);
+}
+
 TEST_F(GraphTest, VarInOutPathEscaping) {
   ASSERT_NO_FATAL_FAILURE(AssertParse(&state_,
 "build a$ b: cat no'space with$ space$$ no\"space2\n"));
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -17,6 +17,8 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+
+#include <algorithm>
 #include <cstdlib>
 
 #ifdef _WIN32
@@ -744,7 +746,8 @@ int NinjaMain::ToolCommands(const Options* options, int argc, char* argv[]) {
   return 0;
 }
 
-void PrintInputs(Edge* edge, set<Edge*>* seen) {
+void CollectInputs(Edge* edge, std::set<Edge*>* seen,
+                   std::vector<std::string>* result) {
   if (!edge)
     return;
   if (!seen->insert(edge).second)
@@ -752,13 +755,41 @@ void PrintInputs(Edge* edge, set<Edge*>* seen) {
 
   for (vector<Node*>::iterator in = edge->inputs_.begin();
        in != edge->inputs_.end(); ++in)
-    PrintInputs((*in)->in_edge(), seen);
+    CollectInputs((*in)->in_edge(), seen, result);
 
-  if (!edge->is_phony())
-    puts(edge->GetBinding("in_newline").c_str());
+  if (!edge->is_phony()) {
+    edge->CollectInputs(true, result);
+  }
 }
 
 int NinjaMain::ToolInputs(const Options* options, int argc, char* argv[]) {
+  // The inputs tool uses getopt, and expects argv[0] to contain the name of
+  // the tool, i.e. "inputs".
+  argc++;
+  argv--;
+  optind = 1;
+  int opt;
+  const option kLongOptions[] = { { "help", no_argument, NULL, 'h' },
+                                  { NULL, 0, NULL, 0 } };
+  while ((opt = getopt_long(argc, argv, "h", kLongOptions, NULL)) != -1) {
+    switch (opt) {
+    case 'h':
+    default:
+      // clang-format off
+      printf(
+"Usage '-t inputs [options] [targets]\n"
+"\n"
+"List all inputs used for a set of targets. Note that this includes\n"
+"explicit, implicit and order-only inputs, but not validation ones.\n\n"
+"Options:\n"
+"  -h, --help   Print this message.\n");
+      // clang-format on
+      return 1;
+    }
+  }
+  argv += optind;
+  argc -= optind;
+
   vector<Node*> nodes;
   string err;
   if (!CollectTargetsFromArgs(argc, argv, &nodes, &err)) {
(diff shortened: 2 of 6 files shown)


jhasse:
LGTM

@naarcini what do you think?

naarcini (approved):
Generally LGTM % a couple of suggestions

src/ninja.cc
```diff
@@ -766,9 +770,30 @@ int NinjaMain::ToolInputs(const Options* options, int argc, char* argv[]) {
...
   }
 
-  set<Edge*> seen;
+  std::set<Edge*> seen;
+  std::vector<std::string> result;
   for (vector<Node*>::iterator in = nodes.begin(); in != nodes.end(); ++in)
-    PrintInputs((*in)->in_edge(), &seen);
+    CollectInputs((*in)->in_edge(), &seen, &result);
+
+  // Make output deterministic by sorting then removing duplicates.
```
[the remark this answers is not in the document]
digit-google: I think I'm confused, I though std::unique was not part of C++03 !? if it is I would gladly use it (the code is based on the implementation example on https://en.cppreference.com/w/cpp/algorithm/unique)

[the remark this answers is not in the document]
naarcini: I found it difficult to find good documentation about what is or isn't in C++03, but I believe std::unique is present. I'm basing this partly on the cppreference page for the algorithm library, which annotates which version of C++ particular functions were added (no annotation for `unique`):
https://en.cppreference.com/w/cpp/algorithm

And partly by clicking through the gcc github repo, which includes an implementation for std::unique in version 2.95 (last edited 1998):
https://github.com/gcc-mirror/gcc/blob/releases/gcc-2.95/libstdc%2B%2B/stl/stl_algo.h

digit-google:
I have added a unit-test for Edge::CollectInputs. It seems there are no good ways to test tools though. Should we want to add these, should we go with regression tests (e.g. scripts that invoke Ninja with pre-determined build plans, and verify behaviour), or refactor ninja.cc to make tools unit-testable? Probably  not in this PR though :-)

jhasse:
Some very basic regression tests are in `misc/output_test.py`, e.g.: https://github.com/ninja-build/ninja/blob/cd967ab21eef51819473b0cea22135fead76ffc2/misc/output_test.py#L117

Making tools unit-testable is probably a good idea though.

digit-google:
Thanks, I could add a regression test to `output_test.py` for the new tool.

I'm still concerned about the fact that we do not include implicit inputs in the the tool's output. The fact that they are not listed in the command line itself doesn't make them less important. And I don't see a use case where you would want only the explicit inputs anyway. should we add an "explicit-inputs" tool for these otherwise?

naarcini:
@digit-google In my specific use case, we only needed the explicit dependencies. At the time, I was trying to improve upon a project in Chromium that audits certain metadata in code (Network Traffic Annotation Auditor). In that case, all I needed was specifically the code files that are used to build a particular target. Since Chromium's tooling does a pretty good job of explicitly listing all dependencies, I found that "explicit" was enough to fulfill this requirement.

I'm not opposed to adding an additional flag or changing the default behavior to include more information. That'd still work fine for my use case and I suspect it'd work fine for most other peoples' use cases too. If I get a vote, I vote for using an additional flag instead of multiple tool names :)

digit-google:
Latest update adds `--type=TYPE` option to the tool, as in `-t inputs --type=TYPE <targets>`. The default is to print explicit + implicit inputs, but this can be changed with `--type=explicit` and `--type=all`. I think that should address all use cases now :)

digit-google:
Friendly ping, anything else to get this merged? I would really prefer if the existing behavior was not part of the next 1.11 release :-/

jhasse:
The `--type` flag needs to be split into its own commit. But before you do that: Is the flag really necessary? I would like to keep the inputs tool simple for the moment.

Also:

> (i.e. this allows us to skip any stale outputs from a previous build, since cleandead will not work in certain cases).

Sounds like a bug of the cleandead tool?

digit-google:
Well, I have a use case where I need all implicit + explicit dependencies, and @naarcini only wants the explicit ones, so a `--type` flag seems the better option here.

This is not really a bug in `cleandead`, it is more a Ninja design issue: the `.ninja_deps` file will record all entries generated by depfiles, and sometimes, some of them will be stale (because a command changed and will no longer use that file to the build and the corresponding depfile). Ninja has no way to detect these cases before re-running the build, so `cleandead` will always keep the stale files in the build directory. This has been causing a variety of breakages in our incremental builds (e.g. https://fxbug.dev/83397 or more recently https://fxbug.dev/85465.

This tool would be one of the different ways we want to work around this problem.

I'll update the PR to make the `--type  flag its own commit.

src/graph_test.cc
```diff
@@ -215,6 +215,53 @@ TEST_F(GraphTest, RootNodes) {
...
+TEST_F(GraphTest, CollectInputs) {
+  ASSERT_NO_FATAL_FAILURE(AssertParse(
+      &state_,
+      "build out$ 1: cat in1 in2 in$ with$ space | implicit || order_only\n"));
+
+  std::vector<std::string> inputs;
+  Edge* edge = GetNode("out 1")->in_edge();
+
+  inputs.clear();
+  edge->CollectInputs(false, InputsType::Default, &inputs);
```
jhasse, line 227: This needs to be part of the second commit.

src/ninja.cc
```diff
@@ -211,29 +213,31 @@ struct Tool {
...
+      "this system]\n"
+      "  -k N     keep going until N jobs fail (0 means infinity) [default=1]\n"
+      "  -l N     do not start new jobs if the load average is greater than N\n"
+      "  -n       dry run (don't run commands but act like they succeeded)\n"
+      "\n"
+      "  -d MODE  enable debugging (use '-d list' to list modes)\n"
+      "  -t TOOL  run a subtool (use '-t list' to list subtools)\n"
+      "    terminates toplevel options; further flags are passed to the tool\n"
+      "  -w FLAG  adjust warnings (use '-w list' to list warnings)\n",
+      kNinjaVersion, config.parallelism);
```
jhasse, line 241: Why this change?

src/ninja.cc
```diff
@@ -744,31 +748,66 @@ int NinjaMain::ToolCommands(const Options* options, int argc, char* argv[]) {
   return 0;
 }
 
-void PrintInputs(Edge* edge, set<Edge*>* seen) {
+void CollectInputs(Edge* edge, set<Edge*>* seen,
```
jhasse: while you're at it, can you add `std::` here?

src/ninja.cc
```diff
@@ -31,13 +31,15 @@
 #include <unistd.h>
 #endif
 
+#include <algorithm>
```
jhasse, line 34: can you move this include to the top?

src/ninja.cc
```diff
@@ -31,13 +31,15 @@
...
+
 #include "browse.h"
 #include "build.h"
 #include "build_log.h"
-#include "deps_log.h"
 #include "clean.h"
 #include "debug_flags.h"
 #include "depfile_parser.h"
+#include "deps_log.h"
```
jhasse, line 42: any reason for this reordering?

src/ninja.cc
```diff
@@ -771,19 +772,36 @@ int NinjaMain::ToolInputs(const Options* options, int argc, char* argv[]) {
...
+      }
+      if (!inputs_type.ParseArg(optarg))
+        Fatal("Invalid --type value, see '-t inputs --type=list'");
+      break;
     case 'h':
     default:
       printf(
           "Usage '-t inputs [options] [targets]\n"
           "\n"
-          "List all inputs used for a set of targets. Note that this ignores\n"
```
jhasse: The "List all inputs used for a set of targets." shouldn't be removed, right?

jhasse (changes requested):
> and @naarcini only wants the explicit ones

If I understood him correctly, he wouldn't mind the implicit ones, but doesn't need them?

The extra `InputsType` struct is quite a lot of code and I only merged the inputs tool because it was such a small change.

digit-google:
I pushed a new version with the changes you requested 10 days ago, but forgot to submit my review comments, which I just did. Sorry. Let me know if there is something else to do to get this submitted. Thanks.

src/ninja.cc
jhasse: I see. Can you still move it back? Best to avoid unnecessary noise in the diff.

src/ninja.cc
```diff
@@ -211,6 +213,7 @@ struct Tool {
 
 /// Print usage information.
 void Usage(const BuildConfig& config) {
+  // clang-format off
```
jhasse, line 217: Please don't add those, clang-format is not enforced in Ninja's code base.

jhasse (commented):
I still think that the tool shouldn't include the `--type` option and always print all inputs, at least for 1.11.0.

digit-google:
Done, let me know what you think :)

digit-google:
I have refactored the two commits so that the first one would print all inputs in a deterministic way (and update the documentation), while the second one adds the `--type=TYPE` option. I'm just waiting for the checks here to complete, and when everything works, I'll move the second commit to a different PR so that the first one can be part of the 1.11 release as you requested.

digit-google:
Done, the current PR only contains the fix to the output. The `--type` option was moved to my Github fork as https://github.com/digit-google/ninja/tree/inputs-type which I will rebase on top of upstream when this one is merged.
Let me know if this requires further changes. This now prints all inputs by default.

jhasse (approved):

jhasse:
Thanks!

PR #2081, opened 2022-02-14
Speedup Pool class implementation

digit-google:
Add proper unit tests for the Pool class, then change the implementation to use an std::priority_queue instead of an std::set.
This data type uses less memory and is generally faster for this specific case.

### DIFF
--- a/src/state.cc
+++ b/src/state.cc
@@ -35,29 +35,35 @@ void Pool::EdgeFinished(const Edge& edge) {
 
 void Pool::DelayEdge(Edge* edge) {
   assert(depth_ != 0);
-  delayed_.insert(edge);
+  delayed_.push(edge);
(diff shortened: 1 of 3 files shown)


digit-google:
Note: I fixed an issue with duplicate edges (and added a relevant unit test for them), as well as measured the speed up with the Fuchsia build (which uses pool), of about 8% (!).

digit-google:
Well, I can't  understand the Windows hang, do not have a Windows machine to reproduce it, and it seems I cannot reproduce my performance differences today vs master (I suspect I messed up the compilation flags between the two branches, or a fluke due to background processes that randomly start on our corporate workstations and eat a decent amount of CPU), so I prefer to close this PR for now. Maybe for another time :)

PR #2088, opened 2022-02-18
Fix ReadFile() handle leak on read error on Windows.

digit-google:
Small change to fix a file handle leak in case of error.
Also return -EIO instead of -1 to signal read errors, since it
is more consistent with what is happening here.

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -350,7 +350,8 @@ int ReadFile(const string& path, string* contents, string* err) {
     if (!::ReadFile(f, buf, sizeof(buf), &len, NULL)) {
       err->assign(GetLastErrorString());
       contents->clear();
-      return -1;
+      ::CloseHandle(f);
+      return -EIO;
     }
     if (len == 0)
       break;


src/util.cc
```diff
@@ -350,7 +350,8 @@ int ReadFile(const string& path, string* contents, string* err) {
     if (!::ReadFile(f, buf, sizeof(buf), &len, NULL)) {
       err->assign(GetLastErrorString());
       contents->clear();
-      return -1;
+      ::CloseHandle(f);
```
jonesmz, line 353: Instead of trying to guess all of the paths where we might return before closing the file, perhaps the right answer is to use std::unique_ptr.
```cpp
std::unique_ptr<std::remove_pointer_t<HANDLE>,
                decltype(&::CloseHandle)> pFile(::CreateFileA(path.c_str(), GENERIC_READ,
                                                              FILE_SHARE_READ, nullptr,
                                                              OPEN_EXISTING, FILE_FLAG_SEQUENTIAL_SCAN, NULL), 
                                                &::CloseHandle);
```

This guarantees that the file is closed when the unique_ptr goes out of scope.

Though, might need to do a lambda for the deleter that checks if the handle == INVALID_HANDLE_VALUE and skips calling ::CloseFIle() on it in that case, i'm not 100% sure if the window API will barf if you pass it INVALID_HANDLE_VALUE.

digit-google: True, but this PR is about fixing a single issue tiny issue, and I had the feeling that @jhasse prefers small changes to avoid breaking existing PRs that are still awaiting review, compared to larger refactorings like the one you suggest.

Also, this definition makes me cry, because std::unique_ptr<> was designed for pointers, it is default-initialized to NULL which is _not_ equal to INVALID_HANDLE_VALUE (which has the numerical value of (DWORD)-1), also you shouldn't need to pass the deleter as an argument to the constructor, this belongs to the type.

A simple ScopedHandle class would be a much better fit, and pretty easy to implement directly. e.g. https://chromium.googlesource.com/chromium/src/base/+/29dc0ca6693ee3d0acb7edf619fd3dd7dc07753c/scoped_handle.h

But I prefer to leave this for another PR, unless @jhasse thinks it's a good idea to introduce now.

jhasse: There's one return path. Anything except this PR is overkill.

digit-google: > There is not one return path, as the ninja CI does not build with exceptions turned off

That's an interesting point, because `configure.py` does turn them off, we might consider disabling them in the `CMakeLists.txt` file though since Ninja does not use them at all. For a different PR?

> Please stop thinking about C++ programming like its still the 80s.
Michael, do you realize that this can be read as offensive? I see that as an undue personal attack on someone you don't really know, and who maintains a project like Ninja on his free time. And it's really hard to take you seriously if you don't engage in constructive criticism. I encourage you to avoid making that kind of statements in the future, in my experience with open-source, it will probably help you get the features and PRs you care about merged more rapidly too. Even when someone is wrong, there are better ways to explain why and how to fix it.

PR #2089, opened 2022-02-18
Build ninja with C++11

randomascii:
In order to allow future use of std::chrono to make the stats code
portable it is desirable to compile with C++11. Doing so also allows use
of std::unordered_map, and reduces the number of #ifdefs in the ninja
source code.

Switching to C++11 requires modifying both CMakeLists.txt and
configure.py, for MSVC and for other build systems. For MSVC the
required change is adding /Zc:__cplusplus to tell the compiler to
give a more accurate value for the __cplusplus macro. For other
platforms the change is to add -std=c++11 or the CMake equivalent.

This change makes some progress towards resolving issue #2004.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,8 @@ endif()
 if(MSVC)
 	set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
 	string(REPLACE "/GR" "" CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+	# Note that these settings are separately specified in configure.py, and
+	# these lists should be kept in sync.
 	add_compile_options(/W4 /wd4100 /wd4267 /wd4706 /wd4702 /wd4244 /GR- /Zc:__cplusplus)
 	add_compile_definitions(_CRT_SECURE_NO_WARNINGS)
 else()
@@ -135,6 +137,8 @@ else()
 		target_link_libraries(libninja PUBLIC "-lperfstat")
 	endif()
 endif()
+
+target_compile_features(libninja PUBLIC cxx_std_11)
 
 #Fixes GetActiveProcessorCount on MinGW
 if(MINGW)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -138,7 +138,9 @@ else()
 	endif()
 endif()
 
-target_compile_features(libninja PUBLIC cxx_std_11)
+if(CMAKE_CXX_STANDARD_DEFAULT EQUAL 98)
+  target_compile_features(libninja PUBLIC cxx_std_11)
+endif()
 
 #Fixes GetActiveProcessorCount on MinGW
 if(MINGW)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -138,9 +138,7 @@ else()
 	endif()
 endif()
 
-if(CMAKE_CXX_STANDARD_DEFAULT EQUAL 98)
-  target_compile_features(libninja PUBLIC cxx_std_11)
-endif()
+target_compile_features(libninja PUBLIC cxx_std_11)
 
 #Fixes GetActiveProcessorCount on MinGW
 if(MINGW)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,8 +15,8 @@ Generally it's the
 a few additions:
 
 * Any code merged into the Ninja codebase which will be part of the main
-  executable must compile as C++03. You may use C++11 features in a test or an
-  unimportant tool if you guard your code with `#if __cplusplus >= 201103L`.
+  executable must compile as C++11. You may use C++14 features in a test or an
+  unimportant tool if you guard your code with `#if __cplusplus >= 201402L`.
 * We have used `using namespace std;` a lot in the past. For new contributions,
   please try to avoid relying on it and instead whenever possible use `std::`.
   However, please do not change existing code simply to add `std::` unless your
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,9 +14,6 @@ Generally it's the
 [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html) with
 a few additions:
 
-* Any code merged into the Ninja codebase which will be part of the main
-  executable must compile as C++11. You may use C++14 features in a test or an
-  unimportant tool if you guard your code with `#if __cplusplus >= 201402L`.
 * We have used `using namespace std;` a lot in the past. For new contributions,
   please try to avoid relying on it and instead whenever possible use `std::`.
   However, please do not change existing code simply to add `std::` unless your
--- a/configure.py
+++ b/configure.py
@@ -305,6 +305,8 @@ if platform.is_msvc():
 else:
     n.variable('ar', configure_env.get('AR', 'ar'))
 
+# Note that build settings are separately specified in CMakeLists.txt and
+# these lists should be kept in sync.
 if platform.is_msvc():
     cflags = ['/showIncludes',
               '/nologo',  # Don't print startup banner.
@@ -320,6 +322,7 @@ if platform.is_msvc():
               # Disable warnings about ignored typedef in DbgHelp.h
               '/wd4091',
               '/GR-',  # Disable RTTI.
+              '/Zc:__cplusplus',
               # Disable size_t -> int truncation warning.
               # We never have strings or arrays larger than 2**31.
               '/wd4267',
@@ -339,6 +342,7 @@ else:
               '-Wno-unused-parameter',
               '-fno-rtti',
               '-fno-exceptions',
+              '-std=c++11',
               '-fvisibility=hidden', '-pipe',
               '-DNINJA_PYTHON="%s"' % options.with_python]
     if options.debug:
--- a/src/hash_map.h
+++ b/src/hash_map.h
@@ -53,7 +53,6 @@ unsigned int MurmurHash2(const void* key, size_t len) {
   return h;
 }
 
-#if (__cplusplus >= 201103L) || (_MSC_VER >= 1900)
 #include <unordered_map>
 
 namespace std {
@@ -68,56 +67,13 @@ struct hash<StringPiece> {
 };
 }
 
-#elif defined(_MSC_VER)
-#include <hash_map>
-
-using stdext::hash_map;
-using stdext::hash_compare;
-
-struct StringPieceCmp : public hash_compare<StringPiece> {
-  size_t operator()(const StringPiece& key) const {
-    return MurmurHash2(key.str_, key.len_);
-  }
-  bool operator()(const StringPiece& a, const StringPiece& b) const {
-    int cmp = memcmp(a.str_, b.str_, min(a.len_, b.len_));
-    if (cmp < 0) {
-      return true;
-    } else if (cmp > 0) {
-      return false;
-    } else {
-      return a.len_ < b.len_;
-    }
-  }
-};
-
-#else
-#include <ext/hash_map>
-
-using __gnu_cxx::hash_map;
-
-namespace __gnu_cxx {
-template<>
-struct hash<StringPiece> {
-  size_t operator()(StringPiece key) const {
-    return MurmurHash2(key.str_, key.len_);
-  }
-};
-}
-#endif
-
 /// A template for hash_maps keyed by a StringPiece whose string is
 /// owned externally (typically by the values).  Use like:
 /// ExternalStringHash<Foo*>::Type foos; to make foos into a hash
 /// mapping StringPiece => Foo*.
 template<typename V>
 struct ExternalStringHashMap {
-#if (__cplusplus >= 201103L) || (_MSC_VER >= 1900)
   typedef std::unordered_map<StringPiece, V> Type;
-#elif defined(_MSC_VER)
-  typedef hash_map<StringPiece, V, StringPieceCmp> Type;
-#else
-  typedef hash_map<StringPiece, V> Type;
-#endif
 };
 
 #endif // NINJA_MAP_H_
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -403,11 +403,7 @@ string EdgeEnv::LookupVariable(const string& var) {
   if (var == "in" || var == "in_newline") {
     int explicit_deps_count = edge_->inputs_.size() - edge_->implicit_deps_ -
       edge_->order_only_deps_;
-#if __cplusplus >= 201103L
     return MakePathList(edge_->inputs_.data(), explicit_deps_count,
-#else
-    return MakePathList(&edge_->inputs_[0], explicit_deps_count,
-#endif
                         var == "in" ? ' ' : '\n');
   } else if (var == "out") {
     int explicit_outs_count = edge_->outputs_.size() - edge_->implicit_outs_;
--- a/src/missing_deps.h
+++ b/src/missing_deps.h
@@ -19,9 +19,7 @@
 #include <set>
 #include <string>
 
-#if __cplusplus >= 201103L
 #include <unordered_map>
-#endif
 
 struct DepsLog;
 struct DiskInterface;
@@ -68,13 +66,8 @@ struct MissingDependencyScanner {
   int missing_dep_path_count_;
 
  private:
-#if __cplusplus >= 201103L
   using InnerAdjacencyMap = std::unordered_map<Edge*, bool>;
   using AdjacencyMap = std::unordered_map<Edge*, InnerAdjacencyMap>;
-#else
-  typedef std::map<Edge*, bool> InnerAdjacencyMap;
-  typedef std::map<Edge*, InnerAdjacencyMap> AdjacencyMap;
-#endif
   AdjacencyMap adjacency_map_;
 };
 
--- a/src/parser.cc
+++ b/src/parser.cc
@@ -31,13 +31,6 @@ bool Parser::Load(const string& filename, string* err, Lexer* parent) {
     return false;
   }
 
-  // The lexer needs a nul byte at the end of its input, to know when it's done.
-  // It takes a StringPiece, and StringPiece's string constructor uses
-  // string::data().  data()'s return value isn't guaranteed to be
-  // null-terminated (although in practice - libc++, libstdc++, msvc's stl --
-  // it is, and C++11 demands that too), so add an explicit nul byte.
-  contents.resize(contents.size() + 1);
-
   return Parse(filename, contents, err);
 }
 


jonesmz:
There's some history with this that reviewers of the PR may be interested in knowing: https://github.com/ninja-build/ninja/issues/1403

configure.py
```diff
@@ -305,6 +305,8 @@ def binary(name):
 else:
     n.variable('ar', configure_env.get('AR', 'ar'))
 
+# Note that build settings are separately specified in CMakeLists.txt and
```
jonesmz, line 308: i think better would be to simply remove this file entirely.

CMakeLists.txt
```diff
@@ -136,6 +138,8 @@ else()
 	endif()
 endif()
 
+target_compile_features(libninja PUBLIC cxx_std_11)
```
jonesmz, line 141: Imho, the better way to handle this is:

```
# Set the explicit standard revision for C and C++ code.
SET(CMAKE_C_STANDARD 11)
SET(CMAKE_C_EXTENSIONS OFF)
SET(CMAKE_C_STANDARD_REQUIRED ON)

SET(CMAKE_CXX_STANDARD 11)
SET(CMAKE_CXX_EXTENSIONS OFF)
SET(CMAKE_CXX_STANDARD_REQUIRED ON)
```

configure.py
randomascii: I would be happy to do that but it is currently used by the OSX builders, and Linux as well I believe. I don't know how such things are configured.

CMakeLists.txt
randomascii: I had something very similar to that in an earlier pull request but in a now-resolved comment (Sarcasm reviewed on Sep 19, 2021) on https://github.com/ninja-build/ninja/pull/2020 it was suggested that this was wrong. I have zero expertise with CMake so I don't know what the best practices are.

jonesmz: Ah. Alright then, nevermind my suggestion. May well be I've been doing it wrong.

configure.py
jonesmz: I don't see anything about python in the osx CI config here: 
https://github.com/ninja-build/ninja/blob/master/.github/workflows/macos.yml
I do see "build-with-python" in the linux CI: https://github.com/ninja-build/ninja/blob/master/.github/workflows/linux.yml
Just remove everything after line 129 in that file.

randomascii: I misread and thought you were suggesting that we remove CMakeLists.txt. Removing configure.py seems like a big step. It is referenced in README.md and in https://github.com/ninja-build/ninja/wiki and it is what I have always used to build ninja. I think that removing configure.py is out-of-scope for this PR.

jhasse: > I think that removing configure.py is out-of-scope for this PR.

Indeed.

CONTRIBUTING.md
```diff
@@ -15,8 +15,8 @@ Generally it's the
 a few additions:
 
 * Any code merged into the Ninja codebase which will be part of the main
-  executable must compile as C++03. You may use C++11 features in a test or an
-  unimportant tool if you guard your code with `#if __cplusplus >= 201103L`.
+  executable must compile as C++11. You may use C++14 features in a test or an
```
jhasse, line 18: As the Google C++ Style Guide contains "Consider portability to other environments before using features from C++14 and C++17 in your project." I think that we can remove this bullet point to keep things simple.

jhasse (commented):
Thanks. If I'm not mistaken you can now also remove this:

https://github.com/ninja-build/ninja/blob/1d576b6b93fdb27ef5cf6f8356a44bcc4d1a4775/src/parser.cc#L34-L39

CONTRIBUTING.md
randomascii: Sounds good. Done.

digit-google:
Very nice PR! @jhasse, can you clarify the criteria needed to officially switch the code base to C++11?
I believe the CI checks already enforces that this compiles for older platforms, i.e.:

- Centos 7, Ubuntu 14.04, Ubuntu 16.04 and Ubuntu 18.04 in .github/workflows/linux.yml
- MacOS 11.0 with a deployment target or 10.12 in .github.workflows/macos.yaml
- Visual Studio 2017 in appveyor.yml

Are we missing something?

nico:
I think C++11 is fine by now. Consider sending a short email to the mailing list asking if this causes problems for anyone.

For what it's worth, LLVM is slightly, but not much, more aggressive in adopting new C++ versions. They switched to C++14 a while ago and they're talking about switching to C++17 now. This is causing some amount of trouble there, since it means to run on old (but still used) systems you either have to statically link libstdc++ or distribute it with the LLVM binaries, neither which is very fun. So I think being very conservative with C++ versions is still a good idea.

But C++11 is probably old enough by now that it shouldn't cause most people headaches.

src/hash_map.h
```diff
@@ -68,56 +67,13 @@ struct hash<StringPiece> {
...
 /// A template for hash_maps keyed by a StringPiece whose string is
 /// owned externally (typically by the values).  Use like:
 /// ExternalStringHash<Foo*>::Type foos; to make foos into a hash
 /// mapping StringPiece => Foo*.
 template<typename V>
 struct ExternalStringHashMap {
-#if (__cplusplus >= 201103L) || (_MSC_VER >= 1900)
   typedef std::unordered_map<StringPiece, V> Type;
-#elif defined(_MSC_VER)
-  typedef hash_map<StringPiece, V, StringPieceCmp> Type;
```
nico, line 117: Have you measured that `unordered_map` has at least as-good perf as hash_map?

CMakeLists.txt
aminya, line 141: Isn't it better to use the most recent C++ standard when it is not specified manually? 

If someone needs an older standard like c++11, they can pass `-D CMAKE_CXX_STANDARD=11`.

https://github.com/cpp-best-practices/project_options/blob/2111be3ee3c1e74ad7d38e25a51f33bff06f3dc1/src/Standards.cmake#L1

aminya (changes requested):

randomascii:
It is a big decision to move ninja to C++11 and I suspect that the ninja developers don't want the try bots compiling with newer versions since that might let C++14 features creep in. That is my impression and I defer to the project maintainers for guidance on this.

CMakeLists.txt
jhasse: Ideally compiling Ninja should work out of the box even on old systems like RHEL 7.

You're right that this might disable some optimizations on newer systems. Too bad CMake doesn't have a way to use the latest standard available?

jhasse:
> [...] I suspect that the ninja developers don't want the try bots compiling with newer versions since that might let C++14 features creep in.

You mean CI? CentOS 7's GCC doesn't support C++14 so that can't happen.

CMakeLists.txt
aminya: > Too bad CMake doesn't have a way to use the latest standard available?

It does if you check the link I gave. The function is part this library:
https://github.com/cpp-best-practices/project_options/

aminya:
I think It is not the responsibility of Ninja to hardcode a C++ standard from 10 years ago and force everyone to lag behind. If someone needs such an old version, they can simply pass `-D CMAKE_CXX_STANDARD=11`.

Instead, Ninja can add a CI job that passes 11 to test if the library can be built with this standard, simply as a backward compatibility test.

jhasse:
@aminya It's added to tell GCC to use C++11 instead of C++03 on CentOS 7 (and other older systems). If you know a way to use the latest standard available I'm all ears.

aminya:
I gave the solution in my comments. This CMake code detects the latest supported standard on the compiler. Here it is again:
https://github.com/cpp-best-practices/project_options/blob/main/src/Standards.cmake

jhasse:
Sorry, GitHub notifications didn't take me to the comment. Let's continue the discussion there.

CMakeLists.txt
jhasse: I was thinking more of a built-in way. In that function the standards are still hard-coded, so it only tells the truth as of right now. What happens when C++23 is released? The CMake code needs to be updated.

Sorry, but I don't like extending the CMakeLists.txt too much. `target_compile_features(libninja PUBLIC cxx_std_11)` is one line which is easy enough to understand. It would be better to tell CMake that we need *at least* C++11, but as there's no easy built-in way to do that we're stuck with that.

aminya: > I was thinking more of a built-in way


The defaults for CMake and compilers are not the best at all. That's the whole reason we created `project_options`. 


> In that function the standards are still hard-coded, so it only tells the truth as of right now. What happens when C++23 is released? 

This is not hard coding. It is providing a range of versions that the authors of the library support. When C++23 is released, first you should try building with it, and once you are sure it works, you can add that to the version range.

> is one line which is easy enough to understand

It is your choice to stick to these sub-optimal (sometimes bad defaults). I don't think a `if` statement is considered complex in the C++ world. 

Worst case scenario, you should at least check if the standard is not set manually by the user:
```suggestion
if("${CMAKE_CXX_STANDARD) STREQUAL "")
	set(CMAKE_CXX_STANDARD 11)
endif()
```

jhasse: I just found [CMAKE_CXX_STANDARD_DEFAULT](https://gitlab.kitware.com/cmake/cmake/-/issues/21115). Maybe we can do the following:

```cmake
if(CMAKE_CXX_STANDARD_DEFAULT EQUAL 98)
  target_compile_features(libninja PUBLIC cxx_std_11)
endif()
```

Haven't tested that, but the idea would be not to mess with the compilers default unless we have to.

neheb:
> > [...] I suspect that the ninja developers don't want the try bots compiling with newer versions since that might let C++14 features creep in.
> 
> You mean CI? CentOS 7's GCC doesn't support C++14 so that can't happen.

It supports a subset of C++14 that it calls 1y. But practically speaking, it doesn't really.

digit-google:
I think that's just another reason to set something like `-std=c++03` or `-std=c++11` when building Ninja on CI: It is the simplest way to guarantee that the code builds and works when compiled against the corresponding language standards, and that features that do not belong to them do not creep up, which is an important goal for the project.

And that does not prevent also checking that it also builds and works on CentOS 7 without that flag, since this is a different issue, which I would label as a convenience rather than a hard requirement (and this applies to other older distributions as well).

While we are on this topic, it looks like Visual Studio 2017 is fully C++11 compliant, starting from 15.7, released in May 2018, as well as Visual Studio 2019 starting from 16.6, released in May 2020. Before that, each compiler has near-complete support for the language, but it is hard to know the exact history.

https://docs.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-170
https://docs.microsoft.com/en-us/visualstudio/releasenotes/vs2017-relnotes-v15.7
https://docs.microsoft.com/en-us/visualstudio/releases/2019/release-notes-v16.6

jhasse:
> It supports a subset of C++14 that it calls 1y.

But that has to be enabled using `--std=c++1y`, right?

neheb:
> > It supports a subset of C++14 that it calls 1y.
> 
> But that has to be enabled using `--std=c++1y`, right?

Yes. Cmake matches on the compiler version and sets 1y in the case of CMAKE_CXX_STANDARD=14

jhasse:
As we don't do that my point still stands: Due to CentOS 7 it won't be possible for C++14 features to creep in.

neheb:
Correct.

jonesmz:
> As we don't do that my point still stands: Due to CentOS 7 it won't be possible for C++14 features to creep in.

FYI: RedHat / CentOS does provide newer compilers for RedHat 7 / CentOS 7.
See: the changes to .github/workflows/linux.yml in this commit: https://github.com/splinter-build/splinter/commit/98f9ca1bb99105c96dbfab94660fdce951f93481 (Ignore changes from other files, they aren't relevant to what I'm saying about compiler support).

This is standard operating procedure for both RedHat and Ubuntu, to provide newer toolchains for older releases, that maintain the appropriate backwards compatibility to support running the resulting program on systems that don't otherwise install newer libs.

As a result, I don't think "CentOS 7 / RedHat 7, out of the box, doesn't have newer compilers" is a meaningful reason not to use a newer version of C++. The barrier to entry for building Ninja w/ C++17 on CentOS 7 is, literally, 3 commands.

```
yum install -y centos-release-scl
yum install -y devtoolset-7-toolchain --enablerepo='centos-sclo-rh'
source /opt/rh/devtoolset-7/enable
// and now your current shell will use the newer compiler.
```

On Ubuntu 14.04, it's slightly less convenient, but relatively straight forward:
```
// assume a relatively up to date package manifest, or otherwise run `apt update`
apt-get install -y software-properties-common
add-apt-repository ppa:ubuntu-toolchain-r/test
apt update
apt install -y g++-7
update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-7 7
update-alternatives --install /usr/bin/c++ c++ /usr/bin/g++-7 7
```

Similarly, you can get C++20 support by installing a newer GCC version that 7, even as far back as Ubuntu 14.04.

randomascii:
Is there a consensus? I'm happy to tweak the CMakeLists.txt change to whatever is deemed appropriate.

aminya:
@randomascii Did you check the discussion here? My solution considers all the parties. No hard coding is needed! If you release a prebuilt binary, you can pass a specific version like 17.

https://github.com/ninja-build/ninja/pull/2089#discussion_r817165194

randomascii:
I saw that, I just wasn't sure if it was the consensus. If so I'm happy to switch to that.

jonesmz:
I don't see how adding a third party collection of cmake stuff is going to be a net improvement to the codebase.

The cmake function [target_compiler_features()](https://cmake.org/cmake/help/latest/command/target_compile_features.html) already has logic to bump the minimum c++ version above what's requested if something else the target is linking against requires it.

If you want something more nuanced than that. Just do a simple if check to see if some command line flag is set or nor.

aminya:
The code is "Unlicensed". You can copy-paste the portion you want and use it.

jhasse:
I prefer this solution:

```cmake
if(CMAKE_CXX_STANDARD_DEFAULT EQUAL 98)
  target_compile_features(libninja PUBLIC cxx_std_11)
endif()
```

randomascii:
That makes sense to me. It's clean and simple. Done.

jhasse:
if I understand https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html correctly, no.

Sarcasm:
I'm surprised with all the noise about around the `target_compile_features`.

The initial version looked semantically better and also simpler:

```cmake
target_compile_features(libninja PUBLIC cxx_std_11)
```

Compared to the new version, which is less standard and more specific / cryptic:

```cmake
if(CMAKE_CXX_STANDARD_DEFAULT EQUAL 98)
  target_compile_features(libninja PUBLIC cxx_std_11)
endif()
```

By reading the CMakeLists.txt, I have no idea why this is gated inside an if statement, no comment.
And reading this PR I'm still not sure?
My guess would be that aminya insisted and this solution seems less heavy-weight than using cpp-best-practices/project_options/src/Standards.cmake but that would not be a good reason.

The second form, with the `if`, has the cons that if "ninja as a library" is added some day, as in https://github.com/ninja-build/ninja/pull/1729, then hiding to CMake the usage requirements that libninja requires `c++ >= 11` may break the transitive dependencies. 

Another cons, is that without the `if`, CMake already takes the compiler default, e.g.:

```
$ grep -C 1 cxx_std_11 CMakeLists.txt
# if(CMAKE_CXX_STANDARD_DEFAULT EQUAL 98)
  target_compile_features(libninja PUBLIC cxx_std_11)
# endif()
$ g++ -x c++ /dev/null -E -dM | grep cplus
#define __cplusplus 201703L
$ mkdir build && cd build
$ cmake -G Ninja ..
$ ninja -v
[1/64] /usr/bin/c++   -Wno-deprecated -fdiagnostics-color -MD -MT CMakeFiles/libninja.dir/src/debug_flags.cc.o -MF CMakeFiles/libninja.dir/src/debug_flags.cc.o.d -o CMakeFiles/libninja.dir/src/debug_flags.cc.o -c /home/papin/dev/github.com/ninja-build/ninja/src/debug_flags.cc
^C
```

So by saying `target_compile_features(libninja PUBLIC cxx_std_11)` CMake already understands it as "take the compiler default" or bump it if it is below C++11 (and error out if `>= c++11` isn't supported).

If the argument is that newer standards are more performant, I'm skeptic, might as well be slightly less performant, depends on the situation.
STL changes are more adding new APIs and preserving backward compatibilities, so it's difficult to improve performances since the data structures used by ninja C++11 will usually be backward compatible. OTOH, compilers do get better at optimizations but these optimizations will also apply to C++11 in many cases, so it's more like "latest toolchain" than "latest standard" that matters.
IMHO, if an user want some specific version, it can and should use the `CMAKE_CXX_{STANDARD,FLAGS,...}*` variables.

aminya:
> The initial version looked semantically better and also simpler:
> target_compile_features(libninja PUBLIC cxx_std_11)

It is not semantically better. Using `PUBLIC`, there is no difference between leaking `-std=c++11` using target_compile_features or setting it using `-DCMAKE_CXX_STANDARD=11`. 

https://cmake.org/cmake/help/latest/command/target_compile_features.html#target-compile-features

The discussion was that when people want to build a library, they should not need to go through each dependency and patch their hardcoded standards. The compiler writers recommend using the latest published standard for the whole build, and anything else is just backward compatibility.

Sarcasm:
"leaking" is not the right term, it is a feature that public usage requirements can be propagated.
Setting `target_compile_features(libninja PUBLIC cxx_std_11)` means that `libninja` public API requires `c++ >= 11` (it does NOT force c++11, it just sets it as a minimum) and CMake will error out if the compiler does not support it.
And if a downstream, specifies `target_link_libraries(foo PUBLIC libninja)`, CMake will ensure that `c++ >= 11` is used.
However, if the downstream specifies a newer standard, e.g.  `target_compile_features(foo PUBLIC cxx_std_17)` CMake will make sure `foo` is built with `c++ >= 17` for this target and for its downstreams.
And if as a packager you want both `libninja` and `foo` to use all the same exact standard, e.g. C++17, thats where you set `-DCMAKE_CXX_STANDARD=17`.
But `libninja` requirements, that it needs `c++ >= 11` is semantically correct and should not be gated inside an `if` statement.

So the statement that people need to go through each dependencies to fix them is plain wrong, you don't have to do this.

---

You link to `target_compile_features` documentation but this page is not helpful to understand how setting the C++ standard works.
Here is a better page:

[cmake.org / cmake-compile-features(7) / Requiring Language Standards](https://cmake.org/cmake/help/latest/manual/cmake-compile-features.7.html#id5)

> In projects that use a large number of commonly available features from a particular language standard (e.g. C++ 11) one may specify a meta-feature (e.g. `cxx_std_11`) that requires use of a compiler mode that is at minimum aware of that standard, but could be greater. This is simpler than specifying all the features individually, but does not guarantee the existence of any particular feature. Diagnosis of use of unsupported features will be delayed until compile time.
> 
> For example, if C++ 11 features are used extensively in a project's header files, then clients must use a compiler mode that is no less than C++ 11. This can be requested with the code:
> 
> ```cmake
> target_compile_features(mylib PUBLIC cxx_std_11)
> ```
> 
> In this example, CMake will ensure the compiler is invoked in a mode of at-least C++ 11 (or C++ 14, C++ 17, ...), adding flags such as `-std=gnu++11` if necessary. This applies to sources within `mylib` as well as any dependents (that may include headers from `mylib`).
> 
> Note: If the compiler's default standard level is at least that of the requested feature, CMake may omit the -std= flag. The flag may still be added if the compiler's default extensions mode does not match the [<LANG>_EXTENSIONS](https://cmake.org/cmake/help/latest/prop_tgt/LANG_EXTENSIONS.html#prop_tgt:%3CLANG%3E_EXTENSIONS) target property, or if the [<LANG>_STANDARD](https://cmake.org/cmake/help/latest/prop_tgt/LANG_STANDARD.html#prop_tgt:%3CLANG%3E_STANDARD) target property is set.
>
> -- https://cmake.org/cmake/help/v3.23/manual/cmake-compile-features.7.html#requiring-language-standards

To be honest, if something is wrong with CMake way of working, it should be reported to them, it's not each CMake project that should work around the build tool they use.
CMake does get better with time, if you reported what you think are issues to them, they may explain why they are correct or could help fix the issue.

---

Finally, here is an example in practice of setting the C++ standard feature and how CMake can upgrade it or error out (disclaimer: I'm the author):

- https://sarcasm.github.io/notes/dev/cmake.html#compiler-options

randomascii:
Is this PR ready to land? Or does it need more changes? Or is it not wanted? I have no opinion on the different ways of implementing this, but I would like to allow C++11 in ninja because that would allow the timing code to be simplified and made portable

CMakeLists.txt
```diff
@@ -136,6 +138,10 @@ else()
 	endif()
 endif()
 
+if(CMAKE_CXX_STANDARD_DEFAULT EQUAL 98)
+  target_compile_features(libninja PUBLIC cxx_std_11)
+endif()
```
hadrielk, line 143: something seems wrong about that - isn't `CMAKE_CXX_STANDARD_DEFAULT` just what the compiler will use if not given a specific option-setting?

What if the user explicitly set the standard version to something higher when invoking this CMake?

So I asked on the CMake Slack channel, and @mathstuf said to just use:
```
target_compile_features(libninja PUBLIC cxx_std_11)
```
and CMake will figure it out.

jhasse:
Yes. Sorry about the back and forth but it looks like `target_compile_features(libninja PUBLIC cxx_std_11)` was correct all along.

aminya:
Because I see some confusion in the comments, I'd like to give a summary of the methods suggested and compare them so you can choose:

1) [Setting the standard to the highest supported by the compiler](https://github.com/cpp-best-practices/project_options/blob/main/src/Standards.cmake?rgh-link-date=2022-03-03T09%3A15%3A16Z) (my suggestion):

This sets the used standard to the highest standard supported by the compiler, which allows for guarding against the build failures caused by the future C++ standards. So the build will be successful for an extended period because you have put a cap on the highest supported number (until you explicitly allow newer standards). 

Since it also has a minimum set in the else clause, If someone uses a compiler that doesn't support C++11, the build will fail (as you intended).

2) using `target_compile_features(libninja PUBLIC cxx_std_11)`. 

This only sets the oldest supported standard by the project. Unlike the above check that sets both a high and a lower bound, if the new C++ standard breaks the build, the project will fail to build if a new C++ standard breaks the built. The user will be forced to debug this **manually** by going back and forth between the standards.

3) Check if `CMAKE_CXX_STANDARD_DEFAULT` is 99.

This is incomplete, in my opinion. It doesn't set a lower bound or a higher bound. It only checks if the standard is 99 and will force 11 in that case. It prevents using newer standards if a compiler exists that defaults to 99 but supports newer standards like C++20.

mathstuf:
> which allows for guarding against the build failures caused by the future C++ standards.

Guarding what from whom?

> So the build will be successful for an extended period because you have put a cap on the highest supported number

This sounds like a benefit to saying "we only require C++11, don't bother us with newer C++ that breaks us", not of using the latest supported by the compiler. There are generally too many moving parts in a C and C++ build as it is; just using the latest standard detected by CMake adds yet another unnecessary flexibility IMO.

> if a new C++ standard breaks the build, the project will fail to build without any guard.

If the compiler is using a newer standard than 11, then something else made it do that. Fixing it is no different than fixing any other fallout from a change.

> so this implementation also misses the opportunity to use the newer standards supported by their compiler.

So what? What C++14 features are being proposed to be used here?

aminya:
> Guarding what from whom?

Build failures caused by future changes in the C++ standard. As you said, C++ has many moving parts. This method sets a high cap preventing from trying those newer features unless you explicitly allow them in the future.

neheb:
This whole thread is so sad.

randomascii:
> Yes, this is still wanted. One change then I will merge: Sorry about the back and forth but it looks like `target_compile_features(libninja PUBLIC cxx_std_11)` was correct all along.

Heh. No worries. You did warn that this might take a while. I returned to the previous CMakeLists.txt implementation. PTAL

randomascii (commented):
Ready to go, I believe

neheb:
*evil laugh*

PR #2093, opened 2022-02-22
configure.py: Support clang-cl with `--platform msvc`

digit-google:
This change allows building Ninja Windows binaries on Linux
with the `clang-cl` toolchain. More specifically:

- `configure.py --platform msvc` will generate a `ninja.build`
  file that invokes `clang-cl` instead of `cl.exe`if the
  right values are defined in CXX and AR environment variables.

- The `--bootstrap` option is not supported. Instead, a host
  version of Ninja is needed to invoke the generated build plan,
  and will compile `ninja.exe` and `ninja_test.exe`.

Note: the generated file works fine under Wine, the `ninja_test.exe`
program fails because mkdtemp() seems to fail in surprising way
under this environment (this looks like a Wine bug, or an issue
with the way Wine is configured on my machine, nothing to do with
Ninja itself).

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -83,12 +83,15 @@ class Platform(object):
     def is_msvc(self):
         return self._platform == 'msvc'
 
-    def msvc_needs_fs(self):
-        popen = subprocess.Popen(['cl', '/nologo', '/help'],
+    def msvc_tool_has_flag(self, tool, flag):
+        popen = subprocess.Popen([tool, '/nologo', '/help'],
                                  stdout=subprocess.PIPE,
                                  stderr=subprocess.PIPE)
         out, err = popen.communicate()
-        return b'/FS' in out
+        if sys.version_info >= (3, 0):
+          return flag in out.decode('utf-8')
+        else:
+          return bytes(flag) in out
 
     def is_windows(self):
         return self.is_mingw() or self.is_msvc()
@@ -269,11 +272,17 @@ if configure_env:
     n.variable('configure_env', config_str + '$ ')
 n.newline()
 
-CXX = configure_env.get('CXX', 'c++')
-objext = '.o'
 if platform.is_msvc():
-    CXX = 'cl'
-    objext = '.obj'
+  objext = '.obj'
+  default_ar = 'link'
+  default_cxx = 'cl'
+else:
+  objext = '.o'
+  default_ar = 'ar'
+  default_cxx = 'c++'
+
+AR = configure_env.get('AR', default_ar)
+CXX = configure_env.get('CXX', default_cxx)
 
 def src(filename):
     return os.path.join('$root', 'src', filename)
@@ -300,10 +309,7 @@ if root == os.getcwd():
 n.variable('root', root)
 n.variable('builddir', 'build')
 n.variable('cxx', CXX)
-if platform.is_msvc():
-    n.variable('ar', 'link')
-else:
-    n.variable('ar', configure_env.get('AR', 'ar'))
+n.variable('ar', AR)
 
 if platform.is_msvc():
     cflags = ['/showIncludes',
@@ -326,7 +332,7 @@ if platform.is_msvc():
               '/DNOMINMAX', '/D_CRT_SECURE_NO_WARNINGS',
               '/D_HAS_EXCEPTIONS=0',
               '/DNINJA_PYTHON="%s"' % options.with_python]
-    if platform.msvc_needs_fs():
+    if platform.msvc_tool_has_flag(CXX, '/FS'):
         cflags.append('/FS')
     ldflags = ['/DEBUG', '/libpath:$builddir']
     if not options.debug:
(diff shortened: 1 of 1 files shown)


digit-google:
Invoking @randomascii to take a look at this

jhasse:
> This change allows building Ninja Windows binaries on Linux with the clang-cl toolchain.

Why not use CMake for that use-case?

digit-google:
CMake already works with `cmake <src_dir> -DWIN32=True` (because for some reason it detects a clang-cl properly, but will not detect that this is targetting Win32 when running on Linux). I prefer to keep both scripts consistent with their support of compilers / toolchains whenever possible.

digit-google:
(And for the record, the CMake-generated ninja_test.exe has exactly the same issue under Wine).

randomascii:
I tested this on Windows to see if I could use it to build ninja with clang-cl and lld-link. I set CXX and AR and then ran:

python configure.py --platform msvc --bootstrap

The first error I hit was that /GL was being passed to the compiler, and that isn't legal. Note that the error message isn't displayed - I had to manually re-run the failed command.

I hacked around that (deleted /GL from configure.py) and then hit a link error. Same deal - I had to manually run the link command to see the error message which was that "subsystem must be defined", so I passed /SUBSYSTEM:CONSOLE,5.02 and then hit "undefined symbol: main" when building ninja.lib. It turns out (since a .lib file was being created) I really needed to put /LIB as the first argument, not /SUBSYSTEM:CONSOLE.

After hacking in /LIB I hit an error running ninja.bootstrap.exe. It says:

    CreateProcess failed: The system cannot find the file specified.

For some reason in the ERROR_FILE_NOT_FOUND case the command is not printed, but I hacked in some printing and it says:

CreateProcess failed. Command attempted:
"CXX='c:\src\chromium3\src\third_party\llvm-build\Release+Asserts\bin\clang-cl.exe' AR='c:\src\chromium3\src\third_party\llvm-build\Release+Asserts\bin\lld-link.exe' python.exe ./configure.py --platform msvc"

Odd. Anyway, maybe running this PR on Windows is out of scope, but being able to build ninja on Windows with clang-cl would be nice.

digit-google:
I should have mentioned that to actually use `clang-cl` on Linux, one has to add plenty of stuff to other environment variables (that are not needed when invoking `configure.py`). For example, here's the `envsetup.sh` script that I use on my Linux workstation:

```
# source this file in your environment to build with clang-cl
# See https://jake-shadle.github.io/xwin/ for reference
_FUCHSIA_PREBUILT_THIRD_PARTY=/work/fuchsia0/prebuilt/third_party
_CLANG_BINPREFIX=/android/clang-15/usr/bin/
_XWIN=/android/xwin
export NINJA=${_FUCHSIA_PREBUILT_THIRD_PARTY}/ninja/linux-x64/ninja
export CL="-Wno-unused-command-line-argument -fuse-ld=lld-link /imsvc ${_XWIN}/crt/include /imsvc ${_XWIN}/sdk/include/ucrt /imsvc ${_XWIN}/sdk/include/um /imsvc ${_XWIN}/sdk/include/shared"
export LINK="/libpath:${_XWIN}/crt/lib/x86_64 /libpath:${_XWIN}/sdk/lib/ucrt/x86_64 /libpath:${_XWIN}/sdk/lib/um/x86_64"
export RUSTFLAGS="-Lnative=/xwin/crt/lib/x86_64 -Lnative=/xwin/sdk/lib/um/x86_64 -Lnative=/xwin/sdk/lib/ucrt/x86_64"
export CXX=${_CLANG_BINPREFIX}clang-cl
export CC=${_CLANG_BINPREFIX}clang-cl
export LD=${_CLANG_BINPREFIX}lld-link
export AR=${_CLANG_BINPREFIX}llvm-lib
export RC=${_CLANG_BINPREFIX}llvm-rc
unset _XWIN
unset _CLANG_BINPREFIX
unset _FUCHSIA_PREBUILT_THIRD_PARTY
```
The change here is only the minimum to get `configure.py` working (and also CMake). Maybe we can add some documentation somewhere to warn about this, though I feel it is a little out of scope for this PR.

digit-google:
To clarify, what I meant is that the same envsetup.sh is needed to build with `cmake .. -DWIN32=True`, the change itself doesn't touch the `CMakeLists.txt` file. Did you try building with CMake on Windows and `clang-cl`, it would be interesting to see if this works (but again, out of scope of this CL :-))

randomascii:
I didn't try building with CMake on Windows - I don't even have it installed.

I really don't have any useful thoughts on this PR, since building ninja for Windows on Linux isn't particularly of interest for me.

jonesmz:
> I didn't try building with CMake on Windows - I don't even have it installed.

cmake comes with visual studio, if you have that intalled

CMake also does not require installation. Simply unbacking the archive file to a folder is sufficent.

digit-google:
No problem with that. Thanks.

PR #2094, opened 2022-02-23
Document C++11 build requirements and enforce them in CI.

digit-google:
Clarify the state of C++03 and C++11 support in CONTRIBUTING.md,
and add a new CI test to ensure that Ninja builds and works
correctly with a C++11 compiler.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -123,6 +123,16 @@ jobs:
       run: pytest-3 --color=yes ../..
       working-directory: build-clang/Release
 
+    - name: Configure (GCC_C++11)
+      run: CXXFLAGS=-std=c++11 cmake -Bbuild-gcc-c++11 -DCMAKE_BUILD_TYPE=Debug -G'Ninja Multi-Config'
+    - name: Build (GCC_C++11, Debug)
+      run: cmake --build build-gcc-c++11 --config Debug
+    - name: Unit tests (GCC_C++11, Debug)
+      run: ./build-gcc-c++11/Debug/ninja_test
+    - name: Python tests (GCC_C++11, Debug)
+      run: pytest-3 --color=yes ../..
+      working-directory: build-gcc-c++11/Debug
+
     - name: clang-tidy
       run: /usr/lib/llvm-10/share/clang/run-clang-tidy.py -header-filter=src
       working-directory: build-clang
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,8 +15,11 @@ Generally it's the
 a few additions:
 
 * Any code merged into the Ninja codebase which will be part of the main
-  executable must compile as C++03. You may use C++11 features in a test or an
-  unimportant tool if you guard your code with `#if __cplusplus >= 201103L`.
+  executable *must* compile as C++03 as well as C++11. And any C++11 specific
+  code path there *must* be guarded in the code with `#if __cplusplus >= 201103L.
+  New Ninja features *must* *not* depend on the language standard used to
+  build the source code. This last restriction does not apply to tests and
+  unimportant tools.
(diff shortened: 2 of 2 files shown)


jhasse:
We already built with C++11 in CI: New GCC releases and MSVC default to it.

I prefer the old sentence in CONTRIBUTION.md as it allows C++11 for unimportant tools.

digit-google:
I see. This is not obvious at all, and if these compilers decide to switch to C++14 next year, the upgrade will be silent and we may miss some breakages (e.g. someone using a C++14 specific feature like variable templates, but guarding it with `#if __cplusplus >= 201103L` incorrectly).

I think it would be better to enforce standards compliance with explicit compiler switches instead (e.g. -std=c++03 and -std=c++11 in this case) to avoid these and make everything completely explicit to everyone looking at the CI build rules. Also in case of failure, it will be easier to see in the log if this is related to a standard feature or not. What do you think?

I have clarified that the last restriction does not apply to tests and unimportant tools (but maybe it would be nice to better describe what these are).

jhasse:
If the compilers decide to switch we won't be affected as the compiler versions are fixed in the CI by using a specific Ubuntu/CentOS version.

Btw have you seen #2089? Given that PR and my general resistance against change (sorry), I'm closing this one.

jhasse:
> However, I think it would be useful for new contributors to properly document the state of C++ standard support (and Python too) and clarify that this is enforced in CI.

And you think that "Any code merged into the Ninja codebase which will be part of the main executable must compile as C++03." is not enough?

digit-google:
> And you think that "Any code merged into the Ninja codebase which will be part of the main executable must compile as C++03." is not enough?

It does not reflect what is already in the code base of the main executable, which already contains a few C++11 specific code path alternatives. Some of them were accepted recently, for good performance-related reasons (moves instead of copies, or better containers). This is surprising to any new contributor.

jhasse:
Can you give an example? auto_ptr for example was removed in C++17 so ifdefs were added to support newer compilers.

digit-google:
Well, just grep for `201103L` in the code base:

- Faster containers in `src/missing_deps.h`

- One to avoid a crash in `src/graph.cc` when `edge_->inputs_` is empty (this is in the commit message, but should really be a comment, so no-one tries to remove this later thinking these are the same things, a better fix would probably not call the function at all if `explicit_deps_count` is 0 though)

jhasse:
missing_deps is an "unimportant tool" and the crash in graph.cc only happens in MSVC and debug mode.

And even with __cplusplus ifdefs the point still stands: The code must compile as C++03.

digit-google:
Ah, I see why I was confused: I thought that "unimportant tools" meant the things that are under `misc/` and which are not part of the main executable. And also in the current case, "missing_deps.cc" is definitely covered by "any code merged into the Ninja codebase which will be part of the main executable". And what is an "important" or "unimportant" tool anyway?

It would be useful to everyone if things could be clarified or made explicit (I don't know if you realize that the project's situation is a little confusing for any new contributor). I'd be happy to provide PRs to update the documentation in this direction, so don't be surprised if I submit a few more similar PRs in the future, and let me know if there are more important tasks there than others.

digit-google:
And there is no use of std::move() though, I was confused because I have local branches that add these to avoid a few extra string copies in certain places, but nothing important that needs to be submitted now :-)

jhasse:
> And also in the current case, "missing_deps.cc" is definitely covered by "any code merged into the Ninja codebase which will be part of the main executable". And what is an "important" or "unimportant" tool anyway?

Good points, I see what you mean now :)

Regarding C++11 I think it would be best to just enable it, that's probably the least confusing situation for new contributors.

BUT: Contributing to Ninja is very frustrating as I don't have time to review most of the PRs. So making anything more attractive to new contributors might give a false impression.

PR #2104, opened 2022-03-15
Add `explicit` specifier for the constructor of `SlidingRateInfo`

ken-matsui:
A constructor with only one argument is commonly suggested to use an `explicit` specifier.

### DIFF
--- a/src/status.h
+++ b/src/status.h
@@ -88,7 +88,7 @@ struct StatusPrinter : Status {
   }
 
   struct SlidingRateInfo {
-    SlidingRateInfo(int n) : rate_(-1), N(n), last_update_(-1) {}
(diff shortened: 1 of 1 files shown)


digit-google:
The point of the `explicit` qualifier is to avoid unexpected implicit conversions, but I don't think there is a good reason to prevent implicit conversions to an int in this case.

ken-matsui:
I intended to bring the good practice that all implicit conversions should be avoided in C++; nevertheless, I respect your opinion as it depends on the ideology of maintainers. Thank you for your review :)

digit-google:
Thanks for your contribution anyway. Btw I am just a contributor, not a maintainer :-)

Issue #2082, opened 2022-02-14
msvc tool issues


digit-google:
The `msvc` tool has minor problems in the current code source. This issue is to track them for future fixes:

- The tool is only defined with `#ifdef _MSVC`, i.e. when compiling with the Microsoft compiler, this should be `#ifdef _WIN32` instead.

- It is also tagged `(EXPERIMENTAL)` in the ninja.cc, but [the Chromium build](https://source.chromium.org/chromium/chromium/src/+/main:build/toolchain/win/BUILD.gn;drc=364c9f71522ca2e7ce2fb1f155cd558bf480a8d5;l=141) relies on it.

- The tool is not documented in doc/manual.asciidoc

- The format of the input file for the `-e ENVFILE` option is a binary block passed directly to the `CreateProcessA()` Win32 function directly, which means it cannot support long file paths. Converting the paths to UTF-16 then using CREATE_UNICODE_ENVIRONMENT in dwCreationFlags would fix this.

- Similarly, CreateProcessW() should be called instead of CreateProcessA(), even if the command-line is currently only available as an 8-bit string, this will simplify future refactors to properly support Windows Unicode file paths.

- The tool modifies the global PATH environment variable, and never resets it to its previous value (see src/msvc_helper_main-win32.cc). It seems it uses it to ensure that CreateProcessA() finds the program using the `PATH` definition found in the `ENVFILE`, but a better way to address this would be to do an explicit search With SearchPathW() instead. This is probably minor until Ninja is turned into a library.

- The tool also modifies the mode of stdout to O_BINARY without resetting it. Also minor until Ninja is turned into a library.

digit-google:
Just to clarify a few things here, the modification of global PATH and stdout mode isn't a practical problem now since this code path is only run as part of a tool. If Ninja is refactored as a library in the future, that tool is likely never going to be part of it.

On the other hand, this tool is typically used in commands invoked by Ninja itself, e.g.:

```
rule cc
   command = ninja -t msvc -e ENVFILE -- cl ....
   
build foo.o: cc foo.c foo.h
```

Since launching a new process on Windows is a costly operation, it might be useful to have Ninja auto-detect this use case and implement the tool directly in the context of its own process (saving an intermediate process creation). This has the potential to speed up every CL invocation. In that specific case, having something that does not change the global state would be required, but that looks like an advanced feature, so let's consider this is not really part of issues with the code for now.

jhasse:
[...]

PR #2106, opened 2022-03-17
Enable msvc tool in cross-compiled windows builds of ninja

sztomi:
Cross-compiled windows builds of ninja are unable to use `-t msvc` because it is ifdefed to `_MSC_VER` instead of `_WIN32`, even though a non-msvc built ninja is perfectly capable of containing and running that code. This is mentioned in https://github.com/ninja-build/ninja/issues/2082. This PR replaces those defines, so that cross-compiled windows executables will not lack this functionality.

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -52,10 +52,12 @@
 
 using namespace std;
 
-#ifdef _MSC_VER
+#ifdef _WIN32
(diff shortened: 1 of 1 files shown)


sztomi:
alright, thanks

PR #2107, opened 2022-03-18
Fix debug C++03 builds with recent GCC versions.

digit-google:
Fix a runtime error that occurs when building Ninja in debug mode
with -std=c++03 and a recent GCC release. E.g.:

```sh
CXXFLAGS='-std=c++03' ./configure.py --debug --bootstrap && ./ninja_test
```

Will fail with:

```
bootstrapping ninja...
wrote build.ninja.
bootstrap complete.  rebuilding...
[31/31] LINK ninja
[10/384] BuildLogTest.MultiTargetEdge/usr/include/c++/11/debug/vector:445:
In function:
    std::__debug::vector<_Tp, _Allocator>::const_reference
    std::__debug::vector<_Tp,
    _Allocator>::operator[](std::__debug::vector<_Tp,
    _Allocator>::size_type) const [with _Tp = Node*; _Allocator =
    std::allocator<Node*>; std::__debug::vector<_Tp,
    _Allocator>::const_reference = Node* const&; std::__debug::vector<_Tp,
    _Allocator>::size_type = long unsigned int]

Error: attempt to subscript container with out-of-bounds index 0, but
container only holds 0 elements.

Objects involved in the operation:
    sequence "this" @ 0x0x55968842c0a0 {
      type = std::__debug::vector<Node*, std::allocator<Node*> >;
    }
Aborted
```

The root of the problem is that an expression like `&(some_vector[0])`
triggers runtime checks in the debug version of recent libstdc++ libraries,
which cause the failure above.

This issue was previously already worked-around for C++11 with the use of data(),
but this wasn't fixed for C++03. It is assumed here that the GCC/libstdc++ version
used on CI/CQ for C++03 verification are just older and do not include this runtime
check.

Fixing the issue allows developers to verify that their local modifications do
pass the C++03 build in debug mode (used during debugging) without having to install
an older toolchain or rely on CI/CQ.

### DIFF
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -403,15 +403,28 @@ string EdgeEnv::LookupVariable(const string& var) {
   if (var == "in" || var == "in_newline") {
     int explicit_deps_count = edge_->inputs_.size() - edge_->implicit_deps_ -
       edge_->order_only_deps_;
+    // In debug builds, de-referencing a past-the-end iterator in expressions
+    // like '&some_vector[0]' or '&(*some_vector.begin())' results in a runtime
+    // error with recent debug libstdc++ versions. To work-around this, use
+    // data() in C++11 and above, or do an explicit check for C++03.
 #if __cplusplus >= 201103L
     return MakePathList(edge_->inputs_.data(), explicit_deps_count,
-#else
-    return MakePathList(&edge_->inputs_[0], explicit_deps_count,
-#endif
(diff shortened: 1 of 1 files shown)


jonesmz:
Is this PR needed after https://github.com/ninja-build/ninja/pull/2089 is merged?

jonesmz (approved):

jhasse:
should be fixed now :)

Issue #2008, opened 2021-07-27
ninja thinks the root directory doesn't exist


comex:
Given a build statement that declares a dependency on the root directory:
```
rule my_rule
    command = echo foo
build foo: my_rule /
```
ninja outputs:
```
ninja: error: '', needed by 'foo', missing and no known rule to make it
```

Reproduced on ninja 1.10.2 and latest master, on macOS.

Yes, I know that depending on the root directory isn't a particularly sensible thing to do. :) In my real use case it's an order-only dependency, so slightly less insane, but still unnecessary; it's just the product of a quirk in my build.ninja generator.  Still, there's no reason it shouldn't work, as far as I know.

PR #2108, opened 2022-03-20
Fix canonicalize path

digit-google:
This patch fixes a few annoying bugs in the implementation of
CanonicalizePath():

- It canonicalized "/" and "/foo/.." incorrectly to "",
  while the correct value should be "/". And this was
  enforced by unit-tests :-(

  This was the root cause for issue #2008.

- In certain cases, it would copy the first byte _after_ the
  path buffer into the destination (assuming in the comment
  that this is a terminating zero), though there is no guarantee
  that this is the case, or that this address is accessible.

  This is important because the function is being called in
  build.cc:xxx and :xxx with arguments coming from StringPiece
  values, which do not guarantee anything about the bytes/memory
  that follow the content they point to. It just happens to work
  due to the way the parser are currently implemented, but could
  break easily in the future if they are modified (e.g.
  to read directly from read-only memory-mapped files).

Performance comparisons with `canonical_perfcompare` shows this
implementation to be slightly faster as well, though benchmarking
ninja with a large Fuchsia build plan doesn't show any difference
overall.

Minimal times reported by canon_perftest (best of 5 runs each):

```
  compiler           BEFORE       AFTER

  g++-11 -O2          120ms    110ms
  g++-11 -O3          123ms    110ms
  clang++-13 -O2      115ms    114ms
  clang++-13 -O3      115ms    110ms
```

Note that the loop had to be unrolled for `gcc -O3` due to some
surprising compiler optimization failure :-(

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -169,6 +169,8 @@ void CanonicalizePath(char* path, size_t* len, uint64_t* slash_bits) {
 #endif
   }
 
+  const char* dst_start = dst;
+
   while (src < end) {
     if (*src == '.') {
       if (src + 1 == end || IsPathSeparator(src[1])) {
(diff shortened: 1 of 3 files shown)


src/util.cc
```diff
@@ -192,22 +194,77 @@ void CanonicalizePath(char* path, size_t* len, uint64_t* slash_bits) {
...
+    // works well with `gcc-11 -O2`, but results in disastrous code with
+    // `gcc-11 -O3`, for some unknown reason. Comparisons of canon_perftest
+    // results (minimal times of 5 runs each):
+    //
+    //   compiler              regular      unrolled
+    //
+    //   clang-13 -O2            114ms        114ms
+    //   clang-13 -O3            110ms        110ms
+    //   gcc-11 -O2              126ms        110ms
+    //   gcc-11 -O3              275ms !!     110ms
```
jonesmz, line 223: IMHO, it would be better to report this as a bug to GCC and keep the less complex code.

PR #2110, opened 2022-03-24
Compile getopt as a C++ file, so that CMake need not look for a C compiler at all

jonesmz:
A project that requires only a C++ compiler, and not also a C compiler, can be supported by weird platforms with slightly less work.

Frequently Linux distributions which offer compatibility packages for older OS releases to support newer compilers break those compat packages out into multiple different packages, with the C and C++ compiler being separate.

With this simple change, the Ninja project can compile with only a C++ compiler, allowing for CMake to skip checking the C compiler entirely.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@ include(CheckIPOSupported)
 
 option(NINJA_BUILD_BINARY "Build ninja binary" ON)
 
-project(ninja)
+project(ninja CXX)
 
 # --- optional link-time optimization
 check_ipo_supported(RESULT lto_supported OUTPUT error)
@@ -128,10 +128,18 @@ if(WIN32)
 		src/getopt.c
 		src/minidump-win32.cc
 	)
+	# Build getopt.c, which can be compiled as either C or C++, as C++
+	# so that build environments which lack a C compiler, but have a C++
+	# compiler may build ninja.
+	set_source_files_properties(src/getopt.c PROPERTIES LANGUAGE CXX)
 else()
 	target_sources(libninja PRIVATE src/subprocess-posix.cc)
 	if(CMAKE_SYSTEM_NAME STREQUAL "OS400" OR CMAKE_SYSTEM_NAME STREQUAL "AIX")
 		target_sources(libninja PRIVATE src/getopt.c)
+		# Build getopt.c, which can be compiled as either C or C++, as C++
+		# so that build environments which lack a C compiler, but have a C++
+		# compiler may build ninja.
+		set_source_files_properties(src/getopt.c PROPERTIES LANGUAGE CXX)
 	endif()
 
 	# Needed for perfstat_cpu_total


neheb (approved):

jhasse:
I think the disadvantages of this change outweigh the advantages, sorry.

jonesmz:
What disadvantages?

If you don't list them, i can't address your concerns.

Note that @neheb is an active contributor to openwrt, which takes heavy advantage of the differences between C and C++ to optimize the platform, and they apparently think this change is an improvement.

jonesmz:
I am apparently not able to re-open this pull request.

If I don't hear back from you in a week or so, I'll open a new one with the same contents so that this discussion can continue.

jhasse:
* it's a change.
* C compiler is faster.
* touches a file which was copied from somewhere else.
* "getopt.cc" is lying about it's content - it's C, not C++.
* not using cmake's default for `project`

jonesmz:
> C compiler is faster.

No, it's not.

The C and C++ compiler front-ends have to do basically the same work, with some additional codepaths that may or may not cause an immeasurable slowdown but won't be activated if the parsed file has only C-compatible syntax. This slowdown however immeasurably tiny is *probably* overcome by a large margin by not having the CPU competing for cache with the c compiler and C++ compiler being alive at the same time anyway.

The backends are the same.

This change allows cmake to not detect the C compiler for the platform, which takes several seconds on a lightly-loaded system. On the build machine's for my work, which are *heavily* loaded, detecting the C compiler for the platform takes cmake > 60 seconds.

> touches a file which was copied from somewhere else.

A file which shouldn't be part of the project in the first place, yes. But changing `.c` to `.cc` doesn't modify the contents of the file, and as such is hardly modifying the file in any meaningful way.

> "getopt.cc" is lying about it's content - it's C, not C++.

This wouldn't be a lie. Changing the extension tells the compiler that it's a C++ file, and not a C file. That's the whole point. The code contained in the file is going to operate exactly the same under either language, since it uses the subset of both the C and C++ languages that are identical, but by compiling it as C++ instead of C, we don't need to engage the C compiler at all, which saves me >60 seconds per build.

> not using cmake's default for project

That's the point, to avoid the >60 seconds that it can take on a heavily loaded system to interrogate the unneeded C compiler for it's feature set. 

CMake's `project()` function only enables C by default due to ages old legacy behavior.

Really, Ninja should have always called `project(NAME C CXX)` to protect against the possibility of CMake changing this legacy behavior in the future.

jonesmz:
> Oh common, you knew what I meant.

I'm not trying to be obstinate, it's a legitimate technical distinction which comes up at my work all the time.

jonesmz:
Here's another consideration.

One of the codebases I work with has it's own toolchain file for cmake. It doesn't have a C compiler, only a C++ compiler.

We have to carry a patch (specifically, this PR, minus the python changes) to be able to build ninja for that project, because we don't have a C compiler.

This change strictly:

- increases platform portability
- decrease cmake configuration times
- decreases or maintains the same compile times

jhasse:
Good points. Can you do it without the .cc extension?

jonesmz:
Yes. I'll have to override cmake's default language detection behavior for that file, but it's possible to get the behavior i'm looking for without changing the file name.

jonesmz:
By only changing how CMake is told to interpret getopt.c, and leaving everything else the same, I accomplish my objective.

I am *immensely* annoyed at this, because it turns out that we've been compiling getopt.c as C++ in the python build script since before commit 5d6249fa8c3d6241f46701f43fd48db576714ae4, which was in April of 2012. I stopped trying to find the exact commit after that.

So compiling getopt.c as C instead of C++ was a regression that was inadvertently introduced into the project when CMake was introduced. This would have been averted entirely if getopt.c had been renamed to getopt.cc like it should have been from day 1.

PR #2115, opened 2022-03-31
Disable exceptions and rtti for cmake builds

jonesmz:


jhasse:
I'm not happy with the string replacing. Until there's a better way to do this in CMake, I think we should stick to the defaults for now.

PR #2118, opened 2022-04-20
Use python3 instead of python

keith:
This is a new version of https://github.com/ninja-build/ninja/pull/2063
for discussion. macOS 12.3 has removed /usr/bin/python entirely, so
there is no longer an executable if you request `/usr/bin/env python`.
So even if you override `--with-python=python3`, you cannot call
`./configure.py` directly, you'd have to run `python3 configure.py`
instead.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,7 +90,7 @@ function(check_platform_supports_browse_mode RESULT)
 
 endfunction()
 
-set(NINJA_PYTHON "python" CACHE STRING "Python interpreter to use for the browse tool")
+set(NINJA_PYTHON "python3" CACHE STRING "Python interpreter to use for the browse tool")
 
 check_platform_supports_browse_mode(platform_supports_ninja_browse)
 
--- a/misc/ninja_syntax.py
+++ b/misc/ninja_syntax.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
 
 # Copyright 2011 Google Inc. All Rights Reserved.
 #
--- a/configure.py
+++ b/configure.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 #
 # Copyright 2001 Google Inc. All Rights Reserved.
 #
--- a/misc/measure.py
+++ b/misc/measure.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # Copyright 2011 Google Inc. All Rights Reserved.
 #
--- a/misc/ninja_syntax_test.py
+++ b/misc/ninja_syntax_test.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # Copyright 2011 Google Inc. All Rights Reserved.
 #
--- a/misc/write_fake_manifests.py
+++ b/misc/write_fake_manifests.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 """Writes large manifest files, for manifest parser performance testing.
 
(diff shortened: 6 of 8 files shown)


jhasse:
> macOS 12.3 has removed /usr/bin/python entirely, so there is no longer an executable if you request /usr/bin/env python.

IMHO this is a bug and should be reported to Apple, no workaround in Ninja or any other software.

jhasse:
> macOS 12.3 has removed /usr/bin/python entirely, so there is no longer an executable if you request /usr/bin/env python.

IMHO this is a bug and should be reported to Apple, no workaround in Ninja or any other software.

jonesmz:
Apple is almost certainly not going to do anything about this. Their release management team is very good, so this was almost certainly a deliberate change with the expectation of breaking software that made "poor" assumptions -- according to Apple's opinion of what constitutes a poor assumption.

keith:
I don't think this should be considered a bug. It was definitely intentional on the Apple side, as it was included in the [release notes for macOS 12.3](https://developer.apple.com/documentation/macos-release-notes/macos-12_3-release-notes):

image

My assumption is that they didn't want to maintain a `python` symlink, while also changing the version, since that would be even more confusing to users than removing it (maybe in the future they will add it back pointing at `python3`, but who knows).

Regardless of Apple's decisions here, the related pep https://peps.python.org/pep-0394/ (which isn't accepted, but seems to have general consensus https://github.com/python/peps/pull/989) states that distributions should have `python2` if they support it, `python3` if they support it, but `python` is optional and up to the distro owner:

image

So my assumption is that over time we will see fewer and fewer users who have a `python` by default, for the same reasons I assume Apple removed it, to reduce version confusion.

In general I think if we believe that python3 will be more widely installed than python2, even if folks did still have this symlink, it makes sense to default to the more widely used version, and folks can continue to manually run `python2 ./configure.py` or `python ./configure.py` if they so choose.

jhasse:
I disagree with the PEP, too :D

misc/ninja_syntax.py
```diff
@@ -1,5 +1,3 @@
-#!/usr/bin/python
```
jonesmz, line 1: I recommend against removing the shebang.

keith: I removed it in this case because this file isn't executable and this file on its own doesn't actually run anything if you it was executable. I can add it back if you're thinking the documentation aspect is worth it?

jonesmz: Personally I think consistency is best.

keith: Done!

ComputerDruid:
Note: it's not just Mac OS that's removing /usr/bin/python by default, debian 11 does that too: https://wiki.debian.org/Python

CMakeLists.txt
```diff
@@ -177,7 +177,7 @@ if(platform_supports_ninja_browse)
 		PROPERTIES
 			OBJECT_DEPENDS "${PROJECT_BINARY_DIR}/build/browse_py.h"
 			INCLUDE_DIRECTORIES "${PROJECT_BINARY_DIR}"
-			COMPILE_DEFINITIONS NINJA_PYTHON="python"
+			COMPILE_DEFINITIONS NINJA_PYTHON="python3"
```
stephan-cr, line 180: Would it make sense to find the Python 3 executable via `find_package(Python3 REQUIRED COMPONENTS Interpreter)`?

keith: I'm not really sure, I was just hoping to do the minimum diff here

jonesmz: Yes, it would. Or the "find_program" function. 

But I agree that this PR should do the minimum. A later PR can do a dynamic search.

jonesmz (approved):

keith:
It looks like debian has officially removed the python 2 packages https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1027108

jhasse:
fixed by 6a17e84370064eec6f22cfb1717ab80cf898d82b

keith:
It looks like there were a few in this PR that aren't in that commit, I submitted https://github.com/ninja-build/ninja/pull/2241

PR #2123, opened 2022-04-30
Introduce NINJA_SHELL environment variable

KarstenB:
For our application we want to capture the resources consumed of the commands being executed by ninja. We came up with the idea of substituting the hard-coded /bin/sh with our script for capturing the resources needed by the spawned process. As this is running in a docker, perf is unfortunately not easily available to us. 

But aside from our use-case I could for-see some other uses-cases like transforming the output of the command into an azure devops compatible  format.

The environment variable for setting a different shell would be NINJA_SHELL.

### DIFF
--- a/src/subprocess-posix.cc
+++ b/src/subprocess-posix.cc
@@ -38,6 +38,12 @@ using namespace std;
 
 Subprocess::Subprocess(bool use_console) : fd_(-1), pid_(-1),
                                            use_console_(use_console) {
+  char* shell_path;
+  shell_path = getenv ("NINJA_SHELL");
+  if (shell_path!=NULL)
+    shell=std::string(shell_path);
+  else
+    shell=std::string("/bin/sh");
 }
 
 Subprocess::~Subprocess() {
@@ -117,8 +123,8 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
   if (err != 0)
     Fatal("posix_spawnattr_setflags: %s", strerror(err));
 
-  const char* spawned_args[] = { "/bin/sh", "-c", command.c_str(), NULL };
-  err = posix_spawn(&pid_, "/bin/sh", &action, &attr,
+  const char* spawned_args[] = { shell.c_str(), "-c", command.c_str(), NULL };
+  err = posix_spawn(&pid_, shell.c_str(), &action, &attr,
         const_cast<char**>(spawned_args), environ);
   if (err != 0)
     Fatal("posix_spawn: %s", strerror(err));
--- a/src/subprocess.h
+++ b/src/subprocess.h
@@ -70,6 +70,7 @@ struct Subprocess {
 #else
   int fd_;
   pid_t pid_;
+  std::string shell;
 #endif
   bool use_console_;
 


atetubou:
Why not put your script in /bin/sh instead when executing Ninja?
Or why not execute your script as Ninja's build actions?

KarstenB:
> Why not put your script in /bin/sh instead when executing Ninja?
Because sub-processes may call/use /bin/sh for themselves. Also substituting such a crucial binary, even in docker, doesn't seem right, but it certainly isn't right on a developer machine. MacOs may even forbid this.

> Or why not execute your script as Ninja's build actions?
The ninja file is automatically generated by cmake and contains lots of commands. I don't know how to tell cmake to prefix every command it puts into the ninja file with a custom command.

Aside from all of that, I think having a hard-coded path in a tool that can't be configured in some way is a good idea.

atetubou:
If you use camke, you can introduce your wrapper via [`CMAKE_<LANG>_COMPILER`](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER.html#variable:CMAKE_%3CLANG%3E_COMPILER)?

KarstenB:
> If you use camke, you can introduce your wrapper via [`CMAKE_<LANG>_COMPILER`](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER.html#variable:CMAKE_%3CLANG%3E_COMPILER)?

No, because we are also using custom_commands, and then you also have to consider the linker and all auxiliary tools of all languages used in the program. It is quite easy to miss something. While not impossible, I think addressing this issue on ninja-level seems more straight-forward.

jhasse:
I'm against using an envronment variable for such a crucial part of Ninja.

You only want to log commands started by Ninja? I think this might still be possible with a custom `sh` executable.

KarstenB:
Replacing /bin/sh is a big no no for me. But if the environment variable is too intransparent, it could also be done with an argument to make the fact more transparent.

hadrielk:
> No, because we are also using custom_commands, and then you also have to consider the linker and all auxiliary tools of all languages used in cmake. It is quite easy to miss something. While not impossible, I think addressing this issue on ninja-level seems more straight-forward.

In case you weren't aware, if you're using CMake you can hook into the compiler _and_ linker executions and custom rules, each separately, using the CMake properties [RULE_LAUNCH_COMPILE](https://cmake.org/cmake/help/latest/prop_gbl/RULE_LAUNCH_COMPILE.html), [RULE_LAUNCH_LINK](https://cmake.org/cmake/help/latest/prop_gbl/RULE_LAUNCH_LINK.html), and [RULE_LAUNCH_CUSTOM](https://cmake.org/cmake/help/latest/prop_gbl/RULE_LAUNCH_CUSTOM.html). 

You can set those properties globally, per directory, and/or or per target.

It works with both Makefile and Ninja generators.

Here's a couple examples from our CMake files, where we use them with the Makefile generator to measure compile and link times:

```
# invokes a shim script, that then invokes the compiler:
set_property(TARGET ${target} PROPERTY RULE_LAUNCH_COMPILE "${CMAKE_HOME_DIRECTORY}/tools/build_tools/measure.sh")

# invokes GNU `time` command:
set_property(TARGET ${target} PROPERTY RULE_LAUNCH_LINK "time --verbose --output=${CMAKE_CURRENT_BINARY_DIR}/${target}.link.measured")
```

jonesmz (approved):

PR #2126, opened 2022-05-06
Really respect the env variable `CLICOLOR_FORCE`

netheril96:
Prior to this commit, ninja disables color output even if CLICOLOR_FORCE is set, if it fails to set VT processing on Windows terminal. This does not respect the user's wish.

One use case of this is to force color on MSYS terminal.

### DIFF
--- a/src/line_printer.cc
+++ b/src/line_printer.cc
@@ -46,17 +46,16 @@ LinePrinter::LinePrinter() : have_blank_line_(true), console_locked_(false) {
   }
 #endif
   supports_color_ = smart_terminal_;
-  if (!supports_color_) {
-    const char* clicolor_force = getenv("CLICOLOR_FORCE");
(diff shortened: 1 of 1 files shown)


jhasse:
Good catch! Could also be fixed by moving the

```cpp
  if (!supports_color_) {
    const char* clicolor_force = getenv("CLICOLOR_FORCE");
    supports_color_ = clicolor_force && string(clicolor_force) != "0";
  }
```

block after the Windows code though, then there's no unnecessary env lookup. I'm not sure though if there's a case where `smart_terminal_` is false, `CLICOLOR_FORCE` is set and `SetConsoleMode` needs to be called?

netheril96:
I recreated the pull request at https://github.com/ninja-build/ninja/pull/2127 because I messed up this branch in my repository.

PR #2127, opened 2022-05-07
Really respect the env variable `CLICOLOR_FORCE`.

netheril96:
Prior to this commit, ninja disables color output even if CLICOLOR_FORCE is set, if it fails to set VT processing on Windows terminal. This does not respect the user's wish.

One use case of this is to force color on MSYS terminal.

### DIFF
--- a/src/line_printer.cc
+++ b/src/line_printer.cc
@@ -46,10 +46,6 @@ LinePrinter::LinePrinter() : have_blank_line_(true), console_locked_(false) {
   }
 #endif
   supports_color_ = smart_terminal_;
-  if (!supports_color_) {
-    const char* clicolor_force = getenv("CLICOLOR_FORCE");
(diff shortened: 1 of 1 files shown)


netheril96:
This is a recreation of https://github.com/ninja-build/ninja/pull/2126. I messed up the original git branch.

src/line_printer.cc
```diff
@@ -61,6 +57,10 @@ LinePrinter::LinePrinter() : have_blank_line_(true), console_locked_(false) {
     }
   }
 #endif
+  if (!supports_color_) {
+    const char* clicolor_force = getenv("CLICOLOR_FORCE");
+    supports_color_ = clicolor_force && string(clicolor_force) != "0";
```
jonesmz, line 62: Please use std:: where appropriate for newly added or modified code.

netheril96: Done

PR #2129, opened 2022-05-15
Add `list` itself to `ninja -d list` and `ninja -t list` output

e-kwsm:


jonesmz (approved):

jhasse:
Debatable. I wouldn't expect it to be listed and would rather keep it as is, sorry.

PR #2130, opened 2022-05-15
improve zsh-completion

e-kwsm:
- add `ninja` prefix to functions
- improve completion of `-d` and `-t`
- stop completion if `-h`, `--help`, or `--version` is supplied
- add missing `--verbose` options

### DIFF
--- a/misc/zsh-completion
+++ b/misc/zsh-completion
@@ -16,7 +16,7 @@
 # Add the following to your .zshrc to tab-complete ninja targets
 #   fpath=(path/to/ninja/misc/zsh-completion $fpath)
 
-__get_targets() {
+(( $+functions[_ninja-get-targets] )) || _ninja-get-targets() {
   dir="."
   if [ -n "${opt_args[-C]}" ];
   then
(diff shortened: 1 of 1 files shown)


e-kwsm:
At the moment completion of `ninja -t` looks like (`|` represents the cursor):
```
$ ninja -t |
Completing tools
browse     cleandead  compdb     graph      recompact  rules               
clean      commands   deps       query      restat     targets
```

This patch improves it as follows:
```
$ ninja -t |
Completing tools
browse     -- browse dependency graph in a web browser
clean      -- clean built files
cleandead  -- clean built files that are no longer produced by the manifest
commands   -- list all commands required to rebuild given targets
compdb     -- dump JSON compilation database to stdout
deps       -- show dependencies stored in the deps log
graph      -- output graphviz dot file for targets
query      -- show inputs/outputs for a path
recompact  -- recompacts ninja-internal data structures
restat     -- restats all outputs in the build log
rules      -- list all rules
targets    -- list targets by their rule or depth in the DAG
```

Note that `list` is missing because `ninja -t list` does not print it (see #2129).

misc/zsh-completion
```diff
@@ -31,42 +31,42 @@ __get_targets() {
   eval ${targets_command} 2>/dev/null | cut -d: -f1
 }
 
-__get_tools() {
-  ninja -t list 2>/dev/null | while read -r a b; do echo $a; done | tail -n +2
+(( $+functions[_ninja-get-tools] )) || _ninja-get-tools() {
+  ninja -t list 2> /dev/null | sed -e '1d;s/^ *//;s/ \+/:/'
```
jonesmz, line 35: While for some people `sed` is a native language, most of us don't find ourselves fluent.

It would be a huge help to people looking at this file in the future if you could clearly document in the code what you're intended outcome is for each of these sed script calls.

e-kwsm: Here `-e '1d;s/^ *//;s/ \+/:/'` is equivalent to `-e '1d'` (remove the first line), `-e 's/^ *//'` (remove leading spaces), and `-e 's/ \+/:/'` (remove multiple spaces with colon)

jonesmz: Sure. Put that in the file so someone looking at it in 3 years has a chance of understanding

e-kwsm: Done.

jonesmz (approved):

PR #2137, opened 2022-05-28
Add an option to avoid building binary when used as a library

ken-matsui:
When Ninja is used as libraries (`libninja` & `libninja-re2c`), we actually need to build them but not a ninja binary. Therefore, I added a CMake variable: `NINJA_NO_BINARY` to prevent building it if users of Ninja's libraries explicitly set it as `ON` (default is `OFF` for backward compatibility).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,8 @@ cmake_minimum_required(VERSION 3.15)
 include(CheckSymbolExists)
 include(CheckIPOSupported)
 
+option(NINJA_BUILD_BINARY "Build ninja binary" ON)
+
 project(ninja)
 
 # --- optional link-time optimization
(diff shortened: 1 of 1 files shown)


jhasse:
You can specify which targets to build, so if you only want to build libninja run `ninja libninja` in the build directory.

jhasse:
Ah I see! Is there any standard CMake way for this or a convention how the variable should be named? In any case I would prefer it if you'd drop the double negative, i.e. name the variable `NINJA_BINARY` and default to 1.

ken-matsui:
@jhasse 
Sure! In this case AFAIK, using `NINJA_BUILD_BINARY` with the [`option command`](https://cmake.org/cmake/help/latest/command/option.html) would be much clearer :)
I have updated the code, but what do you think about this name?

PR #2139, opened 2022-05-31
Support building `libninja-re2c.a` on `configure.py`

ken-matsui:
`configure.py` currently builds only `libninja.a`, but we also want the `libninja-re2c.a` like `CMakeLists.txt` does:

https://github.com/ninja-build/ninja/blob/55f54511d35716c43637dee2bcb5fbc7839f967b/CMakeLists.txt#L47-L50

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -489,16 +489,27 @@ else:
            "changes to src/*.in.cc will not affect your build.")
 n.newline()
 
-n.comment('Core source files all build into ninja library.')
 cxxvariables = []
 if platform.is_msvc():
     cxxvariables = [('pdb', 'ninja.pdb')]
+
+n.comment('Generate a library for `ninja-re2c`.')
+re2c_objs = []
+for name in ['depfile_parser', 'lexer']:
+    re2c_objs += cxx(name, variables=cxxvariables)
+if platform.is_msvc():
+    n.build(built('ninja-re2c.lib'), 'ar', re2c_objs)
+else:
+    n.build(built('libninja-re2c.a'), 'ar', re2c_objs)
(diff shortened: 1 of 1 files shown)


jonesmz:
Could you go into more detail on why this would be a net positive, over removing the configure.py script entirely?

ken-matsui:
Hi @jonesmz,

The `configure.py` is currently used in both [Homebrew](https://github.com/Homebrew/homebrew-core/blob/d9c17c803bae0627449f6d39cab324a91cb9231c/Formula/ninja.rb#L28) and [MacPorts](https://github.com/macports/macports-ports/blob/f69b1ef3f96f783a819f0c9c28f36828c53a6c47/devel/ninja/Portfile#L54) (and might other sources), so I would say deleting the script would bring a big breaking change.

Thus, if this breaking change is acceptable, I think the script should be removed and I am more in favor of that.

hadrielk:
Fedors and others use it too, in their RPM specs.

ken-matsui:
Ah, then this PR should have the least impact, IMO.

jonesmz:
Those projects can stop using configure.py when they version bump to the next release of ninja.
Maintaining two distinct build systems is a waste of everyone's time.

ken-matsui:
Yes, ideally, it would be great, and I would agree with you if I were a maintainer without historical background knowledge. Could you please open a PR that removes `configure.py` rather than updating this PR? Either yours or mine will be merged by maintainers.

jonesmz:
Removing a file doesn't requite a pull request. The project maintainers will remove it when they're ready.

PR #2143, opened 2022-06-07
Flush output after each line

orgads:
The output is not flushed automatically after `\n`, at least on Windows 10.

### DIFF
--- a/src/line_printer.cc
+++ b/src/line_printer.cc
@@ -118,6 +118,7 @@ void LinePrinter::Print(string to_print, LineType type) {
     have_blank_line_ = false;
   } else {
     printf("%s\n", to_print.c_str());
+    fflush(stdout);
   }
 }
 
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -672,6 +672,7 @@ int NinjaMain::ToolRules(const Options* options, int argc, char* argv[]) {
       }
     }
     printf("\n");
+    fflush(stdout);
   }
   return 0;
 }


jonesmz:
Could you go into more detail on why this change is a net benefit?

jonesmz:
I take it that you mean that the video is from a version of ninja without this change?

I use like 6 different command prompts (windows cmd, windows powershell, windows wsl, linux bash + terminology, linux bash + tty, linux bash + tty over UART) and I've never seen the behavior you have in that video.

I strongly suspect there's something wrong with your terminal (mintty? is that from mingw?)

orgads:
Yes, this is before the change.

It doesn't happen on terminals that are considered "smart", see the condition above. On these terminals, `\r` is used.

jonesmz:
> Why do you have support for non-smart terminals at all then?

*shrug* I'm not the person you have to convince to merge this.
 
> By the way, if I run `winpty ninja` it works with `\r`.

That... kind of implies that the problem is with mintty ?

> This also happens in IDEs, such as Qt Creator.

This is a more reasonable justification, imho. IDEs build logs are not interactive (typically...), but we would want to support a convenient experience for them.

orgads:
This is a known limitation with mintty. See [here](https://github.com/git-for-windows/git/wiki/FAQ#some-native-console-programs-dont-work-when-run-from-git-bash-how-to-fix-it).

jonesmz:
So... mintty has known bugs that no one's bothered to fix, but Ninja needs to fix / work around it?

The IDE build log is much more compelling than mintty having problems. If fixing the IDE build log scenerio also happens to make mintty cooperate, that's cool.

orgads:
Do you see a reason *not* to flush?

jonesmz:
Yes. Non-interactive terminal sessions don't need flushing, so flushing is an unnecessary runtime expense, which in some situations can add up to a lot.

I'm not saying I think it's a bad idea to flush every line for interactive sessions, quite the opposite.

I'm just saying that "mintty is broken, lets make ninja work around it" is not a good justification, regardless of whether there's actively a reason not to work around it.

Making IDEs behave properly is, in my opinion, much easier to justify putting a work-around into ninja for, as that's something that an enormous number of people deal with every day. The set of people using mintty is probably very very small.

That being said, I *also* think that IDEs should fix their bugs / poor behavior, and not expect Ninja to do it for them.

orgads:
It's not a bug in the IDE. It just reads from stdout stream rather than implementing a full terminal, which is a very hard task. The application already supports 2 modes (smart/non-smart), but it seems to rely on the C runtime to flush after each line. But unfortunately, it looks like the Windows runtime doesn't do that, at least when the output is not a tty, so let's be explicit.

It is not in the common use-case (smart) code path anyway, so it won't affect most users at all.

jhasse:
Unfortunately this can have quite the performance impact on large builds. Maybe we should limit it to Windows?

Ideally we would remember the last time we flushed and never flush more than 60 time per seonds?

orgads:
Why is it different than "smart terminal", on which you flush after each command?

orgads:
ping

orgads:
Define "faster". I doubt flushing frequently will have more than 0.01% impact on the build time.

If you insist, I'll try to come up with something.

jhasse:
Would be enough if you can quickly test that it's really just 0.01%.

orgads:
I ran a test with llvm build with -n, which should reflect merely the flush difference. My build has 3471 commands.

```bash
time ninja -n > test.txt
```

Ran each version 20 times. The average time is 0.4871 without my change, and 0.489 with my change.

So that's ~0.4% difference, and that's without any real compilation, which obviously takes most of the time.

orgads:
@jhasse ping?

orgads:
Thank you.

Sevenrocks22:
Se7en

PR #2147, opened 2022-06-14
Use `unique_ptr` to safely hold `LogEntry` pointers through `HashMap` when C++11 or later

ken-matsui:
Since it potentially lets memory leak that `Entries` holds `LogEntry` pointers as raw through `HashMap`, I introduced `std::unique_ptr` to safely manage these pointers when C++11 or later is detected.

https://github.com/ninja-build/ninja/blob/1d4034f0ac2b5cfb809b5ab983d47c3cb2c78415/src/build_log.h#L93

### DIFF
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -160,7 +160,7 @@ bool BuildLog::RecordCommand(Edge* edge, int start_time, int end_time,
 #if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
       std::unique_ptr<LogEntry> e(new LogEntry(path));
       log_entry = e.get();
-      entries_.emplace(e->output, std::move(e));
+      entries_.emplace(log_entry->output, std::move(e));
 #else
       log_entry = new LogEntry(path);
       entries_.insert(Entries::value_type(log_entry->output, log_entry));
@@ -352,7 +352,7 @@ LoadStatus BuildLog::Load(const string& path, string* err) {
 #if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
       std::unique_ptr<LogEntry> e(new LogEntry(output));
       entry = e.get();
-      entries_.emplace(e->output, std::move(e));
+      entries_.emplace(entry->output, std::move(e));
 #else
       entry = new LogEntry(output);
       entries_.insert(Entries::value_type(entry->output, entry));
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -382,7 +382,7 @@ LoadStatus BuildLog::Load(const string& path, string* err) {
   return LOAD_SUCCESS;
 }
 
-BuildLog::LogEntryPtrRef BuildLog::LookupByOutput(const string& path) {
+BuildLog::LogEntryPtr BuildLog::LookupByOutput(const string& path) {
   Entries::iterator i = entries_.find(path);
   if (i != entries_.end())
     return i->second;
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -382,7 +382,7 @@ LoadStatus BuildLog::Load(const string& path, string* err) {
   return LOAD_SUCCESS;
 }
 
-BuildLog::LogEntryPtr BuildLog::LookupByOutput(const string& path) {
+BuildLog::LogEntryPtrRef BuildLog::LookupByOutput(const string& path) {
   Entries::iterator i = entries_.find(path);
   if (i != entries_.end())
     return i->second;
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -151,13 +151,9 @@ bool BuildLog::RecordCommand(Edge* edge, int start_time, int end_time,
     Entries::iterator i = entries_.find(path);
     LogEntry* log_entry;
     if (i != entries_.end()) {
-#if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
-      log_entry = i->second.get();
-#else
-      log_entry = i->second;
-#endif
+      log_entry = to_address(i->second);
     } else {
-#if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
+#if NINJA_CPP11
       std::unique_ptr<LogEntry> e(new LogEntry(path));
       log_entry = e.get();
       entries_.emplace(log_entry->output, std::move(e));
@@ -343,13 +339,9 @@ LoadStatus BuildLog::Load(const string& path, string* err) {
     LogEntry* entry;
     Entries::iterator i = entries_.find(output);
     if (i != entries_.end()) {
-#if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
-      entry = i->second.get();
-#else
-      entry = i->second;
-#endif
+      entry = to_address(i->second);
     } else {
-#if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
+#if NINJA_CPP11
       std::unique_ptr<LogEntry> e(new LogEntry(output));
       entry = e.get();
       entries_.emplace(entry->output, std::move(e));
@@ -397,11 +389,7 @@ LoadStatus BuildLog::Load(const string& path, string* err) {
 BuildLog::LogEntry* BuildLog::LookupByOutput(const string& path) {
   Entries::iterator i = entries_.find(path);
   if (i != entries_.end())
-#if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
-    return i->second.get();
-#else
-    return i->second;
-#endif
+    return to_address(i->second);
   return NULL;
 }
 
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -149,16 +149,22 @@ bool BuildLog::RecordCommand(Edge* edge, int start_time, int end_time,
        out != edge->outputs_.end(); ++out) {
     const string& path = (*out)->path();
     Entries::iterator i = entries_.find(path);
-    LogEntryPtr log_entry;
+    LogEntry* log_entry;
     if (i != entries_.end()) {
+#if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
+      log_entry = i->second.get();
+#else
       log_entry = i->second;
+#endif
     } else {
 #if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
-      log_entry = std::make_shared<LogEntry>(path);
+      std::unique_ptr<LogEntry> e(new LogEntry(path));
+      log_entry = e.get();
+      entries_.emplace(e->output, std::move(e));
 #else
       log_entry = new LogEntry(path);
-#endif
       entries_.insert(Entries::value_type(log_entry->output, log_entry));
+#endif
     }
     log_entry->command_hash = command_hash;
     log_entry->start_time = start_time;
@@ -334,17 +340,23 @@ LoadStatus BuildLog::Load(const string& path, string* err) {
     start = end + 1;
     end = line_end;
 
-    LogEntryPtr entry;
+    LogEntry* entry;
     Entries::iterator i = entries_.find(output);
     if (i != entries_.end()) {
+#if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
+      entry = i->second.get();
+#else
       entry = i->second;
+#endif
     } else {
 #if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
-      entry = std::make_shared<LogEntry>(output);
+      std::unique_ptr<LogEntry> e(new LogEntry(path));
+      entry = e.get();
+      entries_.emplace(e->output, std::move(e));
 #else
-      entry = new LogEntry(output);
-#endif
+      entry = new LogEntry(path);
       entries_.insert(Entries::value_type(entry->output, entry));
+#endif
       ++unique_entry_count;
     }
     ++total_entry_count;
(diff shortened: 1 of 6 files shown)


jonesmz:
Keep in mind that std::shared_ptr doesn't have a template parameter to omit the atomic operations involved in the reference counting.

Since Ninja is currently single threaded, that's quite a bit of overhead to pay for.

Might be easier to justify using a hand-rolled ninja::shared_ptr which omits the atomic operations.

src/build_log.h
```diff
@@ -15,6 +15,16 @@
 #ifndef NINJA_BUILD_LOG_H_
 #define NINJA_BUILD_LOG_H_
 
+#if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
+#define NINJA_NULLPTR nullptr
+#else
+#define NINJA_NULLPTR NULL
```
jonesmz, line 21: There's really no value to this typedef. NULL converts to nullptr where appropriate.

Better to just wait for C++11 to be merged, and then switch to nullptr wholesale.

src/build_log.h
```diff
@@ -15,6 +15,16 @@
...
+#endif
+
+#if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
+#include <memory>
```
jonesmz, line 25: Why bother protecting this include? If nothing from it is used, it's a no-op. Just #include <memory> without the ifdefs.

src/build_log.h
```diff
@@ -76,8 +86,14 @@ struct BuildLog {
              int start_time, int end_time, TimeStamp mtime);
   };
 
+#if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
+  typedef std::shared_ptr<LogEntry> LogEntryPtr;
```
jonesmz, line 90: MIght as well use `using` if C++11 is available. It's superior to typedef in everyway.

src/build_log_test.cc
```diff
@@ -63,9 +63,9 @@ TEST_F(BuildLogTest, WriteRead) {
 
   ASSERT_EQ(2u, log1.entries().size());
   ASSERT_EQ(2u, log2.entries().size());
-  BuildLog::LogEntry* e1 = log1.LookupByOutput("out");
+  BuildLog::LogEntryPtr e1 = log1.LookupByOutput("out");
```
jonesmz, line 66: Take the return value by const& to prevent incrementing the refcount unnecessarily.

src/build_log.h
```diff
@@ -76,8 +86,14 @@ struct BuildLog {
...
+#else
+  typedef LogEntry* LogEntryPtr;
+#endif
+
   /// Lookup a previously-run command by its output path.
-  LogEntry* LookupByOutput(const std::string& path);
+  LogEntryPtr LookupByOutput(const std::string& path);
```
jonesmz, line 96: return by const& to avoid incrementing the refcount.

src/build_log.h
```diff
@@ -90,7 +106,7 @@ struct BuildLog {
   bool Restat(StringPiece path, const DiskInterface& disk_interface,
               int output_count, char** outputs, std::string* err);
 
-  typedef ExternalStringHashMap<LogEntry*>::Type Entries;
+  typedef ExternalStringHashMap<LogEntryPtr>::Type Entries;
```
jonesmz, line 109: It *appears* to be the case that this hash map is the "owner" of the log entry.

If that's the case, you could probably use std::unique_ptr, and then return references to that unique ptr where you need to, with the use of std::reference_wrapper<> where appropriate.

src/build_log.h
ken-matsui: @jonesmz 
Thank you for your reviews :)

I tried to use `NULL`, but it seems that `NULL` cannot assign to `std::shared_ptr`.

```console
error: no viable conversion from 'long' to 'BuildLog::LogEntryPtr' (aka 'shared_ptr<BuildLog::LogEntry>')
  BuildLog::LogEntryPtr entry = NULL;
                        ^       ~~~~
```

WDYT?

src/build_log.h
ken-matsui: @jonesmz 

I think it is hard to use `unique_ptr` here because we are using a copy of the pointer in many places:

https://github.com/ninja-build/ninja/blob/1d4034f0ac2b5cfb809b5ab983d47c3cb2c78415/src/build_log.cc#L152-L162

src/build_log.h
ken-matsui: I could not return the pointer as `const&` because the rvalue is returned.

https://github.com/ninja-build/ninja/blob/1d4034f0ac2b5cfb809b5ab983d47c3cb2c78415/src/build_log.cc#L377-L382

src/build_log.h
atetubou: Why we cannot get raw pointer from `unique_ptr` there?

ken-matsui: Ah, totally forgot that, thanks.

src/build_log.h
jonesmz: In those situations, you can wrap those lines with #ifdef, and use nullptr directly.

src/build_log.cc
```diff
@@ -377,7 +397,11 @@ LoadStatus BuildLog::Load(const string& path, string* err) {
 BuildLog::LogEntry* BuildLog::LookupByOutput(const string& path) {
   Entries::iterator i = entries_.find(path);
   if (i != entries_.end())
+#if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
+    return i->second.get();
```
jonesmz, line 401: i think that for many of the places where you are returning `ptr.get()` you would instead want to change the function to return `LogEntryPtr const&`

ken-matsui: Hey @jonesmz,

This function also returns an rvalue, so your suggestion will cause a segmentation fault:

https://godbolt.org/z/a99acs3oe

jonesmz:
> I might be totally wrong, but I thought from reading previous issues about mem leaks, that `ninja` didn't care about them and in fact in some cases it was a conscious decision to leak.
> 
> The rationale being that `ninja` is used as a command not service, so letting the OS reclaim them when `ninja` terminates is reasonable and better performance.

That's true today, but multiple people have said that they use (currently) or intend to use (in the future) ninja as a library.

Applications that "intend to leak" memory for OS cleanup purposes and therefore are written in terms of raw c-style pointers are wrong.

It would be fine those applications wanted to use proper lifetime management techniques, like a std::shared_ptr (or a single-threaded version thereof) or std::unique_ptr, and provide custom deleters that were no-op deleters based on compiler flags so that "command mode" usage could "leak" while "library" usage did not, but that's not what happens in the ninja codebase because std::auto_ptr was a design defect in the standard, and ninja otherwise doesn't have access to "smart" pointers.

src/build_log.cc
jonesmz: That's because functions which return a const& may not return a variable that lives only in the stack of that function.

My recommendation is that for functions which are returning items that live inside of a map, you can return the mapped item as const&, because the map itself won't be deleteing the item until something tells it so.

But you can't change the return type to std::unique_ptr const&, and then return a c-style pointer, because that will result in returning a temporary variable that lives on the stack of the function as a const&, aka a dangling reference.

I wasn't specifically telling you that you *must* do any of the things I commented about. I'm just giving high level design feedback. Don't take my comments as word-of-law, and don't assume that I spent a whole lot of time looking through the code to make sure my advice was 100% correct. It was a drive by recommendation based on years of working with the smart pointers from the standard library.

src/build_log.cc
```diff
@@ -151,10 +151,20 @@ bool BuildLog::RecordCommand(Edge* edge, int start_time, int end_time,
...
+#if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
+      log_entry = i->second.get();
+#else
       log_entry = i->second;
+#endif
     } else {
+#if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
+      std::unique_ptr<LogEntry> e(new LogEntry(path));
+      log_entry = e.get();
+      entries_.emplace(e->output, std::move(e));
```
atetubou, line 163:
```suggestion
      std::string log_output = e->output;
      entries_.emplace(std::move(log_output), std::move(e));
```

In C++, `std::move(e)` might be evaluated before `e->output`.

jonesmz: That's correct, parameter order evaluation is implementation defined.

src/build_log.cc
```diff
@@ -151,10 +151,20 @@ bool BuildLog::RecordCommand(Edge* edge, int start_time, int end_time,
...
     LogEntry* log_entry;
     if (i != entries_.end()) {
+#if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
+      log_entry = i->second.get();
+#else
       log_entry = i->second;
+#endif
     } else {
+#if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
+      std::unique_ptr<LogEntry> e(new LogEntry(path));
```
jonesmz, line 161: Use std::make_unique

src/build_log.cc
```diff
@@ -151,10 +151,20 @@ bool BuildLog::RecordCommand(Edge* edge, int start_time, int end_time,
     Entries::iterator i = entries_.find(path);
     LogEntry* log_entry;
     if (i != entries_.end()) {
+#if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
```
jonesmz, line 154: create a helper function

```
template<typename TYPE_T>
TYPE_T * to_address(TYPE_T * ptr) { return ptr; }
#if CPP11
template<typename TYPE_T>
decltype(auto) to_address(std::unique_ptr<TYPE_T> const& ptr) { return ptr.get(); }
#endif 

So that you don't have all these ugly if-defs everywhere.

File: src/build_log.cc (hunk 9 above)
Comment (ken-matsui) [reply src/build_log.cc:161]: `std::make_unique` is from C++14 :)

File: src/build_log.cc (hunk 7 above)
Comment (ken-matsui) [reply src/build_log.cc:401]: Or much better way would be using `std::optional`, but it is implemented since C++17...

File: src/build_log.cc (hunk 8 above)
Comment (ken-matsui) [reply src/build_log.cc:163]: This is what I was unsure about... Thank you!

Comment (ken-matsui) [reply src/build_log.cc]: This is what I was unsure about. Thank you!

File: src/util.h [hunk 11]
```diff
@@ -49,6 +50,20 @@ NORETURN void Fatal(const char* msg, ...);
...
+#else
+#  define NINJA_CPP11 false
+#endif
+
+template <typename T>
+T* to_address(T* ptr) { return ptr; }
+#if NINJA_CPP11
+template <typename T>
+decltype(auto) to_address(const std::unique_ptr<T>& ptr) { return ptr.get(); }
+#endif
```
jonesmz, line 65: Minor, up to you if you want to do this,

but you could add a check for

#if defined([__cpp_lib_to_address](https://en.cppreference.com/w/cpp/utility/feature_test))

and if true, then replace these with `using std::to_address`, which is the same function, but as part of the standard.

src/util.h
```diff
@@ -49,6 +50,20 @@ NORETURN void Fatal(const char* msg, ...);
 #define NINJA_FALLTHROUGH
 #endif
 
+/// Whether C++11 features are enabled or not.
+#if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
+#  define NINJA_CPP11 true
```
jonesmz, line 55: Minor, newer versions of the standard have changed from just doing monolithic version checks to instead doing #defines based on individual features being supported by that version of the compiler / standard lib.

So since in this situation, you're adding this new definition *specifically* to check if there support for std::unique_ptr, you might consider changing this from NINJA_CPP11, to NINJA_HAS_UNIQUE_PTR, or something of that nature.

That being said, I don't see a problem with what you have here. It's just a minor recommendation.

src/util.h
```diff
@@ -49,6 +50,20 @@ NORETURN void Fatal(const char* msg, ...);
...
+#else
+#  define NINJA_CPP11 false
```
jonesmz, line 57: Also, technically you could change this to simply

#define NINJA_CPP11 __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)

and get the same behavior, without having to do #if / #else / #endif.

src/util.h
ken-matsui: @jonesmz 

Umm... I had found many C++11 checks in different ways, so I wanted to unify them using `NINJA_CPP11` in the next PR.

https://github.com/ninja-build/ninja/search?q=%23if+__cplusplus+%3E%3D+201103L

WDYT?

src/util.h
ken-matsui: Cool. Thank you!

src/util.h
jonesmz: Yea, there's a lot of dispersed CPP11 checks, i agree.

I think my breakdown of preferences, in order, are

1. If your compiler / platform doesn't support C++11, you're using something very broken -- aka, mandatory C++11 minimum
2. Per feature detection. E.g. `std::make_unique` has the feature detection macro: [__cpp_lib_make_unique](https://en.cppreference.com/w/cpp/utility/feature_test) -- I think starting with C++14, basically every feature has a detection macro. If Ninja isn't going to fully adopt C++11, then my personal preference would be to adopt our own C++11 feature detection macros to make more fine-grained support possible for the various broken compilers / platforms out there.
3. Standardize all C++11 features under NINJA_CPP11 like you have here.

jhasse:
Yes, the memory leaks are intentional. Therefore this can't be merged as it potentially degrades performance.

As for using ninja as a library: That's not supported yet.

ken-matsui:
Alright, thank you :) closing.

ken-matsui:
I work on my own fork. Thank you so much for your kind reviews :)

PR #2150, opened 2022-06-15
inja templating engine support for ninja

wilsonwang371:
inja is similar to jinja and using a templating engine can help us reducing a lot of repetitive work in ninja file.

### DIFF
--- a/src/dyndep_parser.cc
+++ b/src/dyndep_parser.cc
@@ -30,7 +30,7 @@ DyndepParser::DyndepParser(State* state, FileReader* file_reader,
     , dyndep_file_(dyndep_file) {
 }
 
-bool DyndepParser::Parse(const string& filename, const string& input,
+bool DyndepParser::Parse(const string& filename, const string& param_filename, const string& input,
                          string* err) {
   lexer_.Start(filename, input);
 
--- a/src/parser.cc
+++ b/src/parser.cc
@@ -14,12 +14,13 @@
 
 #include "parser.h"
 
+#include <string.h>
+#include <exception>
+
 #include "disk_interface.h"
 #include "metrics.h"
-#include "util.h"
-
-#include <string.h>
 #include "thirdparty/inja/inja.hpp"
+#include "util.h"
 
 using namespace std;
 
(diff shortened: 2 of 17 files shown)


jonesmz:
Please used proper std:: namespacing on new or modified code.

You should pretend that the "using namespace std;" lines in some of the cpp files are not actually there.

src/dyndep_parser.cc
```diff
@@ -30,7 +30,7 @@ DyndepParser::DyndepParser(State* state, FileReader* file_reader,
     , dyndep_file_(dyndep_file) {
 }
 
-bool DyndepParser::Parse(const string& filename, const string& input,
+bool DyndepParser::Parse(const string& filename, const string& param_filename, const string& input,
```
jonesmz, line 33: use `StringPiece` over `std::string const&` when practical to do so.

src/parser.cc
```diff
@@ -14,12 +14,25 @@
...
 #include "disk_interface.h"
 #include "metrics.h"
+#include "thirdparty/inja/inja.hpp"
+#include "util.h"
 
 using namespace std;
 
-bool Parser::Load(const string& filename, string* err, Lexer* parent) {
+void registerCallbacks(inja::Environment& env) {
+  env.add_void_callback("warning", 1, [](inja::Arguments& args) {
```
jonesmz, line 28: lambdas are not allowed until https://github.com/ninja-build/ninja/pull/2089 is merged

src/thirdparty/inja/inja.hpp
```diff
@@ -0,0 +1,4443 @@
...
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+*/
+
+#ifndef INCLUDE_INJA_INJA_HPP_
+#define INCLUDE_INJA_INJA_HPP_
+
+#include <nlohmann/json.hpp>
```
jonesmz, line 28: I strongly advice against adding a dependency like this directly into the repository. Use a git submodule, or a CMake "fetch_content" call (personally I prefer git submodule)

jonesmz:
I have to admit, i don't see how this is beneficial. It *appears* to be the case, based on your example, that this inja templating preprocesses the entire ninja file prior to running the dependency graph.

Why should that be built into ninja itself, instead of just run as a pre-processing step on the .ninja file after it's generated?

Ninja is almost always used as a machine-generated intermediate representation, e.g. CMake -> ninja, 

So since CMake, just to use it as an example, already knows how to generate everything based on it's own internal rules, what does this templating syntax provide to cmake that it can't do itself?

PR #2152, opened 2022-06-16
Set _CRT_SECURE_NO_WARNINGS to disable unnecessary and very noisy warnings

jonesmz:
The warnings generated by MSVC related to "insecure" functions from the c runtime are just noise.

This PR will suppress those warnings so that people can focus on real issues.


jonesmz:
Hrmm, i see that this change was added recently to the master branch.

Obsolete.

PR #2162, opened 2022-07-01
also check if stdin is terminal

milahu:
current version of ninja fails to detect the non-terminal environment in the nix-build sandbox

(and nix fails to print ninja's line-cleared output to the terminal)

downstream issue https://github.com/NixOS/nixpkgs/pull/179027

we want to avoid workarounds like

```sh
TERM=dumb ninja
ninja | cat
```

### DIFF
--- a/src/line_printer.cc
+++ b/src/line_printer.cc
@@ -35,7 +35,7 @@ using namespace std;
 LinePrinter::LinePrinter() : have_blank_line_(true), console_locked_(false) {
   const char* term = getenv("TERM");
 #ifndef _WIN32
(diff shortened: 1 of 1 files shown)


milahu:
> I don't understand why we would want to check std**in** whether to **output** in a "smart" way.

yes, actually that makes little sense

> Why does `isatty(1)` return true but `isatty(0)` does not?

in the nix-build sandbox:
stdin is connected to /dev/null 
stdout is a "read only terminal", a pty to get pretty output from builders

we were looking for an upstreamable solution for our downstream issue (ninja output is lost)
but seems like it wont get better than `TERM=dumb ninja`

PR #2165, opened 2022-07-10
miniscule grammar fix in documentation

rjindael:
"If you do this work" -> "If this works for you"


jonesmz:
I think that the point of that sentence was to ask the people who convert build system generators to use ninja to tell the ninja project.

I don't think that the goal of the sentence was to ask people to tell the ninja project about random build tools that happen to work with ninja, as there wouldn't be any that would.

rjindael:
"if you do this work" still doesn't make any grammatical sense

jonesmz:
"If you do this work", aka "If you do the work to adapt a build system to use ninja"

jhasse:
"work" is used as a noun, for me it makes sense.

I'm not a native speaker, but I still believe this has a different meaning than "If you find one that works".

PR #2172, opened 2022-07-26
Windows: assume smart terminal if $TERM is defined and not "dumb"

### LINKED ISSUE #2171: Enable `smart_terminal` functionality outside of the Windows console
Currently `line_printer.cc` contains this code:

```
#ifndef _WIN32
  smart_terminal_ = isatty(1) && term && string(term) != "dumb";
#else
  if (term && string(term) == "dumb") {
    smart_terminal_ = false;
  } else {
    console_ = GetStdHandle(STD_OUTPUT_HANDLE);
    CONSOLE_SCREEN_BUFFER_INFO csbi;
    smart_terminal_ = GetConsoleScreenBufferInfo(console_, &csbi);
  }
#endif
```

which means in practice that on Windows `smart_terminal` is only set to true when outputting to a console. However, some host applications (like Emacs) are perfectly capable of handling Ninja's one-line progress status and color codes as well.

The problem is how to tell Ninja to behave as if in a smart terminal no matter the absence of a console. I propose to set `smart_terminal` to true when `term` is non-empty and different from "dumb". Ditto for `supports_color`. Is this acceptable? Other suggestions?

oscarfv:
On Windows, ninja only uses smart terminal features when outputting to
a Windows console. However, there are applications that act as
terminal emulators and support those features. Emacs and MSYS2's
mintty emulator are some examples.

With this change ninja uses smart terminal features whenever $TERM is
non-empty and its value is different from "dumb".

This fixes #2171

### DIFF
--- a/src/line_printer.cc
+++ b/src/line_printer.cc
@@ -37,8 +37,8 @@ LinePrinter::LinePrinter() : have_blank_line_(true), console_locked_(false) {
 #ifndef _WIN32
   smart_terminal_ = isatty(1) && term && string(term) != "dumb";
 #else
-  if (term && string(term) == "dumb") {
-    smart_terminal_ = false;
+  if (term) {
+    smart_terminal_ = string(term) != "dumb";
   } else {
     console_ = GetStdHandle(STD_OUTPUT_HANDLE);
     CONSOLE_SCREEN_BUFFER_INFO csbi;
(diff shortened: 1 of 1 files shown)


oscarfv:
This wont work: the presence of a value in $TERM does not indicate that the output is going to a (real or emulated) terminal.

For instance, MSYS2 mintty on Windows sets $TERM to `xterm`. When CMake executes ninja at configure time, ninja sees that value in $TERM, enables fancy output and confuses CMake.

Issue #2173, opened 2022-07-28
SIGFPE on system with zero in /cgroup/cpu/cpu.cfs_period_us


lrtfm:
In the code, to calculate the cpu limit, if the value of `/cgroup/cpu/cpu.cfs_period_us` is 0, the program will generate the signal SIGFPE. Should we protect this by checking the value of `period.first`?

code: https://github.com/ninja-build/ninja/blob/d4017a2b1ea642f12dabe05ec99b2a16c93e99aa/src/util.cc#L650

The error information:
```
Program received signal SIGFPE, Arithmetic exception.
0x000000000042d2d5 in ParseCPUFromCGroup() () at ./src/util.cc:650
650       return quota.first / period.first;
```

AyandaRevon:
> *123

PR #1827, opened 2020-08-18
Support cpu limit in docker

okhowang:
close #1824

on windows docker with process isolation, use job api to get cpu limit.
on linux docker, use cgroups to get cpu limit.

PR #2174, opened 2022-07-28
Fixbug: SIGFPE error when cpu.cfs_period_us = 0

lrtfm:
This will fix the issue (#2173).


jhasse:
Thank you very much! In hindsight I shouldn't have merged #1827 as this is a critical issue which needs a bugfix release ...

PR #2179, opened 2022-08-13
Supporting OS400 system

zheddie:
This fix is to make ninja support on IBM i OS for "pip install". Please kindly help to review and merge. thanks.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -69,7 +69,7 @@ class Platform(object):
     def known_platforms():
       return ['linux', 'darwin', 'freebsd', 'openbsd', 'solaris', 'sunos5',
               'mingw', 'msvc', 'gnukfreebsd', 'bitrig', 'netbsd', 'aix',
-              'dragonfly']
+              'dragonfly', 'os400']
 
     def platform(self):
         return self._platform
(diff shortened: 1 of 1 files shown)


zheddie:
@jhasse , Hello, can you help to review this PR ? Thanks.

jonesmz:
You should just use the cmake build script. Configure.py is deprecated.

zheddie:
thanks @jonesmz . I notice that the python package :https://github.com/scikit-build/ninja-python-distributions are still using the configure.py. With this fix , we can make the ninja python package works on our system. Can you help to merge it , if it does not break any other feature/system? thanks.

jonesmz:
I am not a maintainer, so I don't have the ability to merge pull requests.

I strongly discourage making changes to configure.py that aren't also done to cmakelists.txt (if needed) though.

jhasse:
We can't test these changes and I'm against supporting closed source operation systems / software whenever possible (at least when I'm not getting paid). Sorry.

jonesmz:
@zheddie 

Your path forward with this is *probably* to submit a fix to the scikit-build/ninja-python-distributions project to use cmake, instead of configure.py.

CMake, assuming it supports `IBM i OS` or `OS400`, will then let you use ninja for your platform without needing changes to the Ninja codebase.

jonesmz:
> CMake, assuming it supports IBM i OS or OS400, will then let you use ninja for your platform without needing changes to the Ninja codebase.

More closely reviewing the cmake code demonstrates that the Ninja project does support building for OS400 with CMake.

For what it's worth, I agree with jhasse that supporting OS400 is not the Ninja project's problem, and I object to special affordances in the CMakeLists.txt file, but the support is there today anyway.

zheddie:
@jonesmz thanks for your comments and explanation. I would try that way.

PR #2182, opened 2022-08-17
Doc fixes

philg314:
This PR changes all links in the docs to https, fixes the [list continuation](https://asciidoc-py.github.io/userguide.html#X15) of the [Extra Tools: msvc](https://ninja-build.org/manual.html#_extra_tools) section in the manual, and fixes overflowing code blocks.

Currently overflowing code blocks have been rewritten to fit and to prevent future overflows the `min-width` of `<pre>`s is set to `min-content`, which has been [generally supported since about 2020](https://caniuse.com/mdn-css_properties_width_min-content).

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -533,7 +533,7 @@ To create a rule that never rebuilds, use a build rule without any input:
 rule touch
   command = touch $out
 build file_that_always_exists.dummy: touch
-build dummy_target_to_follow_a_pattern: phony file_that_always_exists.dummy
+build dummy_target: phony file_that_always_exists.dummy
 ----------------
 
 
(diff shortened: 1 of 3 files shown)


doc/manual.asciidoc
```diff
@@ -533,7 +533,7 @@ To create a rule that never rebuilds, use a build rule without any input:
 rule touch
   command = touch $out
 build file_that_always_exists.dummy: touch
-build dummy_target_to_follow_a_pattern: phony file_that_always_exists.dummy
+build dummy_target: phony file_that_always_exists.dummy
```
jhasse, line 536: ???

philg314: This line is overflowing its code block.

jhasse: I see. Still I think this should be handled with a scrollbar instead.

jhasse:
Please recreate without the dummy_target_to_follow_a_pattern change.

Issue #2136, opened 2022-05-27
Add a switch to let users decide whether to disable/enable stdout buffering on Windows


shithappens2016:
msys2/MINGW-packages#11711

#2085

jhasse:
I think we should always flush here: https://github.com/ninja-build/ninja/blob/55f54511d35716c43637dee2bcb5fbc7839f967b/src/line_printer.cc#L104

jonesmz:
This seems related to https://github.com/ninja-build/ninja/issues/2164

jhasse:
fixed by #2143

PR #2190, opened 2022-09-03
Enable fancy progress line when $TERM != dumb on Windows (#2171)

oscarfv:


src/line_printer.cc
```diff
@@ -78,6 +84,14 @@ void LinePrinter::Print(string to_print, LineType type) {
 
   if (smart_terminal_ && type == ELIDE) {
 #ifdef _WIN32
+    if (console_ == INVALID_HANDLE_VALUE) {
+      // Assume we are in a terminal emulator.
+      to_print = ElideMiddle(to_print, 80);
```
jhasse, line 89: 80 is wrong though.

src/line_printer.cc
```diff
@@ -51,6 +53,10 @@ LinePrinter::LinePrinter() : have_blank_line_(true), console_locked_(false) {
     supports_color_ = clicolor_force && string(clicolor_force) != "0";
   }
 #ifdef _WIN32
+  // On Windows it is not possible to distinguish a terminal emulator
+  // from a pipe and ANSI escape sequences break applications that
```
jhasse, line 57: What do you mean by "terminal emulator"? Because on Windows it's certainly possible to check if a program is outputting on the console or getting piped.

oscarfv: A terminal emulator collects the output with a pipe, processes it and displays it on a console or some sort of text widget.

So ninja needs some heuristic, because the absence of a console does not mean that the output is not being displayed.

src/line_printer.cc
oscarfv: There is no way of knowing the number of columns of the terminal emulator, 80 looks like a reasonable default to me.

src/line_printer.cc
jhasse: > A terminal emulator collects the output with a pipe, [...]

Well ... that's a pipe then. Emacs needs to provide a real Windows console so programs can call the APIs needed. This is the same as providing a tty on Linux so `ioctl` works.

Have a look at https://github.com/rprichard/winpty which can do most of this for you.

oscarfv: Emacs supports multiple child processes running simultaneously. Some terminal emulators support multiple tabbed "terminals", each with its own process. So having a console (which Emacs has) is of no help. We really need a method of telling Ninja that it is running on a smart terminal regardless the presence or absence of an associated console (because even when the console exists, the output might not meant to be displayed on that console, but elsewhere, so the parent process collects Ninja's output with a pipe).

`winpty` has nothing to do with this PR. `winpty` is a utility for running native Windows interactive text-based programs under a Cygwin/MSYS2 terminal.

oscarfv: Sorry, I missed the part where you say _Emacs needs to provide a real Windows console so programs can call the APIs needed._

Those APIs are of little use when the output is displayed on a terminal emulator because, in general, its characteristics (such as number of columns) and capabilities have no relation to the process' Windows console, when it exists.

By the way, this patch makes Ninja's observable behavior on Windows converge to what the user experiences on *nix, i.e. `smart_terminal_` will be `true` on more scenarios where it is also `true` on *nix.

jhasse: emacs should not forward the calls to its Windows console but answer them itself, i.e. return the number of column of its integrated terminal.

jhasse: > And why are we talking so much about Emacs when it is just an instance of an application that would benefit from this patch?

Because Ninja's smart terminal feature works perfectly fine in Visual Studio Code or other editor's terminal emulators.

> Furthermore: as stated before, the characteristics of the console associated with the host application are not necessarily related to the characteristics of the display area actually used for displaying ninja's output.

Yes, that's why Emacs should not use its associated console but create a new one for the display area where it displays ninja's output.

oscarfv: > Because Ninja's smart terminal feature works perfectly fine in Visual Studio Code or other editor's terminal emulators.

Visual Studio Code on Windows?

As for other editors, some of them work directly on the Windows Console, mostly the text-based ones, or can show the console for the time it takes to compile the code. But other editors work on a GUI interface and can launch multiple build jobs simultaneously, hence the Windows Console is useless.

> Yes, that's why Emacs should not use its associated console but create a new one for the display area where it displays ninja's output.

A process can have only one console:

https://docs.microsoft.com/en-us/windows/console/allocconsole

Besides, the point is to display Ninja's output on Emacs' own window (not a Windows Console window) for which it must capture said output.

jhasse: > does Visual Studio Code support multiple simultaneous build jobs?

Yes.

jhasse: How do you think Windows Terminal supports multiple tabs if a process can only have one console?

oscarfv: > How do you think Windows Terminal supports multiple tabs if a process can only have one console?

I don't know, I just mentioned the official Win32 API documentation.

oscarfv: > > Visual Studio Code on Windows?
> 
> yes
> 
> > does Visual Studio Code support multiple simultaneous build jobs?
> 
> Yes.

Curious. I thought that Visual Studio Code only supports working on one project at a time, like the classic Visual Studio, which had an output window where the build output is echoed and, in case of running multiple simultaneous tasks, the output of those tasks are mixed.

So Visual Studio Code supports multiple simultaneous builds with separate panes for the output of each job. Good to know.

oscarfv: > How do you think Windows Terminal supports multiple tabs if a process can only have one console?

A quick look indicates that the main process creates a subprocess with its associated console for each tab and then the subprocess' console window is embedded in the tab.

This approach is not useful when the main process wants to capture, process and display Ninja's output by itself.

jhasse: > then the subprocess' console window is embedded in the tab.

The output is captured and the main process renders it.

oscarfv: > > then the subprocess' console window is embedded in the tab.
> 
> The output is captured and the main process renders it.

That requires support for the `CreatePseudoConsole` API, which as per [its documentation](https://docs.microsoft.com/en-us/windows/console/createpseudoconsole) was first introduced in late 2018.

Re-engineering the subprocess I/O handling system to implement a complex and convoluted method for dealing with subprocesses' output using new APIs that would make the host application incompatible with any Windows version more than 4 years old, all that for just handling `\r`, is way overblown, and would bring lots of its own problems.

OTOH we have a simple patch to trick Ninja into doing the same it does in the rest of the systems its supports. What's the problem for not accepting it?

jhasse: CreatePseudoConsole makes it a lot easier, but it's not a requirement.

oscarfv: I took a look at what it would require to use `CreatePseudoConsole` for I/O and it is not easy at all and it would create a lot of issues (what if the subprocess decides to use low-level Console I/O, or some other feature that is specific of the Windows console and hence not supported by existing terminal emulators?)

It is unrealistic to expect that terminal emulators such as the one used by the Cygwin/MSYS2 console, or Emacs to redesign their subprocess I/O system to support a complex mechanism for just providing Ninja with the number of columns it must use in `ElideMiddle`.

Right now I'm trying to understand what's the objection to this PR so we can advance with it.

jhasse: > It is unrealistic to expect that terminal emulators such as the one used by the Cygwin/MSYS2 console, [...]

They don't have to. Run ninja on MSYS2 console like this:

```
winpty ninja
```

and the smart terminal features will work.

oscarfv: > > It is unrealistic to expect that terminal emulators such as the one used by the Cygwin/MSYS2 console, [...]
> 
> They don't have to. Run ninja on MSYS2 console like this:
> 
> ```
> winpty ninja
> ```
> 
> and the smart terminal features will work.

Wouldn't it be nice if we the Cygwin/MSYS2 users had not to type `winpty` in front of `ninja`? On Linux it works right away.

Besides, that's a "solution" for just the Cygwin/MSYS2 terminal emulator. And it can be problematic, because Ninja could enable other current (such as color) or future smart terminal features and there is no guarantee that `winpty` will handle those correctly, because it is unmaintained.

So you object to this PR because it uses a hard-coded value for the number of columns in `ElideMidle`, right? What about my previous suggestion of passing a value on an environment variable?

jhasse: > Wouldn't it be nice if we the Cygwin/MSYS2 users had not to type winpty in front of ninja? On Linux it works right away.

It works right away in the Windows terminal, too. If Linux had a terminal which emulated Windows' console APIs a wrapper would be needed, too.

> What about my previous suggestion of passing a value on an environment variable?

Then we're reinventing how console applications work on Windows. Which might be okay, given how much Microsoft's way sucks, but then it shouldn't be a Ninja-only solution.

Code7R:
@jhasse `winpty` is a quick remedy for the problem but sorry, it does not solve it. ATM one has the option:

- use winpty (and manualy detect TERM_PROGRAM==mintty in all calling scripts) - then the inline progress reporing works, and also the buffering does not suck. BUT: you cannot copy-paste long lines easily (typical compiler errors), since newline characters are injected there
- no winpty, then it acts like on a dumb terminal, you can copy-paste compiler errors... but no compact progress anymore, and stupid buffering (not refreshing for a minute and similar fun)

So both options are unpleasant.

jhasse:
> [...] and stupid buffering (not refreshing for a minute and similar fun)

That's #2136 and a regression. I think we should fix this by flushing after each status line, see my comment in the issue.

PR #2202, opened 2022-10-09
fix self cycle bug when using rule variable twice in command

### LINKED ISSUE #1966: Using a rule variable twice in command is considered as a cycle
To reproduce this:

```
~/ninja-lab
> ls
build.ninja  ninja

~/ninja-lab
> ./ninja --version
1.10.2

~/ninja-lab
> cat build.ninja
rule cat
  rspfile = cat.rsp
  rspfile_content = a b c
  command = cat ${rspfile} ${rspfile} > ${out}

build a: cat

~/ninja-lab
> ./ninja
ninja: fatal: cycle in rule variables: rspfile -> rspfile
```

The error message looks incorrect because `rspfile` doesn't form a self cycle. I'd also expect the build to work, and see `rspfile_content` twice in `${out}`.

I've tried this on other rule variables like `description` and got the same error.

Yogaht:
This PR can help fix #1966 @johan PTAL

### DIFF
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -18,6 +18,8 @@
 #include <deque>
 #include <assert.h>
 #include <stdio.h>
+#include <utility>
+#include <unordered_map>
 
 #include "build_log.h"
 #include "debug_flags.h"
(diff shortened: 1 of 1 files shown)


jhasse:
Thanks! Can you add a unit test for this?

Issue #829, opened 2014-10-13
largefile (LFS) issues


glensc:
looks like ninja build on 32bit system is not largefile safe:

```
ninja: error: stat(build/build.o): Value too large for defined data type
ninja: error: stat(build/depfile_parser.o): Value too large for defined data type
```

the files are residing on 64bit-enabled inode (xfs, inode64 enabled since 2.6.35, see [here](http://xfs.org/index.php/XFS_FAQ#Q:_What_is_the_inode64_mount_option_for.3F)):

```
➔ LC_ALL=C ls -li build/build.o build/depfile_parser.o misc/ninja_syntax.py
4846763977 -rw-r--r-- 2 glen users 790700 Oct 14 00:08 build/build.o
4846763974 -rw-r--r-- 2 glen users  58240 Oct 14 00:08 build/depfile_parser.o
4309918606 -rw-r--r-- 1 glen users   5428 Jun 28 02:37 misc/ninja_syntax.py
```

what's worse, build system does not abort but continues with ignoring such errors:

```
+ ninja -v
ninja: error: stat(misc/ninja_syntax.py): Value too large for defined data type
[1/26] src/inline.sh kBrowsePy < src/browse.py > build/browse_py.h
[2/26] re2c -b -i --no-generation-date -o src/depfile_parser.cc src/depfile_parser.in.cc
[3/26] re2c -b -i --no-generation-date -o src/lexer.cc src/lexer.in.cc
...
[13/26] ccache i686-pld-linux-g++ -MMD -MT build/eval_env.o -MF build/eval_env.o.d -g -Wall -Wextra -Wno-deprecated -Wno-unused-parameter -fno-rtti -fno-exceptions -fvisibility=hidden -pipe -Wno-missing-field-initializers '-DNINJA_PYTHON="python"' -O2 -DNDEBUG -DUSE_PPOLL -O2 -fwrapv -pipe -Wformat -Werror=format-security -gdwarf-4 -fno-debug-types-section -fvar-tracking-assignments -g2 -Wp,-D_FORTIFY_SOURCE=2 -fstack-protector --param=ssp-buffer-size=4 -fomit-frame-pointer -march=i686 -fasynchronous-unwind-tables -mtune=pentium4 -gdwarf-4 -fno-debug-types-section -fvar-tracking-assignments -g2  -c src/eval_env.cc -o build/eval_env.o
ninja: error: stat(build/build.o): Value too large for defined data type  
[...]

PR #2203, opened 2022-10-09
support 32bit system like i386 (#829)

imaiguo:
on debian we can create i386 env & recurring problems

create env on amd64:
pbuilder create --mirror http://deb.debian.org/debian --architecture i386   --distribution sid
pbuilder --login

then build ninja on logined chroot mount dir

code for fit stat errors for Value too large ...


UTsweetyfish:
Since 1.12.0 has not released yet, sbuild users could use mount option `xino=off` to force overlay filesystem use 32-bit inode.

Issue #2169, opened 2022-07-19
Add ARM64/ARM precompiled binaries in releases assets on GitHub.


ww898:
Hi, could you please add some additional precompiled binaries for more platforms such as Windows arm64, Linux aarch64 (Glibc / MUSL)?

jonesmz:
Implementing this would need to use a cross compiler. Or build resources donates from outside of githubs actions system.

Anyone wanting to work on this should consider looking at my pull request that tries to modernize the ci system for ninja to provide some inspiration : https://github.com/ninja-build/ninja/pull/2109

jonesmz:
Related / duplicate https://github.com/ninja-build/ninja/issues/2061

martin-g:
I could help with this !
Let me try it!

martin-g:
Linux ARM64 binaries: https://github.com/ninja-build/ninja/pull/2201
I have no idea how to do the same for Windows/MacOS ARM64.

ThadHouse:
It looks like macOS already builds a universal binary, so thats fine. Windows is easy as well, you just pass `-A arm64` to the first cmake command. I can try to PR that to the repo at some point.

PR #2208, opened 2022-10-31
Add build job for Windows Arm64

ThadHouse:
This should be enough to finish https://github.com/ninja-build/ninja/issues/2169 with the next release.


ThadHouse:
I was able to pull the artifact out of my local actions run, and it worked on my windows arm device.

farzonl:
@ThadHouse the windows arm64 build is super easy, however I have a use case where it would be easier if I could just fetch it from  https://github.com/ninja-build/ninja/releases. Is there a plan to update the releases for arm64 binaries?

Issue #2176, opened 2022-08-10
Allow setting the priority of spawned processes


clemenswasser:
This is a feature request for allowing the user to set the priority/niceness of the processes spawned by ninja.
The idea comes from the recently added "Low Priority Builds" feature of MSBuild which they described in a [Blog Post](https://devblogs.microsoft.com/cppblog/msbuild-low-priority-builds-in-visual-studio/).
At least for Windows it should be pretty easy to add one of the [priority classes](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getpriorityclass#return-value) to the [`process_flags` of the `CreateProcessA` call in `subprocess-win32.cc`](https://github.com/ninja-build/ninja/blob/ce700488e01af33bc478bc986e261e306180b993/src/subprocess-win32.cc#L112)

jonesmz:
Related : https://github.com/ninja-build/ninja/issues/1820

PR #2216, opened 2022-11-17
Allow setting the priority of spawned subprocesses

clemenswasser:
This allows setting the priority/niceness of all spawned subprocesses via a new `-p PRIORITY` parameter.
The flag accepts POSIX style integers (typical range is highest: -20 to lowest: 19), on Windows they get translated into PriorityClasses.

Fixes: #2176


clemenswasser:
When not specifying a priority, now no default priority is set. With the previous commit, a ninja process with a lower priority would try to spawn processes with the default priority of 0.

giraldeau:
I just tested this branch on Linux and I can say that this patch does the work and looks good.

Setting low priority to build processes is great and doing it from ninja makes it easier and portable compared to using some launcher, like nice on Linux.

src/subprocess-posix.cc
```diff
@@ -123,6 +127,12 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
   if (err != 0)
     Fatal("posix_spawn: %s", strerror(err));
 
+  if(priority != INT_MIN) {
+    err = setpriority(PRIO_PROCESS, pid_, priority);
```
digit-google, line 131: there is a race condition where the spawned process may have already completed before this code is run. The call would return an error and this would crash Ninja unexpectedly.

Can you try to set the priority before the posix_spawn() call with posix_spawnattr_setschedparam() instead?

src/subprocess.h
```diff
@@ -52,7 +52,7 @@ struct Subprocess {
 
  private:
   Subprocess(bool use_console);
-  bool Start(struct SubprocessSet* set, const std::string& command);
+  bool Start(struct SubprocessSet* set, const std::string& command, int priority);
```
digit-google, line 55: Please document the meaning of `priority` here (including range of valid values and their interpretation, in particular for Posix _and_ Windows).

src/subprocess-posix.cc
```diff
@@ -123,6 +127,12 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
   if (err != 0)
     Fatal("posix_spawn: %s", strerror(err));
 
+  if(priority != INT_MIN) {
```
digit-google, line 130: why does this check exist? It looks like the only values passed by your code are 0 and 1.

src/subprocess-win32.cc
```diff
@@ -74,7 +74,52 @@ HANDLE Subprocess::SetupPipe(HANDLE ioport) {
   return output_write_child;
 }
 
-bool Subprocess::Start(SubprocessSet* set, const string& command) {
+static DWORD POSIXPriorityToWin32(int unixPriority) {
+  if (unixPriority <= -20)
+    return REALTIME_PRIORITY_CLASS;
```
digit-google, line 79: I don't think this is desirable on Windows. I suggest you leave this out entirely.

src/subprocess-win32.cc
```diff
@@ -74,7 +74,52 @@ HANDLE Subprocess::SetupPipe(HANDLE ioport) {
...
+  case 5:
+  case 6:
+  case 7:
+  case 8:
+  case 9:
+  case 10:
+    return BELOW_NORMAL_PRIORITY_CLASS;
+  default:
+    return IDLE_PRIORITY_CLASS;
+  }
```
digit-google, line 118: Can you change this to use normal range checks instead for readability?

src/ninja.cc
```diff
@@ -223,18 +223,19 @@ void Usage(const BuildConfig& config) {
...
 "\n"
-"  -j N     run N jobs in parallel (0 means infinity) [default=%d on this system]\n"
-"  -k N     keep going until N jobs fail (0 means infinity) [default=1]\n"
-"  -l N     do not start new jobs if the load average is greater than N\n"
-"  -n       dry run (don't run commands but act like they succeeded)\n"
+"  -j N           run N jobs in parallel (0 means infinity) [default=%d on this system]\n"
+"  -k N           keep going until N jobs fail (0 means infinity) [default=1]\n"
+"  -l N           do not start new jobs if the load average is greater than N\n"
+"  -n             dry run (don't run commands but act like they succeeded)\n"
+"  -p PRIORITY    set the priority/niceness of spawned processes\n"
```
digit-google, line 233: Please document possible values and their meaning + perform range checks in the parsing code below.

digit-google:
Thank you, this looks very promising :-)

src/ninja.cc
danw: Isn't this highly dependent on your scheduler implementation? On Linux this varies based on the scheduling policy -- in Android builds we run ninja under `SCHED_BATCH`, which doesn't allow changing this.

It looks like the range checks would need to use `sched_get_priority_min` and `sched_get_priority_max` to check the current policy.

src/subprocess-posix.cc
clemenswasser: Somehow, using posix_spawnattr_setschedparam doesn't seem to work. When checking in htop I don't see any change in niceness of the spawned processes.

src/subprocess.h
clemenswasser: Is the documentation I added sufficient?

src/subprocess-win32.cc
clemenswasser: Done

src/subprocess-win32.cc
clemenswasser: Done

src/ninja.cc
clemenswasser: Since this is so platform and scheduler dependent, should I still do the range checks?

src/subprocess-posix.cc
```diff
@@ -75,6 +79,14 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
   if (err != 0)
     Fatal("posix_spawnattr_init: %s", strerror(err));
 
+  if(priority != INT_MIN) {
+    struct sched_param sp{};
+    sp.sched_priority = priority;
+    err = posix_spawnattr_setschedparam(&attr, &sp);
+    if(err != 0)
+      Fatal("setpriority: %s", strerror(errno));
+  }
```
digit-google, line 88: This is correct, but I believe this will be ignored if you do not add `POSIX_SPAWN_SETSCHEDPARAM` to the `flags` value below.

clemenswasser: In the latest commit, I set the POSIX_SPAWN_SETSCHEDPARAM flag, but when setting any priority other than 0 I get `fatal: posix_spawn: Invalid argument`

digit-google: Ah. Maybe initializing the value by calling pthread_getschedparam() instead of zero-ing the struct will solve it?

clemenswasser: Sadly, this doesn't fix it 😢 I've now also looked into the man pages for some of these functions but didn't find anything obviously wrong with my implementation...

src/subprocess.h
digit-google: Yes, thank you.

src/subprocess-posix.cc
digit-google: I do not see this latest commit you are referencing. Can you upload a new version of this commit that sets the flag properly and initialized the params structure with pthread_getschedparam() so we can have a look at it?

Otherwise, the GLibc sources for posix_spawn() do not seem to return EINVAL explicitly, which suggest that one system call is returning this value. 

This maybe the call to `__sched_setparam()` in that source file, which according to the manpage should return EINVAL only if "the argument param does not make sense              for the current scheduling policy" !?

[1] https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/spawni.c;h=e8ed2babb9b1396903466f631da0571504896aa5;hb=143ef68b2aded7c794956beddad495af8c7d3251

[2] https://man7.org/linux/man-pages/man2/sched_setparam.2.html

clemenswasser: Yeah, sorry, forgot to push the change. I've now debugged the execution with rr and stepped back to where it returned `EINVAL`. It's coming from here: __pthread_setcancelstate https://sourceware.org/git/?p=glibc.git;a=blob;f=nptl/pthread_setcancelstate.c;h=18fb42a5c4538678358f744b5f356eca308d8401;hb=HEAD#l29
Also for me the sched_param struct has only the sched_priority field, so using posix_spawnattr_getschedparam first doesn't change anything.

digit-google: Weird, as far as I see, `pthread_setcancelstate()` does not modify errno, and the sources I linked to seem to call `__pthread_setcancelstate()` and never look at the returned value (and the argument seems correct, and would not trigger this EINVAL). Are you sure this is where the return error value is coming?

That doesn't make much sense, this there is no connection to scheduling parameters here at all, so this is extra puzzling :-/

digit-google: I have downloaded your PR and fixed it a little (i.e. you should call `pthread_getschedparam()` not `posix_spawnattr_setschedparam()` to get the current thread's scheduler parameters, then the SubprocessSet.ConsoleTest unit-test needs to have its SubprocessSet::Add() call adjusted for the new `priority` parameter).

I confirm that using a non-0 priority results in an EINVAL error when calling the `sched_setparam()` system call, as seen using `strace -f build-cmake/ninja_test --gtest_filter=SubprocessSet.ConsoleTest` which ends up with:

```
[pid 2490739] rt_sigaction(SIGRT_31, {sa_handler=SIG_DFL, sa_mask=[], sa_flags=SA_RESTORER, sa_restorer=0x7fcc5f97b510}, NULL, 8) = 0
[pid 2490739] rt_sigaction(SIGRT_32, NULL, {sa_handler=SIG_DFL, sa_mask=[], sa_flags=0}, 8) = 0
[pid 2490739] rt_sigaction(SIGRT_32, {sa_handler=SIG_DFL, sa_mask=[], sa_flags=SA_RESTORER, sa_restorer=0x7fcc5f97b510}, NULL, 8) = 0
[pid 2490739] sched_setparam(0, [1])    = -1 EINVAL (Invalid argument)
[pid 2490739] exit_group(127)           = ?
[pid 2490738] <... clone3 resumed>)     = 2490739
[pid 2490739] +++ exited with 127 +++
wait4(2490739, NULL, 0, NULL)           = 2490739
munmap(0x7fcc5ff15000, 36864)           = 0
rt_sigprocmask(SIG_SETMASK, [HUP INT TERM], NULL, 8) = 0
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=2490739, si_uid=28587, si_status=127, si_utime=0, si_stime=0} ---
write(2, "ninja: fatal: ", 14ninja: fatal: )          = 14
write(2, "posix_spawn: Invalid argument", 29posix_spawn: Invalid argument) = 29
```
So it looks like the kernel is complaining when `priority == 1`.

digit-google: Oh, the manpage for sched_get_priority_min() says:

```
Linux allows the static priority value range 1 to 99 for SCHED_FIFO and SCHED_RR and the priority 0 for SCHED_OTHER and SCHED_BATCH. Scheduling priority ranges for the various policies are not alterable.
```

So it looks like 0 is the only valid value for SCHED_OTHER or SCHED_BATCH. I think the solution is simply to restrict the priority passed to the spawn call to the valid range for the current policy in the code. I.e. something like:

```
    int policy = -1;
    err = pthread_getschedparam(pthread_self(), &policy, &sp);
    if (err != 0)
      Fatal("pthread_getschedparam: %s", strerror(err));

   int min_priority = sched_get_priority_min(policy);
   int max_priority = sched_get_priority_max(policy);
   if (priority < min_priority)
       priority = min_priority;
   else if (priority > max_priority)
       priority = max_priority;
```

With this, all tests pass, though I assume the priority value is essentially ignored as it is always clamped to 0 :-/

clemenswasser: Yeah, you are right __pthread_setcancelstate returns EINVAL, but isn't the actual cause for the error of posix_spawn. This seems to happen in the clone3 syscall.

clemenswasser: But if we can't set the priority to something other than 0, why can I set the priority to other values with the nice tool (GNU coreutils)?

digit-google: So in conclusion, there is a drastic difference between `setpriority()` and `sched_setparam()`, and I'm not sure to know why. This Linux source file seems to indicate that the kernel scheduler differentiates between the two:

https://github.com/torvalds/linux/blob/dccb07f2914cdab2ac3a5b6c98406f765acab803/kernel/sched/core.c#L7935

So I guess we should forget about `spawn_attr_set_schedparam()`, sorry for giving you this idea. I recommend you use `getpriority() / setpriority()` before the spawn() to save then change the niceness of the current (Ninja process) before the fork (so it is inherited by the child one), then call `setpriority()` just after that to restore the original niceness. And add one big comment to explain the reason!

What do you think?

clemenswasser:
Ok, I've now reverted to using `setpriority` but ignore `ESRCH` as it could be returned when the process exited before the priority could be set and therefore avoids the previously mentioned race window.

src/subprocess-posix.cc
clemenswasser: I've implemented that, but somehow restoring the previous priority after the posix_spawn results in a "Permission denied" error :cry:

digit-google:
Thank you, thank you for your patience. The latest commit (833033fee228daca03014d9c41564a0311a46d94) looks good to me. @jhasse, wdyt?

jhasse:
LGTM, I'll have a close look in a few months, but in general this is a good feature for 1.13.0.

For the CI failure the package that provides `ps` would need to be installed in the Fedora Docker.

giraldeau:
I did test the feature on Linux. For me the main goal is to launch build jobs in the background, such that my desktop apps are responsive, and it works fine. I don't see a use case for increasing the priority, but I tried it nontheless (requires admin privilege) and it works. I was worried that maybe the minus sign for the negative number could confuse the argument parser, but it's not the case and everything is fine. 

Thanks!

```
# launch ninja with priority 99 and -99
# properly parsed and clamped
ps -eo ni,pid,cmd
 0  461034 ninja -p 99
19  461035 [sh] <defunct>
19  461036 [sh] <defunct>
19  461038 [sh] <defunct>
19  461040 [sh] <defunct>

  0  461466 ninja -p -99
-20  461467 [sh] <defunct>
-20  461470 [sh] <defunct>
-20  461471 [sh] <defunct>
```

giraldeau:
Tests are failing on Fedora because of missing utility `/bin/ps`. 

```
[ RUN      ] SubprocessTest.Priority
/__w/ninja/ninja/src/subprocess_test.cc:172: Failure
Expected equality of these values:
  " NI\n 10\n"
  subproc->GetOutput()
    Which is: "/bin/sh: line 1: ps: command not found\n"
With diff:
@@ -1,2 +1,1 @@
- NI
- 10\n
+/bin/sh: line 1: ps: command not found\n
[  FAILED  ] SubprocessTest.Priority (1 ms)
```

We need to install the package `procps-ng` in the build image, or maybe change the test such that it works without using ps.

jhasse:
add the package here: https://github.com/ninja-build/ninja/blob/dcefb838534a56b2621baf33c7e0583ff63ee35f/.github/workflows/linux.yml#L17

clemenswasser:
I've now opted to use the `/proc` fs directly. procps-ng uses the same anyway, and with that we don't require a new build time dependency.

mpratt14:
I'd like to give `posix_spawnattr_setschedparam()` another try...

giraldeau:
> I'd like to give `posix_spawnattr_setschedparam()` another try...


pthread_setschedparam() is not intended for normal process, but for real-time and other secialized scheduling algorithm. To change niceness for normal process, we should use setpriority() system call.

giraldeau:
> I'd like to give `posix_spawnattr_setschedparam()` another try...

As indicated earlier, this function purpose is not intuitively what the name suggests. This is for realtime scheduling. This is not intended to change the niceness of a process. 

The branch must be rebased to solve the conflicts, but it could be nice (no pun intended) to merge that feature.

giraldeau:
Actually, the main reason of this feature is to decrease the priority of compilation such that the computer stays responsive for interactive usage. I don't see a situation when a user would want to increase the priority of the compilation.

Therefore, I would change the option to "--idle-priority" or "-i". When this switch is enabled, we would set the lower priority to the main ninja process at launch (and possibly the I/O priority too). On Linux and Windows, the lower priority is inherited by the subprocess, so this change is very minimal. 

This is the equivalent of this bash alias on Linux:

```
alias ninja='nice -n 19 ionice -c 3 ninja'
```

giraldeau:
See MR #2708 

The implementation is trivial and it does the job.

PR #2217, opened 2022-11-22
have implicit "#pragma once" when including ninja files

RisingLiberty:
I added sharpmake support for ninja, but the fact that ninja includes files no matter what was causing some annoying bugs
adding this small change, which I don't think would cause any issues with existing code will fix that and allows Sharpmake to generate ninja files now


orenbenkiki:
This is getting towards the bottom of the page whe looking at open pull requests - does this mean that the provided features are of no interest, or that the implementation isn't what you'd like to see?

PR #2219, opened 2022-12-04
fix garbled error message, like Chinese in Windows

Siogian:


Siogian:
Closes #2220

juzi5201314:
This solves my problem. What is the problem that hinders this PR?

PR #2224, opened 2022-12-10
Add overridable parameter for name of Python binary

wjk:
On macOS, Python 2 is no longer available. As such, the `python` command does not exist in the operating system. On macOS with the developer tools installed, Python 3 is available under the command `python3`. However, ninja currently hard-codes the name `python`. This breaks my build on macOS. From what I can tell, `browse.py` appears to be written to run correctly on Python 3. Therefore, I need to introduce this parameter so I can override it with the command `python3` on the CMake command line.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,8 +90,11 @@ function(check_platform_supports_browse_mode RESULT)
 
 endfunction()
 
-if(NOT NINJA_PYTHON)
-	set(NINJA_PYTHON python)
+option(USE_PYTHON3 "Use python3 command to run browse.py" OFF)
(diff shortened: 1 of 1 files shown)


wjk:
Note also that `configure.py` exhibits the same bug.

jhasse:
Please use CMake's [option](https://cmake.org/cmake/help/latest/command/option.html) instead.

jhasse:
Ah sorry, forgot how much CMake sucks and that option doesn't support strings. This should be the way to go:

```cmake
set(NINJA_PYTHON "python" CACHE STRING "Python interpreter to use for the browse tool")
```

PR #2225, opened 2022-12-12
Add longPathAware manifest to enable long paths on Windows

cristianadam:
Fixes: #1900 when the registry key `Computer\HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\FileSystem\LongPathsEnabled (Type: REG_DWORD)` is set to `1`

See https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation for more details.

### DIFF
--- a/windows/ninja.manifest
+++ b/windows/ninja.manifest
@@ -3,6 +3,7 @@
   <application>
     <windowsSettings>
       <activeCodePage xmlns="http://schemas.microsoft.com/SMI/2019/WindowsSettings">UTF-8</activeCodePage>
+      <longPathAware  xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">true</longPathAware>
     </windowsSettings>
   </application>
 </assembly>


jhasse:
wait a minute ... wasn't this supposed to only work with the unicode WinAPI functions ending with W? We're using the *A functions. I remember testing this a few years ago and it didn't have any effect. Has something changed?

cristianadam:
> wait a minute ... wasn't this supposed to only work with the unicode WinAPI functions ending with W? We're using the *A functions. I remember testing this a few years ago and it didn't have any effect. Has something changed?

Above you have my test case. With the manifest entry and the registry key `ninja` was able to read the `build.ninja` file. Without it failed to find the file.

The same thing works also for MinGW GCC https://gcc.gnu.org/bugzilla/show_bug.cgi?id=107974 
Visual C++ doesn't like it, as seen at https://developercommunity.visualstudio.com/t/compiler-cant-find-source-file-in-path/10221576 

Ninja is the missing part from having a working setup with GCC or Clang on Windows.

PR #2227, opened 2022-12-17
Changed deps tool output format to make syntax

stnuessl:
The new output format contains the same information and is almost exactly the same as before but it adheres to the makefile syntax which increases its operability. This change allows the generated output to be directly used in makefiles or
other tools capable of dealing with makefile syntax.


stnuessl:
Also check the output of clang-scan-deps which uses the makefile dependency file syntax. 
For some reason ninja chose to deviate slightly from that.

PR #2231, opened 2022-12-25
Fix apostrophe in printed output

asashour:
I should be better to use ' instead of `, to match the closing one.


jhasse:
This was done intentionally to match Make's output exactly.

PR #2236, opened 2023-01-14
Replace virtual with override everywhere

orgads:


jhasse:
Thanks for the PR. We don't want to do such hole repo cleanups though. We'd rather want to use override for new code but leave everything else untouched to e.g. retain Git line history.

PR #2237, opened 2023-01-14
Use inline member initialization and default for ctor/dtors

orgads:


jhasse:
Thanks, but we don't want to modernize the existing code right now.

orgads:
Why not?

jhasse:
Why would we? Any code change can be a regression and increases Git line history. So the question is never "why not?" but always "why?".

src/status.h
```diff
@@ -88,7 +88,7 @@ struct StatusPrinter : Status {
   }
 
   struct SlidingRateInfo {
-    SlidingRateInfo(int n) : rate_(-1), N(n), last_update_(-1) {}
```
orgads, line 91: `rate` here is initialized with int, while it's a double.

orgads (commented):
Found 1 missing initialization, and one with an implicit cast.

orgads:
> > Prevents missed initializations
> 
> Theoretically or did you actually find one? This should be covered by unit tests + valgrind anyway.

Yes, I found one. That was my initiative for this change.

> Big con btw:
> 
> * Values are moved from cc to h so changes result in bigger recompiles.

Init values are hardly ever changed, so this shouldn't have any impact in real-life scenarios.

> Most of the changes to Ninja are bugfixes for real world issues where I (or the Owners) care.

Does anybody else maintain this project? You're the only one I find here.

PR #2239, opened 2023-01-15
Support newlines and carriage returns in variable bindings

Colecf:
This PR adds a new escape sequence, $|, to represent a newline in a ninja variable binding. In addition, it also allows carriage returns in variable bindings, which previously gave a lexing error.

It's reasonably common for build systems to want to write out files during the build. The easiest way to do that with ninja would be to make a rule that has an rspfile with the desired content, and then moves the rspfile to the destination location. However, currently, the rspfile can't contain newlines, and thus some form of escaping needs to be done, and the rspfile will need to be copied, doing the unescaping during the copy, instead of moved. This hurts build performance. An alternative implementation would be to use echo commands to write the file, but that can also have slight performance penalties due to needing to split up long command lines into multiple echo commands, and merge the resulting files afterwards. Using echo commands also requires more sophisticated shell escaping than just ninja escaping.

After this PR, it's possible to use any byte except for a null (0) byte in a variable binding. Newlines and dollars must be escaped.

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -831,6 +831,9 @@ paths, where a space would otherwise separate filenames.  See below.)
 `$:` :: a colon.  (This is only necessary in `build` lines, where a colon
 would otherwise terminate the list of outputs.)
 
+`$|`:: a newline.  (This becomes a literal newline in the variable
+assignment.)
+
 `$$`:: a literal `$`.
 
 A `build` or `default` statement is first parsed as a space-separated
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -813,7 +813,11 @@ Comments begin with `#` and extend to the end of the line.
 
 Newlines are significant.  Statements like `build foo bar` are a set
 of space-separated tokens that end at the newline.  Newlines and
-spaces within a token must be escaped.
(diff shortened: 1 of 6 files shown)


doc/manual.asciidoc
```diff
@@ -821,6 +821,9 @@ behaviors:
 `$` followed by a newline:: escape the newline (continue the current line
 across a line break).
 
+`$|`:: a newline. Can be used to put literal newlines into ninja variables
```
lamontj, line 828: I think this belongs between `$:` and `$$` below, rather than here.  To be consistent with the other escapes, maybe something like:

`$|`:: a newline.  (This becomes a literal newline in the variable assignment.)

lamontj, line 828: I think this belongs between `$:` and `$$` below, rather than here. To be consistent with the other escapes, maybe something like:

```
`$|`:: a newline. (This becomes a literal newline in the variable assignment.)
```

Colecf: Done

lamontj (approved):
LGTM

jhasse:
Sorry I forgot about this and merged #2606 now. Does that cover your use-cases?

PR #2240, opened 2023-01-17
Increment ninja_log version

Jwata:
The timestamp format changed in https://github.com/ninja-build/ninja/pull/1219 without incrementing the log version number.

This mismatch caused a bug in Chromium that Ninja mis-recognizes some targets fresh, and skips them unexpectedly. 
https://crbug.com/1406628 

Changes 
- Increment `kCurrentVersion`
- Increment `kOldestSupportedVersion` as the current format is not compatible with the version 5.
- Fix `bulid_log_test.cc`

### DIFF
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -344,14 +344,9 @@ LoadStatus BuildLog::Load(const string& path, string* err) {
     entry->start_time = start_time;
     entry->end_time = end_time;
     entry->mtime = mtime;
-    if (log_version >= 5) {
-      char c = *end; *end = '\0';
-      entry->command_hash = (uint64_t)strtoull(start, NULL, 16);
-      *end = c;
-    } else {
-      entry->command_hash = LogEntry::HashCommand(StringPiece(start,
-                                                              end - start));
-    }
+    char c = *end; *end = '\0';
+    entry->command_hash = (uint64_t)strtoull(start, NULL, 16);
+    *end = c;
   }
   fclose(file);
 
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -279,11 +279,16 @@ LoadStatus BuildLog::Load(const string& path, string* err) {
     if (!log_version) {
       sscanf(line_start, kFileSignature, &log_version);
 
-      if (log_version < kOldestSupportedVersion ||
-          log_version > kCurrentVersion) {
-        *err =
-            ("build log version invalid, perhaps due to being too old or "
-             "too new; starting over");
+      bool invalid_log_version = false;
+      if (log_version < kOldestSupportedVersion) {
+        invalid_log_version = true;
+        *err = "build log version is too old; starting over";
+
+      } else if (log_version > kCurrentVersion) {
+        invalid_log_version = true;
+        *err = "build log version is too new; starting over";
+      }
+      if (invalid_log_version) {
         fclose(file);
         unlink(path.c_str());
         // Don't report this as a failure.  An empty build log will cause
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -279,9 +279,11 @@ LoadStatus BuildLog::Load(const string& path, string* err) {
     if (!log_version) {
       sscanf(line_start, kFileSignature, &log_version);
 
-      if (log_version < kOldestSupportedVersion) {
-        *err = ("build log version invalid, perhaps due to being too old; "
-                "starting over");
+      if (log_version < kOldestSupportedVersion ||
+          log_version > kCurrentVersion) {
+        *err =
+            ("build log version invalid, perhaps due to being too old or "
+             "too new; starting over");
         fclose(file);
         unlink(path.c_str());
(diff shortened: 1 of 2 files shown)


jhasse:
oh not good. Isn't this already too late though? Hires timestamps were introduced in 1.9.0.

Jwata:
Chromium had been using v1.8 until last month...

So some developers may build against an old revision with v1.8.
If that happens, their build dir get corrupted, and it would be difficult to debug.

jhasse:
@atetubou @digit-google @nico Can one of you share your thoughts on this?

digit-google:
I think this change is fine. I.e. looking at the code, my understanding of the various formats being used is that:

- Format 4 used a 32-bit restat low-resolution decimal, written as a decimal, and copied the build command as-is in each log entry.

- Format 5 changed the entry format by storing a 64-bit hexadecimal hash for the command instead.

- The change in #1219 changed to a 64-bit high-resolution restat decimal timestamp. Also there is no code left to convert an old 32-bit low-resolution timestamp into the corresponding 64-bit high-resolution value, so bumping the minimum kOldestSupportedVersion value here to 6 seems to be necessary.

Wondering what will happen if an older version of Ninja is being invoked with a build log in format 6 (which could happen when performing an incremental build after checking out an older Chromium source revision, for example while bisecting)?

It looks like the old code will consider that the format is valid (i.e. build_log.cc only checks for `log_version < kOldestSupportedVersion`), and will likely read invalid timestamps with atol(), the worse that can happen is rebuilding too many things iiuc, which seems acceptable here.

This change might be an opportunity to remove the special code branch to deal with format 4 on build_log.cc:352 though since it will now be dead.

Jwata:
> Wondering what will happen if an older version of Ninja is being invoked with a build log in format 6 (which could happen when performing an incremental build after checking out an older Chromium source revision, for example while bisecting)?
> It looks like the old code will consider that the format is valid (i.e. build_log.cc only checks for log_version < kOldestSupportedVersion), and will likely read invalid timestamps with atol(), the worse that can happen is rebuilding too many things iiuc, which seems acceptable here.

I think this case happened to us, and caused an incident. 
The situation was worse than rebuilding too many things. It skipped build steps unexpectedly maybe because the timestamps made Ninja to recognize some files as updated. 

I realized that this PR doesn't prevent this case. 
But at least, using a different ninja log version in the latest Ninja helps people debug this type of issue when happened.

> This change might be an opportunity to remove the special code branch to deal with format 4 on build_log.cc:352 though since it will now be dead.

Removed the code.

digit-google:
> I think this case happened to us, and caused an incident. The situation was worse than rebuilding too many things. It skipped build steps unexpectedly maybe because the timestamps made Ninja to recognize some files as updated.
> 
Oh, this is pretty bad :-(

> I realized that this PR doesn't prevent this case. But at least, using a different ninja log version in the latest Ninja helps people debug this type of issue when happened.
> 
This PR cannot fix the issue which is in the older Ninja source base, but we may want to update the current one to avoid it in the future. I am thinking about changing the checks in build_log.cc:282 from:

```
      if (log_version < kOldestSupportedVersion) {
        *err = ("build log version invalid, perhaps due to being too old; "
                "starting over");
         ...
      }
```

To something like:

```
      if (log_version < kOldestSupportedVersion || log_version > kCurrentVersion) {
        *err = ("build log version invalid, perhaps due to being too old or too new; "
                "starting over");
         ...
      }
```

Wdyt?

> > This change might be an opportunity to remove the special code branch to deal with format 4 on build_log.cc:352 though since it will now be dead.
> 
> Removed the code.

Thanks!

Jwata:
> To something like:
>       if (log_version < kOldestSupportedVersion || log_version > kCurrentVersion) {
>        *err = ("build log version invalid, perhaps due to being too old or too new; "
>                "starting over");
>         ...
>      }
> Wdyt?

Sounds good. Done.

atetubou (approved):
This looks good to me.

src/build_log.cc
```diff
@@ -279,9 +279,11 @@ LoadStatus BuildLog::Load(const string& path, string* err) {
...
       sscanf(line_start, kFileSignature, &log_version);
 
-      if (log_version < kOldestSupportedVersion) {
-        *err = ("build log version invalid, perhaps due to being too old; "
-                "starting over");
+      if (log_version < kOldestSupportedVersion ||
+          log_version > kCurrentVersion) {
+        *err =
+            ("build log version invalid, perhaps due to being too old or "
+             "too new; starting over");
```
jhasse, line 286: Can you turn this into two error messages? One for too old and one for too new.

Jwata: I updated it. PTAL.

digit-google:
ping, the latest version looks fine to me, can we submit this?

Jwata:
I don't have a permission to submit.
Can anybody do that?

digit-google:
That would be @jhasse I believe, when he has the time.

PR #2244, opened 2023-01-25
CI: semantic version. CMake DRY refactor

scivision:
CI: avoid nuisance warnings about outdated libraries by using semantic versions in *.yml

CMake: DRY (don't repeat yourself) by refactoring to inspect target sources and set CXX property in one place.

CMake: avoid nuisance extra CTest targets and use consistent variable name NINJA_BUILD_TESTING by using enable_testing()

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -13,7 +13,7 @@ jobs:
       image: rockylinux:8
     steps:
     - uses: actions/checkout@v4
-    - uses: codespell-project/actions-codespell@master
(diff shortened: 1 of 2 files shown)


jhasse:
Hm ... isn't `BUILD_TESTING` supposed to be the clean solution to disable one's tests? Introducing our own variable sounds like a bad idea and not like the clean CMake way (not that there is one).

kbdharun (approved):

jhasse:
-> conflicts now

If still needed, open individual PRs for each issue.

PR #2246, opened 2023-01-25
Delete memory that will not be added so it does not leak

AtariDreams:
Otherwise the memory will end up leaked.

### DIFF
--- a/src/manifest_parser.cc
+++ b/src/manifest_parser.cc
@@ -146,8 +146,10 @@ bool ManifestParser::ParseRule(string* err) {
   while (lexer_.PeekToken(Lexer::INDENT)) {
     string key;
     EvalString value;
-    if (!ParseLet(&key, &value, err))
+    if (!ParseLet(&key, &value, err)) {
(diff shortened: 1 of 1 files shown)


src/manifest_parser.cc
```diff
@@ -146,8 +146,10 @@ bool ManifestParser::ParseRule(string* err) {
   while (lexer_.PeekToken(Lexer::INDENT)) {
     string key;
     EvalString value;
-    if (!ParseLet(&key, &value, err))
+    if (!ParseLet(&key, &value, err)) {
+      delete rule;
```
Osyotr, line 150: Why not just replace it with `std::unique_ptr`?

AtariDreams: Seems like overkill, but maybe?

jhasse:
The leaks are intentional as ninja exits soon after anyway. Could be documented better (or maybe wrap the pointers in a type which leaks, but doesn't when building with a special define?).

PR #2247, opened 2023-01-25
Prefer for-each loop over iterators

AtariDreams:


jhasse:
Thanks, but we want to keep Git line history clean and better not touch too much instead of modernizing the code. It's a trade-off.

PR #2248, opened 2023-01-25
Use size_t instead of int where appropriate

AtariDreams:


jhasse:
Thanks, but we don't want to touch these lines for such cleanup changes.

PR #2251, opened 2023-01-30
Clean up Windows workflow and add an x86 build

ndabas:
The [Windows workflow](https://github.com/ninja-build/ninja/blob/master/.github/workflows/windows.yml) has a lot of redundant code to support the arm64 version, which is identical to the default x64 build, but without tests.

This can be cleaned up with a matrix build and conditionals. While in there I added in an x86 build as well, because why not.

The current artifact filename, ninja-win.zip, is preserved for the x64 build. The arm64 build also retains the existing suffix.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -14,8 +14,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-        - arch: 'win32'
-          suffix: 'x86'
         - arch: 'x64'
           suffix: ''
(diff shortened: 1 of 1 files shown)


jhasse:
Since x86 is about its way out, I wouldn't want to add a CI job for it.

ndabas:
Okay, I have removed the x86 config from the job. So now we're just removing duplicated/redundant code from the workflow, with no functional change.

PR #2252, opened 2023-02-02
Order clean by path length to ensure directories can be deleted

mjansson:
My ninja build generator is setup to create intermediate build directories and keep track of these as build outputs in order to properly clean up the intermediate directory structure with the clean ninja target.

It is cumbersome to have to track the directory hierarchy as implicit dependencies, but not doing so results in the clean target not being able to correctly order the deletion, causing directory deletions to fail due to still having files (which will be deleted later by the clean target).

A simple solution is to defer all deletions, sort by path length and perform all deletions starting with the deepest path. That guarantees that any directory is deleted after all the files in that directory.

### DIFF
--- a/src/clean.cc
+++ b/src/clean.cc
@@ -39,6 +39,20 @@ int Cleaner::RemoveFile(const string& path) {
   return disk_interface_->RemoveFile(path);
 }
 
+void Cleaner::RemoveAllPending() {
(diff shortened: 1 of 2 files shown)


jhasse:
Needs a test, otherwise LGTM.

PR #2256, opened 2023-02-15
refactor: NULL => nullptr

scivision:
this was done with IDE find and replace `NULL` => `nullptr`
To me, there were too many lints that had accumulated to use clang-tidy directly.

if this makes a PR conflict, it may only be on a couple lines, and they could likewise do find/replace.


.clang-tidy
```diff
@@ -6,6 +6,7 @@ Checks: '
   ,readability-redundant-string-cstr,
   ,readability-redundant-string-init,
   ,readability-simplify-boolean-expr,
+  ,modernize-use-nullptr
```
atetubou, line 9: sort alphabetically?

digit-google:
This is nice! I would suggest two small things though:

1) Provide clear instructions on how to reproduce this locally, so that developers can easily update their own patches on top of this change once it is submitted.

2) Apply `git clang-format` just after the `clang-tidy` step, and add that in the commit message as well.

For example, on my Debian/testing machine, I could do the following:

```
clang-tidy-14 --checks=-*,modernize-use-nullptr --fix-errors src/*.cc src/*.h -- std=c++11
git clang-format
``` 
Using "-*" at the start of `--checks` is necessary to disable all the checks in the .clang-tidy file, since there are tons of unrelated errors trigerred by the source code if I do not use this.

jhasse:
The change to .clang-tidy seems to be gone?

Also can you edit lexer.cc and disk_interface.cc without the whitespace changes?

PR #2260, opened 2023-02-19
Add GNU make jobserver master support

### LINKED ISSUE #1139: Add GNU make jobserver client support

stefanb2:
Part 2 in the implementation series for PR #1139: with these changes ninja can take the same role as GNU make at the root of a load balanced build.

* add new command line option `-m` / `--tokenpool-master`
* sets up ninja as tokenpool master, i.e. ninja and jobserver clients in the build will cooperate to run at maximun `-j N` jobs in parallel
* ninja will export the `MAKEFLAGS` environment variable to its child processes, which communicates the jobserver parameters to them.

https://www.gnu.org/software/make/manual/html_node/Job-Slots.html#Job-Slots

Fixes #1139


kepstin:
Is there any possibility of having the selection between jobserver and client being done automatically, instead of requiring a command line option? In most cases, I'll want ninja to provide a jobserver by default (for integration with e.g. gcc lto or rustc), and the extra option is easy to forget. It's also extra work to pass the option when ninja is being used as a backend for e.g. cmake or meson.

The behaviour I'm hoping for is:
* If ninja is passed explicit command line options that prevent it from using an external jobserver, it will become a jobserver.
* If ninja detects via `MAKEFLAGS` a jobserver that it's able to connect to, it'll use it as a client.
* Otherwise ninja becomes a jobserver.

stefanb2:
The project has decided to close PR #1140 on which this PR is based on. Therefore this PR can never be merged.

PR #2263, opened 2023-02-22
Add GNU make jobserver style "fifo" support

### LINKED ISSUE #1139: Add GNU make jobserver client support

stefanb2:
Part 3 in the implementation series for PR #1139: add support for the POSIX jobserver style `fifo`, introduced in GNU make 4.4, that uses a named pipe as communication channel between master & clients.

* add an optional argument to the command line option `-m` / `--tokenpool-master` to specify the jobserver style.
  * POSIX supports `fifo` (default) and `pipe`
  * Win32 supports `sem` (default)
  * specifying an unknown style will cause ninja to abort with a fatal error
* update client & server parts of `GNUmakeTokenPoolPosix` to support the `fifo` style
* add tests for `fifo` style

https://www.gnu.org/software/make/manual/html_node/Job-Slots.html#Job-Slots

Fixes #1139


src/build.h
```diff
@@ -147,15 +151,19 @@ struct CommandRunner {
     bool success() const { return status == ExitSuccess; }
   };
   /// Wait for a command to complete, or return false if interrupted.
-  virtual bool WaitForCommand(Result* result) = 0;
+  /// If more_ready is true then the optional TokenPool is monitored too
+  /// and we return when a token becomes available.
+  virtual bool WaitForCommand(Result* result, bool more_ready) = 0;
```
Ext3h, line 156: The naming of  that method is no longer exactly accurate with that extension.

But not just that - something feels fishy about merging these functionalities into a single method. It's two distinct states mangled into one - *either* you are waiting for jobs to complete as you have an empty work-queue *or* you are waiting for for jobs to complete as you need fresh tokens.

It just happens that those two states share a single wait condition when there is no external job server, but they are simply not the same when there is.

So wrapping those two states into distinct, named wait methods would improve readability.

src/tokenpool-gnu-make.cc
```diff
@@ -0,0 +1,140 @@
...
+
+  return false;
+}
+
+bool GNUmakeTokenPool::SetupMaster(bool verbose,
+                                   int parallelism,
+                                   double max_load_average,
+                                   const char* style) {
+  // no need to set up token pool for serial builds
+  if (parallelism == 1)
```
Ext3h, line 79: That's not true - you still need to setup a job server, as it's the responsibility of the job server to communicate the serial nature to child processes.

src/build.cc
```diff
@@ -628,45 +684,54 @@ bool Builder::Build(string* err) {
...
   while (plan_.more_to_do()) {
-    // See if we can start any more commands.
-    if (failures_allowed && command_runner_->CanRunMore()) {
-      if (Edge* edge = plan_.FindWork()) {
-        if (edge->GetBindingBool("generator")) {
+    // See if we can start any more commands...
+    bool can_run_more =
+        failures_allowed   &&
+        plan_.more_ready() &&
+        command_runner_->CanRunMore();
```
Ext3h, line 691: This `CanRunMore()` is kind of wrong.  If there is a token pool, it's the *sole* authority on parallelism.

If you have an `AcquireToken()` now, then you should only be calling `CanRunMore()` as an implementation detail within `AcquireToken`  in the fallback case when a token pool is absent.

src/build.cc
```diff
@@ -628,45 +684,54 @@ bool Builder::Build(string* err) {
...
           result.status == ExitInterrupted) {
         Cleanup();
         status_->BuildFinished();
         *err = "interrupted by user";
         return false;
       }
 
+      // We might be able to start another command; start the main loop over.
+      if (result.status == ExitTokenAvailable)
+        continue;
```
Ext3h, line 733: That's not looking right - you restarted the loop, but you didn't *steal* the token from the already completed job yet. Instead the next iteration will do yet another fresh `AcquireToken()` / `Reserve()` even though you had still held on to an already available token.

src/build.h
Ext3h: Looks like not making that distinction led you to fusing both sets of FDs into one list in `SubprocessSet`.

You didn't need to merge those sets, and thus could just have kept this hidden within `TokenPool`.

src/build.cc
```diff
@@ -471,34 +504,57 @@ vector<Edge*> RealCommandRunner::GetActiveEdges() {
...
+         !subprocs_.IsTokenAvailable()) {
+    bool interrupted = subprocs_.DoWork(more_ready ? tokens_ : NULL);
     if (interrupted)
       return false;
   }
 
+  // token became available
+  if (subproc == NULL) {
+    result->status = ExitTokenAvailable;
+    return true;
```
Ext3h, line 551: This can't be right... `subprocs_.NextFinished()` in that loop above has popped *any* number of completed processes.

But `tokens_->Release()` is only being called *conditionally* and only once at most.

So this is leaving tokens incorrectly marked as "in-use". You would have needed to yield the token together with `finished_.push(*i);`.

src/tokenpool-gnu-make.cc
```diff
@@ -0,0 +1,140 @@
...
+
+void GNUmakeTokenPool::Return() {
+  if (ReturnToken())
+    available_--;
+}
+
+void GNUmakeTokenPool::Release() {
+  available_++;
+  used_--;
+  if (available_ > 1)
```
Ext3h, line 131: You shouldn't do that synchronously on every `Release()`. Return tokens in bulk when you start ***blocking*** on sub-process completion, but before that, keep holding on!

src/build.h
Ext3h: ... not sure if what I assumed is correct. As it's this process which has to return a token in the case of blocking itself, it does need to await both sets at once after..

src/tokenpool-gnu-make.cc
```diff
@@ -0,0 +1,140 @@
...
+                                   double& max_load_average) {
+  const char* value = GetEnv("MAKEFLAGS");
+  if (!value)
+    return false;
+
+  // GNU make <= 4.1
+  const char* jobserver = strstr(value, "--jobserver-fds=");
+  if (!jobserver)
+    // GNU make => 4.2
+    jobserver = strstr(value, "--jobserver-auth=");
```
Ext3h, line 44: Assume `--jobserver-auth=` has precedence over `--jobserver-fds=` please.

src/tokenpool-gnu-make.cc
```diff
@@ -0,0 +1,140 @@
...
+
+void GNUmakeTokenPool::Release() {
+  available_++;
+  used_--;
+  if (available_ > 1)
+    Return();
+}
+
+void GNUmakeTokenPool::Clear() {
+  while (used_ > 0)
```
Ext3h, line 136: If you got it right, you should *always* end up with `used_ == 0` at this point, given that all sub-processes must have been terminated.

src/tokenpool-gnu-make-posix.cc
```diff
@@ -0,0 +1,368 @@
...
+
+    if (dup_rfd_ != -1) {
+      struct sigaction act, old_act;
+      int ret = 0;
+      char buf;
+
+      // Temporarily replace SIGCHLD handler with our own
+      memset(&act, 0, sizeof(act));
+      act.sa_handler = CloseDupRfd;
+      if (sigaction(SIGCHLD, &act, &old_act) == 0) {
```
Ext3h, line 306: Why?... Unlike `make`, `ninja` doesn't monitor the child as a child process via `SIGCHLD`, but merely tracks the pipe opened to it. Meaning it's not even subject to the same race conditions which `make` had to handle.

Ext3h: Ah... this code used to provide non-blocking reads in a legacy portable way, not recognizing `O_NONBLOCK` as a legit way of doing this.

The `SIGCHLD` part is pretty much entirely irrelevant though - it's merely an optimization.

src/tokenpool-gnu-make-posix.cc
```diff
@@ -0,0 +1,368 @@
...
+  // First check if read() would succeed without blocking.
+#ifdef USE_PPOLL
+  pollfd pollfds[] = {{rfd_, POLLIN, 0}};
+  int ret = poll(pollfds, 1, 0);
+#else
+  fd_set set;
+  struct timeval timeout = { 0, 0 };
+  FD_ZERO(&set);
+  FD_SET(rfd_, &set);
+  int ret = select(rfd_ + 1, &set, NULL, NULL, &timeout);
```
Ext3h, line 287: Duplicated code between the code updating `SubprocessSet::token_available_` and this select here.

src/build.cc
```diff
@@ -471,34 +504,57 @@ vector<Edge*> RealCommandRunner::GetActiveEdges() {
...
+  }
+
+  // command completed
+  if (tokens_)
+    tokens_->Release();
```
Ext3h, line 556: This has to be done *after* `result->status = subproc->Finish()` - the process is still running, *only* the pipe was broken yet.

Unrelated - but this is also a spot for optimization, as Ninja will get stalled here if a long-running process detaches from terminal early.

src/build.cc
Ext3h: You might have wanted to put it in the `else` to the above `if`, and put a comment on the loop explaining the non-trivial exit condition.

src/tokenpool-gnu-make-posix.cc
```diff
@@ -0,0 +1,368 @@
...
+  int wfd = open(filename, O_WRONLY);
+  if (wfd < 0) {
+    close(rfd);
+    free(filename);
+    return false;
+  }
+
+  free(filename);
+
+  rfd_ = rfd;
```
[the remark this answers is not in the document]
Ext3h: If you *did*,  then you should have also specified `--jobserver-fds=` for downward compatibility.

src/build.cc
Ext3h: Also the methods used here are definitely overdue for renaming, their names don't reflect what they do.

stefanb2:
The project has decided to close PR #1140 on which this PR is based on. Therefore this PR can never be merged.

Issue #2218, opened 2022-11-28
Use /proc/loadavg on Linux


reddwarf69:
GNU Make uses /proc/loadavg on Linux to decide whether to start another task: https://lwn.net/Articles/913253/ / https://git.savannah.gnu.org/cgit/make.git/commit/src/job.c?id=d8728efc80b720c630e6b12dbd34a3d44e060690. I would like Ninja to do the same.

Since /proc/loadavg provides "number of processes currently runnable (running or on ready queue)" (https://www.kernel.org/doc/html/latest/filesystems/proc.html), Ninja would be able to take better decisions that by looking only at the info provided by getloadavg().

reddwarf69:
To have some numbers, I have a Jenkins slave with 64 cores. When pushing to a repository, multiple jobs are triggered. Those jobs are independent, and there is no easy way to have a common "jobserver" to limit the number of parallel tasks. All the makes/ninjas are run with "-j 64 -l 128".

With GNU Make 4.4 the peak 1 minute average load is 132, and the peak memory usage 55 GiB.
With Ninja the peak 1 minute average load is 562, and the peak memory usage 222 GiB.

reddwarf69:
I'm not creating a PR because I'm not sure about the licence implications of me "basing" it on the GPL make patch. Maybe /proc/loadavg should be kept open the whole time instead of opening and closing in continuously, the parsing could maybe be made more robust, etc.

But this does the trick.

```
--- a/src/util.cc
+++ b/src/util.cc
@@ -818,6 +818,38 @@ double GetLoadAverage() {
 double GetLoadAverage() {
     return -0.0f;
 }
+#elif defined(linux)
+double GetLoadAverage() {
+  int fd = open("/proc/loadavg", O_RDONLY);
+  if(fd == -1) {
+    return -0.0f;
+  }
+
+  char avg[65];
+  int size = read(fd, avg, 64);
+  close(fd);
+  if(size == -1) {
+    return -0.0f;
+  }
+  avg[size] = '\0';
+
+  const char *p = strchr(avg, ' ');
+  if (p) {
+    p = strchr(p+1, ' ');
+  }
+  if (p) {
+    p = strchr(p+1, ' ');
+  }
+  if (!p) {
+    return -0.0f;
+  }
+
+  if((unsigned)p[1] - '0' > 9) {
+    return -0.0f;
+  }
+
+  return atoi(p+1);
+}
 #else
 double GetLoadAverage() {
[...]

PR #2268, opened 2023-03-06
On Linux, try to use /proc/loadavg to get instaneous load (#2218)

entrope:
The /proc/loadavg file includes the traditional 1-, 5- and 15-minute loads plus an instantaneous count of processes that are ready to run. This lets us respond faster to new processes being launched by other processes.


entrope:
This is draft because I am not sure whether having a `static` variable is considered bad form.  The Windows function CalculateProcessorLoad() already uses static variables, so perhaps it is okay here?

src/util.cc
```diff
@@ -825,6 +825,36 @@ double GetLoadAverage() {
...
 #else
 double GetLoadAverage() {
+# if defined(__linux__)
+  // Only try to open the file once.
+  static int fd = -1;
+  if (fd == -1) {
+    fd = open("/proc/loadavg", O_RDONLY);
+    if (fd < 0) {
+      fd = -2;
+    }
```
ptsneves, line 835: Why set this if it is still negative and thus ignored by the `fd >=0` check?

ptsneves (changes requested):

src/util.cc
```diff
@@ -825,6 +825,36 @@ double GetLoadAverage() {
 }
 #else
 double GetLoadAverage() {
+# if defined(__linux__)
+  // Only try to open the file once.
+  static int fd = -1;
```
ptsneves, line 830: what is the reason to make the variable static?

src/util.cc
```diff
@@ -825,6 +825,36 @@ double GetLoadAverage() {
...
+    }
+  }
+
+  if (fd >= 0) {
+    char buf[65], *p;
+    ssize_t nbr = pread(fd, buf, sizeof(buf) - 1, 0);
+    // Load average should be at least "1 5 9 n/".
+    if (nbr < 8) {
+      return -0.0f;
+    }
```
ptsneves, line 844: If `nbr < 8` we should just assume there is something wrong and let the non linux code try to get the load average.

src/util.cc
```diff
@@ -825,6 +825,36 @@ double GetLoadAverage() {
...
+    p = strchr(buf, ' ');
+    if (p) {
+      p = strchr(p + 1, ' ');
+    }
+    if (p) {
+      p = strchr(p + 1, ' ');
+    }
+    if (p && (unsigned(p[1]) - '0' <= 9)) {
```
ptsneves, line 852: how do you know `p+1 != NULL`?

ptsneves (changes requested):

src/util.cc
entrope: "fd" is retained across calls.  Setting it to -2 means we already tried to open() it once and failed for whatever reason -- so we should fall back to  the original behavior.

src/util.cc
entrope: This reduces the number of system calls per call to this function: Rather than open(), read(), close() every time, we open() once and then read() each following time.  Caching the file descriptor seemed appropriate in this case.

src/util.cc
entrope: Agreed, I will change this.

src/util.cc
entrope: Would it be sufficient to set buf[nbr] = '\0' after checking that nbr is large enough?  p+1 == NULL only if the kernel gives us a corrupted buffer from pread() and the stack happened to be near the top of the address stack and nothing higher on the stack was either '\0' or ' '.

But if the kernel gives us a corrupted buffer, we could read past the end of buf[], so it is safer to explicitly stick a '\0' at the end.  Then we know that at this point, (p == NULL) || (p[0] == ' ' and p[0] has at least '\0' afterwards).

src/util.cc
ptsneves: Oh I see. This is a cool trick. I am fine with this

src/util.cc
entrope: Updated in 367fbf8d8.

Flowdalic:
This changes the API of ninja's `double GetLoadAverage()` on Linux to return the currently running number of kernel-level threads (tasks) instead of the damped 1-minute average.

Frankly, besides me being skeptical that this function should return a different value depending on the operating system, using the instantaneous value also has drawbacks. The major one is that it, essentially being the value returned by Linux's `nr_running()`, does not account for uninterruptible tasks, i.e., tasks that are likely waiting for an I/O-operation to complete, while those are included in the loadaverage (on Linux). You can read more about it in Brendan Gregg's excellent blog post: https://www.brendangregg.com/blog/2017-08-08/linux-load-averages.html

This poses the risk of overloading the system as those I/O-waiting tasks are likely to become runnable again.

And yes, if they become not runnable again soon, then this would mean that no new ninja jobs are started, even though there are spare CPU resources. However, in this case, the system is likely already I/O-bound and starting to compilation jobs would be counterproductive anyways.

In summary, I am not sure if this change is sensible.

entrope:
The underlying problem this was meant to address (#2218) is actually the opposite: The current approach takes too long to notice when lots of processes are creating, leading to CPU oversubscription and potentially VM (and thus swap -> IO) oversubscription.  Do you propose a different way to solve the same problem?

Flowdalic:
> Do you propose a different way to solve the same problem?

Yes, see https://github.com/ninja-build/ninja/pull/1949, which got recently merged.

cmorve-te:
Flowdalic makes valid points.

But using ninja 1.12.1 with "-l 20", in an otherwise idle system, I have had 126 running tasks. For 19 consecutive seconds the number of running tasks was over 40.

I didn't double check. But I think what happened is that a bunch of parallel LTO linking started, so the load increased very fast. https://github.com/ninja-build/ninja/pull/1949 says

> We assume that every started edge increases the load by one

but in this case the started edges increased the load by a lot more (the number of CPUs in the system, each, I guess).

More in general, even with https://github.com/ninja-build/ninja/pull/1949, without this patch we have the problem that we may not notice quick load increases until it's too late.

Naturally, this is only a big problem when the tasks started by ninja happen to also increase the load a lot. And, once Ninja start to support the GNU Make jobserver, those parallel LTO will not increase the load that much.

Flowdalic:
You also make very good points. I would still a bit cautious to change the default behavior though. Mostly because of the different accounting of handling uninterruptible tasks between the 1-min load average and . That said, I It may be nice to be able to select the used "load" value via a command-line argument and get some experience with it.


> And, once Ninja start to support the GNU Make jobserver, those parallel LTO will not increase the load that much.

Could you elaborate why that would be the case? Does ninja obtain multiple job slots from the jobserver when starting a parallel LTO job? If not, then on a N-core machine, using `--jobs N`, ninja may as well start N parallel LTO tasks, each creating N kernel-level threads, leading to an overload situation with an increased number of involuntary context switches.

cmorve-te:
> It may be nice to be able to select the used "load" value via a command-line argument and get some experience with it.

Not sure how I would call that argument. But yeah, it would be nice to have.


> Could you elaborate why that would be the case? Does ninja obtain multiple job slots from the jobserver when starting a parallel LTO job? If not, then on a N-core machine, using --jobs N, ninja may as well start N parallel LTO tasks, each creating N kernel-level threads, leading to an overload situation with an increased number of involuntary context switches.

The option has grown quite a bit, but it's explained in https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-flto


> You can also specify -flto=jobserver to use GNU make’s job server mode to determine the number of parallel jobs. This is useful when the Makefile calling GCC is already executing in parallel. You must prepend a ‘+’ to the command recipe in the parent Makefile for this to work. This option likely only works if MAKE is GNU make. Even without the option value, GCC tries to automatically detect a running GNU make’s job server.
> 
> Use -flto=auto to use GNU make’s job server, if available, or otherwise fall back to autodetection of the number of CPU threads present in your system. 

So, once ninja can become a GNU make's job server, gcc's LTO will not create N kernel-level threads itself (unless it's the only remaining task).

AJIOB:
Is this PR ready for merge?

Or we're waiting for any of discussion results?

PR #2273, opened 2023-03-19
Fixed manual layout

Volker-Weissmann:
https://ninja-build.org/manual.html currently looks like this:
image

This looks broken. This PR should fix it.


jhasse:
fixed by #2282

PR #2281, opened 2023-03-28
emacs: improve ninja-mode

Hi-Angel:
1st commit fixes a minor visual bug where inside variable declaration like `foo = bar = buzz` the `bar` was highlighted as a variable even though it's not, the complete `bar = buzz` part is just a string that gets assigned to `foo`, which is the only variable there.

2nd commit adds support for indentation. I have successfully used this indentation support while writing and editing a `build.ninja` manually. Additionally, the support is tested by re-indenting two `build.ninja` files, one is generated by Meson *(1790 lines)*, and another by CMake *(7777 lines)*, and seeing that the buffer is left unchanged because all calculated indentation levels match ones already used in the files.

### DIFF
--- a/misc/ninja-mode.el
+++ b/misc/ninja-mode.el
@@ -19,15 +19,21 @@
 ;;; Commentary:
 
 ;; Simple emacs mode for editing .ninja files.
-;; Just some syntax highlighting for now.
 
 ;;; Code:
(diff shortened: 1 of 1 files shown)


Hi-Angel:
Ping?

PR #2282, opened 2023-03-31
Fix formatting for `msvc` tool documentation

arxanas:
The formatting currently looks like this:

Screenshot 2023-03-31 at 14 00 03

The code block is broken, and I think the text following the code block should be included in the `msvc` section.


jhasse:
thx!

PR #2289, opened 2023-04-21
Use *W versions of win32 file functions

parbo:
So that the long paths work on windows with registry LongPathsEnabled=1

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -340,7 +340,8 @@ int ReadFile(const string& path, string* contents, string* err) {
   // This makes a ninja run on a set of 1500 manifest files about 4% faster
   // than using the generic fopen code below.
   err->clear();
-  HANDLE f = ::CreateFileA(path.c_str(), GENERIC_READ, FILE_SHARE_READ, NULL,
+  std::wstring wpath = ::toWide(path);
+  HANDLE f = ::CreateFileW(wpath.c_str(), GENERIC_READ, FILE_SHARE_READ, NULL,
                            OPEN_EXISTING, FILE_FLAG_SEQUENTIAL_SCAN, NULL);
   if (f == INVALID_HANDLE_VALUE) {
     err->assign(GetLastErrorString());
@@ -476,6 +477,13 @@ void Win32Fatal(const char* function, const char* hint) {
   } else {
     Fatal("%s: %s", function, GetLastErrorString().c_str());
   }
+}
+
+std::wstring toWide(const string& s) {
+  int wchars_num = MultiByteToWideChar(CP_UTF8, 0, s.c_str(), -1, NULL, 0);
+  std::wstring w(wchars_num, ' ');
+  MultiByteToWideChar(CP_UTF8, 0, s.c_str(), -1, (LPWSTR)w.data(), w.size());
+  return w;
 }
 #endif
 
--- a/src/disk_interface.cc
+++ b/src/disk_interface.cc
@@ -25,7 +25,7 @@
 #ifdef _WIN32
 #include <sstream>
 #include <windows.h>
-#include <direct.h>  // _mkdir
+#include <fileapi.h>
 #else
 #include <unistd.h>
 #endif
@@ -56,7 +56,8 @@ string DirName(const string& path) {
 
 int MakeDir(const string& path) {
 #ifdef _WIN32
-  return _mkdir(path.c_str());
+  std::wstring wpath = ::toWide(path);
+  return CreateDirectoryW((LPCWSTR)wpath.c_str(), NULL) ? 0 : -1;
 #else
   return mkdir(path.c_str(), 0777);
 #endif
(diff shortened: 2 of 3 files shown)


parbo:
@emmenlau yes, at least for my case. We use response files, and ninja was unable to create the files without this patch. Now my next problem is that `cl` is unable to open the created response file.. 😢

emmenlau:
> Now my next problem is that `cl` is unable to open the created response file.. cry

We happily switched to `clang-cl.exe`, which in my eyes is the much better compiler, and is trivially simple to switch from `cl.exe`. It works with long paths, too.

parbo:
I think the problem seems to be that the response file that gets created is empty, so there might be some issue with my patch.

emmenlau:
@cristianadam It seems reasonable that `*W` methods should be used in combination with the manifest. Could this be the missing link why the previous PR alone did not solve the problem?

Looks legit to me.

parbo:
The docs explicitly state the functions that the manifest allows long paths for in conjunction with `LongPathsEnabled`. It's only the *W functions, no *A functions mentioned.

emmenlau:
> The docs explicitly state the functions that the manifest allows long paths for in conjunction with `LongPathsEnabled`. It's only the *W functions, no *A functions mentioned.

I agree. However @cristianadam tested his change in https://github.com/ninja-build/ninja/pull/2225#issuecomment-1347407419 and found that it works for him. It did work for me on our desktop computer, too. It only failed on our CI machine for me. So there seems to be at least a small step forward, but it is unclear why it is not a full solution.

Maybe this is the missing link, I'm keeping fingers crossed!

parbo:
I'm running into this issue with `cl.exe`: https://developercommunity.visualstudio.com/t/clexe-command-line-error-d8022-cannot-open-path-re/1478916

But ninja itself seems to work fine with this patch.

src/util.cc
```diff
@@ -477,6 +478,13 @@ void Win32Fatal(const char* function, const char* hint) {
     Fatal("%s: %s", function, GetLastErrorString().c_str());
   }
 }
+
+std::wstring toWide(const string& s) {
+  int wchars_num = MultiByteToWideChar(CP_UTF8, 0, s.c_str(), -1, NULL, 0);
+  std::wstring w(wchars_num, ' ');
+  MultiByteToWideChar(CP_UTF8, 0, s.c_str(), -1, (LPWSTR)w.data(), w.size());
```
Osyotr, line 485: Are you sure this is `CP_UTF8` and not `CP_ACP`?

bruchar1:
It this supposed to fix long paths issues in Windows? It tried it on my project, and I still get the following error:

> ninja: error: WriteFile(src/monorepo/lib/LineMasterUI/LineMasterUI-vc143-mt-gd-x64-u.dll.p/MainWindow_AdministratorOptions_AdvancedOptions_AutomaticDispositionChange_AutomaticDispositionChangeWidget.cpp.obj.rsp): Unable to create file. The system cannot find the path specified.
ninja: build stopped: .

makslevental:
@Osyotr are you still working on this? I.e., comfortable reviewing this PR if I take it over?

Osyotr:
> @Osyotr are you still working on this? I.e., comfortable reviewing this PR if I take it over?

I am not a maintainer of this project. I removed my review comment because I found the answer myself: ninja uses UTF-8 on windows:
https://github.com/ninja-build/ninja/blob/dcefb838534a56b2621baf33c7e0583ff63ee35f/windows/ninja.manifest#L5

makslevental:
@jhasse would you accept this PR if I got it in shape?

jhasse:
No, sorry.

emmenlau:
This is a forever nightmare. I guess the only real "solution" is to maintain a fork with this PR, and make binary releases for people to use.

We currently maintain this inside our company, but it's obvious that more people need this.

Volunteers? I can also chime in after my holidays in two weeks...

makslevental:
> No, sorry.

I don't get it - you think the current state (where everyone has to move their build dir to `C:\b`) is the correct state?

mpratt14:
you should follow up the "no, sorry" with an alternative that would be considered acceptable... by the way what's the reason? is it compatibility?

here's a random idea:

1. catch the error and put the files that error into a completely different directory than the build directory, call it something like `C:\ninja-extra-<project>\`

2. catch the error again if it happens again and rename the file

emmenlau:
@makslevental and @mpratt14 , I share your sentiment! But all of this has been discussed before, and already a few times. Just read the (multiple) corresponding issues and pull requests.

Short summary: The core maintainer wants this to be solved in Microsoft layer, not in this project.

There is no point in discussing this over and over, unless there is new insight. Either provide a PR that solves it in the Microsoft layer (like the already merged PR tries to do) or fork the project and maintain a patch with some other kind of workaround.

 Just my two cents.

cristianadam:
Please do try the latest ninja release (1.21.1) and make sure your Windows has `Computer\HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\FileSystem\LongPathsEnabled (Type: REG_DWORD)` set to `1`.

Ninja has now the manifest entry mentioned at https://github.com/ninja-build/ninja/pull/2225 and it should be able to read and write files longer than 260 characters.

So there is no need for a `*W` patch.

There is a BIG issue though, Windows cannot execute processes in paths longer than 260 characters (due to `SetCurrentDirectory`). I've mentioned this at  https://developercommunity.visualstudio.com/t/compiler-cant-find-source-file-in-path/10221576 and it looks like this:

windows11-longpaths-fail

Windows has historical limitations and you need to use `subst` or _Junctions_ to get things to work. Or just cross-compile with `clang-cl` from a different operating system.

mpratt14:
Nice explanation

I see that `longPathAware` is also added already, which I was just reading about.

I was not suggesting that this PR is good, but like I said, there must be *something* we can do.

jhasse:
closing in favor of #2552

Issue #1821, opened 2020-07-28
CMake build doesn't enable the use of ppoll on platforms that support it


petrhosek:
On platforms that support it, `configure.py` enables the use of `ppoll` by defining `USE_PPOLL`. CMake build doesn't replicate this logic which means that Ninja built with CMake is always going to use the less efficient `pselect`.

00003481:
nothing complicated ...

just add a test if the header "poll.h" is present 

if present add the definition -DHAVE_POLL_H  and  -DUSE_PPOLL

jhasse:
@ghost PR would be very welcome :)

PR #2290, opened 2023-04-21
Enable ppoll() usage when available.

digit-google:
This patch modifies the CMakeLists.txt file to probe for ppoll() on the target system, and define -DUSE_PPOLL=1 if it is available.

This can be disabled by setting -DNINJA_FORCE_PSELECT=ON when invoking CMake.

This matches the default behavior of the configure.py script (and its `--force-pselect` option). Note that there is no noticeable performance difference before build commands are launched, so this
change is very hard to benchmark properly.

Fix for issue #1821

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,7 @@ include(CheckSymbolExists)
 include(CheckIPOSupported)
 
 option(NINJA_BUILD_BINARY "Build ninja binary" ON)
+option(NINJA_FORCE_PSELECT "Use pselect() even on platforms that provide ppoll()" OFF)
 
 project(ninja CXX)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -35,6 +36,16 @@ else()
...
 		add_compile_options(-fdiagnostics-color)
 	endif()
+
+	if(NOT NINJA_FORCE_PSELECT)
+		# Check whether ppoll() is usable on the target platform.
+		# Set -DUSE_PPOLL=1 if this is the case.
+		include(CheckSymbolExists)
+		check_symbol_exists(ppoll poll.h HAVE_PPOLL)
+		if(HAVE_PPOLL)
+			add_compile_options(-DUSE_PPOLL=1)
```
petrhosek, line 46: `add_compile_definitions(USE_PPOLL=1)` is more idiomatic.

PR #2291, opened 2023-04-24
Allow duplicate rule variable usage.

digit-google:
This fixes #1996 by removing the variable name from the lookups stack once the recursive lookup call has been performed. Without this, any previously expanded variable could no longer be referenced in the command, as Ninja would (incorrectly) complain about a cyclical dependency.

### DIFF
--- a/misc/output_test.py
+++ b/misc/output_test.py
@@ -112,6 +112,19 @@ red
 \x1b[31mred\x1b[0m
 ''')
 
+    def test_issue_1966(self):
+        self.assertEqual(run(
+'''rule cat
+  command = cat $rspfile $rspfile > $out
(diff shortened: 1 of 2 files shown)


jhasse:
You mean #1966 instead of #1996, right?

digit-google:
Oh absolutely, good catch :-) I have uploaded a new commit with the right number in the message. Fortunately, the test already has the correct number.

jhasse:
Is it possible to create a more lightweight unit test for this? (if too complicated don't bother)

Have you seen #2202 and can comment on the differences?

digit-google:
friendly ping. The pull request was updated to include an explanation.

PR #2300, opened 2023-06-14
Match concurrency to available CPU bandwidth

maxim-kuvyrkov:
This change allows ninja to throttle number of parallel tasks based on feedback from Linux kernel's PSI (Pressure Stall Information) interfacts.  It extends "-l" parameter to accept negative values; "-l-NN" means that ninja should limit concurrency when processes in current cgroup spend more than NN% of their time stalled on CPU.

E.g., running "ninja -j100 -l-10" on a 32-core machine will quickly settle on parallelism of 32-34.

This option is designed to make ninja use all CPU bandwidth available to a cgroup-based container, while not starting excessive number of processes, which could eat up all RAM.

The motivation for this feature is too automatically reduce parallelism when the system is about to run out of RAM. If the system has swap enabled, "ninja -l-10" will dance with parallelism on the edge of just using a bit of swap.  As soon as a process starts swapping, that increases CPU "stalled" cycles, and parallelism is reduced.  The same argument works when a process is waiting for its turn to use IO and/or network.


src/util.cc
```diff
@@ -835,6 +837,94 @@ double GetLoadAverage() {
...
+  // The "total" units are microseconds, but documentation does not say
+  // whether it's cumulative across all CPUs or not.  Apparently, it's
+  // not cumulative.  IIUC, on an 8-core system if we have 6 processes
+  // running at 100% and another 2 stalled at 100% -- then every second
+  // the "total" stalled count will be increased by 1000000 [microseconds].
+  // The count will be increased by the same 1000000 [microseconds] if all
+  // 8 processes are 100% stalled.
+
+  ifstream cpupressure("/sys/fs/cgroup/cpu.pressure", ifstream::in);
+  string token;
```
Flowdalic, line 863: Shouldn't this first check `cpupressure.fail()` to detect the failure case? For example, when the kernel is build without PSI.

Flowdalic (commented):
Nice idea using PSI information. This is probably what you want build systems to dynamically control their parallelism. This is similar to what I did in https://github.com/ninja-build/ninja/pull/1949, although slightly less portable.

src/util.cc
maxim-kuvyrkov: If the cpu.pressure file does not exist, then "cpupressure >> token" will return "false" and the while loop will exit without setting "psi_ok = true", which will cause early exit with "return -1.0f;".
Or am I missing something?

oscardssmith:
any chance this can get revived? I'm currently commenting on my phone because my desktop is completely stuck using all 32 gb of RAM (and 48gb of swap) because I built LLVM without limiting parallelism sufficiently

src/util.cc
```diff
@@ -835,6 +837,94 @@ double GetLoadAverage() {
 }
 #endif // _WIN32
 
+double GetCPUWaitRatio(size_t subproc_number, int parallelism) {
```
DominiquePelle-TomTom, line 840: According to Google C++ style  guide, it should rather be called `GetCpuWaitRatio(...)`

Quoting https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

> When an abbreviation is used, prefer to capitalize it as a single "word", e.g., StartRpc() rather than StartRPC()

TrentHouliston:
I mean I'm not a maintainer but I wanted to comment that I've been interested in this for ages. Sad that it still hasn't happened!

PR #2307, opened 2023-07-05
Zos support

alexgubanow:
it compiles and runs, but Lexer fails to work. IBM ZOS has EBCIDIC not a ASCII, that breaks Lexer inner logic.


jhasse:
what's the point if it still doesn't work?

alexgubanow:
In order to get help with generating lex tables for EBCDIC, since the once in code today are ascii based and not working on zos, unless you enable ascii mode, which isn't preferable option.
Does anyone knows what generator was used for those tables? sounds like we could just use bison for that.

configure.py
```diff
@@ -382,8 +384,8 @@ def search_system_path(file_name):
         cflags.append('-D_LARGE_FILES')
     if platform.is_zos():
         cflags += ['-D_POSIX_SOURCE', '-D_POSIX_C_SOURCE=200112L', '-D_XOPEN_SOURCE_EXTENDED=1', '-D_UNIX03_THREADS']
-        cflags.remove('-pipe')
-        cflags.remove('-std=c++11')
+        # cflags.remove('-pipe')
+        # cflags.remove('-std=c++11')
```
digit-google, line 388: I would leave this flag here to ensure that the Ninja sources always exclusively use C++11 constructs, instead of whatever the compiler supports by default (which could be C++14 or anything more recent).

digit-google:
Thank you. This seems like great news.

Can you provide a slightly more descriptive commit message (e.g. "Use ibm-clang++ to build on zOS").

Also please rebase and clean up your patcheset to avoid pushing your merge to the upstream repository. This is more noise than signal and makes looking at the project's history painful once it is submitted.

digit-google:
Which tables are you refering to exactly? We use `re2c` to convert `lexe.in.cc` into `lexer.cc`, and `depfile_parser.in.cc` into `depfile_parser.cc`, though the outputs are checked into the source tree for people who do not have this tool in their PATH. If you do, it should be picked up by the build and used automatically.

See `has_re2c()` definition in `configure.py` to see how that works. The solution might be to change the command arguments when the tool is invoked on zOS.

Issue #2083, opened 2022-02-15
Add fail fast mode that immediately stops execution of running jobs after failures.


kuchara:
Currently there is option `-k N` which does not start new jobs when there were `N` errors. However sometimes users would like to immediately stop execution after failures, and free computing resources -- this is especially useful for big projects (thousands of jobs, distributed using tools like `distcc`). 

My proposal is to add fail fast option (e.g. `-K`), which would immediately stop further execution of all running jobs after reaching `-k` amount of failures.

Minimal example:

    $ cat build.ninja
    rule delay
      command = sleep 5
    
    rule failure
      command = exit 1
    
    build foo1: delay
    build foo2: failure

    $ time ninja
    [1/2] exit 1
    FAILED: foo2
    exit 1
    [2/2] sleep 5
    ninja: build stopped: subcommand failed.
    
    real    0m5.070s
    user    0m0.000s
    sys     0m0.078s

I have already been experimenting with source code, and easiest change to me was to send a SIGINT to itself in Builder::Build:

    $ time ninja/build-cmake/ninja -K
    [1/2] exit 1
    FAILED: foo2
    exit 1
    Fail fast mode activated. No more failures allowed. SIGINT self...
    : No such file or directory
    ninja: build stopped: interrupted by user.
    
    real    0m0.046s
    user    0m0.000s
    sys     0m0.031s

I am eager to create a PR for this, however please suggest if such feature can be accepted, and direction of the implementation.

jhasse:
Not sure if SIGINTing jobs is always safe. In my Ninja frontend I implemented that the remaining jobs run in the background (and output is suppressed): https://bixense.com/ja/
While this won't free up resources, it at least lets me use my terminal again immediately on a failure.

kuchara:
Could you explain why it may not be safe? I believe it is the same as "Ctrl+C" of a ninja execution, or?

However now, as I am reviewing signal list, maybe SIGTERM is better option here (specifically to inform process to terminate)? If some subprocesses ignore this, well... build tool should not bother...

This `ja` is interesting! Thanks for sharing.

jhasse:
In the same way that Ctrl+C'ing ninja might not be safe: Jobs just might just not handle it correctly.

Now that I think of it, the frontend coulds just Ctrl+C ninja on the first failure. So I'm tagging this as a frontend bug.

kuchara:
Thanks!

By the way, I've found out that `-k` feature is a small white lie. ;-)
[...]

PR #2308, opened 2023-07-07
Add failfast mode

kuchara:
Commit adds new command line option, -K, that aborts all running jobs after first failed job.

This implements #2083.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -19,6 +19,12 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <functional>
+#ifdef _WIN32
+#include <windows.h>
+#else
+#include <signal.h>
+#include <unistd.h>
+#endif
 
 #if defined(__SVR4) && defined(__sun)
 #include <sys/termios.h>
@@ -675,6 +681,17 @@ bool Builder::Build(string* err) {
       }
 
       if (!result.success()) {
+        if (config_.failfast_mode){
+          Warning("at least one subcommand failed and failfast mode activated. Sending Ctrl+C (SIGINT) to itself.");
+#ifdef _WIN32
+          if(! GenerateConsoleCtrlEvent(CTRL_C_EVENT,
+                                        GetCurrentProcessId())
+            Win32Fatal("GenerateConsoleCtrlEvent");
+#else
+          kill(getpid(), SIGINT);
+#endif
+        }
+
         if (failures_allowed)
           failures_allowed--;
       }
--- a/src/build.h
+++ b/src/build.h
@@ -156,7 +156,8 @@ struct CommandRunner {
 /// Options (e.g. verbosity, parallelism) passed to a build.
 struct BuildConfig {
   BuildConfig() : verbosity(NORMAL), dry_run(false), parallelism(1),
-                  failures_allowed(1), max_load_average(-0.0f) {}
+                  failures_allowed(1), max_load_average(-0.0f),
+                  failfast_mode(false)	{}
 
   enum Verbosity {
     QUIET,  // No output -- used when testing.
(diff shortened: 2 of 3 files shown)


emmenlau:
Reporting on the current status of this PR:

On Linux, it works flawless! The build stops on first error, which is great! Very nice work!

On Windows, it stops on first error, but I do get a terminal prompt:
```
6 errors generated.
Terminate batch job (Y/N)? 
```
I am running `ninja` inside `CMD`, which in turn has been started from Cygwin `bash`. I am not certain whether `CMD`, Cygwin or `bash` have an effect on this. The prompt seems to originate from `CMD`, so maybe the kill signal should not be passed to `CMD`?

emmenlau:
> As I read some userexchange discussions, this message on Windows CMD is "by design". :-( https://superuser.com/a/35701/982937

Yep, after some googling I was also under the impression that CTRL-C is not really awesome for these kind of things.

> Do you have any suggestions how to continue here?

Well, I'm really no expert, but sending CTRL-C to oneself seems anyways a "hard" way to end the whole process tree. Are these even separate processes or "just" threads, or even something else? What other ways does Win32 provide to kill them? Do we even need to kill, or can we just send an "interrupt" signal and catch that on the receiving side to exit the processes?

kuchara:
Hello @emmenlau !

I'd love to finish this. 

Finally was able to spent some more time to inspect ninja source code, and I've noticed that regular Clear() function done in build.cc is actually sending signals to its children, so decided to give it a try. 

Could you try the newest change I've just uploaded?

If that does not work, my only suggestion would be to enable this feature only for posix systems, until there will be someone who could help with Windows case.
In my company we already use patched ninja for some time with very good results (cha-ching!).

emmenlau:
Awesome @kuchara , this works very nicely on Windows! I have not tested Linux and MacOSX yet, but the change looks very promising! Great work!

src/subprocess-posix.cc
```diff
@@ -366,3 +366,8 @@ void SubprocessSet::Clear() {
     delete *i;
   running_.clear();
 }
+
+void SubprocessSet::Abort() {
+  SetInterruptedFlag(SIGINT);
+  Clear();
```
digit-google, line 372: Thanks for your patch. Can you ensure that SIGINT is also sent to the console subprocess, if there is one?
Otherwise if the failure comes from a non-console command, the console subprocess will continue to run and block Ninja.

Let me clarify:

Ninja launches subprocesses in their own process group, with the exception of commands in the "console" pool. These are launched in the same process group as Ninja itself. This is done to ensure that when Ctrl-C is pressed, both Ninja and the console subprocess receive a SIGINT (I suspect this is required for proper terminal handling too, but I don't remember the details).

By default, the console subprocess receiving the SIGINT would exit, while Ninja would catch the SIGINT, report it when exiting DoWork() to the Builder(), which ends up invoking Clear(). The latter relays the SIGINT to all non-console subprocesses to stop them as well (hence the use_console check in that function).

With your code, if the failure comes from a non-console subprocess, calling SetInterruptedFlag(SIGINT) will tell the next DoWork() call to report an interrupt, but will not send SIGINT to the console process which will continue running. And because it also shared the stdout / stderr descriptors, it will continue printing to the terminal. And Ninja will likely be stuck waiting for it in the Subprocess destructor (which ends up calling waitpid()).

It looks like something similar is true for Win32.

Also, because this is very subtle, I strongly recommend adding a robust unit test for this behavior to your PR.

For example one with a long console subprocess that sleeps for 10 seconds, and another non-console one that fails immediately. The test should verify that Ninja exits immediately with -K. You should also test the case where the console process fails first of course.

Hope this helps

PR #2319, opened 2023-08-30
Allow 'ninja -t compdb' accept targets

### LINKED ISSUE #1544: [Feature Request] Allow 'ninja -t compdb' accept one target

csmoe:
Fix #1544 

Cherry-picked https://github.com/ninja-build/ninja/pull/1546 and addressed the comments.
May I have your code review @jhasse ? Thanks

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -19,7 +19,10 @@
 #include <string.h>
 
 #include <algorithm>
+#include <cassert>
 #include <cstdlib>
+#include <set>
+#include <sstream>
 
 #ifdef _WIN32
 #include "getopt.h"
@@ -932,6 +935,32 @@ void printCompdb(const char* const directory, const Edge* const edge,
   printf("\"\n  }");
 }
 
+struct DependentEdgeCollector {
+  bool CollectFrom(Node* node, std::vector<Edge*>* depend_edges) {
+    assert(node);
+    assert(depend_edges);
+
+    if (!visited_nodes_.insert(node).second)
+      return true;
+
+    Edge* edge = node->in_edge();
+    if (!edge || !visited_edges_.insert(edge).second)
+      return true;
+
+    depend_edges->push_back(edge);
+
+    for (Node* input_node : edge->inputs_)
+      if (!CollectFrom(input_node, depend_edges))
+        return false;
+
+    return true;
+  }
+
+ private:
+  std::set<Node*> visited_nodes_;
+  std::set<Edge*> visited_edges_;
+};
+
 int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
                                        char* argv[]) {
   // The compdb tool uses getopt, and expects argv[0] to contain the name of
@@ -943,11 +972,42 @@ int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
 
   optind = 1;
   int opt;
-  while ((opt = getopt(argc, argv, const_cast<char*>("hx"))) != -1) {
+  const option long_options[] = {
+      { "target", required_argument, nullptr, 't' },
+      { nullptr, 0, nullptr, 0 }
+  };
+
+  std::vector<Edge*> edges_to_process = state_.edges_;
+  while ((opt = getopt_long(argc, argv, "hx:", long_options, nullptr)) != -1) {
     switch(opt) {
       case 'x':
         eval_mode = ECM_EXPAND_RSPFILE;
         break;
+
+      case 't': {
+        std::istringstream iss(optarg);
+        std::string target;
+        std::vector<Edge*> user_interested_edges;
+        DependentEdgeCollector deps_edge_collector;
+
+        while (std::getline(iss, target, ',')) {
+          string err;
+          Node* node = CollectTarget(target.c_str(), &err);
+
+          if (!node) {
+            Error("%s", err.c_str());
+            return 1;
+          }
+          if (!deps_edge_collector.CollectFrom(node, &user_interested_edges)) {
+            return 1;
+          }
+        }
+        if (!user_interested_edges.empty()) {
+          edges_to_process = user_interested_edges;
+        }
+
+        break;
+      }
 
       case 'h':
       default:
@@ -955,7 +1015,8 @@ int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
             "usage: ninja -t compdb [options] [rules]\n"
             "\n"
             "options:\n"
-            "  -x     expand @rspfile style response file invocations\n"
+            "  -x                  expand @rspfile style response file invocations\n"
+            "  --target TARGETS    generate compilation database for given targets, separated by commas\n"
             );
         return 1;
     }
@@ -978,23 +1039,22 @@ int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
   }
 
   putchar('[');
-  for (vector<Edge*>::iterator e = state_.edges_.begin();
-       e != state_.edges_.end(); ++e) {
-    if ((*e)->inputs_.empty())
+  for (Edge* e : edges_to_process) {
+    if (e->inputs_.empty())
       continue;
     if (argc == 0) {
       if (!first) {
         putchar(',');
       }
-      printCompdb(&cwd[0], *e, eval_mode);
+      printCompdb(&cwd[0], e, eval_mode);
       first = false;
     } else {
       for (int i = 0; i != argc; ++i) {
-        if ((*e)->rule_->name() == argv[i]) {
+        if (e->rule_->name() == argv[i]) {
           if (!first) {
             putchar(',');
           }
-          printCompdb(&cwd[0], *e, eval_mode);
+          printCompdb(&cwd[0], e, eval_mode);
           first = false;
         }
       }


src/ninja.cc
```diff
@@ -920,19 +955,51 @@ int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
...
+
+  std::vector<Edge*> edges_to_process = state_.edges_;
+  while ((opt = getopt_long(argc, argv, "x:", long_options, NULL)) != -1) {
     switch(opt) {
       case 'x':
         eval_mode = ECM_EXPAND_RSPFILE;
         break;
 
+      case 't': {
+        vector<Node*> targets;
```
jhasse, line 971: unused?

src/ninja.cc
```diff
@@ -909,6 +912,38 @@ void printCompdb(const char* const directory, const Edge* const edge,
...
+    return GetAllDependentEdges(node, depend_edges, &visited_nodes_, &visited_edges_);
+  }
+
+  assert(node);
+  assert(depend_edges);
+
+  if (visited_nodes->count(node)) {
+    return true;
+  }
+  visited_nodes->insert(node);
```
digit-google, line 930: nit: use the result of insert() to avoid the count() call above and perform one lookup instead of two, e.g.:

```
if (!visited_nodes->insert(node).second)
  return true;
```

src/ninja.cc
```diff
@@ -909,6 +912,38 @@ void printCompdb(const char* const directory, const Edge* const edge,
...
+
+  Edge* edge = node->in_edge();
+  // Leaf node
+  if (!edge || visited_edges->count(edge)) {
+    return true;
+  }
+  visited_edges->insert(edge);
```
digit-google, line 937: same here:

```
if (!edge ||! visited_edges->insert(edge).second))
  return true;
```

src/ninja.cc
```diff
@@ -920,19 +955,50 @@ int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
...
+
       case 'h':
       default:
         printf(
             "usage: ninja -t compdb [options] [rules]\n"
             "\n"
             "options:\n"
-            "  -x     expand @rspfile style response file invocations\n"
+            "  -x           expand @rspfile style response file invocations\n"
+            "  --target     generate compilation database for given targets, separated by commas\n"
```
digit-google, line 1001: nit: specify arguments to --target for clarity, as in:

```
-x                              expand @rspfile style...
--target TARGETS   generate compilation database for a comma-separated list of targets.
```

src/ninja.cc
```diff
@@ -955,8 +1021,8 @@ int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
   }
 
   putchar('[');
-  for (vector<Edge*>::iterator e = state_.edges_.begin();
-       e != state_.edges_.end(); ++e) {
+  for (vector<Edge*>::iterator e = edges_to_process.begin();
+       e != edges_to_process.end(); ++e) {
```
digit-google, line 1025: since we're C++11 now, can we switch to proper for-loops now when modifying the code? This seems a good opportunity to do that here.

src/ninja.cc
```diff
@@ -920,19 +955,50 @@ int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
...
+  std::vector<Edge*> edges_to_process = state_.edges_;
+  while ((opt = getopt_long(argc, argv, "x:", long_options, NULL)) != -1) {
     switch(opt) {
       case 'x':
         eval_mode = ECM_EXPAND_RSPFILE;
         break;
 
+      case 't': {
+        std::istringstream iss(optarg);
+        string target;
```
digit-google, line 972: nit: can we use std::string for all new code, since we care about removing all `using namespace std` in the future.

src/ninja.cc
```diff
@@ -909,6 +912,38 @@ void printCompdb(const char* const directory, const Edge* const edge,
...
 }
 
+bool GetAllDependentEdges(Node* node, std::vector<Edge*>* depend_edges,
+                          std::set<Node*>* visited_nodes=NULL,
+                          std::set<Edge*>* visited_edges=NULL) {
+  if (visited_nodes == NULL || visited_edges == NULL) {
+    std::set<Node*> visited_nodes_;
+    std::set<Edge*> visited_edges_;
+    return GetAllDependentEdges(node, depend_edges, &visited_nodes_, &visited_edges_);
+  }
```
digit-google, line 922: It looks like this function is called in a loop with default arguments, each call will start with empty visited sets, so edges that are common to multiple calls will end up being appended to |depend_edges| multiple times. Is this what you really want here or a bug?

I suggest creating a small dedicated class to hold the visited state as an implementation detail instead, and make lifetime analysis clearer. E.g.:

```
struct DependentEdgeCollector {
  void CollectFrom(Node* node, std::vector<Edge*>*  out_edges) {
    ...
  }
  
  // Reset visited sets (only if this is needed, omit it otherwise).
  void Reset() {
    visited_nodes_.clear();
    visited_edges_.clear();
  }
 private:
  std::set<Node*> visited_nodes_;
  std::set<Edge*> visited_edges_;
};
```

digit-google:
This looks good to me, but will require @jhasse 's approval.

Can I ask you to add some regression test in misc/output_test.py for this feature though?

src/ninja.cc
```diff
@@ -955,23 +1016,22 @@ int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
   }
 
   putchar('[');
-  for (vector<Edge*>::iterator e = state_.edges_.begin();
-       e != state_.edges_.end(); ++e) {
-    if ((*e)->inputs_.empty())
+  for (Edge* e: edges_to_process) {
```
jhasse, line 1019: Please add a space before `:` (or does our .clang-format suggest otherwise?).

src/ninja.cc
```diff
@@ -920,19 +949,51 @@ int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
...
+
+          if (!node) {
+            Error("%s", err.c_str());
+            return 1;
+          }
+          if (!deps_edge_collector.CollectFrom(node, &user_interested_edges)) {
+            return 1;
+          }
+        }
+        if (user_interested_edges.size() > 0) {
```
jhasse, line 982: Please use `!user_interested_edges.empty()` instead.

src/ninja.cc
```diff
@@ -920,19 +949,51 @@ int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
 
   optind = 1;
   int opt;
-  while ((opt = getopt(argc, argv, const_cast<char*>("hx"))) != -1) {
+  const option long_options[] = {
+      {"target", required_argument, NULL, 't'},
```
jhasse, line 953: Please use `nullptr` instead of `NULL`.

src/ninja.cc
```diff
@@ -920,19 +949,51 @@ int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
...
+      {NULL, 0, NULL, 0}
+  };
+
+  std::vector<Edge*> edges_to_process = state_.edges_;
+  while ((opt = getopt_long(argc, argv, "x:", long_options, NULL)) != -1) {
```
jhasse, line 958: Why `"x:"` instead of `"hx"`?

src/ninja.cc
```diff
@@ -17,6 +17,9 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sstream>
+#include <set>
+#include <assert.h>
```
jhasse, line 22: Please use `#include <cassert>` instead.

jhasse (changes requested):
Thanks for having another go at this :)

JamesWidman:
fwiw, i really like this feature and look forward to this PR being merged! 

one obvious use case for C/C++ users is to produce the database consumed by clangd. It means i get more concise & relevant search hits when e.g. asking clangd for all references to some symbol `x` in some program `foo`.

it also means the CPU fans in my PC no longer go crazy as a consequence of clangd's attempt to background-index a large number of source files that i didn't even want to browse (because they don't participate in building the executable `foo`).

thank you, @csmoe!

JamesWidman:
side note (for anyone who, like me, only learned about [the jq command](https://jqlang.github.io/jq/) recently): in addition to compile commands, `-t compdb --target foo` will produce entries for commands that don't translate C/C++ code to machine code (e.g. linker invocations, generation of header files, etc), so in case anyone wants to spare clangd from attempting to read object files, you could do something like: 
```sh
ninja -t compdb --target foo | jq '[.[] | select( .command | test("/bin/clang.* -c"))]' > compile_commands.json
```
(or `test("/bin/your_compiler_here.* -c")`

JamesWidman:
> ```shell
> ninja -t compdb --target foo | jq '[.[] | select( .command | test("/bin/clang.* -c"))]' > compile_commands.json
> ```

In some future PR, it might be nice if the compilation database format were extended to allow a field like `"command_type"` as in:
```json
  { "directory": "/home/user/llvm/build",
    "command": "clang-cl.exe [args...] /c /o t.o t.cc",
    "command_type": "compile_CXX",
    "file": "t.cc" },
```
...which generator programs might communicate to Ninja in the form of a `rule` variable:
```ninja
 rule some_cxx_rule
   command = clang-cl.exe [args...] /o $out /c $in
   command_type = compile_CXX
   description = Building CXX object $out
```
...so that our `jq` query would be a little more portable: 
```shell
ninja -t compdb --target foo | jq '[.[] | select( .command_type | test("^compile_(CXX|C)$"))]' > compile_commands.json
```
(But again, this would be a *separate* PR, and afaict the current PR is complete.)

csmoe:
https://github.com/ninja-build/ninja/actions/runs/10105777720/job/27951482409?pr=2319 @jhasse sorry for bothering you here, this error seems not caused by my patches.

mpratt14:
Would it be better to shift out arguments so that no flag is required for the targets list?

so instead of 

`-t compdb --target foo,bar`

you would have

`-t compdb foo bar`

or if a flag is really necessary to break out of flag-only parsing, use syntax like in `git`

`-t compdb -- foo bar`

so they can be space separated instead of comma separated

JamesWidman:
> Would it be better to shift out arguments so that no flag is required for the targets list?
> 
> so instead of
> 
> `-t compdb --target foo,bar`
> 
> you would have
> 
> `-t compdb foo bar`

Prior to the commits in this PR, the output of `ninja -t compdb -h` is:
```
usage: ninja -t compdb [options] [rules]

options:
```

so `ninja -t compdb foo bar` already means "emit every compile command that uses foo or bar", and we probably don't want to break that.

JamesWidman:
> or if a flag is really necessary to break out of flag-only parsing, use syntax like in `git`
> 
> `-t compdb -- foo bar`
> 
> so they can be space separated instead of comma separated

i guess this would work: 
```
ninja -t compdb [options] [rules] -- [targets]
```
if this is happening in a bash script, and if you've defined an array parameter where each element is the name of a target, then this alternative syntax might save users from learning how to join strings in bash:
```bash
ninja -t compdb --target $(targets=( $(cat $WORKSPACE/active-executables.txt)); echo $(IFS=','; echo "${targets[*]}"))
```

JamesWidman:
so, boiling down the above, it seems like the only open question is whether we want:

- option 1 (same old command form, but new option for [options]):
```
ninja -t compdb --target TARGETS [rules]
```
...where TARGETS is a comma-separated list of names of targets; 

or:

- option 2 (new command form):
```
ninja -t compdb [options] [rules] [-- targets]
```

i prefer the first form (as already implemented), since the second form requires me to remember that `-t compdb foo -- bar` means "*rule* foo and *target* bar", as opposed to "*target* foo and *rule* bar".

@mcprat, @csmoe, what do you think?

JamesWidman:
minor nit: if we go with option 1 (`--target TARGETS`), can we change the spelling of the option to `--targets` (plural)? (I already accidentally used the plural form a couple times.)

JamesWidman:
in case we need more motivation to merge this PR: i just wrote a small extension for [Telescope](https://github.com/nvim-telescope/telescope.nvim) (which is [pretty straightforward](https://github.com/nvim-telescope/telescope.nvim/blob/master/developers.md)) that reads a compilation database and hands the list of all primary source files from that database over to Telescope.

For example, if i cd to the build directory produced by cmake for building ninja itself and then run:
```zsh
ninja -t compdb --target ninja > compile_commands.json
```
...here's what i see when i invoke my Telescope extension:

image

For anyone starting to read the Ninja source code for the first time, this seems like a good starting point. (And as soon as i hit `<enter>` to load any source file in that list, [clangd](https://clangd.llvm.org/) indexes all translation units named by the same `compile_commands.json`, and *none* of the translation units *not* named therein. This means that e.g. the LSP command to list all references to a symbol will include only search hits from the correct set of translation units.)

By contrast, when i use an IDE to load project files generated by cmake, it is not at all obvious which source files are linked into the executable in which i'm currently interested.

To me, this means that i'll have an easier time browsing a C/C++ program `foo` with  `-t compdb --target foo` and Telescope than i would with an IDE. And it's not because i've developed more muscle memory with these tools; it's because this set of tools is actually better!

csmoe:
I like `ninja -t compdb --targets TARGETS [rules]`, @JamesWidman Are you interested to take over this pr? I might not always in radar.

Feel free to modify anything or recreate one, but preserve [lkchen](https://github.com/lk-chen)'s contribution record, he completed the main part.

JamesWidman:
> @JamesWidman Are you interested to take over this pr? I might not always in radar.

@csmoe sure! How should we proceed? (Does github's PR interface have something like a 'bequeath' button so that you can bequeath the PR to me, or should i open a new PR like you did, or...?)

csmoe:
github seems not support that, so create your pr😃

Issue #2266, opened 2023-02-28
Still getting long pathname issues under Windows with today's master version.


grebdioZ:
We were running into the "long pathname issue" on Windows #1900 with version 1.11.0 and hoped to overcome it with the fixes from #2225. If I understand correctly, those fixes should be available in the `master` branch, of which I built today's version. However, I still get the `Filename longer than 260 characters` error.

Configuring seems to work:

```ps1
PS C:\temp\test> cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -S C:\dev\cmake\fme\FMEwork\Internal\Projects\MLTestExceedinglyLongProjectPath\TestExceedinglyLongProjectPath_TestExceedinglyLongProjectPath_TestExceedinglyLongProjectPath_TestExceedinglyLongProjectPath_TestExceedinglyLongProjectPath\TestExceedinglyLongProjectPath_TestExceedinglyLongProjectPath_TestExceedinglyLongProjectPath\ -B bld --fresh
-- The CXX compiler identification is MSVC 19.34.31942.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files/Microsoft Visual Studio/2022/Professional/VC/Tools/MSVC/14.34.31933/bin/Hostx64/x64/cl.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- CMAKE_INSTALL_PREFIX initialized to default setting 'C:/Temp/test/bld/install'

[... internal stuff that did not contain relevant warnings or errors]

-- Performing Test COMPILER_HAS_DEPRECATED_ATTR
-- Performing Test COMPILER_HAS_DEPRECATED_ATTR - Failed
-- Performing Test COMPILER_HAS_DEPRECATED
-- Performing Test COMPILER_HAS_DEPRECATED - Success
-- Configuring done
-- Generating done
-- Build files have been written to: C:/Temp/test/bld
```

But then:

```ps1
PS C:\temp\test> ninja -C bld
[...]

PR #2321, opened 2023-09-07
Fix check of filename length

brondani:
Fix check of filename length in `RealDiskInterface::Stat` when long paths support in Windows is enabled.
Add the application manifest to the `ninja_test` executable target to enable long paths support during tests.
Add the test case `DiskInterfaceTest.StatExistingFileWithLongPath`.
Address https://github.com/ninja-build/ninja/issues/2266 (`Filename longer than 260 characters` error).

### DIFF
--- a/src/disk_interface.cc
+++ b/src/disk_interface.cc
@@ -26,6 +26,7 @@
 #include <sstream>
 #include <windows.h>
 #include <direct.h>  // _mkdir
+#include <atlcore.h>
 #else
 #include <unistd.h>
 #endif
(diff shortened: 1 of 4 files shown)


src/disk_interface.cc
```diff
@@ -157,13 +158,27 @@ bool DiskInterface::MakeDirs(const string& path) {
 }
 
 // RealDiskInterface -----------------------------------------------------------
+RealDiskInterface::RealDiskInterface() 
+#ifdef _WIN32
+: use_cache_(false), long_paths_enabled_(false) {
+  setlocale(LC_ALL, "");
```
orgads, line 164: How is this related? It causes a regression. See #2435.

brondani: It seems to be a leftover, thanks for reporting!

PR #2329, opened 2023-09-20
Fix Mingw cross-compilation.

digit-google:
The previous pull request broke compilation with the Mingw toolchain, which does not provide <atlcore.h>. Moreover, this massive header was only used to expand a macro that generates a static volatile variable under the hood (which is not necessary here).

Replace it with with a simpler and more portable code path to fix the build.

Also, ensure that a broken DiskInterfaceTest, which was only disabled when using the Microsoft compiler, is also disabled for Mingw (note that the whole test code is inside a #ifdef _WIN32 .. #endif block).

After this patch, a `ninja_test.exe` generated on Linux with the Mingw toolchain runs and succeeds properly on a Windows machine.

### DIFF
--- a/src/disk_interface.cc
+++ b/src/disk_interface.cc
@@ -23,10 +23,10 @@
 #include <sys/types.h>
 
 #ifdef _WIN32
-#include <sstream>
-#include <windows.h>
 #include <direct.h>  // _mkdir
-#include <atlcore.h>
+#include <windows.h>
+
+#include <sstream>
 #else
 #include <unistd.h>
 #endif
@@ -162,10 +162,16 @@ RealDiskInterface::RealDiskInterface()
 #ifdef _WIN32
 : use_cache_(false), long_paths_enabled_(false) {
   setlocale(LC_ALL, "");
-  IFDYNAMICGETCACHEDFUNCTIONTYPEDEF(L"ntdll", BOOLEAN(WINAPI*)(),
-    "RtlAreLongPathsEnabled",
-    RtlAreLongPathsEnabled) {
-    long_paths_enabled_ = RtlAreLongPathsEnabled();
+
+  // Probe ntdll.dll for RtlAreLongPathsEnabled, and call it if it exists.
+  HINSTANCE ntdll_lib = ::GetModuleHandleW(L"ntdll");
+  if (ntdll_lib) {
+    typedef BOOLEAN(WINAPI FunctionType)();
+    auto* func_ptr = reinterpret_cast<FunctionType*>(
+        ::GetProcAddress(ntdll_lib, "RtlAreLongPathsEnabled"));
+    if (func_ptr) {
+      long_paths_enabled_ = (*func_ptr)();
+    }
   }
 }
 #else


digit-google:
+cc @brondani

src/disk_interface_test.cc
```diff
@@ -171,7 +171,7 @@ TEST_F(DiskInterfaceTest, StatCache) {
   EXPECT_GT(disk_.Stat("subdir/subsubdir", &err), 1);
   EXPECT_EQ("", err);
 
-#ifndef _MSC_VER // TODO: Investigate why. Also see https://github.com/ninja-build/ninja/pull/1423
+#if 0  // TODO: Investigate why. Also see https://github.com/ninja-build/ninja/pull/1423
```
jhasse, line 174: should be `#ifndef _WIN32` then, otherwise we would disable it for Linux, too.

jhasse (changes requested):

src/disk_interface_test.cc
digit-google: Ok, I'll do it, but note that there is already an #ifndef _WIN32 just before this test declaration, because this is (currently) the only platform with a StatCache.

jhasse: Ah I see! Then we probably shouldn't touch this line at all, should we? The test succeeded on AppVeyor for MinGW: https://ci.appveyor.com/project/nico/ninja/builds/47865014/job/eg5b0uic6jgrrida

digit-google: The test definitely fails when building with my Debian mingw installation, and running on a Windows 10 machine though, which is why I had to disable it there too. I have no idea why.

```
[3/398] MSVCHelperTest.NoReadOfStderrto stderr 
[207/398] DiskInterfaceTest.StatCache
*** Failure in /work2/github-digit/ninja/src/disk_interface_test.cc:175
disk_.Stat("subdir", &err) == disk_.Stat("subdir/.", &err)
*** Failure in /work2/github-digit/ninja/src/disk_interface_test.cc:178
disk_.Stat("subdir", &err) == disk_.Stat("subdir/subsubdir/..", &err)  
[398/398] BuildLogRecompactTest.Recompact
```

I suspect the issue is with the MSVCRT.DLL runtime being used, not the compiler that builds the code.

jhasse: I see. Still it runs fine in some cases and one of them is the CI. That's why I would like to keep it enabled.

digit-google: All right, we'll keep a local patch for our own CI then. I have removed the change to disk_interface_test.cc here.

jhasse:
Have you seen this warning?
```
./src/disk_interface.cc:171:62: warning: cast between incompatible function types from 'FARPROC' {aka 'long long int (*)()'} to 'BOOLEAN (*)()' {aka 'unsigned char (*)()'} [-Wcast-function-type]
  171 |         ::GetProcAddress(ntdll_lib, "RtlAreLongPathsEnabled"));
      |  
```

digit-google:
I was not aware of this warning. I have pushed https://github.com/digit-google/ninja/commit/f6e83fa80acec76ffc16fbc9c257d2ddf9da6114 as a work-around.

Searching online for an explanation only finds a few pages that state that this is a GCC 8+ bug, without proof however.
The warning itself does not seem to make sense, the WINAPI macro is used to ensure the right standard call convention is used, so I am not sure I understand what's going on here.

Issue #1583, opened 2019-05-28
BuildTest.WrongOutputInDepfileCausesRebuild leaves build_log and ninja_deps behind


jhasse:
Running `./ninja_test --gtest_filter=BuildTest.WrongOutputInDepfileCausesRebuild` leaves those two files behind.

PR #2330, opened 2023-09-22
Improve EXPECT_XXX and ASSERT_XXX output in case of failure.

digit-google:
Ensure that binary comparisons (e.g. EXPECT_EQ) display the actual values of the left and right side operands, and not just their source expression, as in:

```
Failure on file.cc:10:  expected != actual
Left:  10
Right: 20
```

Also ensure that boolean comparisons (e.g. EXPECT_TRUE) display the expected and actual values too as in:

```
Failure on file.cc:10: vec.size()
Expected: false
Actual:   1
```

This uses a small amount of C++11 SFINAE, without needing to change any of the test macro invocations. If a value is not supported, then "<UNPRINTABLE>" will be used instead.

### DIFF
--- a/src/build_log_test.cc
+++ b/src/build_log_test.cc
@@ -295,7 +295,7 @@ TEST_F(BuildLogTest, VeryLongInputLine) {
   ASSERT_EQ("", err);
 
   BuildLog::LogEntry* e = log.LookupByOutput("out");
-  ASSERT_EQ(NULL, e);
+  ASSERT_NULL(e);
 
   e = log.LookupByOutput("out2");
   ASSERT_TRUE(e);
--- a/src/build_test.cc
+++ b/src/build_test.cc
@@ -84,7 +84,7 @@ TEST_F(PlanTest, Basic) {
 
   ASSERT_FALSE(plan_.more_to_do());
   edge = plan_.FindWork();
-  ASSERT_EQ(0, edge);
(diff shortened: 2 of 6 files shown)


digit-google:
Oh, note that the AppVeyor build fails because of the <atlcore.h> include which is fixed by PR #2329 , and has nothing to do with this patch.

digit-google:
I had to change the patch after realizing that the expressions were evaluated twice in case of failure, which would result in incorrect runtime behavior. For example, consider:

```
EXPECT_EQ(0, fclose(file))
```

Which does happen in the source code, the `fclose()` would be called twice. I tried solving this by using lambdas to capture the value once with a const reference.

digit-google:
Mmm, there are too many subtle compilation issues here. I', going to close this PR for now until I can weed out everything, if that is possible. Sorry about the noise.

jhasse:
IMHO #1562 is the way forward.

digit-google:
Thanks, I was not aware that this was still a thing. The latest version of my branch at https://github.com/digit-google/ninja/commit/5e6a105d3619bc0b2278c01cc8e158aa09bd9b56 passes all tests now. I'll keep it as a local change in our own Ninja version for now.

Do you know if anyone is working on #1562 at the moment?

jhasse:
I would continue working on it when #1583 gets fixed. No one is working on that though.

PR #2340, opened 2023-10-09
Replace pipes.quote with shlex.quote in configure.py 

bc-lee:
Python 3.12 deprecated the pipes module and it will be removed
in Python 3.13. In configure.py, I have replaced the usage of `pipes.quote`
with `shlex.quote`, which is exactly the same function as `pipes.quote`.

For more details, refer to PEP 0594: https://peps.python.org/pep-0594


kloczek:
It would be rally good to release new version because this commit.

PR #2346, opened 2023-10-17
Add `NINJA_USE_SYSTEM_RE2C` Option

sunshaoce:
In some systems, the provided version of re2c is too low to build the latest release, so an option needs to be provided. For example, on CentOS7, the re2c version is `re2c 0.14.3`.


jhasse:
I would prefer that we check the version of re2c and automatically use it if high enough. We could use https://github.com/potassco/aspcud/blob/master/cmake/FindRE2C.cmake as a reference

PR #2352, opened 2023-11-06
Remove phony edges for nodes created by a dependency loader.

digit-google:
This patch simplifies Ninja internals without modifying its behavior. It removes the creation (and removal) of phony edges as producers for nodes loaded by dependency loaders, i.e. coming from depfiles, dyndep files or the deps log.

These edges were only used to ensure the build did not abort when these files are missing, unlike regular source inputs. This can be easily checked by adding a new flag to the Node class instead. This makes it easier to reason about how Ninja works internally.

More specifically:

- Move the generated_by_dep_loader_ flag from the Edge class to the Node class. The flag is true by default to minimize changes to the source code, since node instances can be first created by reading the deps or build logs before the manifest itself.

- Modify Plan::AddSubTarget() to avoid aborting the build when a generated-by-deploader node is missing. Instead the function exits immediately, which corresponds to what happened before.

- State::AddOut(), State::AddIn(), State::AddValidation(): Ensure that nodes added by these methods, which are only called from the manifest parser and unit-tests set the |generated_by_dep_loader_| flag to false, to indicate that these are regular input / output nodes.

- ManifestParser::ParseEdge(): Add an assertion verifying that the dyndep file is marked as a regular input.

- DyndepLoader::UpdateEdge(): Remove code path that looked for phony in-edges and ignored them.

- DepLoader::CreatePhonyInEdge() is removed as no longer necessary.

+ Update a few places in unit-tests that were checking for the creation of the phony edges.


digit-google:
Summonning @bradking and @jdrouhard who worked on this part of the code for their feedback.

bradking:
Putting the `generated_by_dep_loader_` field on the node itself, and dropping the phony input edge, makes sense to me.  It more accurately represents that no edge explicitly produces the node, unless and until one is added by a dyndep binding.

I just built `ninja` from this PR and verified that CMake's entire test suite still passes with it, including Fortran and C++ modules test cases that exercise ninja's dyndep bindings.

src/build.cc
```diff
@@ -95,12 +95,18 @@ bool Plan::AddTarget(const Node* target, string* err) {
...
+    // file. In the first case, a dirty flag means the file is missing,
+    // and the build should stop. In the second, do not do anything here
+    // since there is no producing edge to add to the plan.
+    if (node->dirty() && !node->generated_by_dep_loader()) {
       string referenced;
       if (dependent)
         referenced = ", needed by '" + dependent->path() + "',";
-      *err = "'" + node->path() + "'" + referenced + " missing "
+      *err = "'" + node->path() + "'" + referenced +
+             " missing "
```
bradking, line 109: This line wrap seems like a spurious hunk in the diff.

digit-google: Agreed, fixed, thanks :)

digit-google:
Also thank you Brad for verifying that this passes the whole CMake test suite. I just completed a run of it on my local machine (which is failing on two tests because I don't have all pre-requisites like `hg` installed). Impressive work, I'll certainly use it to verify my other Ninja changes.

jhasse:
Thanks to both of you!

PR #2358, opened 2023-12-05
CanonicalizePath: Remove kMaxComponents limit

digit-google:
This patch refactors the `CanonicalizePath()` function to fix two issues and improve performance. This is achieved through the following:

- Remove the kMaxComponents limit entirely, which fixes ninja-build#1732, by dropping the `components` array entirely, in favor of back-tracking the destination pointer.

- Properly handle `/` and `\\` which were incorrectly converted into an empty string. This fixes ninja-build#2008.

- Skip initial `../` components in relative paths, as these are common when referencing source files in build plans, and doing so make the loop after this step run faster in practice since most source files do not need adjustments.

- Simplify the inner loop logic by handling the last component (which is not followed by a trailing directory separator) separately.

  This noticeably improves performance because the inner loop becomes smaller with less branch mis-predictions in the general case.

- Never access or copy the character after the end of the input string.

- Use `memchr()` to find the next `/` on Posix, which allows the use of SIMD implementations provided by the C runtime (e.g. through IFUNC functions on Linux), resulting in very noticeable speedup. This is also why a statically Ninja executable will be slower than one that links to the C library dynamically :-/

- Avoid performing any writes when the input path doesn't need any adjustment, which is also quite common.

Note that this patch does _not_ remove the 64-bit limit for the `slash_bits`  value, which is only used on Win32.

Benchmarking was done in several ways:

- On Linux, running `hyperfine canon-perftest` to run the canonicalization benchmark program and measure its total running time. Three compilers were used to generate dynamically-linked executables.

  ```
               BEFORE (ms)  AFTER (ms)

  GCC 13.2.0      651         369
  Clang 14.0      591         402
  Clang 18,0      653         400
  ```

- On Windows, running `canon-perftest` 5 times and keeping the best reported average result. The number are slower since they only measure the benched function.

  ```
                  BEFORE (ms)  AFTER (ms)

  Mingw64 GCC 12      246         195
  ```

- On Linux, run `hyperfine ninja -C out/default -n --quiet` on a large Fuchsia build plan, once with 70000+ pending commands, and once after the build (i.e. `ninja: no work to do`).

  ```` 
              BEFORE (s)  AFTER (s)

  pre_build       8.789    8.647
  post_build      6.703    6.590
  ```

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -143,20 +143,19 @@ void CanonicalizePath(char* path, size_t* len, uint64_t* slash_bits) {
     return;
   }
 
-  const int kMaxPathComponents = 60;
-  char* components[kMaxPathComponents];
-  int component_count = 0;
-
   char* start = path;
   char* dst = start;
+  char* dst_start = dst;
   const char* src = start;
   const char* end = start + *len;
+  const char* src_next;
 
+  // For absolute paths, skip the leading directory separator
+  // as this one should never be removed from the result.
   if (IsPathSeparator(*src)) {
 #ifdef _WIN32
-
-    // network path starts with //
-    if (*len > 1 && IsPathSeparator(*(src + 1))) {
+    // Windows network path starts with //
+    if (src + 2 <= end && IsPathSeparator(src[1])) {
       src += 2;
       dst += 2;
     } else {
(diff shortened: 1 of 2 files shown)


digit-google:
Note that this PR replaces and deprecates https://github.com/ninja-build/ninja/pull/2108 that I had uploaded a long time ago to only fix one of the issues handled here.

src/util_test.cc
```diff
@@ -357,7 +446,7 @@ TEST(CanonicalizePath, NotNullTerminated) {
   len = strlen("foo/../file");
   CanonicalizePath(&path[0], &len, &unused);
   EXPECT_EQ(strlen("file"), len);
-  EXPECT_EQ("file ./file bar/.", string(path));
+  EXPECT_EQ("file../file bar/.", string(path));
```
jhasse, line 449: Hm ... should the unit test really care about the rest of the string?

src/util.cc
```diff
@@ -167,62 +166,138 @@ void CanonicalizePath(char* path, size_t* len, uint64_t* slash_bits) {
...
-        *c = '/';
-        NINJA_FALLTHROUGH;
-      case '/':
-        bits_mask <<= 1;
+    case '\\':
+      bits |= bits_mask;
+      *c = '/';
+      NINJA_FALLTHROUGH;
+    case '/':
+      bits_mask <<= 1;
```
jhasse, line 300: Can you keep this as before? (although it's formatted wrong)

jhasse (changes requested):

src/util_test.cc
digit-google: Yes, the point is to verify that this does not touch the rest of the string, including from the input (hence the caller should add a terminating zero if it needs it). I've added a comment to clarify.

src/util.cc
digit-google: Sure, it was all `clang-format` but I've reverted these line changes. :-)

digit-google:
Ping. I believe the requested changes have been applied 3 weeks ago. Let me know if you need more.

PR #2360, opened 2023-12-08
[FIX] compile: gcc version > 11.3, treat -Wmaybe-uninitialized as error

Simonhancrew:
my gcc version
```bash
gcc (SUSE Linux) 13.2.1 20231130 [revision 741743c028dc00f27b9c8b1d5211c1f602f2fddd]
Copyright (C) 2023 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE
```
compile ninja, 1 error occurs
```bash
ninja/build/_deps/googletest-src/googletest/src/gtest-death-test.cc:1301:24: error: ‘dummy’ may be used uninitialized [-Werror=maybe-uninitialized]
 1301 |   StackLowerThanAddress(&dummy, &result);
      |   ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~
```
just try to ignore this error

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -252,6 +252,16 @@ if(BUILD_TESTING)
     if (NOT TARGET GTest::gtest)
       message(STATUS "Defining GTest::gtest alias to work-around bug in older release.")
       add_library(GTest::gtest ALIAS gtest)
+
+      # NOTE: gtest uninit some variables, gcc >= 1.11.3 may cause error on compile.
+      # Remove this comment and six lines below, once ninja deps gtest-1.11.0 or above.
+      if (CMAKE_COMPILER_IS_GNUCXX AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER "1.11.3")
+        check_cxx_compiler_flag(-Wno-maybe-uninitialized flag_no_maybe_uninit)
+        if (flag_no_maybe_uninit)
+          target_compile_options(gtest PRIVATE -Wno-maybe-uninitialized)
+        endif()
+      endif()
+
     endif()
   endif()
 


CMakeLists.txt
```diff
@@ -37,6 +37,11 @@ else()
 		add_compile_options(-fdiagnostics-color)
 	endif()
 
+	check_cxx_compiler_flag(-Wno-maybe-uninitialized flag_no_maybe_uninit)
+	if (flag_no_maybe_uninit)
+	add_compile_options(-Wno-maybe-uninitialized)
```
digit-google, line 42: Can you use `target_compile_options` to ensure this applies to the GTest library only. It is better to keep the warning for Ninja and fix the errors in the source code directly instead.

Also please add a comment that explains why this section exists in the first place, e.g. it can be removed in the future, for example once googlegtest is upgraded to a newest release that doesn't have this problem. Ideally a version check here would be a good idea too, but I don't know if this is possible here.

Simonhancrew: hh, thanks for your advice, works before looks aggresive a little bit.

CMakeLists.txt
```diff
@@ -237,6 +237,18 @@ if(BUILD_TESTING)
     )
     FetchContent_MakeAvailable(googletest)
 
+		# NOTE: gtest uninit some variables, gcc >= 1.11.3 may cause error on compile. 
+		# remove this func and func call, once ninja deps gtest-1.12.0 or above  
+		function(maybe_add_compile_flags)
```
digit-google, line 242: Thank you for the latest change. As an additional nit, I would suggest naming this `maybe_add_gtest_compile_flags` for clarity :-)

Node that I don't have write access to the repository, invoking @jhasse here for his feedback.

digit-google (approved):

CMakeLists.txt
jhasse: I'm wondering why using a function at all? Couldn't this code be inlined below?

Simonhancrew: cause i wanna make inside part looks more clear

jhasse: hm ... I would rather keep the added lines at a minimum. Can you inline it and also remove the ` message`?

Simonhancrew: sure

jhasse:
Awesome :) Can you replace the tabs in your change with 2 spaces? Then this can be merged.

Simonhancrew:
> Awesome :) Can you replace the tabs in your change with 2 spaces? Then this can be merged.
sorry, my bad QAQ

Simonhancrew:
> Awesome :) Can you replace the tabs in your change with 2 spaces? Then this can be merged.

sorry, my bad, QAQ

CMakeLists.txt
```diff
@@ -252,6 +252,16 @@ if(BUILD_TESTING)
...
       add_library(GTest::gtest ALIAS gtest)
+
+      # NOTE: gtest uninit some variables, gcc >= 1.11.3 may cause error on compile.
+      # Remove this comment and six lines below, once ninja deps gtest-1.12.0 or above.
+      if (CMAKE_COMPILER_IS_GNUCXX AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER "1.11.3")
+        check_cxx_compiler_flag(-Wno-maybe-uninitialized flag_no_maybe_uninit)
+        if (flag_no_maybe_uninit)
+          target_compile_options(gtest PRIVATE -Wno-maybe-uninitialized)
+        endif()
+      endif()
```
digit-google, line 263: Can you put this section after the endif() for the GTest::gtest check? Otherwise it will not be interpreted when compiling against a version of GTest that is more recent than 1.10 (such as 1.11 which would still need this)

Simonhancrew: actually, 1.11 is fixed. see [this pr]( https://github.com/google/googletest/pull/3024/commits/4679637f1c9d5a0728bdc347a531737fad0b1ca3)

digit-google: Ah, good to know. In this case please update the comment to reference 1.11 instead of 1.12 :)

Simonhancrew: sure, it's ok now

PR #2385, opened 2024-02-18
added missing include guards in order to support easy amalgamation

icebreaker:
These changes make it easy to build an amalgamated `ninja.cc` that can be used to bootstrap `ninja` with just a working C++ compiler, without the need for any third-party tools like `cmake` or `python`.

## *nix

```bash
c++ -O2 src/ninja_amalgamated.cc -o ninja
```

## osx-cross

```bash
x86_64-apple-darwin19-c++ -O2 src/ninja_amalgamated.cc -o ninja
```

## mingw

```bash
x86_64-w64-mingw32-c++ -O2 src/ninja_amalgamated.cc -o ninja.exe
```

## msvc

```bash
cl.exe /nologo /Ox /GR- src\ninja_amalgamated.cc /out:ninja.exe
```

### DIFF
--- a/src/includes_normalize.h
+++ b/src/includes_normalize.h
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#ifndef INCLUDES_NORMALIZE_H_
+#define INCLUDES_NORMALIZE_H_
+
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


digit-google:
Thank you, this looks good to me. @jhasse what do you think?

jhasse:
I would prefer `#pragma once`, otherwise LGTM.

icebreaker:
I initially wanted to go with `#pragma once`, but since all the other headers do it the old fashioned way, I decided to stay consistent.

PR #2390, opened 2024-03-13
Gracefully handle outdated .ninja_log during '-t recompact'

### LINKED ISSUE #2048: Recompact fail with "failed recompaction: No such file or directory"
In rare cases, on our builds we got this error during `recompact` step:

```
'C:/ProgramData/chocolatey/bin/ninja.exe' '-C' '<SOME_PATH>' '-t' 'recompact'
    failed with:
      ninja: warning: bad deps log signature or version; starting over
    ninja: error: failed recompaction: No such file or directory  
```

Trying to trace back why this is failing, we see that:

- `DepsLog::Load` delete `ninja_deps` and return `LOAD_SUCCESS` if version mismatch or headers are corrupted.
- `OpenDepsLog(/*recompact_only=*/ true)` doesn't stop because on`LOAD_SUCCESS`
- `DepsLog::Recompact` fail because of the missing `ninja_deps` file.

In our case we need to find why the file loading fail (corrupted file ?) but I wonder if this is intentionnal to fail in this case ? 

- if no, returning a `LOAD_NOT_FOUND` instead of `LOAD_SUCCESS` will skip the recompaction since the file was deleted.
- if yes, a more explicit error message could help

thanks,

jheydebrand:
Fixes #2048

I ran into a similar issue when migrating existing workspaces from ninja 1.9.0 to the latest revision from Git.

This change will cause `NinjaMain::OpenBuildLog()` to not call `build_log_.Recompact()` after `build_log_.Load()` just deleted that file because it contained an invalid log version.

### DIFF
--- a/misc/output_test.py
+++ b/misc/output_test.py
@@ -127,6 +127,29 @@ build a: cat
         self.assertEqual(run('', flags='-t recompact'), '')
         self.assertEqual(run('', flags='-t restat'), '')
 
+    def test_issue_2048(self):
+        with tempfile.TemporaryDirectory() as d:
+            with open(os.path.join(d, 'build.ninja'), 'w'):
+                pass
+
+            with open(os.path.join(d, '.ninja_log'), 'w') as f:
+                f.write('# ninja log v4\n')
+
+            try:
+                output = subprocess.check_output([NINJA_PATH, '-t', 'recompact'],
+                                                 cwd=d,
+                                                 env=default_env,
+                                                 stderr=subprocess.STDOUT,
+                                                 text=True
+                                                 )
+
+                self.assertEqual(
+                    output.strip(),
+                    "ninja: warning: build log version is too old; starting over"
+                )
+            except subprocess.CalledProcessError as err:
+                self.fail("non-zero exit code with: " + err.output)
+
     def test_status(self):
         self.assertEqual(run(''), 'ninja: no work to do.\n')
         self.assertEqual(run('', pipe=True), 'ninja: no work to do.\n')
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -291,9 +291,9 @@ LoadStatus BuildLog::Load(const string& path, string* err) {
       if (invalid_log_version) {
         fclose(file);
         unlink(path.c_str());
-        // Don't report this as a failure.  An empty build log will cause
+        // Don't report this as a failure. A missing build log will cause
         // us to rebuild the outputs anyway.
-        return LOAD_SUCCESS;
+        return LOAD_NOT_FOUND;
       }
     }
 


digit-google:
Thank you. The fix looks good to me, but can you add a regression test in misc/output_test.py to verify that this works correctly?

E.g.: perform a build, then overwrite the version bytes in the build log, then run the recompact tool and verify it does not report an error?

jheydebrand:
I added a test as suggested, @digit-google
However, I could run it only on Windows locally - hopefully test test runs fine on Linux, too.

misc/output_test.py
```diff
@@ -130,6 +130,31 @@ def test_pr_1685(self):
         self.assertEqual(run('', flags='-t recompact'), '')
         self.assertEqual(run('', flags='-t restat'), '')
 
+    def test_pr_2390(self):
```
digit-google, line 133: Can you renamed this `test_issue_2048` instead? Using an issue number makes more sense than a pull request one in this context.

misc/output_test.py
```diff
@@ -130,6 +130,31 @@ def test_pr_1685(self):
...
+            with open('build.ninja', 'w') as f:
+                pass
+
+            with open('.ninja_log', 'w') as f:
+                f.write('# ninja log v4\n')
+
+            try:
+                output = subprocess.check_output([NINJA_PATH, '-t', 'recompact'],
+                                                 shell=True,
+                                                 env=default_env,
```
digit-google, line 146: nit: use `text = True` here to avoid the `decode('utf-8')` below.

misc/output_test.py
```diff
@@ -130,6 +130,31 @@ def test_pr_1685(self):
...
+
+            with open('build.ninja', 'w') as f:
+                pass
+
+            with open('.ninja_log', 'w') as f:
+                f.write('# ninja log v4\n')
+
+            try:
+                output = subprocess.check_output([NINJA_PATH, '-t', 'recompact'],
+                                                 shell=True,
```
digit-google, line 145: nit: `shell = True` is not necessary here.

misc/output_test.py
```diff
@@ -130,6 +130,31 @@ def test_pr_1685(self):
         self.assertEqual(run('', flags='-t recompact'), '')
         self.assertEqual(run('', flags='-t restat'), '')
 
+    def test_pr_2390(self):
+        with tempfile.TemporaryDirectory() as d:
+            os.chdir(d)
```
digit-google, line 135: FYI, there is a subtle issue when changing the current directory to the temporary one, as in reality this prevents the temporary directory from being removed on scope exit! Also this breaks pytest when adding other _test.py files in the misc/ directory.

This problem already exists in the run() function above, and I'll submit a patch soon to fix this. Can I ask you to avoid this os.chdir() completely here by doing the following:

1) Use `os.path.join(d, 'build.ninja')` and `os.path.join(d, `.ninja_log`)` when writing these files below.
2) Use `cwd=d` when invoking subprocess.check_output()? This works because `NINJA_PATH` is always an absolute path (see line 21 above).

digit-google:
Thank you, this looks good to me. @jhasse what do you think?

jheydebrand:
Should I try to work around Python 3.6 in AppVecyoe not knowing `text=True`?

jhasse:
Dropping Ubuntu 14.04, 16.04 and 18.04 in favor of 20.04, 22.04 and 24.04 is something I would want to do soon, so you might also just wait. But I think we'd also have to run these tests on RHEL 7 (or 8 soon), I don't know what Python version they are using.

jheydebrand:
Looks like RHEL 7 supports Python 3.8 per https://access.redhat.com/documentation/en-us/red_hat_software_collections/3/html-single/3.5_release_notes/index#sect-RHSCL-Changes-python

Depending on your timeline for dropping the older distros, I'd offer to get the test to work with the older Python version to wrap this PR up.

digit-google:
Aargh, I was not aware of this limitation, remove the `text=True` for now then, this is not a good reason to block this PR. Thanks.

jhasse:
See #2399

jhasse:
Can you rebase against / merge master?

jheydebrand:
Looking good now :)

src/build_log.cc
```diff
@@ -293,7 +293,7 @@ LoadStatus BuildLog::Load(const string& path, string* err) {
         unlink(path.c_str());
         // Don't report this as a failure.  An empty build log will cause
```
jhasse, line 294: This comments needs to be updated / removed now, right?

jhasse (changes requested):

src/build_log.cc
jheydebrand: I think the comment is still correct when comparing it with the other `return LOAD_NOT_FOUND`. Maybe change "empty" into "missing"?

digit-google:
LGTM now, thank you.

PR #2391, opened 2024-03-13
Support empty depfiles.

### LINKED ISSUE #2357: depfile consisting exclusively of newlines causes an error
I'm currently working on adding support for using depfiles with Vala compiler in meson. I got it mostly working - however, if files just so happen to have no dependencies (unlikely in real projects but happens for two of meson's tests), `valac` generates a depfile consisting exclusively of two newline characters (I've double-checked with a hex editor that it's really just that).

Ninja fails to parse such a file with the following error:

```
expected ':' in depfile
```

digit-google:
Some tools generate completely empty depfiles when there are no implicit inputs. This patch modifies the depfile parser to support them. Fixes #2357

### DIFF
--- a/src/depfile_parser.cc
+++ b/src/depfile_parser.cc
@@ -54,6 +54,7 @@ bool DepfileParser::Parse(string* content, string* err) {
   bool have_target = false;
   bool parsing_targets = true;
   bool poisoned_input = false;
+  bool is_empty = true;
   while (in < end) {
     bool have_newline = false;
     // out: current output point (typically same as in, but can fall behind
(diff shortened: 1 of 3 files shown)


jhasse:
Thanks!

Can you change not_empty to empty to avoid double negative?

Do we have any tests for the "expected ':' in depfile" error message?

digit-google:
Good points, changed the variable name to `is_empty` and added a unit-test for an invalid but non-empty file.

jhasse:
Thanks! :)

PR #2392, opened 2024-03-13
lexer: allow TABS for indentation

### LINKED ISSUE #1598: Is there a good reason to err on tabs in build files after all?
Seen the closed issue #394 as "intended behavior", but even more significantly the open one (from 4 years later): #952. Quoting @Alhadis's comment from there:

> You might consider lifting the ban on tabs, then, because enforcing a specific style 
> of formatting is just repeating Make's mistake.

Yes, please. Or at least avoid "silently" erring on tabs (by "silently" I mean confusing error messages like "a line with 'command =' is expected", for a line that apparently reads just that; it was my first obstacle (well, second, after a "linting error" for a BOM) that I stumbled across in my hello-world attempt to evaluate whether I should consider switching to Ninja (until switching to redo :) ). 

BTW, realizing that the tab might be the problem was made even less obvious by the explicit notion that Ninja is (trying to be) a close kin of make.

And regarding @Alhadis's other point: "The lexer should raise an error only if tabs and spaces are mixed within the same file, which (of course) should be forbidden." -- considering Evan's comment of

> Re 3, ... nested scopes of indenting (like Haskell "where" clauses) and
> then it never ended up being useful. I agree the docs could remove all mention of 
> parent and simplify to just "indented or not".

then even mixing tabs and spaces should be none of Ninja's business, as far as I understand. Until e.g. nested scopes may get introduced, of course. But even then, only an indent string of mixed tabs and spaces at line-start should be a problem (not any mixing in general).

Thanks, and a nice job overall!

_(Note: I'm all new to Ninja, and ended up here only due to my own tab affair, as the [manual](https://ninja-build.org/manual.html) is still silent about tabs -- which is fine, as in this case possibly many of us would prefer changing the code, not the docs.)_

pkitszel:
Allow TAB character to be used for indentation.

This is useful to have TAB character used as indentation, especially when parts of build.ninja are hand-written as HEREDOCs in otherwise TAB-indented file (either mandated by style for other part of project, or required by language itself).

Changing lexer is easy thanks to the use of re2c, syntax is perhaps a bit too permissive now, but that is job of the parser to reject use of mixed indentation.

Let's stop complaining that:
ninja: error: build.ninja:3: expected 'command =' line when it is exactly:
	command = cc $cflags -c $in -o $out

Closes #1598
Signed-of-by: Przemek Kitszel <pkitszel@gmail.com>

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -861,7 +861,13 @@ one_word_with_no_space = foo$
 Other whitespace is only significant if it's at the beginning of a
 line.  If a line is indented more than the previous one, it's
 considered part of its parent's scope; if it is indented less than the
-previous one, it closes the previous scope.
+previous one, it closes the previous scope. Note that the physical line that is
+a line continuation is not considered as a separate line for the scope
+determination purposes.
+
+_Available since Ninja 1.13._
+Tab character can be used for indentation. Earlier Ninja versions allow only
+space character to be used for such purpose.
 
 [[ref_toplevel]]
 Top-level variables
--- a/misc/output_test.py
+++ b/misc/output_test.py
@@ -191,6 +191,30 @@ out1
 out2
 ''')
 
+    def test_tabs_indent(self):
+        content = '''
+rule exec
+<TAB>command = $cmd
+
+var_hello = hell$
+<TAB>o
+
+build foo: exec
+<TAB>cmd = touch foo
+
+build bar: exec $
+<TAB>foo
+<TAB>cmd = touch bar
+
+build $var_hello: exec
+<TAB>cmd = touch $var_hello
+
+build baz: exec $
+<TAB>bar $var_hello
+<TAB>cmd = touch baz
+'''.replace('<TAB>', '\t')
+        run(content)
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/src/lexer.in.cc
+++ b/src/lexer.in.cc
@@ -104,12 +104,9 @@ const char* Lexer::TokenErrorHint(Token expected) {
 
 string Lexer::DescribeLastError() {
   if (last_token_) {
-    switch (last_token_[0]) {
-    case '\t':
-      return "tabs are not allowed, use spaces";
-    }
+	return "lexing error <last token:"+string(last_token_)+">";
   }
-  return "lexing error";
+  return "lexing error (EOF?)";
 }
 
 void Lexer::UnreadToken() {
@@ -133,10 +130,10 @@ Lexer::Token Lexer::ReadToken() {
     simple_varname = [a-zA-Z0-9_-]+;
     varname = [a-zA-Z0-9_.-]+;
 
-    [ ]*"#"[^\000\n]*"\n" { continue; }
-    [ ]*"\r\n" { token = NEWLINE;  break; }
-    [ ]*"\n"   { token = NEWLINE;  break; }
-    [ ]+       { token = INDENT;   break; }
+    [ \t]*"#"[^\000\n]*"\n" { continue; }
+    [ \t]*"\r\n" { token = NEWLINE;  break; }
+    [ \t]*"\n"   { token = NEWLINE;  break; }
+    [ \t]+       { token = INDENT;   break; }
     "build"    { token = BUILD;    break; }
     "pool"     { token = POOL;     break; }
     "rule"     { token = RULE;     break; }
@@ -175,7 +172,7 @@ void Lexer::EatWhitespace() {
   for (;;) {
     ofs_ = p;
     /*!re2c
-    [ ]+    { continue; }
+    [ \t]+  { continue; }
     "$\r\n" { continue; }
     "$\n"   { continue; }
     nul     { break; }
@@ -241,10 +238,10 @@ bool Lexer::ReadEvalString(EvalString* eval, bool path, string* err) {
       eval->AddText(StringPiece(" ", 1));
       continue;
     }
-    "$\r\n"[ ]* {
+    "$\r\n"[ \t]* {
       continue;
     }
-    "$\n"[ ]* {
+    "$\n"[ \t]* {
       continue;
     }
     "${"varname"}" {
--- a/src/lexer_test.cc
+++ b/src/lexer_test.cc
@@ -17,12 +17,20 @@
 #include "eval_env.h"
 #include "test.h"
 
-using namespace std;
+std::string tok(Lexer::Token t) {
+  const char *str = Lexer::TokenName(t);
+  if (!str)
+    return "TokenOutOfRange: " + std::to_string(t);
+  return str;
+}
+
+#define EXPECT_EQ_TOK(t1, t2) \
+  EXPECT_EQ(tok(t1), tok(t2))
 
 TEST(Lexer, ReadVarValue) {
   Lexer lexer("plain text $var $VaR ${x}\n");
   EvalString eval;
-  string err;
+  std::string err;
   EXPECT_TRUE(lexer.ReadVarValue(&eval, &err));
   EXPECT_EQ("", err);
   EXPECT_EQ("[plain text ][$var][ ][$VaR][ ][$x]",
@@ -32,7 +40,7 @@ TEST(Lexer, ReadVarValue) {
 TEST(Lexer, ReadEvalStringEscapes) {
   Lexer lexer("$ $$ab c$: $\ncde\n");
   EvalString eval;
-  string err;
+  std::string err;
   EXPECT_TRUE(lexer.ReadVarValue(&eval, &err));
   EXPECT_EQ("", err);
   EXPECT_EQ("[ $ab c: cde]",
@@ -41,7 +49,7 @@ TEST(Lexer, ReadEvalStringEscapes) {
 
 TEST(Lexer, ReadIdent) {
   Lexer lexer("foo baR baz_123 foo-bar");
-  string ident;
+  std::string ident;
   EXPECT_TRUE(lexer.ReadIdent(&ident));
   EXPECT_EQ("foo", ident);
   EXPECT_TRUE(lexer.ReadIdent(&ident));
@@ -56,12 +64,12 @@ TEST(Lexer, ReadIdentCurlies) {
   // Verify that ReadIdent includes dots in the name,
   // but in an expansion $bar.dots stops at the dot.
   Lexer lexer("foo.dots $bar.dots ${bar.dots}\n");
-  string ident;
+  std::string ident;
   EXPECT_TRUE(lexer.ReadIdent(&ident));
   EXPECT_EQ("foo.dots", ident);
 
   EvalString eval;
-  string err;
+  std::string err;
   EXPECT_TRUE(lexer.ReadVarValue(&eval, &err));
   EXPECT_EQ("", err);
   EXPECT_EQ("[$bar][.dots ][$bar.dots]",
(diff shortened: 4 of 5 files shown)


src/lexer.in.cc
```diff
@@ -133,10 +130,10 @@ Lexer::Token Lexer::ReadToken() {
...
     varname = [a-zA-Z0-9_.-]+;
 
-    [ ]*"#"[^\000\n]*"\n" { continue; }
-    [ ]*"\r\n" { token = NEWLINE;  break; }
-    [ ]*"\n"   { token = NEWLINE;  break; }
-    [ ]+       { token = INDENT;   break; }
+    [ \t]*"#"[^\000\n]*"\n" { continue; }
+    [ \t]*"\r\n" { token = NEWLINE;  break; }
+    [ \t]*"\n"   { token = NEWLINE;  break; }
+    [ \t]+       { token = INDENT;   break; }
```
digit-google, line 136: Thanks. I need this is not enough since you also need to handle line continuations.

It looks like this requires also modifying the definitions of EatWhiteSpace and ReadEvalString below, for example modify `$\r\n"[ ]*` into `$\r\n"[ \t]*` and `$\n"[ ]*` into `$\r\n"[ ]*`.

src/lexer_test.cc
```diff
@@ -89,10 +89,10 @@ TEST(Lexer, CommentEOF) {
...
 TEST(Lexer, Tabs) {
   // Verify we print a useful error on a disallowed character.
-  Lexer lexer("   \tfoobar");
+  Lexer lexer("\tcommand = foo $in");
   Lexer::Token token = lexer.ReadToken();
   EXPECT_EQ(Lexer::INDENT, token);
+
   token = lexer.ReadToken();
-  EXPECT_EQ(Lexer::ERROR, token);
-  EXPECT_EQ("tabs are not allowed, use spaces", lexer.DescribeLastError());
```
digit-google, line 97: Please add tests to also verify that:

- line continuations with tabs actually work. E.g.:

```
build foo: rule_name $
<tab>input1 $
<tab>input2
```
- Tabs that appear after indentation are *still* not allowed, except when they are in comments. E.g.

rule correct_rule
<tab>command = foo bar

rule incorrect_rule
<tab>command = foo<tab>bar

build incorrect_build: rule_name<tab>input1<tab>input2
```

The most important reason to avoid tabs is to avoid using them in `command = ` definitions because this will be equally confusing to users.

Finally, please update manual.doc appropriately to document the new behavior.

Comment (pkitszel) [reply src/lexer_test.cc:97]: Thank you, I have indeed tweak the indentation after a linebreak.

However,  even without my changes ` command = foo<tab>bar` is fine to use.

File: src/lexer_test.cc [hunk 3]
```diff
@@ -19,6 +19,16 @@
 
 using namespace std;
 
+string tok(Lexer::Token t) {
```
digit-google, line 22: nit: Ninja sources are now using std:: namespace prefixes, can you change this to use `std::string` and `std::to_string` here and below?

src/lexer.in.cc
```diff
@@ -104,12 +104,9 @@ const char* Lexer::TokenErrorHint(Token expected) {
 
 string Lexer::DescribeLastError() {
   if (last_token_) {
-    switch (last_token_[0]) {
-    case '\t':
-      return "tabs are not allowed, use spaces";
-    }
+	return "lexing error <"+string(last_token_)+">";
```
digit-google, line 107: You might want to add a static Lexer::TokenToString() method (which would implement the same code as the one you added for the test in the second commit) and use this here instead.

digit-google:
Thank you, I added two small comments. Can you also squash the result into a single commit, this will keep git history simpler for something small like this.

I am not sure about your local test file under misc/, since it is not part of this PR. Normally `pytest-3` is invoked in the CMake build directory, pointing to the root source directory, and all `*_test.py` will be invoked automatically, including those under `misc/` (this is not ideal imho, since it can pick up broken test files from third-party dependencies). See the content of the files under `.github/workflows/` for details.

Finally, I don't know about `tpm` files, what are these?

src/lexer.in.cc
pkitszel: this is for error (reason), not token (type), IOW valid only for Token==ERROR

src/lexer_test.cc
pkitszel: sure

src/lexer.in.cc
digit-google: Oh, I see, thanks for the clarification then :)

pkitszel:
> Thank you, I added two small comments. Can you also squash the result into a single commit, this will keep git history simpler for something small like this.

Sure, the bigger commit already does two things, which is a tid bit above my liking, but I agree that the smaller one does not really decouple anything substantial :)
 
> I am not sure about your local test file under misc/, since it is not part of this PR. Normally `pytest-3` is invoked in the CMake build directory, pointing to the root source directory, and all `*_test.py` will be invoked automatically, including those under `misc/` (this is not ideal imho, since it can pick up broken test files from third-party dependencies). See the content of the files under `.github/workflows/` for details.
> 
> Finally, I don't know about `tpm` files, what are these?

ugh, sorry, 'tmp' files, for temporary

the file would be `misc/tabs-test.ninja` , and the passing criteria is that `./ninja -v -f ./misc/tabs-test.ninja` runs fine

I would be happy to extend current PR to add this test (perhaps without temporaries even) if those "misc/*.ninja" tests are already executed somehow (I didn't found that).

But ninja_syntax.py is a generator, that does not need to use TABs, so no point in changing it (and it's existing tests should cover what we have in this PR).

src/lexer_test.cc
pkitszel: sorry for double push, forgot about that the first time

misc/output_test.py
```diff
@@ -162,6 +162,30 @@ def test_tool_inputs(self):
 out2
 ''')
 
+    def test_tabs_indent(self):
+        content = '''
```
digit-google, line 166: nit: It is really difficult to see the tabs in this input. I suggest you use explicit `<TAB>` markers and `str.replace()` to replace them with real tabs, as in:

```
    # For clarity, tabs are written as <TAB> in the literal string below, then
    # replaced with a real "\t" value.
    content = '''
    rule exec
    <TAB>command = $cmd
    ....
    """.replace("<TAB>", "\t")
```

src/lexer_test.cc
digit-google: Ack, then no need to test for this then. Thank you for double-checking.

digit-google:
No worries for the double-push. Can you modify `doc/manual.asciidoc` to mention that TABs are allowed since 1.12 (next planned release IIRC).

pkitszel:
> No worries for the double-push. Can you modify `doc/manual.asciidoc` to mention that TABs are allowed since 1.12 (next planned release IIRC).

I have no idea if it is safe to left $<TAB> escape rules not specified or I should go further this route ;)

OTOH I have spot a problem that some (TAB containing) variables that are used without qouting are passed as two params to the final command, sounds like a problem. (Not present if there is space instead of TAB)


so, please do not merge yet

pkitszel:
> so, please do not merge yet

I was somehow testing with a system `ninja` instead of local build - turns out that code is fine as is.

I did a rebase + applied <TAB> suggestion in test to make it visible.

doc/manual.asciidoc
```diff
@@ -861,7 +861,13 @@ one_word_with_no_space = foo$
 Other whitespace is only significant if it's at the beginning of a
 line.  If a line is indented more than the previous one, it's
 considered part of its parent's scope; if it is indented less than the
-previous one, it closes the previous scope.
+previous one, it closes the previous scope. Note that the physical line that is
+a line continuation is not considered as a separate line for the scope
+determination purposes.
+
+_Available since Ninja 1.12._
+Horizontal Tab character (0x09) could be used for indentation. Earlier Ninja
```
digit-google, line 869: nit: s/could/can/ here :)

digit-google:
Nicely done, thanks.

pkitszel:
BTW, is there any prefered way to measure performance impact of any change/PR?

doc/manual.asciidoc
```diff
@@ -861,7 +861,13 @@ one_word_with_no_space = foo$
...
 line.  If a line is indented more than the previous one, it's
 considered part of its parent's scope; if it is indented less than the
-previous one, it closes the previous scope.
+previous one, it closes the previous scope. Note that the physical line that is
+a line continuation is not considered as a separate line for the scope
+determination purposes.
+
+_Available since Ninja 1.12._
+Horizontal Tab character (0x09) can be used for indentation. Earlier Ninja
+versions allow only space (0x20) character to be used for such purpose.
```
jhasse, line 870: Hm ... while technically correct, I find these unnecessary. Every programmer (that is writing Ninja files by hand) should know what "tabs" and "spaces" are. I would simplify it:

"Horizontal Tab character (0x09)" -> Tabs
"space (0x20) character" -> spaces

Opinions?

doc/manual.asciidoc
```diff
@@ -861,7 +861,13 @@ one_word_with_no_space = foo$
 Other whitespace is only significant if it's at the beginning of a
 line.  If a line is indented more than the previous one, it's
 considered part of its parent's scope; if it is indented less than the
-previous one, it closes the previous scope.
+previous one, it closes the previous scope. Note that the physical line that is
+a line continuation is not considered as a separate line for the scope
+determination purposes.
+
+_Available since Ninja 1.12._
```
jhasse, line 868: Can you change this to 1.13 now?

doc/manual.asciidoc
pkitszel: sure

doc/manual.asciidoc
pkitszel: sure, sorry for the delay

jhasse:
I'm not sure if this is a good idea yet, since it will result in breaking backwards compatibility (all build.ninja files with tabs won't work with older ninja versions). Please correct me if I'm wrong.

My suggestion would be to still put out a warning when there are tabs, but not fail. Then in the next release remove the warning.

pkitszel:
> I'm not sure if this is a good idea yet, since it will result in breaking backwards compatibility (all build.ninja files with tabs won't work with older ninja versions). Please correct me if I'm wrong.
> 
> My suggestion would be to still put out a warning when there are tabs, but not fail. Then in the next release remove the warning.

I believe that users/generators should emit minimal version required.

jhasse:
> I believe that users should describe/generators should emit the minimal version required.

Yes, but we would need to implement a minimal parser that just reads that value and doesn't choke on anything else he doesn't know.

jhasse:
Just thought again about this: If ninja_required_version is the first line it doesn't matter of an older version can't parse the rest of the file. So I think we should allow tabs if and only if ninja_required_version is set to 1.13 (or probably 1.14 since I want to get 1.13 ready soon). Or maybe put out a warning that using tab will make the script incompatible with older Ninja versions and suggest to add `ninja_required_version = 1.13`.

pkitszel:
So, in general it would be best if users specify the required version always (and likely they would for anything substantial (and generators will do too)). For "small/PoC/throwaway" projects it likely would not matter (as there will be just one user/version).

(discussion/rant):
Adding a warning would tempt users to convert back to spaces, and I think that would be unfortunate (and not the point of the PR). I guess that spaces came from initial implementation being in python, but that should not matter much for users, especially that tool is often used for C-like world, and often as a make replacement ;)

jhasse:
Can you fix the conflicts and change 1.13 to 1.14?

jhasse:
I want to release 1.14 soon, so I have to move this to the next milestone for now.

PR #2394, opened 2024-03-14
Fix crash when FormatMessageA sets lpBuffer to nullptr

### LINKED ISSUE #2393: Crash in GetLastErrorString(): FormatMessageA may set out-param lpBuffer to NULL
The call to https://github.com/ninja-build/ninja/blob/ab510c7a8cccbea0ea2c82531dc23893b551d55e/src/util.cc#L533 may set the local variable `msg_buf` to nullptr. This leads to an access violation and subsequent crash when that char pointer is assigned to the std::string (because it's constructor will try to perform strlen on that nullptr).

This happend to me on one of my machines. However, I did not bother to try to find out which Windows component is (presumably) missing that would restore the expected FormatMessageA behaviour.

If `FormatMessageA` was unable to provide a usable message, ninja should probably just format the `DWORD err` into a string.
I will create a pull request with this proposed fix.

grafik

jheydebrand:
Fixes #2393

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -540,6 +540,13 @@ string GetLastErrorString() {
         (char*)&msg_buf,
         0,
         NULL);
+
+  if (msg_buf == nullptr) {
+    char fallback_msg[128] = {0};
+    snprintf(fallback_msg, sizeof(fallback_msg), "GetLastError() = %d", err);
+    return fallback_msg;
+  }
+
   string msg = msg_buf;
   LocalFree(msg_buf);
   return msg;


src/util.cc
```diff
@@ -540,6 +540,13 @@ string GetLastErrorString() {
         (char*)&msg_buf,
         0,
         NULL);
+
+  if (msg_buf == nullptr) {
+    char fallback_msg[128] = {0};
+    sprintf(fallback_msg, "GetLastError() = %d", err);
```
digit-google, line 546: nit: simply return GetLastErrorString() which is defined in util.h

PR #2395, opened 2024-03-14
Minor fix to output_test.py

digit-google:
Do not use os.chdir() to change the current directory inside the run() function, as doing this prevents the temporary directory from being removed.

Moreover, this breaks pytest invocations when adding new regression test scripts in this directory (as done in other forks).

+ Use dict.pop() to undefine environment variables in `default_env` dictionary.


jhasse:
Thanks! TIL

PR #2398, opened 2024-03-15
CanonicalizePath: fix 'a/b/.._foo' -> 'a'  replacement

paboldin:


digit-google:
Thank you, the change looks good. Can you add a unit-test to util_test.cc to ensure we never regress in the future if we change this function?

paboldin:
Thanks for the feedback! Updated.

paboldin:
Can you please take a look at this once again?

digit-google:
Thank you, this looks good to me. The Windows/x64 failures seems to be a flake when downloading re2c. Not sure how to work around this? @jhasse do you have an idea?

jhasse:
Wait a few hours and then restart CI :)

PR #2215, opened 2022-11-12
Dyndep patches

dfgtyx:
This request contains two patches. The first adjusts "ninja -t cleandead" to account for dynamic dependencies. By neglecting dynamic dependencies, Ninja spuriously flags artifacts as "dead" thereby increasing the cost of rebuilds down the road. The second patch improves the "ninja clean" performance for projects with dynamic dependencies. With this patch, on one of my projects I have observed runtimes for "ninja clean" drop from 5 minutes to 25 seconds.

PR #2406, opened 2024-04-02
Fix cleanded with dyndeps

dcbaker:
Correctly account for dyndep files when cleaning dead, so that dynamic outputs aren't considered dead when they are not.

This is basically the same as #2215. I've used the commit messages that @bradking suggested in that PR for these commits.

### DIFF
--- a/src/clean.cc
+++ b/src/clean.cc
@@ -127,6 +127,7 @@ int Cleaner::CleanAll(bool generator) {
 int Cleaner::CleanDead(const BuildLog::Entries& entries) {
   Reset();
   PrintHeader();
+  LoadDyndeps();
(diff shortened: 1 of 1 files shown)


dcbaker:
I did take a swing at writing a unittest for this, but I'm a little stuck on how to get the dynamic information into the `Cleaner::CleanDead` method, after creating the dyndeps file it seems to load them, but the BuildLog doesn't list them so the cleaner doesn't even know that they exist and I can't figure out how to get them into the BuildLog

jhasse:
While a unittest would be nice, we will go without one for now. Especially around clean/cleandead the unittest coverage isn't good. Maybe some integration tests (i.e. misc/output_test.py) would work. Anyway, that's a different issue.

Thank you for the PR!

PR #2407, opened 2024-04-06
Implement type hints in `ninja_syntax.py`

Repiteo:
Deliberately left all functionality untouched, instead focusing on hinting the system that currently exists.

### DIFF
--- a/misc/ninja_syntax.py
+++ b/misc/ninja_syntax.py
@@ -23,37 +23,54 @@ use Python.
 
 import re
 import textwrap
+from io import TextIOWrapper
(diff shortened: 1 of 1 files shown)


digit-google:
Thank you, this looks fine to me. I can run `mypy` succesfully on that file with this PR applied locally. @jhasse, do you think we could add a `mypy` pass to the Github action script (or even to the CMake-based test suite?)

digit-google:
This doesn't change the code in any way, it just allows it to be checked with `mypy`, the Python type checker (running `mypy misc/ninja_syntax.py`, which reports no problems). So I would say go for it :-)

PR #2410, opened 2024-04-11
Properly support for Windows long file paths through Unicode Win32 API calls.

digit-google:
This is an attempt to properly support Windows long file paths by only using Win32 Unicode API calls, and getting rid of the ANSI ones (some of which are still plagued by MAX_PATH limitations, even when long paths support is enabled on the machine).

This is achieved in the following ways (see individual commits for details):

- Add `UTF8ToWin32Unicode()` and `Win32UnicodeToUTF8()` functions to util.h

- For simplicity, introduce SystemDiskInterface and NullDiskInterface, and make RealDiskInterface a derived class of SystemDiskInterface that adds a caching layer.
(this opens the door to other caching implementations on Posix).

- Augment the `DiskInterface` API to add `RenameFile()` and `OpenFile()` methods to ... rename a file (just like `rename() / _rename()`), and open an stdio FILE instance (just like `fopen()`).

- Modify the DiskInterface implementation to only use Unicode Win32 APIs.

- Modify sources that rely on direct calls to `fopen()`, `unlink()`, `rename()` to use a DiskInterface instance. This requires injecting such an instance in the constructors of BuildLog and DepsLog btw.

- Modify other misc Win32-specific sources to use Win32 Unicode API calls directly (e.g. subprocess-win32.cc)

NOTE: VirtualFileSystem::OpenFile() only supports read access for now. Unlike its real SystemDiskInterface implementation, it doesn't perform \r\n to \n conversion on Windows (do we really want this?). That is why many tests (e.g. build_log_test.cc) do not use it, but rely on a real DiskInterface instance. This can be fixed by another PR after this one.

This should fix issue #1900 once and for all (crossing fingers here :-))

### DIFF
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -21,12 +21,12 @@
 #endif
 
 #include "build_log.h"
-#include "disk_interface.h"
 
-#include <cassert>
 #include <errno.h>
 #include <stdlib.h>
 #include <string.h>
+
+#include <cassert>
 
 #ifndef _WIN32
 #include <inttypes.h>
@@ -72,8 +72,8 @@ BuildLog::LogEntry::LogEntry(const string& output, uint64_t command_hash,
     start_time(start_time), end_time(end_time), mtime(mtime)
 {}
 
-BuildLog::BuildLog()
-  : log_file_(NULL), needs_recompaction_(false) {}
+BuildLog::BuildLog(DiskInterface& disk_interface)
+    : disk_interface_(&disk_interface) {}
 
 BuildLog::~BuildLog() {
   Close();
@@ -137,7 +137,7 @@ bool BuildLog::OpenForWriteIfNeeded() {
   if (log_file_ || log_file_path_.empty()) {
     return true;
   }
-  log_file_ = fopen(log_file_path_.c_str(), "ab");
+  log_file_ = disk_interface_->OpenFile(log_file_path_, "ab");
   if (!log_file_) {
     return false;
   }
@@ -211,7 +211,7 @@ struct LineReader {
 
 LoadStatus BuildLog::Load(const string& path, string* err) {
   METRIC_RECORD(".ninja_log load");
-  FILE* file = fopen(path.c_str(), "r");
+  FILE* file = disk_interface_->OpenFile(path, "r");
   if (!file) {
     if (errno == ENOENT)
       return LOAD_NOT_FOUND;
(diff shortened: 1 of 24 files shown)


JRStolle:
I tried this fork, but I still get the ninja error "Filename longer than 260 characters" on my jenkins when compiling my cmake project. 
Is there any way, I could supply some more meaningful debug-infos for this?

marius-alex-tache:
I've also tested this in an attempt to fix the issue when building a gn project, but it doesn't work. Same question as @JRStolle.

markh-discord:
During early testing this fix seems to work for us (although other tooling outside of ninja may eventually fail too). However we did bump into an issue where many dependency targets were incorrectly re-run after the first invocation.

I added a fix on our side to resolve: https://github.com/discord/ninja/commit/c5996dce0fbc0a501a324f359c9e65bdd1a9243a

digit-google:
Thank you @markh-discord , I rebased and integrated a similar fix for the weird GetFullPathnameW behavior. Let me know if this works for you.

@jhasse, would you prefer to split this PR into a series of smaller ones for easier review?

jochenmuell:
> To convince me I would like to know exactly in which cases the current solution (UTF-8 and largeFileSupport in manifest) does NOT work?

For me, it is the Qt build using clang. The Qt build has very long paths in the clang-cl output. The output seems to be handled using the max path length restrictions for the maximum buffer size of a path.

digit-google:
I think we need exact reproduction steps that confirm that the problem is with Ninja and not some other tooling.
Can you provide these for your Qt build example, with exact details on how you setup your workspace, and what goes wrong?

maxhmv:
@jhasse Ok that sounds good. Just a question because I don't fully understand which part of this PR you are opposed to. From my understanding fixing the issue with _MAX_PATH in [includes_normalize-win32.cc](https://github.com/ninja-build/ninja/blob/master/src/includes_normalize-win32.cc) would more or less require all the changes in [includes_normalize-win32.cc](https://github.com/ninja-build/ninja/blob/master/src/includes_normalize-win32.cc) @digit-google is suggesting here. Is there something you are opposed to in these changes?

Apart of _MAX_PATH there is also _MAX_DIR which also seems to be limited to 256 and could potentially cause problems. Should we fix this as well?

Thanks you!

jhasse:
> Is there something you are opposed to in these changes?

I don't want to use UTF-16 versions of the WinAPI like GetFullPathNameW, but keep the UTF-8 versions like GetFullPathNameA.

> Apart of _MAX_PATH there is also _MAX_DIR which also seems to be limited to 256 and could potentially cause problems. Should we fix this as well?

Yes, I think so.

digit-google:
I actually uploaded PR #2552 to fix the includes normalize problem. I just pushed a more recent version.
Regarding this specific PR (#2410) I think we can probably drop it.

jhasse:
okay let's continue in #2552

PR #2414, opened 2024-04-14
fix: add explict braces to avoid ambiguous else

heitbaum:
Fix to address the following compiler warning:

../src/build_test.cc: In function 'void TestPhonyUseCase(BuildTest*, int)': ../src/build_test.cc:1366:8: warning: suggest explicit braces to avoid ambiguous 'else' [-Wdangling-else]
 1366 |     if (!builder_.AlreadyUpToDate())
      |        ^

### DIFF
--- a/src/build_test.cc
+++ b/src/build_test.cc
@@ -1363,8 +1363,9 @@ void TestPhonyUseCase(BuildTest* t, int i) {
     // Build number 1
     EXPECT_TRUE(builder_.AddTarget("test" + ci, &err));
     ASSERT_EQ("", err);
-    if (!builder_.AlreadyUpToDate())
+    if (!builder_.AlreadyUpToDate()) {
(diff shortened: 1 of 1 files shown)


src/build_test.cc
```diff
@@ -1364,7 +1364,9 @@ void TestPhonyUseCase(BuildTest* t, int i) {
     EXPECT_TRUE(builder_.AddTarget("test" + ci, &err));
     ASSERT_EQ("", err);
     if (!builder_.AlreadyUpToDate())
+    {
```
jhasse, line 1367: should be on the previous line

jhasse (changes requested):

src/build_test.cc
heitbaum: Updated

PR #2415, opened 2024-04-14
util: Fix compiling with Android NDK

Biswa96:
This change fixes the following compiler error.
    
    util.cc:899:18: error: variable has incomplete type 'struct sysinfo'
      struct sysinfo si;
                     ^
    
    C++ compiler does not define 'linux` macro with -std=c++11 which was
    added in a4c24a33c1ed32d9d51c8df763ec6ad574587d02 commit. So, 'linux`
    is replaced with '__linux__`. See the difference in following output.
    
    $ clang++ -xc++ -dM -E - < /dev/null | grep linux
    > #define __gnu_linux__ 1
    > #define __linux 1
    > #define __linux__ 1
    > #define linux 1
    
    $ clang++ -xc++ -std=c++11 -dM -E - < /dev/null | grep linux
    > #define __gnu_linux__ 1
    > #define __linux 1
    > #define __linux__ 1

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -47,7 +47,7 @@
 #include <sys/loadavg.h>
 #elif defined(_AIX) && !defined(__PASE__)
 #include <libperfstat.h>
-#elif defined(linux) || defined(__GLIBC__)
+#elif defined(__linux__) || defined(__GLIBC__)
 #include <sys/sysinfo.h>
 #include <fstream>
 #include <map>
@@ -589,7 +589,7 @@ string StripAnsiEscapeCodes(const string& in) {
   return stripped;
 }
 
-#if defined(linux) || defined(__GLIBC__)
+#if defined(__linux__) || defined(__GLIBC__)
 std::pair<int64_t, bool> readCount(const std::string& path) {
   std::ifstream file(path.c_str());
   if (!file.is_open())
@@ -789,7 +789,7 @@ int GetProcessorCount() {
 #else
   int cgroupCount = -1;
   int schedCount = -1;
-#if defined(linux) || defined(__GLIBC__)
+#if defined(__linux__) || defined(__GLIBC__)
   cgroupCount = ParseCPUFromCGroup();
 #endif
   // The number of exposed processors might not represent the actual number of


finagolfin:
Hold off on this, I'm looking into it.

truboxl:
I believe whoever compiles ninja 1.12.0 for uclibc will also see this issue.
A proper fix should be done by upstream as the use of `-std=c++11` can cause `#if defined(linux)` to be false.
This is true for g++ and clang++.
https://github.com/ninja-build/ninja/blob/4a669954c47d9feb211a69488aee9ad0c8d7c674/src/util.cc#L50
https://github.com/ninja-build/ninja/blob/4a669954c47d9feb211a69488aee9ad0c8d7c674/src/util.cc#L592
https://github.com/ninja-build/ninja/blob/4a669954c47d9feb211a69488aee9ad0c8d7c674/src/util.cc#L792
https://github.com/ninja-build/ninja/blob/4a669954c47d9feb211a69488aee9ad0c8d7c674/src/util.cc#L897

finagolfin:
I looked into why it's not defined, in the clang source. It is set with [the `DefineStd()` method here in the clang source](https://github.com/llvm/llvm-project/blob/a4fe6a183883812fa8268490c987798a42c0d34b/clang/lib/Basic/Targets/OSTargets.h#L319), but that method does not set `linux`, ie without underscores, [if the GNU standard mode is not set](https://github.com/llvm/llvm-project/blob/ea9ec80b7aa1bd34edd5b544cd41450573f2cbff/clang/lib/Basic/Targets.cpp#L66), which the comment there notes will not be for `-std=c99`.

I think the mistake here is checking for `linux`, which I see almost no C++ code elsewhere doing. The correct checks in this file should be for `__linux__`.

finagolfin:
@truboxl, if you want to prepare a patch to change those three `linux` checks to `__linux__` in Termux, that should fix the build for Android.

@Biswa96, if you would change this patch also to do the same, we can get @jhasse to review.

Biswa96:
> if you would change this patch also to do the same,

That would be cheating. The idea is yours. Please feel free to create another pull request here. I shall close this one.

finagolfin:
Nah, this is a minor issue and you already have a pull open. Please just modify your patch here.

finagolfin:
Thanks, @Biswa96.

@jhasse, this fixed the Android build for us.

digit-google:
LGTM. I confirm that `linux` is deprecated and that `__linux__` should always be used instead.

jhasse:
Thanks! Would releasing a 1.12.1 hotfix tag help you in any way?

Biswa96:
> Would releasing a 1.12.1 hotfix tag help you in any way?

No hurry. Termux follows rolling releases for packages. This fix can be added in normal ninja release cycle.

Issue #2412, opened 2024-04-13
[Ninja-1.12.0] CMakeLists.txt: xx (cmake_minimum_required): Warning ...!!!


rumas84:
'CMakeLists.txt' Files need to be updated to avoid these warnings:
```
$ cmake -Bbuild-cmake
-- The CXX compiler identification is GNU 11.4.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- IPO / LTO enabled
-- Performing Test flag_no_deprecated
-- Performing Test flag_no_deprecated - Success
-- Performing Test flag_color_diag
-- Performing Test flag_color_diag - Success
-- Looking for ppoll
-- Looking for ppoll - found
-- Looking for fork
-- Looking for fork - found
-- Looking for pipe
-- Looking for pipe - found
-- Could NOT find GTest (missing: GTEST_LIBRARY GTEST_INCLUDE_DIR GTEST_MAIN_LIBRARY)
CMake Warning (dev) at /usr/share/**cmake-3.29**/Modules/FetchContent.cmake:1352 (message):
  The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is
  not set.  The policy's OLD behavior will be used.  When using a URL
  download, the timestamps of extracted files should preferably be that of
  the time of extraction, otherwise code that depends on the extracted
  contents might not be rebuilt if the URL changes.  The OLD behavior
  preserves the timestamps from the archive instead, but this is usually not
  what you want.  Update your project to the NEW behavior or specify the
  DOWNLOAD_EXTRACT_TIMESTAMP option with a value of true to avoid this
  robustness issue.
Call Stack (most recent call first):
  CMakeLists.txt:233 (FetchContent_Declare)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Deprecation Warning at build-cmake/_deps/googletest-src/CMakeLists.txt:4 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


-- The C compiler identification is GNU 11.4.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
[...]

PR #2418, opened 2024-04-15
Set CMake policy CMP0135 to fix gtest warning

Arghnews:
Partial fix for #2412 - remove warning about [DOWNLOAD_EXTRACT_TIMESTAMP](https://cmake.org/cmake/help/latest/policy/CMP0135.html) when building tests and downloading gtest, by using newer cmake behaviour. This sets "timestamps of all extracted contents to the time of the extraction" rather than the "timestamps in the archive".

Can be removed once cmake_minimum_version >=3.24

Fixes/removes this:
```
CMake Warning (dev) at /usr/local/share/cmake-3.29/Modules/FetchContent.cmake:1352 (message):
  The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is
  not set.  The policy's OLD behavior will be used.  When using a URL
  download, the timestamps of extracted files should preferably be that of
  the time of extraction, otherwise code that depends on the extracted
  contents might not be rebuilt if the URL changes.  The OLD behavior
  preserves the timestamps from the archive instead, but this is usually not
  what you want.  Update your project to the NEW behavior or specify the
  DOWNLOAD_EXTRACT_TIMESTAMP option with a value of true to avoid this
  robustness issue.
Call Stack (most recent call first):
  CMakeLists.txt:233 (FetchContent_Declare)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -227,6 +227,12 @@ endif()
 
 include(CTest)
 if(BUILD_TESTING)
+
+  # Can be removed if cmake min version is >=3.24
+  if (POLICY CMP0135)
+    cmake_policy(SET CMP0135 NEW)
+  endif()
+
   find_package(GTest)
   if(NOT GTest_FOUND)
     include(FetchContent)


digit-google:
Thanks, this LGTM. @jhasse  should know when we can update GTest to a better version. Probably in another PR.

Arghnews:
> #2422 would make this obsolete?

jhasse:
closing in favor of #2422

PR #2419, opened 2024-04-16
FindGTest: ignore Conda prefix with incompatible lib

scivision:
fixes issue introduced by 836abffe5eb34485e755a6976815656d28984e92 that CTest fails with SIGABRT at runtime due to incompatible Conda compiler ABI of their GTest.


CMakeLists.txt
```diff
@@ -227,6 +227,9 @@ endif()
 
 include(CTest)
 if(BUILD_TESTING)
+  if(DEFINED ENV{CONDA_PREFIX})
```
digit-google, line 230: Thank you, but can you be more specific about the issue here?
Ideally, there would be a comment here explaining the issue, how it can be reproduced, and more importantly under which conditions this special code-path could be removed.

This looks like a Conda-specific bug, I am not sure Ninja is the right place to deal with this (since it likely affects tons of other projects).

PR #2421, opened 2024-04-16
CMake: avoid unneeded color diagnostic check

scivision:
many developer machines will have CMake >= 3.24, but if not, keep existing logic.

On Windows or networked file system build directory, each CMake `check_*()` avoided can save 100s of milliseconds configure time.


CMakeLists.txt
```diff
@@ -32,9 +32,15 @@ else()
 	if(flag_no_deprecated)
 		add_compile_options(-Wno-deprecated)
 	endif()
-	check_cxx_compiler_flag(-fdiagnostics-color flag_color_diag)
-	if(flag_color_diag)
-		add_compile_options(-fdiagnostics-color)
+
+	if(NOT DEFINED ENV{CMAKE_COLOR_DIAGNOSTICS})
+	    set(CMAKE_COLOR_DIAGNOSTICS ON)
+	endif()
```
digit-google, line 38: wouldn´t it make more sense to put this in an elif() statement after the if() block below? E.g.:

```
# Enable color diagnostics by default. These are always enabled with CMake < 3.24
# or disabled setting CMAKE_COLOR_DIAGNOSTICS=OFF with newer versions.
if(CMAKE_VERSION VERSION_LESS 3.24)
     ...
elif(NOT DEFINED ENV{CMAKE_COLOR_DIAGNOSTICS})
    set(CMAKE_COLOR_DIAGNOSTICS ON)
endif()
```

PR #2427, opened 2024-04-23
Simplify ComputeCriticalPath() function.

digit-google:
Using simple unordered sets is enough to perform the same computation with less complexity, smaller and faster code.

On a large Fuchsia build plan, this reduces the
ComputeCriticalPath metric from 2.6s to 2.1s!

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -16,11 +16,13 @@
 
 #include <assert.h>
 #include <errno.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
+
 #include <climits>
-#include <stdint.h>
 #include <functional>
+#include <unordered_set>
 
 #if defined(__SVR4) && defined(__sun)
 #include <sys/termios.h>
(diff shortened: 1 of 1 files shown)


digit-google:
Invoking @bradking here too.

bradking:
Since we were not relying on any ordering of `std::set`, switching to `std::unordered_set` should be fine, and C++11 is now allowed AFAIK.

digit-google:
Thanks.

On a related note, I must admit that personally, I find the extra 2.1s a productivity killer for incremental builds (especially since my version of Ninja has [persistent mode](https://fuchsia.dev/fuchsia-src/development/build/ninja_fuchsia_improvements?hl=en) to speed these significantly).

What do you think about adding an option, or environment variable to disable this (not in this PR of course).
@jhasse, do you have an opinion on this?

jhasse:
Thanks!

I agree that 2.1s are still too much. Instead of an option we should fallback to a simpler, naive algorithm above a certain threshold?

benoit-pierre:
> (especially since my version of Ninja has [persistent mode](https://fuchsia.dev/fuchsia-src/development/build/ninja_fuchsia_improvements?hl=en) to speed these significantly).

That `NINJA_STATUS_MAX_COMMANDS` feature to show the status of running commands is really sweet, any plan on upstreaming it?

PR #2428, opened 2024-04-27
Correctly handle ANSI Escape Codes in ElideMiddle, fix #713

### LINKED ISSUE #713: Support color in NINJA_STATUS

jhasse:
I find it a bit weird, that even in the old implementation, passing 10 will result in a string of size 9. Isn't this a bug? @evmar do you remember if there was a reason for that? I think d82e806729bed17bf8fba738ed63eb740863c79e is the first commit which adds the 10 -> 9 unit test.

Maybe that should be fixed in the old implementation first.

Regarding this implementation: Using std::regex might be too slow, but it's a start and can be used as a reference implementation.

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -26,6 +26,7 @@
 #include <assert.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <regex>
 #include <stdarg.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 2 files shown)


evmar:
Ha, looking at your diff I think I just had it wrong.  EIther an off-by-one or maybe a rounding mistake in the `int elide_size = (width - kMargin) / 2` bit?

jhasse:
elide_size rounds down and if the width is even (let's say 12), this means that the result would be:

`[elide_size][kMargin][elide_size] == 4 + 3 + 4`

To fix this we'd have to show another character either in the beginning or the end (my preference). I think this only works with an explicit `(width - kMargin) % 2 == 1)` check.

If you think this was unintentional I would have a go at it in another PR :)

mathstuf (approved):
Seems to work here. Thanks!

mathstuf (changes requested):
Ok, playing around with it more, it seems to not work quite right. Here is a copy of the `tmux` buffer on a pane that is narrow enough to require elision (executed from with `nvim`, but the elision code is definitely involved here).

<details>
<summary>`tmux` buffer state</summary>

```
                                                                                                           [-Wunused-variable]                                                                                         3577 |     FlagClassification cls = FlagClassification::PrivateFlag;                                                                                                                                                 [-Wunused-variable]                                                                                         3582 |     FlagClassification cls = FlagClassification::LocationFlag;                                                                                                                                                [-Wunused-variable]                                                                                         3583 |     FlagKind kind = FlagKind::BuildSystem;                                                                                                                                                                     0-5->13/17@0.5] Linking CXX executable bin/ctest                                                                |              ^~~~                                                                                                                                                                                              0-4->14/17@0.5] Linking CXX executable bin/cmake                                                          [15-1->2/17@0.2] Linking CXX static library Source/libCMakeLib.a                                                                                                                                                       0-3->15/17@0.5] Linking CXX executable bin/ccmake                                                         [5-10->3/17@0.3] Linking CXX executable Tests/CMakeLib/testAffinity                                                                                                                                                    0-2->16/17@0.5] Linking CXX executable bin/ctresalloc                                                     [4-10->4/17@0.3] Linking CXX executable Tests/CMakeLib/testUVProcessChainHelper                                                                                                                                        0-1->17/17@0.6] Linking CXX executable Tests/CMakeLib/CMakeLibTests                                       [3-10->5/17@0.4] Linking CXX executable Tests/CMakeLib/runcompilecommands                                                                                                                                                                                                                                                        [2-10->6/17@0.5] Linking CXX executable Tests/CMakeLib/testDebuggerNamedPipe                                                                                                                                          
[1-10->7/17@0.6] Linking CXX executable Tests/CMakeLib/PseudoMemcheck/valgrind                                                                                                                                        
[0-10->8/17@0.7] Linking CXX executable Tests/CMakeLib/PseudoMemcheck/purify                                                                                                                                          
[0-9->9/17@0.7] Linking CXX executable Tests/CMakeLib/PseudoMemcheck/cuda-memcheck                                                                                                                                    
[0-8->10/17@0.8] Linking CXX executable Tests/CMakeLib/PseudoMemcheck/memcheck_fail                                                                                                                                   
[0-7->11/17@0.9] Linking CXX executable Tests/CMakeLib/PseudoMemcheck/BC                                                                                                                                              
[0-6->12/17@0.5] Linking CXX executable bin/cpack                                                                                                                                                                     
[0-5->13/17@0.5] Linking CXX executable bin/ctest                                                                                                                                                                     
[0-4->14/17@0.5] Linking CXX executable bin/cmake                                                                                                                                                                     
[0-3->15/17@0.5] Linking CXX executable bin/ccmake                                                                                                                                                                    
[0-2->16/17@0.5] Linking CXX executable bin/ctresalloc                                                                                                                                                                
[0-1->17/17@0.6] Linking CXX executable Tests/CMakeLib/CMakeLibTests                                                                                                                                                  
                                                                                                                                                                                                                      

```

</details>

It looks like newlines are being truncated somewhere.

PR #2429, opened 2024-04-27
ElideMiddle: Don't reduce size by 1 if width is even

jhasse:
see #2428


digit-google:
this looks good to me.

PR #2433, opened 2024-05-04
Small DynDep code improvements

gruenich:


src/dyndep.cc
```diff
@@ -43,9 +43,7 @@ bool DyndepLoader::LoadDyndeps(Node* node, DyndepFile* ddf,
 
   // Update each edge that specified this node as its dyndep binding.
   std::vector<Edge*> const& out_edges = node->out_edges();
-  for (std::vector<Edge*>::const_iterator oe = out_edges.begin();
-       oe != out_edges.end(); ++oe) {
-    Edge* const edge = *oe;
+  for (Edge* edge : out_edges) {
```
digit-google, line 46: nit: use `const Edge*` here. Same concern for other loops below.

src/dyndep.cc
```diff
@@ -65,10 +63,9 @@ bool DyndepLoader::LoadDyndeps(Node* node, DyndepFile* ddf,
   }
 
   // Reject extra outputs in dyndep file.
-  for (DyndepFile::const_iterator oe = ddf->begin(); oe != ddf->end();
-       ++oe) {
-    if (!oe->second.used_) {
-      Edge* const edge = oe->first;
+  for (auto& oe : *ddf) {
```
digit-google, line 66: Apart from using `const auto&`, can we also rename the variable for clarity here?
E.g. use `dyndep_output` instead of the cryptic `oe`?

digit-google:
Thank you, this looks good (if you can address the nits). @jhasse, wdyt?

src/dyndep.cc
gruenich: Your proposal does not work, `map->find` complains no match as it "would lose const qualifier".

src/dyndep.cc
gruenich: I renamed the variable.
I added `const` and double-checked all the other loops and applied it where possible. Thanks for the review!

src/dyndep.cc
digit-google: Weird, what did the original const_iterator do then?

digit-google: Oh I see, this was doing the equivalent of `Edge* const&`, so using `Edge*` here is equivalent. Let's keep your code though, sorry for the noise :)

gruenich:
Clang-tidy was not happy, I added two more commits.

src/dyndep_parser.cc
```diff
@@ -95,9 +95,9 @@ bool DyndepParser::ParseDyndepVersion(string* err) {
 bool DyndepParser::ParseLet(string* key, EvalString* value, string* err) {
   if (!lexer_.ReadIdent(key))
     return lexer_.Error("expected variable name", err);
-  if (!ExpectToken(Lexer::EQUALS, err) || !lexer_.ReadVarValue(value, err))
-    return false;
-  return true;
+  if (ExpectToken(Lexer::EQUALS, err) && lexer_.ReadVarValue(value, err))
+    return true;
+  return false;
```
digit-google, line 100: can these three lines can be further simplified as a simple return statement? wdyt?

gruenich: Yeah, clang-tidy forced me to do so, too.

jhasse:
I'm not sure about [Use C++11 range-based for loops instead of iterator style](https://github.com/ninja-build/ninja/pull/2433/commits/0b6df95ae6cec373885c5d49e85a5f48863ad9e1). In general I'm more in favor of preserving Git (blame) history than such modernizations. But the dyndep code is quite new so I think it's fine.

PR #2437, opened 2024-05-04
RealDiskInterface: Do *not* set locale to an empty string

### LINKED ISSUE #2435: [REG 1.11->1.12] Screen updates are extremely slow on Windows
With Ninja 1.12, the progress updates are horribly slow, and it looks like the cursor passes through all the characters one by one. It looks like running on 80386...

With 1.11 it worked much better.

Run example:
ninja-1 12

orgads:
It causes the cursor handling to be extremely slow on MinGW.

Added in #2321.

Fixes #2435.


orgads:
cc @brondani

jhasse:
Thanks! I think this justifies a 1.12.1 bugfix release.

PR #2440, opened 2024-05-05
fix: don't attempt to write and stat the lock file during dry runs

### LINKED ISSUE #2431: Dry run (-n) error if the builddir directory does not exist in 1.12.0
Hi. I'm using Ninja 1.12.0 on Windows.

Since 1.12.0 dry run commands will produce an error such as `error: WriteFile(ninja_builddir/.ninja_lock): Unable to create file. No such file or directory` if the builddir directory does not already exist. I reproduced the error with a simple build.ninja file and running `ninja -n`:

```
builddir = ninja_builddir

rule cc
    command = clang -c $in -o $out

build test.obj: cc test.cpp
```

Thanks!

jdrouhard:
Fixes #2431

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -846,7 +846,7 @@ bool Builder::StartEdge(Edge* edge, string* err) {
 
   status_->BuildEdgeStarted(edge, start_time_millis);
 
-  TimeStamp build_start = -1;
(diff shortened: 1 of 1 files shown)


jhasse:
Thanks! We're really should have more (integration) tests for this stuff ...

PR #2445, opened 2024-05-08
Use single string::replace instead of two string::substr

gruenich:
Small performance improvement, as creation of two string objects is avoided.
Found by Cppcheck (uselessCallsSubstr): "Ineffective call of function 'substr' because a prefix of the string is assigned to itself. Use replace() instead."


jhasse:
Hm ... I think the new version is less readable. Is the performance impact measurable?

gruenich:
Yes, it is measurable. At least with my limited benchmarking skills. With 1,000,000 repeats, a recent `g++`, `-O3`, my somewhat older laptop, and openSuse Tumbleweed:
Existing code: 0.13 sec
My version: 0.08 sec

What do you think, would additional comments explaining the formula of the second argument help with understanding the code?

digit-google:
Yes, this will definitely avoid unnecessary string allocations + concatenations, but I agree some comments should clarify the formula being used, or possibly renaming the variables, e.g.:

```
if (result.size() > width) {
  // Keep |width- kMargin| visible characters from the input, divided into two spans, separated by "...".
  size_t visible_size = width - kMargin;
  size_t left_span_size = visible_size / 2;
  size_t right_span_size = visible_size - left_span;
  
  // Replace the middle gap in the input string with "..."
  size_t gap_start = left_span_size;
  size_t gap_end = result.size() - right_span_size;
  result.replace(gap_start, gap_end - gap_start, "...");
}
```

gruenich:
I fixed a wrong variable and adopted the rest. Now it is better readable than the initial code.

digit-google:
Thanks, looks good to me!

PR #2448, opened 2024-05-13
configure.py: Support --gtest-source-dir to build tests.

### LINKED ISSUE #2447: Tests can't be built with configure.py
`configure.py` doesn't support building/running tests since https://github.com/ninja-build/ninja/pull/1562.

This is tricky for us downstream, as to run tests (which is something we care about, especially for a critical package like ninja), we would need to use the CMake build - but then we get a circular dependency with e.g. ninja->cmake->jsoncpp->ninja.

Could test support please be restored to `configure.py`?

This is similar to https://github.com/ninja-build/ninja/pull/2423.

digit-google:
Allow the Ninja build plan generated by configure.py to build `ninja_test` by compiling GoogleTest from source if the path to the library if passed through the new option `--gtest-source-dir` or the GTEST_SOURCE_DIR environment variable.

For simplicity, probing for an installed version of the library, and linking to it, is not supported (use the CMake build for this).

This also removes the obsolete `--gtest-dir` option.

+ Update README.md

Fixes #2447

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,19 @@ via CMake. For more details see
 This will generate the `ninja` binary and a `build.ninja` file you can now use
 to build Ninja with itself.
 
+If you have a GoogleTest source directory, you can build the tests
+by passing its path with `--gtest-source-dir=PATH` option, or the
+`GTEST_SOURCE_DIR` environment variable, e.g.:
+
+```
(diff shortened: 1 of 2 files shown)


jhasse:
Thanks!

PR #2449, opened 2024-05-14
manifest_parser.cc: Fix three minor memory leaks

gruenich:
Add missing delete in case of an error to prevent
memory leaks.
Found by clang tidy (clang-analyzer-cplusplus.NewDeleteLeaks)

### DIFF
--- a/src/manifest_parser.cc
+++ b/src/manifest_parser.cc
@@ -313,8 +313,10 @@ bool ManifestParser::ParseEdge(string* err) {
   while (has_indent_token) {
     string key;
     EvalString val;
-    if (!ParseLet(&key, &val, err))
+    if (!ParseLet(&key, &val, err)) {
+      delete env;
       return false;
+    }
 
     env->AddBinding(key, val.Evaluate(env_));
     has_indent_token = lexer_.PeekToken(Lexer::INDENT);
--- a/src/manifest_parser.cc
+++ b/src/manifest_parser.cc
@@ -18,6 +18,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 
+#include <memory>
 #include <vector>
 
 #include "graph.h"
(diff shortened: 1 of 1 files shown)


src/manifest_parser.cc
```diff
@@ -148,14 +148,17 @@ bool ManifestParser::ParseRule(string* err) {
   while (lexer_.PeekToken(Lexer::INDENT)) {
     string key;
     EvalString value;
-    if (!ParseLet(&key, &value, err))
+    if (!ParseLet(&key, &value, err)) {
+      delete rule;
```
digit-google, line 152: It seems you forgot lines 168 and 170 below :-/

I suggest using an `std::unique_ptr<Rule>` local variable definition on line 146 above, then passing its ownership on line 175 with `env->AddRule(rule.release())` to prevent similar issues to crop up in the future.

src/manifest_parser.cc
```diff
@@ -311,8 +314,10 @@ bool ManifestParser::ParseEdge(string* err) {
   while (has_indent_token) {
     string key;
     EvalString val;
-    if (!ParseLet(&key, &val, err))
+    if (!ParseLet(&key, &val, err)) {
+      delete env;
```
digit-google, line 318: Technically this is correct here, but the reason why is quite subtle.

Also, these new BindingEnv instances are always leaked because there is no single owner for them (several edge->env_ values can point to the same instance).

I suggest you make State be the owner of all BindingEnv instances (for example with a new `std::vector<std::unique_ptr<BindingEnv>> envs_;` member, and just call a method like `state_->NewBindingEnv(env_)` to get a new instance here. In case of parse error, we don't really care about one dangling instance, and we can remove all existing ones in State::Reset() as well.

What do you think about this?

src/manifest_parser.cc
gruenich: Great idea! This also reflect the to-do comment from 2010 to use a `scoped_ptr`.

src/manifest_parser.cc
gruenich: Sounds complicated. Wouldn't a `shared_ptr` instead of a pointer be enough?

digit-google:
I think we already call _exit() instead of exit() to ensure the memory is not automatically release when Ninja exits, so leaking on purpose seems to be of little value.

And we really want to make it the default, let's introduce a dedicated class template like LeakedPointer<T> to make this explicit in the code

(Meanwhile, in my fork, which implements persistent mode, I had to fix these leaks too anyway since the same build graph is reused over multiple build invocations. But it's ok if you don't want to do that upstream).

PR #2451, opened 2024-05-19
Remove global `explanations_` variable

digit-google:
This is a followup for PR #2067 which cleans up the code base by removing the global variable, which was updated by calling record_explanation() from random places of the code. See [the comments on the original commits](https://github.com/ninja-build/ninja/pull/2067/commits/8e6c741a4b6c4f5cab4695016be2b0a8183bab83#r814232149) for the reasons to do so.

This does not change behavior in any way.

This also helps expose the extra coupling that the PR introduced, e.g. the StatusPrinter instance now needs to access the map of recorded explanations, so a new `Status::SetExplanations()` method must be added to the base `Status` virtual class to pass an appropriate pointer.

This could be avoided by ensuring the explanations are printed slightly earlier / later by the Builder class, but for now this will keep everything consistent.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -260,6 +260,7 @@ if(BUILD_TESTING)
     src/disk_interface_test.cc
     src/dyndep_parser_test.cc
     src/edit_distance_test.cc
+    src/explanations_test.cc
     src/graph_test.cc
     src/json_test.cc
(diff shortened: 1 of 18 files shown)


jhasse:
Awesome stuff, thanks!

PR #2455, opened 2024-05-24
cmake: disable min() and max() macros in windows.h

hundeboll:
Using std::min() and/or std::max() in sources where windows.h has been include fails with

    error C2589: '(': illegal token on right side of '::' [D:\a\ninja\ninja\build\libninja.vcxproj]
    error C2062: type 'unknown-type' unexpected [D:\a\ninja\ninja\build\libninja.vcxproj]
    error C2059: syntax error: ')' [D:\a\ninja\ninja\build\libninja.vcxproj]
    error C2589: '(': illegal token on right side of '::' [D:\a\ninja\ninja\build\libninja.vcxproj]
    error C2059: syntax error: ')' [D:\a\ninja\ninja\build\libninja.vcxproj]

Avoid this by defining NOMINMAX for windows builds, which causes the windows.h header to skip defining those two macros.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -163,6 +163,11 @@ if(WIN32)
 	# so that build environments which lack a C compiler, but have a C++
 	# compiler may build ninja.
 	set_source_files_properties(src/getopt.c PROPERTIES LANGUAGE CXX)
+
+	# windows.h defines min() and max() which conflict with std::min()
+	# and std::max(), which both might be used in sources. Avoid compile
+	# errors by telling windows.h to not define those two.
+	add_compile_definitions(NOMINMAX)
 else()
 	target_sources(libninja PRIVATE src/subprocess-posix.cc)
 	if(CMAKE_SYSTEM_NAME STREQUAL "OS400" OR CMAKE_SYSTEM_NAME STREQUAL "AIX")


CMakeLists.txt
```diff
@@ -163,6 +163,11 @@ if(WIN32)
 	# so that build environments which lack a C compiler, but have a C++
 	# compiler may build ninja.
 	set_source_files_properties(src/getopt.c PROPERTIES LANGUAGE CXX)
+
+	# windows.h defines min() and max() which conflict with std::min()
+	# and std::max(), which both might be used in sources. Avoid compile
+	# errors by telling windows.h to not define those two.
+	add_definitions(-DNOMINMAX)
```
stephan-cr, line 170:
```suggestion
	add_compile_definitions(NOMINMAX)
```
I think `add_compile_definitions` is preferred nowadays and it would be consistent to the rest of the file.

hundeboll: done

jhasse:
> New in version 3.26: Any leading -D on an item will be removed.

https://cmake.org/cmake/help/latest/command/add_compile_definitions.html

If I interpret this correctly this will therefore break with CMake < 3.26. Can you remove the leading `-D`?

hundeboll:
> > New in version 3.26: Any leading -D on an item will be removed.
> 
> https://cmake.org/cmake/help/latest/command/add_compile_definitions.html
> 
> If I interpret this correctly this will therefore break with CMake < 3.26. Can you remove the leading `-D`?

Done

jhasse:
Thanks.

Windows sucks.

PR #2468, opened 2024-06-21
Use greg7mdp/parallel-hashmap for faster execution

atetubou:
This implements option 1 in https://github.com/ninja-build/ninja/pull/2462#issuecomment-2172740047.

This uses https://github.com/greg7mdp/parallel-hashmap instead of unordered_map for faster ninja execution.
Some more context is in https://crbug.com/346910589.

This makes noop build for chromium
* 12% faster on Linux
* 2% slower on Windows
* 96% faster on macOS

# Linux
```
$ hyperfine --warmup 1 \
   '~/ninja_master -C out/non_component build:buildflag_header_h' \
   '~/ninja_faster_hash -C out/non_component build:buildflag_header_h'
Benchmark 1: ~/ninja_master -C out/non_component build:buildflag_header_h
  Time (mean ± σ):      5.251 s ±  0.027 s    [User: 4.641 s, System: 0.608 s]
  Range (min … max):    5.222 s …  5.309 s    10 runs
 
Benchmark 2: ~/ninja_faster_hash -C out/non_component build:buildflag_header_h
  Time (mean ± σ):      4.689 s ±  0.022 s    [User: 4.055 s, System: 0.633 s]
  Range (min … max):    4.648 s …  4.737 s    10 runs
 
Summary
  ~/ninja_faster_hash -C out/non_component build:buildflag_header_h ran
    1.12 ± 0.01 times faster than ~/ninja_master -C out/non_component build:buildflag_header_h
```

# Windows
```
Benchmark 1: C:\Users\tikuta\source\repos\ninja-build\ninja\ninja_master -C out/non_component build:buildflag_header_h
  Time (mean ± σ):     13.161 s ±  0.105 s    [User: 10.657 s, System: 2.194 s]
  Range (min … max):   13.005 s … 13.328 s    10 runs

Benchmark 2: C:\Users\tikuta\source\repos\ninja-build\ninja\ninja_faster_hash -C out/non_component build:buildflag_header_h
  Time (mean ± σ):     13.206 s ±  0.210 s    [User: 10.653 s, System: 2.194 s]
  Range (min … max):   12.951 s … 13.578 s    10 runs

Summary
  C:\Users\tikuta\source\repos\ninja-build\ninja\ninja_master -C out/non_component build:buildflag_header_h ran
    1.00 ± 0.02 times faster than C:\Users\tikuta\source\repos\ninja-build\ninja\ninja_faster_hash -C out/non_component build:buildflag_header_h
```

# macOS
```
$ hyperfine --warmup 1 '~/ninja_master -C out/non_component build:buildflag_header_h' '~/ninja_faster_hash -C out/non_component build:buildflag_header_h'
Benchmark 1: ~/ninja_master -C out/non_component build:buildflag_header_h
  Time (mean ± σ):     14.360 s ±  0.706 s    [User: 9.006 s, System: 1.681 s]
  Range (min … max):   12.832 s … 15.297 s    10 runs

Benchmark 2: ~/ninja_faster_hash -C out/non_component build:buildflag_header_h
  Time (mean ± σ):      7.340 s ±  0.260 s    [User: 6.263 s, System: 0.828 s]
  Range (min … max):    6.948 s …  7.799 s    10 runs

Summary
  ~/ninja_faster_hash -C out/non_component build:buildflag_header_h ran
    1.96 ± 0.12 times faster than ~/ninja_master -C out/non_component build:buildflag_header_h
```

### DIFF
--- a/third_party/parallel_hashmap/download.sh
+++ b/third_party/parallel_hashmap/download.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+set -eu
+
+cd $(dirname $0)
+
+revision=d88c5e15079047777b418132ece5879e7c9aaa2b
+
+for file in LICENSE \
+              parallel_hashmap/phmap.h \
+              parallel_hashmap/phmap_base.h \
+              parallel_hashmap/phmap_bits.h \
+              parallel_hashmap/phmap_config.h \
+              parallel_hashmap/phmap_fwd_decl.h \
+              parallel_hashmap/phmap_utils.h
+do
+  mkdir -p $(dirname $file)
+  curl --silent https://raw.githubusercontent.com/greg7mdp/parallel-hashmap/$revision/$file > $file
+done
--- a/.clang-format-ignore
+++ b/.clang-format-ignore
@@ -1,0 +1,1 @@
+/third_party/parallel_hashmap/parallel_hashmap/*
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -45,6 +45,7 @@ jobs:
     - uses: codespell-project/actions-codespell@master
       with:
         ignore_words_list: fo,wee,addin,notin
+        skip: ./third_party
     - name: Install dependencies
       run: |
         dnf install -y make gcc-c++ libasan clang-analyzer cmake dnf-plugins-core epel-release
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,6 +61,8 @@ else()
 	endif()
 endif()
 
+include_directories(third_party/parallel_hashmap)
+
 # --- optional re2c
 set(RE2C_MAJOR_VERSION 0)
 find_program(RE2C re2c)
--- a/configure.py
+++ b/configure.py
@@ -345,6 +345,7 @@ if platform.is_msvc():
               # Disable size_t -> int truncation warning.
               # We never have strings or arrays larger than 2**31.
               '/wd4267',
+              '/Ithird_party/parallel_hashmap',
               '/DNOMINMAX', '/D_CRT_SECURE_NO_WARNINGS',
               '/D_HAS_EXCEPTIONS=0',
               '/DNINJA_PYTHON="%s"' % options.with_python]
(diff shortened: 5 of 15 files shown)


atetubou:
@digit-google @jhasse how do you think about this PR instead of https://github.com/ninja-build/ninja/pull/2462?

digit-google:
Thanks, I have tested this CL with a local Fuchsia build on Linux, I get a consistent 2% (x1.02) speedup there.

Can you fuse all commits into a single one though? I'll add a few inline comments too.

I suggest adding a top-level `.clang-format-ignore` file to avoid editors to reformat `third_party/*`. Unfortunately, it seems that this file is ignored by `git-clang-format` for now, so that's not a hard-requirement.

third_party/parallel_hashmap/download.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+cd $(dirname $0)
+
+revision=d88c5e15079047777b418132ece5879e7c9aaa2b
+
+for file in phmap.h phmap_base.h phmap_bits.h phmap_config.h phmap_fwd_decl.h phmap_utils.h
```
digit-google, line 7: I would suggest respecting the upstream file layout, even if only download a few of them. This will be easier long-term for maintenance, and allows adding a few more things (e.g. the `phmap_{gdb,lldb}.py` files might be handy while debugging Ninja).

In this case, this would mean, copying the headers to `third_party/parallel_hashmap/parallel_hashmap` and adding `third_party/parallel_hashmap` instead of `third_party` to the include directory. Wdyt?

atetubou: Thanks, I updated to use upstream layout.

atetubou:
> Can you fuse all commits into a single one though? I'll add a few inline comments too.

Done.

> I suggest adding a top-level .clang-format-ignore file to avoid editors to reformat third_party/*.

Done.

digit-google:
Thanks, the latest commit is much better, but please change the commit message to something informative. This is what will be submitted into the final git repository, so the current message will be useless for people looking at the git log in the future. And the github conversation here, and its description message at the top, will essentially be lost in the wind after that.

@jhasse , what do you think?

atetubou:
Thank you for your review, I updated commit message.

jhasse:
That it's slower on Windows is a bit unfortunate. I'm not sure if adding a third party hash map is worth it for a 12 % speed up - the macOS speed up is nice though.

jhasse:
Not sure what the current situation is now, but since this conflicts now and I'm still not sure, closing for now. Please discuss on the mailing list if you still think this should be merged.

PR #2474, opened 2024-08-10
Implement GNU jobserver posix client support

### LINKED ISSUE #1139: Add GNU make jobserver client support

mcprat:
a rework of #2450 supporting all versions of GNU Make, but without Windows support
(I'm not able to test for Windows, and I have doubts with proposed Windows support)

resolves #1139 for posix systems

thanks to @hundeboll for much of the work with this newer implementation

ping @jhasse @digit-google 

significant differences:
 - no changes to any function parameters
 - no new intermediary functions
 - instantiate client support in real_main() instead of Plan
 - pass ~~references~~ pointers to the jobserver class into other classes
 - use a constructor to initialize jobserver client
 - release all tokens on any fatal error
 - calculate a value for "load capacity" instead of returning `SIZE_MAX`
 - supports both fifo and simple pipe file descriptors from Make
 - detect invalid or closed pipe and inform user about the most likely reason

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -163,6 +163,16 @@ Edge* Plan::FindWork() {
     return NULL;
 
   Edge* work = ready_.top();
+
+  // Only initiate work if the jobserver client can acquire a token.
+  if (builder_ && builder_->jobserver_ &&
+      builder_->jobserver_->Enabled()) {
+    int job_tokens = builder_->jobserver_->Tokens();
+    work->job_token_ = builder_->jobserver_->Acquire();
+    if (job_tokens == builder_->jobserver_->Tokens())
+      return NULL;
+  }
+
   ready_.pop();
   return work;
 }
@@ -198,6 +208,10 @@ bool Plan::EdgeFinished(Edge* edge, EdgeResult result, string* err) {
   if (directly_wanted)
     edge->pool()->EdgeFinished(*edge);
   edge->pool()->RetrieveReadyEdges(&ready_);
+
+  // If jobserver is used, return the token for this job.
+  if (builder_ && builder_->jobserver_)
+    builder_->jobserver_->Release(&edge->job_token_);
 
   // The rest of this function only applies to successful commands.
   if (result != kEdgeSucceeded)
@@ -592,14 +606,18 @@ void Plan::Dump() const {
 }
 
 struct RealCommandRunner : public CommandRunner {
-  explicit RealCommandRunner(const BuildConfig& config) : config_(config) {}
+  explicit RealCommandRunner(const BuildConfig& config, Jobserver* jobserver) :
+      config_(config), jobserver_(jobserver) {}
+
   size_t CanRunMore() const override;
   bool StartCommand(Edge* edge) override;
   bool WaitForCommand(Result* result) override;
   vector<Edge*> GetActiveEdges() override;
+  void ClearJobTokens(const std::vector<Edge*>&) override;
   void Abort() override;
 
   const BuildConfig& config_;
+  Jobserver* jobserver_;
   SubprocessSet subprocs_;
   map<const Subprocess*, Edge*> subproc_to_edge_;
 };
@@ -612,7 +630,13 @@ vector<Edge*> RealCommandRunner::GetActiveEdges() {
   return edges;
 }
 
+void RealCommandRunner::ClearJobTokens(const std::vector<Edge*> &edges) {
+  for (Edge* edge : edges)
+    jobserver_->Release(&edge->job_token_);
+}
+
 void RealCommandRunner::Abort() {
+  ClearJobTokens(GetActiveEdges());
   subprocs_.Clear();
 }
 
@@ -627,6 +651,14 @@ size_t RealCommandRunner::CanRunMore() const {
     if (load_capacity < capacity)
       capacity = load_capacity;
   }
+
+  // When initialized, behave as if the implicit token is acquired already.
+  // Otherwise, this happens after a token is released but before it is replaced,
+  // so the base capacity is represented by job_tokens + 1 when positive.
+  // Add an extra loop on capacity for each job in order to get an extra token.
+  int job_tokens = jobserver_->Tokens();
+  if (job_tokens)
+    capacity = abs(job_tokens) - subproc_number + 2;
 
   if (capacity < 0)
     capacity = 0;
@@ -667,10 +699,10 @@ bool RealCommandRunner::WaitForCommand(Result* result) {
   return true;
 }
 
-Builder::Builder(State* state, const BuildConfig& config, BuildLog* build_log,
-                 DepsLog* deps_log, DiskInterface* disk_interface,
-                 Status* status, int64_t start_time_millis)
-    : state_(state), config_(config), plan_(this), status_(status),
+Builder::Builder(State* state, const BuildConfig& config, Jobserver* jobserver,
+                 BuildLog* build_log, DepsLog* deps_log, DiskInterface* disk_interface,
+                 Status* status, int64_t start_time_millis) : state_(state),
+      config_(config), jobserver_(jobserver), plan_(this), status_(status),
       start_time_millis_(start_time_millis), disk_interface_(disk_interface),
       explanations_(g_explaining ? new Explanations() : nullptr),
       scan_(state, build_log, deps_log, disk_interface,
@@ -775,7 +807,7 @@ bool Builder::Build(string* err) {
     if (config_.dry_run)
       command_runner_.reset(new DryRunCommandRunner);
     else
-      command_runner_.reset(new RealCommandRunner(config_));
+      command_runner_.reset(new RealCommandRunner(config_, jobserver_));
   }
 
   // We are about to start the build process.
--- a/src/build.h
+++ b/src/build.h
@@ -24,6 +24,7 @@
 #include "depfile_parser.h"
 #include "exit_status.h"
 #include "graph.h"
+#include "jobserver.h"
 #include "util.h"  // int64_t
 
 struct BuildLog;
@@ -161,6 +162,7 @@ struct CommandRunner {
   virtual bool WaitForCommand(Result* result) = 0;
 
   virtual std::vector<Edge*> GetActiveEdges() { return std::vector<Edge*>(); }
+  virtual void ClearJobTokens(const std::vector<Edge*>&) {}
   virtual void Abort() {}
 };
 
@@ -187,9 +189,9 @@ struct BuildConfig {
 
 /// Builder wraps the build process: starting commands, updating status.
 struct Builder {
-  Builder(State* state, const BuildConfig& config, BuildLog* build_log,
-          DepsLog* deps_log, DiskInterface* disk_interface, Status* status,
-          int64_t start_time_millis);
+  Builder(State* state, const BuildConfig& config, Jobserver* jobserver,
+          BuildLog* build_log, DepsLog* deps_log, DiskInterface* disk_interface,
+          Status* status, int64_t start_time_millis);
   ~Builder();
 
   /// Clean up after interrupted commands by deleting output files.
@@ -224,6 +226,7 @@ struct Builder {
 
   State* state_;
   const BuildConfig& config_;
+  Jobserver* jobserver_;
   Plan plan_;
   std::unique_ptr<CommandRunner> command_runner_;
   Status* status_;
--- a/src/jobserver-posix.cc
+++ b/src/jobserver-posix.cc
@@ -1,0 +1,160 @@
+// Copyright 2024 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "jobserver.h"
+
+#include <fcntl.h>
+#include <unistd.h>
+
+#include <cassert>
+#include <cstring>
+
+#include "util.h"
+
+// Declare complete type for static constants in class.
+constexpr char const Jobserver::kAuthKey[];
+constexpr char const Jobserver::kFdsKey[];
+constexpr char const Jobserver::kFifoKey[];
+
+PosixJobserverClient::PosixJobserverClient() {
+  assert(!Enabled());
+
+  // Set name, type of pipe, and if non-parallel from MAKEFLAGS.
+  Parse();
+
+  const char* jobserver = jobserver_name_.c_str();
+
+  // Warn if jobserver type is unknown (neither fifo nor pipe).
+  if (!jobserver_fifo_ && sscanf(jobserver, "%d,%d", &rfd_, &wfd_) != 2)
+    if (!jobserver_name_.empty())
+      Warning("invalid jobserver auth: '%s'", jobserver);
+
+  // Open FDs to the pipe if needed, read must be non-blocking.
+  // If passed FDs are blocking on read, force non-parallel build.
+  if (jobserver_fifo_) {
+    rfd_ = open(jobserver + strlen(kFifoKey), O_RDONLY | O_NONBLOCK);
+    wfd_ = open(jobserver + strlen(kFifoKey), O_WRONLY);
+  } else if (Enabled() && (fcntl(rfd_, F_GETFL) & O_NONBLOCK) == 0) {
+    jobserver_closed_ = true;
+  }
+
+  // Exit on failure to open FDs, build non-parallel for invalid passed FDs.
+  if (Enabled())
+    Info("using jobserver: %s", jobserver);
+  else if (jobserver_fifo_ && (rfd_ == -1 || wfd_ == -1))
+    Fatal("failed to open jobserver: %s: %s", jobserver, strerror(errno));
+  else if (!jobserver_name_.empty())
+    jobserver_closed_ = true;
+
+  // Signal that we have initialized but do not have a token yet.
+  if (Enabled())
+    token_count_ = -1;
+}
+
+void PosixJobserverClient::Parse() {
+  // Return early if no makeflags are passed in the environment.
+  const char* makeflags = std::getenv("MAKEFLAGS");
+  if (makeflags == nullptr || strlen(makeflags) == 0)
+    return;
+
+  std::string::size_type flag_char = 0;
+  std::string flag;
+  std::vector<std::string> flags;
+
+  // Tokenize string to characters in flag, then words in flags.
+  while (flag_char < strlen(makeflags)) {
+    while (flag_char < strlen(makeflags) &&
+           !isblank(static_cast<unsigned char>(makeflags[flag_char]))) {
+      flag.push_back(static_cast<unsigned char>(makeflags[flag_char]));
+      flag_char++;
+    }
+
+    if (!flag.empty())
+      flags.push_back(flag);
+
+    flag.clear();
+    flag_char++;
+  }
+
+  // --jobserver-auth=<val>
+  for (size_t n = 0; n < flags.size(); n++)
+    if (flags[n].find(kAuthKey) == 0)
+      flag = flags[n].substr(strlen(kAuthKey));
+
+  // --jobserver-fds=<val>
+  if (flag.empty())
+    for (size_t n = 0; n < flags.size(); n++)
+      if (flags[n].find(kFdsKey) == 0)
+        flag = flags[n].substr(strlen(kFdsKey));
+
+  // -j 1
+  if (flag.empty())
+    for (size_t n = 0; n < flags.size(); n++)
+      if (flags[n].find("-j") == 0)
+        jobserver_closed_ = true;
+
+  // Check for fifo pipe.
+  if (flag.find(kFifoKey) == 0)
+    jobserver_fifo_ = true;
+
+  jobserver_name_.assign(flag);
+}
+
+bool PosixJobserverClient::Enabled() const {
+  return (rfd_ >= 0 && wfd_ >= 0) || jobserver_closed_;
+}
+
+unsigned char PosixJobserverClient::Acquire() {
+  unsigned char token = '\0';
+
+  // The first token is implicitly handed to a process.
+  // Fallback to non-parallel if jobserver-capable parent has no pipe.
+  if (token_count_ <= 0 || jobserver_closed_) {
+    token_count_ = 1;
+    return token;
+  }
+
+  int ret = read(rfd_, &token, 1);
+  if (ret < 0 && errno != EAGAIN && errno != EWOULDBLOCK) {
+    jobserver_closed_ = true;
+    if (!jobserver_fifo_)
+      Warning("pipe closed: %d (mark the command as recursive)", rfd_);
+    else
+      Fatal("failed to read from jobserver: %d: %s", rfd_, strerror(errno));
+  }
+
+  if (ret > 0)
+    token_count_++;
+
+  return token;
+}
+
+void PosixJobserverClient::Release(unsigned char* token) {
+  if (token_count_ < 0)
+    token_count_ = 0;
+  if (token_count_ > 0)
+    token_count_--;
+
+  // The first token is implicitly handed to a process.
+  // Writing is not possible if the pipe is closed.
+  if (*token == '\0' || jobserver_closed_)
+    return;
+
+  int ret = write(wfd_, token, 1);
+  if (ret != 1) {
+    Fatal("failed to write to jobserver: %d: %s", wfd_, strerror(errno));
+  }
+
+  *token = '\0';
+}
--- a/src/jobserver.h
+++ b/src/jobserver.h
@@ -1,0 +1,112 @@
+// Copyright 2024 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <string>
+
+/// The GNU jobserver limits parallelism by assigning a token from an external
+/// pool for each command. On posix systems, the pool is a fifo or simple pipe
+/// with N characters. On windows systems, the pool is a semaphore initialized
+/// to N. When a command is finished, the acquired token is released by writing
+/// it back to the fifo or pipe or by increasing the semaphore count.
+///
+/// The jobserver functionality is enabled by passing --jobserver-auth=<val>
+/// (previously --jobserver-fds=<val> in older versions of Make) in the MAKEFLAGS
+/// environment variable and creating the respective file descriptors or objects.
+/// On posix systems, <val> is 'fifo:<name>' or '<read_fd>,<write_fd>' for pipes.
+/// On windows systems, <val> is the name of the semaphore.
+///
+/// The classes parse the MAKEFLAGS variable and opens an object handle if needed.
+/// Once enabled, Acquire() must be called to acquire a token from the pool.
+/// If a token is acquired, a new command can be started.
+/// Once the command is completed, Release() must be called to return a token.
+/// The token server does not care in which order a token is received.
+struct Jobserver {
+  /// Return current token count or initialization signal if negative.
+  int Tokens() const { return token_count_; }
+
+  /// Read MAKEFLAGS environment variable and process the jobserver flag value.
+  virtual void Parse() {};
+
+  /// Return true if jobserver functionality is enabled and initialized.
+  virtual bool Enabled() const { return false; }
+
+  /// Implementation-specific method to acquire a token from the external pool
+  /// which is called for all tokens but returns early for the first token.
+  /// This method is called every time Ninja needs to start a command process.
+  /// Return a non-NUL char value on success (token acquired), and '\0' on failure.
+  /// First call always succeeds. Ninja is aborted on read errors from a fifo pipe.
+  /// The return is the token character to be saved for release after work is done.
+  virtual unsigned char Acquire() { return '\0'; }
+
+  /// Implementation-specific method to release a token to the external pool
+  /// which is called for all tokens but returns early for the last token.
+  /// The parameter is the token returned by Acquire() to be sent to the token server.
+  /// A token with the default value of '\0' will not be sent to the token server.
+  /// After sent, the token that the pointer parameter points to is cleared to '\0'.
+  /// It must be called for each successful call to Acquire() after the command exits,
+  /// even if subprocesses fail or in the case of errors causing Ninja to exit.
+  /// Ninja is aborted on write errors, and calls always decrement token count.
+  virtual void Release(unsigned char*) {};
+
+ protected:
+  /// The number of currently acquired tokens, or a status signal if negative.
+  /// This is used to estimate the load capacity for attempting to start a new job,
+  /// and when the implicit (first) token has been acquired (initialization).
+  /// -1: initialized without a token
+  ///  0: uninitialized or disabled
+  /// +n: number of tokens in use
+  int token_count_ = 0;
+
+  /// Whether a pipe to the jobserver token pool is closed
+  /// when it is expected to be open based on MAKEFLAGS
+  /// (e.g. subcommands not marked recursive, environment passed),
+  /// or the pipe is closed when expected to be closed
+  /// but when the parent process is jobserver-capable
+  /// (e.g. the parent jobserver process build is non-parallel).
+  bool jobserver_closed_ = false;
+
+  /// String of the parsed value of the jobserver flag passed to environment.
+  std::string jobserver_name_;
+
+  /// Substrings for parsing MAKEFLAGS environment variable.
+  static constexpr char const kAuthKey[] = "--jobserver-auth=";
+  static constexpr char const kFdsKey[]  = "--jobserver-fds=";
+  static constexpr char const kFifoKey[] = "fifo:";
+};
+
+struct PosixJobserverClient : public Jobserver {
+  /// Parse the MAKEFLAGS environment variable to receive the path / FDs
+  /// of the token pool, and open the handle to the pool if it is an object.
+  /// If a jobserver argument is found in the MAKEFLAGS environment variable,
+  /// and the handle is successfully opened, later calls to Enable() return true.
+  /// If a jobserver argument is found, but the handle fails to be opened,
+  /// the ninja process is aborted with an error, or, when the FDs provided are bad
+  /// the build falls back to non-parallel building and the client does not read.
+  explicit PosixJobserverClient();
+
+  void Parse() override;
+  bool Enabled() const override;
+  unsigned char Acquire() override;
+  void Release(unsigned char*) override;
+
+ private:
+  /// Whether the type of jobserver pipe supplied to ninja is named.
+  bool jobserver_fifo_ = false;
+
+  /// File descriptors to communicate with upstream jobserver token pool.
+  int rfd_ = -1;
+  int wfd_ = -1;
+};
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -84,8 +84,8 @@ struct Options {
 /// The Ninja main() loads up a series of data structures; various tools need
 /// to poke into these, so store them as fields on an object.
 struct NinjaMain : public BuildLogUser {
-  NinjaMain(const char* ninja_command, const BuildConfig& config) :
-      ninja_command_(ninja_command), config_(config),
+  NinjaMain(const char* ninja_command, const BuildConfig& config, Jobserver* jobserver) :
+      ninja_command_(ninja_command), config_(config), jobserver_(jobserver),
       start_time_millis_(GetTimeMillis()) {}
 
   /// Command line used to run Ninja.
@@ -93,6 +93,9 @@ struct NinjaMain : public BuildLogUser {
 
   /// Build configuration set from flags (e.g. parallelism).
   const BuildConfig& config_;
+
+  /// Client for jobserver to allow a parent process to control parallelism.
+  Jobserver* jobserver_;
 
   /// Loaded state (rules, nodes).
   State state_;
@@ -267,7 +270,8 @@ bool NinjaMain::RebuildManifest(const char* input_file, string* err,
   if (!node)
     return false;
 
-  Builder builder(&state_, config_, &build_log_, &deps_log_, &disk_interface_,
+  Builder builder(&state_, config_, jobserver_,
+                  &build_log_, &deps_log_, &disk_interface_,
                   status, start_time_millis_);
   if (!builder.AddTarget(node, err))
     return false;
@@ -1355,7 +1359,8 @@ int NinjaMain::RunBuild(int argc, char** argv, Status* status) {
 
   disk_interface_.AllowStatCache(g_experimental_statcache);
 
-  Builder builder(&state_, config_, &build_log_, &deps_log_, &disk_interface_,
+  Builder builder(&state_, config_, jobserver_,
+                  &build_log_, &deps_log_, &disk_interface_,
                   status, start_time_millis_);
   for (size_t i = 0; i < targets.size(); ++i) {
     if (!builder.AddTarget(targets[i], &err)) {
@@ -1542,6 +1547,14 @@ NORETURN void real_main(int argc, char** argv) {
 
   Status* status = Status::factory(config);
 
+  // Client of jobserver to manage job slots assigned to ninja.
+// TODO: jobserver client support for Windows
+#ifdef _WIN32
+  Jobserver jobserver;
+#else
+  PosixJobserverClient jobserver;
+#endif
+
   if (options.working_dir) {
     // The formatting of this string, complete with funny quotes, is
     // so Emacs can properly identify that the cwd has changed for
@@ -1558,14 +1571,14 @@ NORETURN void real_main(int argc, char** argv) {
   if (options.tool && options.tool->when == Tool::RUN_AFTER_FLAGS) {
     // None of the RUN_AFTER_FLAGS actually use a NinjaMain, but it's needed
     // by other tools.
-    NinjaMain ninja(ninja_command, config);
+    NinjaMain ninja(ninja_command, config, &jobserver);
     exit((ninja.*options.tool->func)(&options, argc, argv));
   }
 
   // Limit number of rebuilds, to prevent infinite loops.
   const int kCycleLimit = 100;
   for (int cycle = 1; cycle <= kCycleLimit; ++cycle) {
-    NinjaMain ninja(ninja_command, config);
+    NinjaMain ninja(ninja_command, config, &jobserver);
 
     ManifestParserOptions parser_opts;
     if (options.phony_cycle_should_err) {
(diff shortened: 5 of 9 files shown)


mpratt14:
I forgot I have to adapt to a windows build even if I'm not going to support it on windows...

mpratt14:
The CI for Windows is happy now, but it would be nice to have a tester for Windows...

src/jobserver.h
```diff
@@ -0,0 +1,107 @@
...
+  void Clear();
+
+private:
+  /// The number of currently acquired tokens, or the jobserver status if negative.
+  /// Used to verify that all acquired tokens have been released before exiting,
+  /// and when the implicit (first) token has been acquired (initialization).
+  /// -1: initialized without a token
+  ///  0: uninitialized or disabled
+  /// +n: number of tokens in use
+  int token_count_ = 0;
```
mpratt14, line 87: I want to point out this concept and ask whether or not this usage of an int is too unusual or non-standard.
This coincides with the last line of the constructor Jobserver::Jobserver() and the value of capacity in CanRunMore() using the absolute value function.

It's pretty easy to rework this. I just happened to have this idea first, to use the token number in place of Enabled() when there actually are no tokens yet.

mpratt14:
found and fixed some minor mistakes, added some commit tags

mpratt14:
sorry I didn't realize that there were "builder" constructors for the test suite, I built and ran the test suite this time.

mpratt14:
oops... I went too fast and everything is still "green" on segfault...

jhasse:
> We're very wary of changes that increase the complexity of Ninja, so a PR that implements both methods while one of them is technical superior and results in less code in Ninja (and to my understand that's the case for fifo), is very unlikely to get merged.

mpratt14:
The current struggle is regarding the creation of the Jobserver object in `real_main()`, nothing to do with the new files and new functionality. I was trying to avoid making functions that call Jobserver functions through another class and instead pass references to the object wherever it's needed, but I can always go back to the other way of creating the object within the Plan struct.

mpratt14:
ah, I see what you mean, it errored on readability...

but the other thing I have doubts about

```
Run ctest -C Release -vv
CMake Error: Unknown argument: -vv
CMake Error: Run 'ctest --help' for all supported options.
Error: Process completed with exit code 1.
```

mpratt14:
the bool now defaults to false, and set true with a simple if instead of a ternary, like in the rest of the project 👍🏼

mpratt14:
I believe that all the minor issues caught by the CI are handled now...

mpratt14:
some simplification:

I was reading the Google Style Guide and saw this
> ...we never allow non-­const reference parameters.

so changes in the last push are:

1. Converted all new references to pointers, to comply with the style guide for Jobserver being a non-const object.

Then I realized that I no longer need to create a Jobserver object for `build_test.cc` (at least in this commit), so

2. Pass NULL for Jobserver* in Builder instantiations in `build_test.cc`, check for null dereferencing Jobserver in `build.cc`

then finally, another opportunity to save lines:

3. Create Jobserver object in NinjaMain instead of real_main() since I realized that only 1 NinjaMain is created in a single process run.

mpratt14:
updated commit message

mpratt14:
@jhasse can we run the CI workflow again?

mcprat:
small organization update:
 - comment rewriting
 - style formatting
 - save some more lines
 - line wrapping
 - moved new `const` functions to header

mcprat:
reworded comment and alphabetical order in CMakeLists.txt / configure.py

src/build.h
```diff
@@ -224,6 +225,7 @@ struct Builder {
 
   State* state_;
   const BuildConfig& config_;
+  Jobserver* jobserver_;
```
kalvdans, line 228: Document what it means for `jobserver_` to be NULL.

src/jobserver.h
```diff
@@ -0,0 +1,107 @@
...
+
+  /// File descriptors to communicate with upstream jobserver token pool.
+  int rfd_ = -1;
+  int wfd_ = -1;
+#endif
+
+  /// Helpers for initialization.
+  std::string::size_type flag_char_ = 0;
+  std::string flag_;
+  std::vector<std::string> flags_;
```
kalvdans, line 106: Move these to be local variables in the constructor instead of member variables.

src/build.h
mcprat: It's never NULL when running the real `ninja`, only for the `ninja_test` binary

src/build.cc
```diff
@@ -164,8 +164,26 @@ Edge* Plan::FindWork() {
   if (ready_.empty())
     return NULL;
 
+// TODO: jobserver client support for Windows
+#ifndef _WIN32
```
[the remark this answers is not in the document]
mcprat: It's never NULL right now, but actually, I can make it NULL on Windows in order to remove some of them...

src/jobserver.h
mcprat: I can try, but earlier attempts to do something similar before I came up with this resulted in segfault when using the value of a local variable to define a member variable...

src/ninja.cc
```diff
@@ -94,6 +94,9 @@ struct NinjaMain : public BuildLogUser {
   /// Build configuration set from flags (e.g. parallelism).
   const BuildConfig& config_;
 
+  /// Client of jobserver to manage job slots assigned to ninja.
+  Jobserver jobserver_;
```
kalvdans, line 98: This declaration calls the constructor of Jobserver, but jobserver-posix.cc is never compiled for windows. Why don't we get a linker error on Windows?

src/build.h
kalvdans: Great! Add as a comment:

```suggestion
  Jobserver* jobserver_;  // Never NULL running the real ninja, only for the ninja_test binary
```

hwti:
When forgetting to mark the command as recursive (`+` in the makefile), there is a warning followed by an error.
Shouldn't it default to either -j1 (like make) or to the default ninja behavior ?

With make 4.3, building ninja itself from a Makefile :
```
ninja: using jobserver: 3,4
[1/33] CXX build/browse.oninja: warning: pipe closed: 3 (mark the command as recursive)
ninja: fatal: failed to read from jobserver: 3: Bad file descriptor
make: *** [Makefile:2: all] Error 1
```

hwti:
There is a `ninja: warning: invalid jobserver value: ''`  when there is no jobserver :
 - with just `make` (ie `MAKEFLAGS=`) : keeping the default ninja behavior seems logic here, but the warning should probably be avoided as it's a perfectly valid case
 - with `make -j1` (ie `MAKEFLAGS= -j1`), or `make -j` (ie `MAKEFLAGS= -j`) : shouldn't ninja behave as if the option was given to ninja ?
This is weird to have `-j2` limit to 2 subprocesses, but `-j1` still building in parallel.

src/ninja.cc
mcprat: constructor/destructor is blocked off by preprocessor macros for Windows in the header, but like we discussed I can assign NULL pointer here instead

src/jobserver.h
```diff
@@ -0,0 +1,107 @@
...
+// limitations under the License.
+
+#pragma once
+
+#include <string>
+#include <vector>
+
+#define AUTH_KEY  "--jobserver-auth="
+#define FDS_KEY   "--jobserver-fds="
+#define FIFO_KEY  "fifo:"
```
kalvdans, line 22: Move these to the .cc file, they are not used anywhere else.

src/jobserver.h
mcprat: segfault would always happen in all cases if the variables are not declared/instantiated correctly, and would never happen if they are.

src/jobserver.h
mcprat: they would potentially be used in the future `jobserver-win32.cc` for windows

mcprat:
rebase first...

src/build.cc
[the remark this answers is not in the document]
mcprat: Ultimately, even if the pointer to the Jobserver object is NULL, compilation would still need a definition of the functions after the check for NULL pointer as if it would pass even though it never would, so I have to put the preprocessor macros back. Now we have even more...

mcprat:
I addressed all review comments so far in one way or another, and more small changes:
 - "initialization helpers" that were in the Jobserver class are now local variables (with no segfault problems, I don't remember the exact segfault problem I had before but it must have been different).
 - relaxed the errors in the cases of being passed an invalid set of FDs (-2,-2) and failing to read from the pipe when it is non-fifo. the build continues as non-parallel, and a new `bool` facilitates that.
 - moved `Enabled()` function back from the header because, even though it is a `const` function, it is implementation-specific.
 - "TODO" comments are now also in ninja.cc, and for Windows, pass `NULL` into Builder object instead of creating the object to do nothing.
 - changed an error message and added more comments

mcprat:
a few more things:

 - catch length 0 of MAKEFLAGS and stop initializing if so
 - removed the warning for empty value, as the way it is currently written there is no distinction between the jobserver flag being missing (`MAKEFLAGS=''`) vs. the jobserver flag being present but with no value (`MAKEFLAGS='--jobserver-auth=""'`), and there is no need to make the distinction.
 - removed warning for failing to write to specifically a non-fifo pipe, since the read would fail first in practically all cases, and the message would be irrelevant if the cause is something else.
 - using !empty() instead of size() > 0

mcprat:
> * with `make -j1` (ie `MAKEFLAGS= -j1`), or `make -j` (ie `MAKEFLAGS= -j`) : shouldn't ninja behave as if the option was given to ninja ?

@hwti The purpose of being a client of the jobserver is to allow `make` (or in the future, another instance of `ninja`) to decide the number of parallel jobs *dynamically* alongside other subprocesses and other limitations like load average, where `-j N` passed to `make` is the maximum number of jobs that any subprocess of `make` can run at a time. That idea would be OK if and only if one subprocess of `make` ran at a time, but `make` attempts to run it's own set of subprocesses in parallel at the same time, because some builds may have subprocesses that do not have their own subprocesses (which is called non-recursive commands). In this case, the jobserver method instead balances parallelism between `make`'s other subprocesses and `ninja`'s subprocesses, so it is sometimes called a "job slot pool" since any process can join in no matter how many process parents up `make` is.

hwti:
> > * with `make -j1` (ie `MAKEFLAGS= -j1`), or `make -j` (ie `MAKEFLAGS= -j`) : shouldn't ninja behave as if the option was given to ninja ?
> 
> @hwti The purpose of being a client of the jobserver is to allow `make` (or in the future, another instance of `ninja`) to decide the number of parallel jobs _dynamically_ alongside other subprocesses and other limitations like load average, where `-j N` passed to `make` is the maximum number of jobs that any subprocess of `make` can run at a time. That idea would be OK if and only if one subprocess of `make` ran at a time, but `make` attempts to run it's own set of subprocesses in parallel at the same time, because some builds may have subprocesses that do not have their own subprocesses (which is called non-recursive commands). In this case, the jobserver method instead balances parallelism between `make`'s other subprocesses and `ninja`'s subprocesses, so it is sometimes called a "job slot pool" since any process can join in no matter how many process parents up `make` is.

I know, but with `-j1` or `-j`, make doesn't start a jobserver.
With `-j1`, make will only start one subprocess, so if ninja handle this particular value from `MAKEFLAGS`, this will enforce a non-parallel build.
For `-j -l 2`, there can be several make subprocesses, but they will all check the load average. This is not precise, but ninja could do the same too.

hwti:
> * relaxed the errors in the cases of being passed an invalid set of FDs (-2,-2) and failing to read from the pipe when it is non-fifo. the build continues as non-parallel, and a new `bool` facilitates that.

This means that with make 4.3, when ninja launches itself (like what I found with WebKit, when a CMake project reconfigures itself), there is a `ninja: warning: pipe closed: 3 (mark the command as recursive)` and the build is non-parallel.
The pipe shouldn't be closed, or if there is no other way maybe the `--jobserver-auth=` / `--jobserver-fds=` should be removed from `MAKEFLAGS`.

src/jobserver.h
kalvdans: I split out the text parsing to its own function, and added a unit test in https://github.com/ninja-build/ninja/commit/8580d5142eb2b01b9e327d52d86d8de454713472
What do you think? I can clean up the code if you agree that this is the way to go.

src/jobserver.h
```diff
@@ -0,0 +1,102 @@
...
+  /// The number of currently acquired tokens, or a status signal if negative.
+  /// Used to verify that all acquired tokens have been released before exiting,
+  /// and when the implicit (first) token has been acquired (initialization).
+  /// -1: initialized without a token
+  ///  0: uninitialized or disabled
+  /// +n: number of tokens in use
+  int token_count_ = 0;
+
+  /// String of the parsed value of the jobserver flag passed to environment.
+  std::string jobserver_name_;
```
kalvdans, line 87: Not used after the constructor, remove this member.

src/jobserver-posix.cc
```diff
@@ -0,0 +1,160 @@
...
+
+  // Return early if no makeflags are passed in the environment.
+  char* makeflags = std::getenv("MAKEFLAGS");
+  if (makeflags == nullptr || strlen(makeflags) == 0)
+    return;
+
+  std::string::size_type flag_char_ = 0;
+  std::string flag_;
+  std::vector<std::string> flags_;
+
```
kalvdans, line 36: Drop the trailing underscore, now when these are local variables instead of member variables.

mcprat:
> I know, but with `-j1` or `-j`, make doesn't start a jobserver.

I understand what you mean now... I had been used to testing with a build system that already uses ninja with `-j 1`, and I have been looking for parallelism when adding jobserver support. However, the opposite case is possible, without any options, if the jobserver client stops initialization early because of a problem, ninja will use the default parallelism of `nproc + 2`.

This is a tricky situation but I think I got it... I'm going to review what I just pushed again for a possible regression, and then do the variable renames later...

summary:
 - Most return statements are replaced with more uses of the new `bool` for falling back to non-parallel building.
 - "invalid value" warning now only applies to non-empty strings
 - specifically set FDs to 0 when falling back to non-parallel build (but FD 0 will never be read or written to)

mcprat:
> if there is no other way maybe the `--jobserver-auth=` / `--jobserver-fds=` should be removed from `MAKEFLAGS`.

It is the responsibility of the job token server to adjust the environment including the MAKEFLAGS variable. Make versions 4.3.90+ and 4.4.x do this by appending `--jobserver-auth=-2,-2` to the end of MAKEFLAGS. The warning is appropriate for the cases where this does not happen, as the environment variable suggests that the pipe will be open when it is never actually opened.

hwti:
> > if there is no other way maybe the `--jobserver-auth=` / `--jobserver-fds=` should be removed from `MAKEFLAGS`.
> 
> It is the responsibility of the job token server to adjust the environment including the MAKEFLAGS variable. Make versions 4.3.90+ and 4.4.x do this by appending `--jobserver-auth=-2,-2` to the end of MAKEFLAGS. The warning is appropriate for the cases where this does not happen, as the environment variable suggests that the pipe will be open when it is never actually opened.

Here make is fine, it's about the make -j2 => ninja (OK, runs a rule regenerating build.ninja, then later pipe closed).
I thought ninja executed a second instance, since I see `ninja: using jobserver: 3,4` twice, but it doesn't seem to be the case.
It seems to "restart" after having regenerated the build.ninja, closing the pipe fds before the the second `using jobserver` trace.

The same case (ninja restarting after the CMake reconfigure) works when the jobserver uses a fifo (make 4.4).

When trying to reproduce with a manually written ninja file (simulating a buildsystem which generates `build.ninja` from `depfile`), I managed to get a case which even fails writing to the pipe.

`build.ninja` :
```
rule DUMMY
  command = echo dummy

rule REGENERATE
  command = touch build.ninja

build all: phony all1 all2

build all1: DUMMY
build all2: DUMMY

build build.ninja: REGENERATE depfile
```
`Makefile` :
```
all:
        touch depfile
        +ninja
```
=>
```
$ make -j2
touch depfile
ninja
ninja: using jobserver: 3,4
[1/1] touch build.ninja
ninja: using jobserver: 3,4
[1/3] echo dummy
dummy
ninja: fatal: failed to write to jobserver: 4: Bad file descriptor
make: *** [Makefile:3: all] Error 1
```

avikivity:
Suggest to update the manual, so users are aware that the feature exists and how to operate it.

src/build.cc
```diff
@@ -614,6 +645,10 @@ vector<Edge*> RealCommandRunner::GetActiveEdges() {
 
 void RealCommandRunner::Abort() {
   subprocs_.Clear();
+// TODO: jobserver client support for Windows
+#ifndef _WIN32
+  jobserver_->Clear();
+#endif
```
avikivity, line 651: Will this be called after SIGTERM/SIGINT? The specification recommends so.

mcprat: yes, that's exactly the reason I put it there.

mcprat:
> I thought ninja executed a second instance, since I see `ninja: using jobserver: 3,4` twice, but it doesn't seem to be the case.
> It seems to "restart" after having regenerated the build.ninja, closing the pipe fds before the the second `using jobserver` trace

Earlier, I did a small refactor:

> Create Jobserver object in NinjaMain instead of real_main() since I realized that only 1 NinjaMain is created in a single process run.

It sounds like that observation I made is wrong, so I'll put the client instantiation back in `real_main()`. I should have known that's the reason why you sometimes see 2 subprocesses that are numbered as `1/N` targets...

mcprat:
update summary:
 - small regression: `Acquire()` must return false when pipe is closed
 - moved instantiation of the client back to `real_main()`

src/jobserver-posix.cc
```diff
@@ -0,0 +1,170 @@
...
+#include <cassert>
+#include <cstring>
+
+#include "util.h"
+
+Jobserver::Jobserver() {
+  assert(!Enabled());
+
+  // Return early if no makeflags are passed in the environment.
```
digit-google, line 28: Please move all parsing logic to a separate static method that can be unit-tested properly. Also be aware that the GNU Make documentation states explicitly: [Be aware that the MAKEFLAGS variable may contain multiple instances of the --jobserver-auth= option. Only the last instance is relevant](https://www.gnu.org/software/make/manual/html_node/Job-Slots.html). Hence this should be implemented properly (and tested).

src/jobserver-posix.cc
```diff
@@ -0,0 +1,170 @@
...
+    return;
+
+  std::string::size_type flag_char_ = 0;
+  std::string flag_;
+  std::vector<std::string> flags_;
+
+  // Tokenize string to characters in flag_, then words in flags_.
+  while (flag_char_ < strlen(makeflags)) {
+    while (flag_char_ < strlen(makeflags) &&
+           !isblank(static_cast<unsigned char>(makeflags[flag_char_]))) {
```
digit-google, line 40: nit: isblank() is locale-dependent, which leads to surprises and is generally slow. It is easier to just compare with `' '` and `'
t'` in this case.

src/jobserver-posix.cc
```diff
@@ -0,0 +1,170 @@
...
+  if (rfd_ >= 0 && wfd_ >= 0)
+    Info("using jobserver: %s", jobserver);
+  else if (jobserver_fifo_ && (rfd_ == -1 || wfd_ == -1))
+    Fatal("failed to open jobserver: %s: %s", jobserver, strerror(errno));
+  else
+    jobserver_closed_ = true;
+
+  // Fallback to non-parallel if jobserver-capable parent has no jobserver.
+  if (jobserver_closed_)
+    rfd_ = wfd_ = 0;
```
digit-google, line 97: did you mean -1 here?

src/jobserver.h
```diff
@@ -0,0 +1,102 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <string>
+#include <vector>
```
digit-google, line 18: nit: this include is no longer necessary.

src/jobserver.h
digit-google: If you really want to expose these, convert these to static constexpr members instead for proper scoping, e.g.:

```
// in .h file
struct JobServer {
   static constexpr char kAuthKey[] = "--jobserver-auth=";
   static constexpr char kFdsKey[] = "--jobserver-fds=";
   static constexpr char kFifoKey[] = "fifo:";
};

// in .cc file, to appease picky compilers.
constepxr char JobServer::kAuthKey[];
constexpr char JobServer::kFdsKey[];
constexpr char JobServer::kFifoKey[];
```

This will avoid polluting the preprocessor state of any source or header file that includes this transitively.

src/build.h
digit-google: To make the code easier to follow, I recommend to instead make JobServer a base class that does nothing (never enabled), then add a derived PosixJobServer  class. This will allow you to add a Win32JobServer class later, and remove the need for all the `#ifdef _WIN32` section sprinkled over the code.

mcprat: @digit-google I understand class inheritance can work, but in this case I think I would have to fill the base class with no-op functions to get rid of the preprocessor stuff. Would that not be equally ugly?

src/jobserver-posix.cc
mcprat: setting the FDs to 0 when the pipe is closed is to ensure that Enabled() still returns true so that the default parallelism of `processors + 2` or whatever the user sets for `-j` is blocked. 

I was trying to avoid changing the definition of Enabled() to anything other than the FD numbers, but if you think it looks better the function can be defined as:

```
bool Jobserver::Enabled() const {
  return (rfd_ >= 0 && wfd_ >= 0) || jobserver_closed_;
}
```

then there would be no need to set the FDs here, although that might also look counterintuitive...

src/jobserver.h
mcprat: the vector include just needs to be moved to the cc file

src/jobserver.h
```diff
@@ -0,0 +1,102 @@
...
+
+  /// Return current token count or initialization signal if negative.
+  int Tokens() const { return token_count_; }
+
+  /// Implementation-specific method to acquire a token from the external pool
+  /// which is called for all tokens but returns early for the first token.
+  /// This method is called every time Ninja needs to start a command process.
+  /// Return true on success (token acquired), and false on failure (no tokens
+  /// available). First call always succeeds. Ninja is aborted on read errors.
+  bool Acquire();
```
digit-google, line 64: Unfortunately, I think this interface needs work. The GNU Make specification states that [It’s important that when you release the job slot, you write back the same character you read. Don’t assume that all tokens are the same character;](https://www.gnu.org/software/make/manual/html_node/POSIX-Jobserver.html), which means this method must at least return a token job byte value, that should be passed to Release().

src/jobserver-posix.cc
digit-google: Btw, see also to implement the following requirement:

```
Your tool may also examine the first word of the MAKEFLAGS variable and look for the character n. If this character is present then make was invoked with the ‘-n’ option and your tool may want to stop without performing any operations
```

From https://www.gnu.org/software/make/manual/html_node/POSIX-Jobserver.html

src/build.h
digit-google: Yes, you would need no-op methods for the base class, but this leads to code that is easier to read and maintain over time. You can have test-specific implementations if you want as well. Since this class is not performance critical, you can do something like that:

1) Add an `std::unique_ptr<Jobserver> jobserver_` member to this class, and default-initialize it with a base/null instance that does nothing. This avoids modifying tests that don't care about the feature at all.

2) Add a `SetJobserver(std::unique_ptr<Jobserver> jobserver)` method to replace the instance's jobserver with a new value. This can be called after constructing a Builder instance where it is needed only. This passes ownership to avoid lifecycle management issues.

3) The rest of the code here can use `jobserver_->xxxx()` without worrying about the rest.

src/jobserver-posix.cc
```diff
@@ -0,0 +1,170 @@
...
+
+#include <fcntl.h>
+#include <unistd.h>
+
+#include <cassert>
+#include <cstring>
+
+#include "util.h"
+
+Jobserver::Jobserver() {
```
digit-google, line 25: I would recommend splitting this PR into multiple commits, i.e.:

1) One that adds the Jobserver class, and its Posix implementation + appropriate unit-tests for it. Also try to make the class as independent from the rest of Ninja as possible (e.g. to not call Warning() or Info() in the parser function, leave that to clients).

2) One that adds usage of the class to build.cc / ninja.cc.

src/jobserver-posix.cc
digit-google: Frankly, this is the kind of clever coding that ends up creating confusion or subtle bugs over time. At a minimum there should be a comment over any subtle code paths when they are really required (e.g. for performance reason).

In this specific case, I recommend you use_ an `enabled_` boolean flag instead, and set the descriptors to -1.

src/jobserver-posix.cc
```diff
@@ -0,0 +1,170 @@
...
+
+  char token;
+
+  int ret = read(rfd_, &token, 1);
+  if (ret < 0 && errno != EAGAIN && errno != EWOULDBLOCK) {
+    jobserver_closed_ = true;
+    if (!jobserver_fifo_)
+      Warning("pipe closed: %d (mark the command as recursive)", rfd_);
+    else
+      Fatal("failed to read from jobserver: %d: %s", rfd_, strerror(errno));
```
digit-google, line 139: Is there a way to fallback gracefully to the usual mode when this happens instead?

In general, it is better to avoid calling Fatal() in methods like these, because these conditions cannot be properly unit-tested.

src/jobserver-posix.cc
```diff
@@ -0,0 +1,170 @@
...
+    flag_.clear();
+    flag_char_++;
+  }
+
+  // --jobserver-auth=<val>
+  for (size_t n = 0; n < flags_.size(); n++)
+    if (flags_[n].find(AUTH_KEY) == 0)
+      flag_ = flags_[n].substr(strlen(AUTH_KEY));
+
+  // --jobserver-fds=<val>
```
digit-google, line 57: Note that this option is never going to work unless you can modify subprocess-posix.cc and subprocess-win32.cc to pass these extra file descriptors to all child processes. Which is a very non-trivial change. Do you have a commit planned for this? Otherwise, I would leave this out of the current PR, explaning why it's not implemented yet.

(Alternatively, support it in the parser function, but ensure that the client recognizes this is not implemented yet, and ignore it).

src/jobserver-posix.cc
fogti: it feels somewhat bad (harder to detect errors, or leading to potentially unusual output) when using numbers in FD variables which are strictly valid file descriptors. It is probably better to use an explicit negative magic constant instead.

src/jobserver-posix.cc
```diff
@@ -0,0 +1,170 @@
...
+  if (flag_.empty())
+    for (size_t n = 0; n < flags_.size(); n++)
+      if (flags_[n].find(FDS_KEY) == 0)
+        flag_ = flags_[n].substr(strlen(FDS_KEY));
```
digit-google, line 61: These two successive loops means that `--jobserver-fds=4,5 --jobserver-auth=pipe:<name>` will not use the pipe method, even though it should. It would be nicer if this was handled properly.

Also maybe add a comment explaning what `--jobserver-fds`  is, because it is not on the official GNU Make documentation, since it was supposed to be used internally until GNU Make 4.2 renamed it to `--jobserver-auth`.

https://lists.gnu.org/archive/html/info-gnu/2016-05/msg00013.html

src/jobserver.h
digit-google: Technically Ninja is C++11 only, so no `std::optional<char>` is possible, but it would be easy to have a dedicated class such as:

```
struct Jobserver {
  ...
  struct Token {
    bool has_value() const { return value_ >= 0; }
    unsigned char value() const {
      assert(has_value() && "Cannot take value from invalid token");
      return static_cast<unsigned char>(value_);
    }
   private:
    int char value_ = -1;
  };
};
```

Which is easier to reason about than a magic integer.

src/build.h
mcprat: I split the class into a basic pair of base and derived classes, now it is down to one `#ifdef _WIN32`. The rest of these tips revolving around `unique_ptr` is a little advanced for me, so I'm pushing what I have and we can take some more time to figure out that bit and how useful it would be.

src/jobserver.h
mcprat: > I split out the text parsing to its own function, and added a unit test in [8580d51]

its a nice start for unit tests, but I don't think we want to introduce tuples to ninja when it's not necessary. ***maybe*** a `std::pair` for the FDs but I still don't see it as necessary.

src/jobserver.h
mcprat: > Not used after the constructor, remove this member.

now it is

src/jobserver-posix.cc
mcprat: > Btw, see also to implement the following requirement:
> 
> 
> > Your tool may also examine the first word of the MAKEFLAGS variable and look for the character n. If this character is present then make was invoked with the ‘-n’ option and your tool may want to stop without performing any operations
> 
> 
> From https://www.gnu.org/software/make/manual/html_node/POSIX-Jobserver.html

I wish they didn't mention this in that part of the manual. It's completely out of scope for the jobserver. The jobserver client should not be responsible for determining whether ninja does a "dry run" even though it happens to be parsing MAKEFLAGS which can have many different flags for many different reasons... 

This should be implemented in a separate commit (and separate PR) and directly with the BuildConfig object instead of involving the Jobserver objects. Traditionally, ninja does not rely on the environment for any of it's configuration flags, so that's another conversation as well.

src/jobserver-posix.cc
mcprat: > did you mean -1 here?

this has been reworked, and comments added for `jobserver_closed_` in the header.

src/jobserver.h
mcprat: > Unfortunately, I think this interface needs work. The GNU Make specification states that 
> > It’s important that when you release the job slot, you write back the same character you read. Don’t assume that all tokens are the same character

Another thing I wish the manual didn't say...

This was discussed a little bit in #2450. As far as I can tell, there are no signs that this part of the jobserver design has ever or will ever be implemented anywhere. It may have been part of the original draft of how the jobserver will work, but then they realized they lacked a reason for one token to mean something different that another one. Current GNU Make source code uses the `+` and even mentions that it is ambiguous since success is determined by the read or write itself and not the content.

```
/* Token written to the pipe (could be any character...)  */
static char token = '+';
```

I think that it should be very easy to implement in the future if we *really* need to, so better to forget about it. However, if it makes you more comfortable of course I can do it now.

src/jobserver-posix.cc
mcprat: the remaining calls to Fatal() are for the following cases:

1. MAKEFLAGS from environment told ninja that a fifo object needs to have file descriptors open for it. The open() syscall is ran and errored. 
2. File descriptors have been opened to the fifo object, but reading from it has failed not due to blocking. This likely means that the fifo object has been deleted while we have file descriptors that point to nothing, or maybe a problem with the filesystem itself.
3. File descriptors have been opened to the fifo object, but writing to it has failed, again suggesting that the fifo object has been deleted while we have file descriptors that point to nothing or a problem with the filesystem.

They can become warnings, but I think they protect execution from continuing when something is extremely wrong. It should be exceedingly rare for the program to step into these Fatal() calls and they are probably out of scope for unit tests anyway, but I'll let you decide.

src/jobserver-posix.cc
mcprat: > Note that this option is never going to work unless you can modify subprocess-posix.cc and subprocess-win32.cc to pass these extra file descriptors to all child processes. 

Are you sure about that? I don't see a difference in functionality between Make 4.1 and Make 4.2 for the jobserver. There was a lot of refactoring going on, but the commit that renames the flag is literally just a rename. I'll do some testing regarding this though.

> Do you have a commit planned for this? Otherwise, I would leave this out of the current PR, explaning why it's not implemented yet.

No, I don't have a commit for that. I imagine that any problem with the simple pipe with invalid FDs passed will fallback to non-parallel building.

src/jobserver-posix.cc
mcprat: > These two successive loops means that `--jobserver-fds=4,5 --jobserver-auth=pipe:<name>` will not use the pipe method, even though it should. It would be nicer if this was handled properly.

It is handled properly 👍🏼 

This is one of the first changes I made to #2450 so it has the most testing behind it so far.

First, a match to `--jobserver-auth` is searched and if found, `flag` is set. Then, only if `flag` is empty, search for the deprecated name. Then (again), only if `flag` is empty, look for the possibility of Make being used with `-j 1`.

> Also maybe add a comment explaning what `--jobserver-fds` is, because it is not on the official GNU Make documentation, since it was supposed to be used internally until GNU Make 4.2 renamed it to `--jobserver-auth`.

This is quickly mentioned in the comments of the header but I can expand on that if you want.

mcprat:
big update...

summary:
 - the destructor has been removed. I realized that it never gets run at all. Also, it is not the responsibility of the client to close the pipe. Emergency return of tokens before exiting still happens during call to `Clear()`.
 - the constructor has been split into a `Parse()` function for the majority of the parsing again.
 - the Jobserver class is split into base and derived classes, reducing preprocessor `#if`'s to 1.
 - `Enabled()` uses `jobserver_closed_` instead of using `jobserver_closed_` to put fake FDs.
 - macros for constant strings have been converted to static constexpr's as suggested.
 - removed underscores from local variables that were previously member variables.
 - moved `<vector>` include from the header to the TU.
 - small rewrites to comments and warnings.

src/jobserver.h
mcprat: I worked a little on saving the value of the token itself, and it actually allows for a little bit of simplification in some spots for existing files, so after I make sure it works right I might as well include it here. However, it makes the definition of the new `Clear()` function a little more complex...

src/jobserver-posix.cc
mcprat: @digit-google funny enough, instead of finding a problem between Make 4.1 and Make 4.2, I found a problem between Make 4.2.1 and Make 4.3. This bug makes any process into a zombie if and only if they try to read a token from the token server and the token is not available AND there is no other subprocess that will finish so that a token is delivered back into the pool to be redirected to the zombie process so it can run. In my testing use case this makes `make -j 3` fail while `make -j` without a number works.

This is because the pipe used to be blocking on the write side of the server, the case described by POSIX as:

> If some process has the pipe open for writing and O_NONBLOCK is clear, read() will block the calling thread until some data is written or the pipe is closed by all processes that had the pipe open for writing.

One way we can adapt is to check the NONBLOCK flag and force non-parallel if it is cleared. However, most of the responsibility here should be to the user who should know the limitations of Make 4.2.1 and earlier.

bug:
https://savannah.gnu.org/bugs/?51159

fixed with commit "[SV 51159] Use a non-blocking read with pselect to avoid hangs.":
https://git.savannah.gnu.org/cgit/make.git/commit/?id=b552b05251980f693c729e251f93f5225b400714

mcprat: A little more context, the way the client is currently written, ninja tries to get a second token before returning the first one it just got, so this Make bug causes a ninja build to hang even if it is not multiple recursive levels deep.

mcprat:
smaller update, this might be considered mergeable now.
hopefully all major issues are handled so we can focus on style and nitpicking...

summary:

 - now handling and storing the actual token character instead of just a count. this allowed for some simplification so there is a line decrease. this is written assuming a char is 1 byte.
 - added a case to fallback to non-parallel build, when the token server provides an FD to read tokens from that is blocking (Make versions 4.2.1 and earlier).
 - some light cleanup and rewording
 - tested all the way back to Make 4.0, which is when Windows support for jobserver was added. anything earlier than that can be considered "ancient" enough to ignore...

src/jobserver-posix.cc
```diff
@@ -0,0 +1,164 @@
...
+  if (ret != 1) {
+    Fatal("failed to write to jobserver: %d: %s", wfd_, strerror(errno));
+  }
+
+  *token = '\0';
+}
+
+void PosixJobserverClient::Clear(std::vector<Edge*> edges) {
+  for (size_t n = 0; n < edges.size(); n++)
+    Release(&edges[n]->job_token_);
```
digit-google, line 163: Can you avoid puting this code here to avoid un-necessary coupling (i.e. the JobServer class doesn't need to know anything about the Edge class). There is also no reason to pass a vector by value here, which will perform a copy, use a const reference instead, and consider using for-loop constructs, e.g.:

```
void ClearJobTokens(const std::vector<Edge*> edges, JobServer* jobserver) {
  for (const Edge* edge : edges)
    jobserver->Release(&edge->job_token_);
}
```

src/jobserver.h
```diff
@@ -0,0 +1,116 @@
...
+  /// available). First call always succeeds. Ninja is aborted on read errors.
+  virtual unsigned char Acquire() { return '\0'; }
+
+  /// Implementation-specific method to release a token to the external pool
+  /// which is called for all tokens but returns early for the last token.
+  /// Return a previously acquired token to the external token pool.
+  /// It must be called for each successful call to Acquire() after the command
+  /// even if subprocesses fail or in the case of errors causing Ninja to exit.
+  /// Ninja is aborted on write errors, and otherwise calls always succeed.
+  virtual void Release(unsigned char*) {};
```
digit-google, line 62: Please explain what the pointer being passed as argument here should point to? E.g. is it ok to call this with a pointer that points to a value that was never the result of a previous Acquire() code (as suggested by your Clear() function implementation). In which case, what is the default value to be used (also apparently `\0` from the code, but you should make that clear in the documentation). Also will the function modify the pointed value or not?

Ideally, users of the API should not have to guess these details by looking at the source code.

I suggest writing a dedicated move-only Token class, even if trivial, to better encapsulate these semantics.

```
/// A wrapper for token values acquired or released to the pool.
/// A default instance has no value, and is used to indicate that
/// no token is available.
struct Token {
  // Default constructor builds a value-less token.
  Token() = default;
  
  // Explicit constructor for a Token with a value from the pipe.
  explicit Token(uint8_t value) : value_(static_cast<int>(value)) {}
  
  // Move operations are allowed.
  Token(Token&& other) noexcept : value_(other.value_) { other.value_ = -1; }
  Token& operator=(Token&&) noexecpt = default;
  
  // Copy operations are forbidden.
  Token(const Token&) = delete;
  Token& operator=(const Token&) noexcept = default;
  
  /// Returns true if this instance contains a value received from the pipe.
  bool HasValue() const { return value_ != -1; }
  
  /// Return underlying value. It is a runtime error to call this method
  /// if HasValue() returns false.
  uint8_t  GetValue() const {
    assert(HasValue());
    return static_cast<uint8_t>(value_); 
  }
  
  int value_ = -1;
};
```

Then you can have `Acquire()`  return a Token by value, and `Release()` take a token by value, and forget about pointers entirely, e.g.:

```
/// Try to acquire a token from the pool. A value-less token instance is returned
/// if no token is available.
virtual Token Acquire() { return Token(); }

/// Release a previously acquire token to the pool. Does nothing if the
/// token argument has no value.
virtual void Release(Token token) {}
```

mcprat: I rewrote comments in the header to reflect the change in functionality and new returns and arguments.

Making a whole class just to manage the value of the tokens seems complicated and overkill to me. It makes the function declarations read better, but the Token class itself is not very readable and adds a lot of object constructs for not much benefit...

Is there any benefit to avoiding pointers here? or avoiding `char`?

What if we just made a simple typedef for the tokens?

I'm very confident that a jobserver will never return a NUL char unless it's to indicate "no tokens available" just as I am using it to mean. I'm also pretty confident that Ninja will never receive anything other than a `'+'` when being used with Make, probably even with forks of Make...

mcprat:
applied some review comments

summary:

 - moved new Clear() function to the RealCommandRunner class as ClearJobTokens()
 - ClearJobTokens() function uses range-based for loop statement
 - ClearJobTokens() function takes a const reference for the vector
 - removed headers for vectors and Edge class from jobserver.h
 - reworded comments in jobserver.h

mcprat:
> * Add an `std::unique_ptr<Jobserver> jobserver_` member to this class, and default-initialize it with a base/null instance that does nothing. This avoids modifying tests that don't care about the feature at all.
> * Add a `SetJobserver(std::unique_ptr<Jobserver> jobserver)` method to replace the instance's jobserver with a new value. This can be called after constructing a Builder instance where it is needed only. This passes ownership to avoid lifecycle management issues.


> nit: isblank() is locale-dependent, which leads to surprises and is generally slow. It is easier to just compare with ' ' and ' t' in this case.

> I suggest writing a dedicated move-only Token class, even if trivial, to better encapsulate these semantics.

@digit-google for these comments from you, would you consider any of them critical that I must do? Or perhaps keeping things simple may be better at this point?

Presence of `#ifdef _WIN32` is down to 1, and I don't think that last one can be removed, or can it?

digit-google:
Oh, I realize I was completely wrong in one of my earlier comments where I stated that Ninja would not pass file descriptor to command sub-processes. It actually totally does, so using `--jobserver-auth=R,W` works fine on Posix. I tested that actually with this patch. Sorry for the noise.

(Turns out that posix_spawn() doesn't close heritable file descriptors as I mistakenly thought).

mcprat:
No problem, I had to double-check because I wasn't sure myself, but at least documentation on file descriptors is pretty clear. There is a lot of confusion about how `O_CLOEXEC` is used by Make, but it has nothing to do with the external pipes.

mcprat:
- added `const` to makeflags variable again

src/jobserver-posix.cc
```diff
@@ -0,0 +1,160 @@
...
+
+    if (!flag.empty())
+      flags.push_back(flag);
+
+    flag.clear();
+    flag_char++;
+  }
+
+  // --jobserver-auth=<val>
+  for (size_t n = 0; n < flags.size(); n++)
```
neheb, line 91: these loops look like they're missing break;

src/jobserver-posix.cc
```diff
@@ -0,0 +1,160 @@
...
+
+  std::string::size_type flag_char = 0;
+  std::string flag;
+  std::vector<std::string> flags;
+
+  // Tokenize string to characters in flag, then words in flags.
+  while (flag_char < strlen(makeflags)) {
+    while (flag_char < strlen(makeflags) &&
+           !isblank(static_cast<unsigned char>(makeflags[flag_char]))) {
+      flag.push_back(static_cast<unsigned char>(makeflags[flag_char]));
```
neheb, line 79: I wonder if emplace_back and no cast is better here.

src/jobserver-posix.cc
mcprat: lack of `break` is intentional, so only the last matching value is the one that takes effect

neheb: Sounds like it would be better to reverse the loop then.

src/jobserver-posix.cc
mcprat: I'm not sure exactly what the difference is, I would have to read about it...

the cast is not strictly necessary, it's just protection against invalid characters that I ran into while reading about tokenizing strings.

I saw something about using a `std::vector<uint8_t>` instead of `std::string` but that might be confusing...

src/jobserver-posix.cc
mcprat: sure it can be reversed, but since this will only be run once per execution of ninja it's not performance critical, so I lean towards the simplest way to write it with less lines and no brackets, etc...

src/jobserver-posix.cc
mcprat: apparently after optimization, `push_back` and `emplace_back` are exactly the same, unless you are doing fancy constructors for the elements in the vector.

mcprat: > ... no cast is better here.

@neheb I got it from the example here

https://en.cppreference.com/w/cpp/string/byte/isblank

mcprat:
@digit-google @neheb would this be better? to make the range explicitly equal to `isgraph()`:

```
  // Tokenize string to characters in flag, then words in flags.
  while (flag_char < strlen(makeflags)) {
    while (flag_char < strlen(makeflags) &&
           ' ' < makeflags[flag_char] && makeflags[flag_char] <= '~') {
      flag.push_back(makeflags[flag_char]);
      flag_char++;
    }

...
...
```

nanonyme:
Is this now obsolete with https://github.com/ninja-build/ninja/pull/2506 merged?

jhasse:
yes

PR #2041, opened 2021-11-03
CI: remove unused params, clarify steps

scivision:
* remove unneeded CMake download from 2 years ago I made in 5ae2ce8
* clarify configure/build logs by making each separate steps

PR #2479, opened 2024-08-19
GitHub Actions: Prevent ctest invocation error on macOS

mpratt14:
ping @jhasse @nico

It seems that new versions of CTest now require case-sensitive arguments.

To prevent errors related to case-sensitivity of arguments, use the long option form for verbosity, and fallback to printing out the help message and running normally.

While at it, print the version in all cases for reference.


mpratt14:
@nico @jhasse @digit-google I know you all must be very busy, sorry to be impatient on this one, but this is just 1 line that fixes the CI test. Can we do it?

My other PR's CI test states:

```
Run ctest -C Release -vv
CMake Error: Unknown argument: -vv
CMake Error: Run 'ctest --help' for all supported options.
```

mpratt14:
a similar fix is part of #2041

jhasse:
I would prefer a version which avoids ctest all together. Not sure what the advantages of it are besides issue likes this.

There were other issues with ctest in the past and my general opinion of it is, that it should be avoided at all costs and nuked from existance.

mpratt14:
@jhasse ok. no more `ctest`

mpratt14:
I'm not going to bother changing the title of the PR, if that's ok...

mpratt14:
@jhasse I dug deeper and found out that `--config` is only for multi-config generators

When `ctest` was used with the "Xcode" generator, it should have been used directly in the subdirectory without a `-C` option

mpratt14:
for context

https://stackoverflow.com/questions/24460486/cmake-build-type-is-not-being-used-in-cmakelists-txt

mpratt14:
nevermind after reading more documentation I see it is considered a multi-config generator even though many people do not use it that way.

so now, just correcting the path

digit-google:
Thank you for this, I am glad we can get rid of CTest too. @jhasse, can we submit this asap to unblock all other PRs that fail their checks because of this issue?

digit-google:
Thanks!

PR #2483, opened 2024-08-28
Add 'dupbuild' as deprecated warning

dendy:
Fixup after 8f47d5aa33c6c303a71093be2eac02672dfb2966

Since v1.12 ninja breaks compatibility with existing generators, particularly bob build system, which always invokes ninja with -w dupbuild=err

See: https://github.com/ARM-software/bob-build/blob/master/bob.bash

This patch should also be applied to the 'release' branch.

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -1230,6 +1230,10 @@ bool WarningEnable(const string& name, Options* options) {
   } else if (name == "phonycycle=warn") {
     options->phony_cycle_should_err = false;
     return true;
+  } else if (name == "dupbuild=err" ||
+             name == "dupbuild=warn") {
+    Warning("deprecated warning 'dupbuild'");
+    return true;
   } else if (name == "depfilemulti=err" ||
              name == "depfilemulti=warn") {
     Warning("deprecated warning 'depfilemulti'");


tsniatowski:
This would be nice to have, I think it's unfortunate that #2356 causes a hard breakage for people that were already opting-in to the warning.

digit-google:
PR #2479 was submitted to fix the MacOS issue, please rebase on top of current `origin/master` and re-upload to get the checks to pass.

dendy:
Rebased, all tests passed.

mcprat:
commit message doesn't need the last line about backporting

and if you want to be very formal `See:` can be replaced with `Link:`

dendy:
> commit message doesn't need the last line about backporting

Agree, was hesitating about that, but decided to add that line for better visibility. Will remove.

> and if you want to be very formal See: can be replaced with Link:

Don't want to be over formal, unless there are strict rules. Any reason to prepend "Link:", isn't it obvious that link is a link? :-) I know people who have a habit to always begin commit body with "Description:" which is silly.

mcprat:
I agree that is silly... I just mentioned `Link:` because that's de-facto linux standard tagging and makes it greppable

PR #2485, opened 2024-08-28
Fix `inputs` tool logic and add new formatting options.

### LINKED ISSUE #2482: Ninja `-t inputs` output is incorrect for phony targets
The `-t inputs` tool does not report correct results when phony targets exist in the build plan, for example, with:

```
rule cat
  command = cat $in > $out
build out1: cat in1
build out2: cat in2
build all: phony out1 out2
```

Then `ninja -t inputs all` will return

```
in1
in2
```

Which misses the `out1` and `out2` targets which are real inputs for `all`.

The problem is the implementation of the stand-alone CollectInputs() function in src/ninja.cc.

digit-google:
This fixes the logic used by the `inputs` tool to report correctly the inputs of phony targets.

- Move the logic into its own InputsCollector class, and add proper unit-tests to it.

- Add `--no-shell-escape`, `--dependency-order` and `--print0` options to the tool to alter the order and format of the result for easier processing by tools.

- Add new integration tests for the problematic cases.

Fixes #2482


mcprat:
would it be too much to ask for this to be split into 3 commits?

digit-google:
I have split the PR into three commits. I don't really need to `--print0` option and can remove it for simplicity if you prefer.
@jhasse, can we submit this now? The correctness of this tool's output is very important for the Fuchsia build, as it is used after the build to pickup artifacts to upload to cloud storage, while filtering out stale files from previous incremental builds.

jhasse:
Thanks!

PR #2487, opened 2024-08-28
Faster elide middle implementation

digit-google:
Drastically speed up the implementation of the ElideMiddle() function by getting rid of `std::regex` usage abd unnecessary string allocations / reallocations. This also removes about 65 kiB from the stripped Linux binary.

- Move the function to its own source file and add a performance test.
- Fix a bug in the implementation that removed color sequences from elided results + avoid allocations (x1.8 speedup)
- Remove std::regex usage completely, using custom parsing classes (x5.8 speedup)
- Add fast path to avoid ANSI sequence stripping when not needed.

Overall performance test speedup is x10.6 (from 159ms to 15,0ms) on my machine


jhasse:
ElideMiddle is only used in the unit tests? Can you move the declaration there then?

I'm also not sure if it's a good idea to move so much code around because it breaks the Git history. Especially for the unit tests it would be helpful to see, that your changes don't change any behavior just by looking if you touched any tests.

jhasse:
Forgot to say: Big thanks! I was thinking that std::regex would probably to slow and we should replace it before releasing 1.13.0.

digit-google:
I can keep the unit-test at the same location, but I would rather move the function to its own source file, given its complexity (especially after the refactor).

Another option is to make the unit-tests more readable first, using string macros to encode the ANSI sequences, so things look like:

```
#define MAGENTA "\x1B[0;35m"
#define NOTHING "\33[m"
#define RED "\x1b[1;31m"
#define RESET "\x1b[0m"

TEST(ElideMiddle, ElideAnsiEscapeCodes) {
  std::string input = "012345" MAGENTA "67890123456789";
  EXPECT_EQ("012..." MAGENTA "6789", ElideMiddle(input, 10));
  EXPECT_EQ("012345" MAGENTA "67...23456789", ElideMiddle(input, 19));

  EXPECT_EQ("Nothing " NOTHING " string.", ElideMiddle("Nothing "NOTHING" string.", 18));
  EXPECT_EQ("0" NOTHING "12...6789", ElideMiddle("0" NOTHING "1234567890123456789", 10));

  input = "abcd" RED "efg" RESET "hlkmnopqrstuvwxyz";
  EXPECT_EQ("", ElideMiddle(input, 0));
  EXPECT_EQ(".", ElideMiddle(input, 1));
  EXPECT_EQ("..", ElideMiddle(input, 2));
  EXPECT_EQ("...", ElideMiddle(input, 3));
  EXPECT_EQ("..." RED RESET "z", ElideMiddle(input, 4));
  EXPECT_EQ("a..." RED RESET "z", ElideMiddle(input, 5));
  EXPECT_EQ("a..." RED RESET "yz", ElideMiddle(input, 6));
  EXPECT_EQ("ab..." RED RESET "yz", ElideMiddle(input, 7));
  EXPECT_EQ("ab..." RED RESET "xyz", ElideMiddle(input, 8));
  EXPECT_EQ("abc..." RED RESET "xyz", ElideMiddle(input, 9));
  EXPECT_EQ("abc..." RED RESET "wxyz", ElideMiddle(input, 10));
  EXPECT_EQ("abcd" RED "..." RESET "wxyz", ElideMiddle(input, 11));
  EXPECT_EQ("abcd" RED "..." RESET "vwxyz", ElideMiddle(input, 12));

  EXPECT_EQ("abcd" RED "ef..." RESET "uvwxyz" ElideMiddle(input, 15));
  EXPECT_EQ("abcd" RED "ef..." RESET "tuvwxyz", ElideMiddle(input, 16));
  EXPECT_EQ("abcd" RED "efg" RESET "...tuvwxyz", ElideMiddle(input, 17));
  EXPECT_EQ("abcd" RED "efg" RESET "...stuvwxyz", ElideMiddle(input, 18));
  EXPECT_EQ("abcd" RED "efg" RESET "h...stuvwxyz", ElideMiddle(input, 19));

  input = "abcdef" RED "A" RESET "BC";
  EXPECT_EQ("..." RED RESET "C", ElideMiddle(input, 4));
  EXPECT_EQ("a..." RED RESET "C", ElideMiddle(input, 5));
  EXPECT_EQ("a..." RED RESET "BC", ElideMiddle(input, 6));
  EXPECT_EQ("ab..." RED RESET "BC", ElideMiddle(input, 7));
  EXPECT_EQ("ab..." RED "A" RESET "BC", ElideMiddle(input, 8));
  EXPECT_EQ("abcdef" RED "A" RESET "BC", ElideMiddle(input, 9));
}
```

Which is easier to read, then move to a different file. Wdyt?

jhasse:
I like it :) And since we now have to touch all lines anyway, moving it to its own file is fine.

jhasse:
Thanks!

digit-google:
Thanks!

PR #2490, opened 2024-09-04
Don't force coloured output

kalvdans:
emacs can't parse the filename and line number in compiler errors, if the output contains escape sequences.

Instead, let the C++ compiler itself decide coloring based on the TERM environment variable, or let the users set CMAKE_COLOR_DIAGNOSTICS environment variable before they run `cmake`.


mohkale:
FYI @kalvdans you can get Emacs to colorize compilation output.

```elisp
(use-package ansi-color
  :autoload ansi-color-compilation-filter
  :hook (compilation-filter . ansi-color-compilation-filter))
```

jhasse:
The TERM environment variable doesn't help because ninja pipes the compiler. Therefore we want to have color by default by setting that variable. You should be able to disable it for emacs by setting `CMAKE_COLOR_DIAGNOSTICS ` to 0. Or my recommendation: Stop using emacs ;)

PR #2494, opened 2024-09-06
fix typos

spaette:
opened as a Draft as `netherless` might be replaced with another string

```
$ sed -n '3264p' ninja/src/build_test.cc
  // Validate that out is rebuilt netherless, as deps are missing.
$ 
```

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -526,7 +526,7 @@ void Plan::ComputeCriticalPath() {
   for (Edge* edge : sorted_edges)
     edge->set_critical_path_weight(EdgeWeightHeuristic(edge));
 
-  // Second propagate / increment weidghts from
+  // Second propagate / increment weights from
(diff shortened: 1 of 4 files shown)


spaette:
@digit-google

> replaced with another string

Would that be `nonetheless` or `nevertheless`?

digit-google:
Thanks, this looks good to me. I think we should go with `nevertheless`, from definitions I am finding online. @jhasse any idea here?

spaette:
@digit-google 

used `nevertheless`
made a soft reset
workflow may need to be approved again

mcprat:
consider `regardless`

spaette:
> `regardless`

that's understandable, but I've already replaced the string with `nevertheless`

jhasse:
not a native speaker, but looks good to me. Thanks :)

PR #2496, opened 2024-09-08
Check progress status format string validity first and fall back to default if it's invalid

adamjoer:
In my opinion it shouldn't be fatal if the progress format string configured with `$NINJA_STATUS` is invalid. Especially now that we have placeholders that have only been available since Ninja 1.12.

This PR therefore changes the way the progress status format string is validated, so that it is checked in the `StatusPrinter` constructor, and if it's invalid the default `"[%f/%t] "` format is used instead. It also adds a small unit test to ensure this functionality.


mcprat:
I don't think it's necessary to make a new function for this... there should be a simple way to do it

adamjoer:
> I don't think it's necessary to make a new function for this... there should be a simple way to do it

I agree that it's a clunky way to do it.

Would it make more sense to add a boolean out parameter to `StatusPrinter::FormatProgressStatus` that indicates if the format is invalid? Then the check could look something like this:
```cpp
progress_status_format_ = getenv("NINJA_STATUS");
bool is_format_valid = true;
FormatProgressStatus(progress_status_format_, 0, &is_format_valid)
if (!progress_status_format_ || !is_format_valid)
  progress_status_format_ = "[%f/%t] ";
```

src/status.h
```diff
@@ -39,9 +39,9 @@ struct Status {
   /// (which is the default).
   virtual void SetExplanations(Explanations*) = 0;
 
-  virtual void Info(const char* msg, ...) = 0;
-  virtual void Warning(const char* msg, ...) = 0;
-  virtual void Error(const char* msg, ...) = 0;
+  virtual void Info(const char* msg, ...) const = 0;
```
digit-google, line 42: These are unrelated changes, they are useful but please move them to their own commit instead. Thanks.

src/status_printer.h
```diff
@@ -36,9 +36,11 @@ struct StatusPrinter : Status {
...
   void BuildFinished() override;
 
-  void Info(const char* msg, ...) override;
-  void Warning(const char* msg, ...) override;
-  void Error(const char* msg, ...) override;
+  void Info(const char* msg, ...) const override;
+  void Warning(const char* msg, ...) const override;
+  void Error(const char* msg, ...) const override;
+
+  bool IsValidProgressStatus(const char* progress_status_format) const;
```
digit-google, line 43: nit: Please make this method static.

digit-google: Also, document it properly and make it return an error string, instead of calling Warning(), i.e. let the caller decide what to do, e.g.:

```
/// Validate the status format string passed as an argument to verify it only contains
/// valid escape sequences. On success return true, on failure, return false and sets \arg error.
static bool IsValidProgressStatusFormat(const char* status_format, std::string* err);
```

src/build_test.cc
```diff
@@ -2287,6 +2287,12 @@ TEST_F(BuildTest, StatusFormatReplacePlaceholder) {
             status_.FormatProgressStatus("[%%/s%s/t%t/r%r/u%u/f%f]", 0));
 }
 
+TEST_F(BuildTest, StatusFormatValidator) {
+  EXPECT_TRUE(status_.IsValidProgressStatus("[%f/%t] "));
+  EXPECT_FALSE(status_.IsValidProgressStatus("[%f/%X] "));
+  EXPECT_EQ("", status_.FormatProgressStatus("[%f/%X] ", 0));
+}
+
```
digit-google, line 2295: It would make more sense to move this to a new `status_printer_test.cc` file.

mcprat:
I have not looked deeper into this but at first glance, the goal should be able to be achieved with less than 10 lines changed... I don't understand how we are passing 100 lines...

Find the Fatal() call, turn it into a Warning(), and then set the format value.

am I wrong?

mcprat:
@digit-google would it be acceptable to make `FormatProgressStatus()` non-const? that would make it take only 9 lines to do this.

src/status_printer.cc
```diff
@@ -49,8 +49,14 @@ StatusPrinter::StatusPrinter(const BuildConfig& config)
     printer_.set_smart_terminal(false);
 
   progress_status_format_ = getenv("NINJA_STATUS");
-  if (!progress_status_format_)
+
+  string error_output;
```
digit-google, line 53: nit: use `std::string` as we are trying to remove `using namespace std` from the sources gradually.

src/status_printer.cc
```diff
@@ -49,8 +49,14 @@ StatusPrinter::StatusPrinter(const BuildConfig& config)
...
+  if (!progress_status_format_ ||
+      !IsValidProgressStatus(progress_status_format_, &error_output)) {
+    if (!error_output.empty())
+      Warning("%s", error_output.c_str());
     progress_status_format_ = "[%f/%t] ";
+  }
```
digit-google, line 59: nit: For clarity use two conditions and a common variable such as:

```
std::string error;
if (progress_status_format_ && !IsValidProgressStatus(progress_status_format_, &error)) {
    Warning("%s", error.c_str());
    progress_status_format_ = nullptr;
}
if (!progress_status_format_) {
  // Enforce default value if unspecified or invalid.
  progress_status_format_ = kDefaultStatusFormat;
}
```

src/build_test.cc
```diff
@@ -2258,47 +2258,6 @@ TEST_F(BuildTest, DepsGccWithEmptyDepfileErrorsOut) {
   ASSERT_EQ(1u, command_runner_.commands_ran_.size());
 }
 
-TEST_F(BuildTest, StatusFormatElapsed_e) {
```
digit-google, line 2261: Good, but can you put this commit before the previous one, this will make the addition of `BuildTest.StatusFormatValidator` clearer.

digit-google:
@mcprat, to be honest, I don't think it would be a good idea here. A formatting function that changes the state of its object is always confusing for the reader.

Also a single validation function that is called once before anything heavy work is done is very simple to understand.
Trying to detect the invalid value on each status output, and them modifying the formatting string there is introducing un-needed mutations. And unexpected mutations are the source of ... lots of bugs.

src/status_printer.h
```diff
@@ -100,7 +100,10 @@ struct StatusPrinter : Status {
   Explanations* explanations_ = nullptr;
 
   /// The custom progress status format to use.
-  const char* progress_status_format_;
+  const char* progress_status_format_ = nullptr;
+
+  /// The progress status format to use by default
+  static constexpr const char* kDefaultProgressStatusFormat = "[%f/%t] ";
```
digit-google, line 106: nit: `static constexpr` members need a definition in the corresponding .cc file or some compilers will complain that it is missing (especially when compiling with `-O0`). In this case you can simply add `constexpr const char* StatusPrinter::kDefaultProgressStatusFormat;` (with no value) to `status_printer.cc`.

However, we do not need to expose the value publicly, so for simplicity, just define it as a const variable in the constructor that uses it.

Also `const char kName[] = "..."` is preferable to `static const char* kName = "..."` because the latter introduces an unnecessary relocation in the generated code / data segment.

```
StatusPrinter::StatusPrinter(...) {
    ...
   const char kDefaultProgressStatusFormat[] = "[%f/%t] ";
   ...
}
```

Sorry for the nitpicks, this is a great PR :-)

adamjoer: It's no problem at all, this is great feedback. :)

adamjoer:
@jhasse @digit-google Any chance that this will get merged?

digit-google:
The code LGTM as-is, though I would appreciate if you could address the latest review commits. I'll let @jhasse decide on the merge itself :)

src/status_printer.cc
```diff
@@ -42,15 +42,24 @@ Status* Status::factory(const BuildConfig& config) {
...
     : config_(config), started_edges_(0), finished_edges_(0), total_edges_(0),
-      running_edges_(0), progress_status_format_(NULL),
-      current_rate_(config.parallelism) {
+      running_edges_(0), current_rate_(config.parallelism) {
   // Don't do anything fancy in verbose mode.
   if (config_.verbosity != BuildConfig::NORMAL)
     printer_.set_smart_terminal(false);
 
+  // The progress status format to use by default
+  static const char kDefaultProgressStatusFormat[] = "[%f/%t] ";
```
jhasse, line 51: Why this variable at all?

src/status_printer.cc
```diff
@@ -257,6 +266,45 @@ void StatusPrinter::BuildFinished() {
...
+    case 'P':  // The percentage (in ppp% format) of time elapsed out of
+               // predicted total runtime.
+    case '%':  // A plain `%` character.
+      break;
+
+    default:
+      if (error_output) {
+        char buffer[64];
+        snprintf(buffer, sizeof(buffer),
+                 "unknown placeholder '%%%c' in $NINJA_STATUS", *s);
```
jhasse, line 299: What if `*s` is `'\0'`?

src/status_printer_test.cc
```diff
@@ -0,0 +1,70 @@
...
+
+#include "status_printer.h"
+
+#include "build.h"
+#include "test.h"
+
+struct StatusPrinterTest : public testing::Test {
+  StatusPrinterTest() : config_(MakeConfig()), status_(config_) {}
+
+  virtual void SetUp() { status_.BuildStarted(); }
```
jhasse, line 23: Just put this into the ctor body

jhasse:
closing due to inactivity

PR #2497, opened 2024-09-20
Introduce a new tool: `ninja -t compdb-targets`

### LINKED ISSUE #1544: [Feature Request] Allow 'ninja -t compdb' accept one target

JamesWidman:
This fixes #1544.

Here we use the same commit as in #2319, plus one more commit so that the option is spelled `--targets`.

Many thanks to @lk-chen for the original work in #1546, and to @csmoe for carrying it forward in #2319!

--- 
update 2024-10-04:

The most recent push implements the new tool, `ninja -t compdb-targets`.

todo:
- [x] add unit tests
- [x] update manual to mention the new tool (next to the entry for `compdb`)
- [x] review coding-style guidelines & fix any violations

### DIFF
--- a/doc/style.css
+++ b/doc/style.css
@@ -53,3 +53,16 @@ div.chapter {
 p {
     margin-top: 0;
 }
+
+/* The following applies to the left column of a [horizontal] labeled list: */
+table.horizontal > tbody > tr > td:nth-child(1) {
+
+    /* prevent the insertion of a line-break in the middle of a label: */
+    white-space: nowrap;
+
+    /* insert a little horizontal padding between the two columns: */
+    padding-right: 1.5em;
+
+    /* right-justify labels: */
+    text-align: end;
+}
--- a/misc/output_test.py
+++ b/misc/output_test.py
@@ -12,13 +12,48 @@ import sys
 import tempfile
 import unittest
 from textwrap import dedent
-from typing import Dict
+import typing as T
 
 default_env = dict(os.environ)
 default_env.pop('NINJA_STATUS', None)
 default_env.pop('CLICOLOR_FORCE', None)
 default_env['TERM'] = ''
 NINJA_PATH = os.path.abspath('./ninja')
+
+def remove_non_visible_lines(raw_output: bytes) -> str:
+  # When running in a smart terminal, Ninja uses CR (\r) to
+  # return the cursor to the start of the current line, prints
+  # something, then uses `\x1b[K` to clear everything until
+  # the end of the line.
+  #
+  # Thus printing 'FOO', 'BAR', 'ZOO' on the same line, then
+  # jumping to the next one results in the following output
+  # on Posix:
+  #
+  # '\rFOO\x1b[K\rBAR\x1b[K\rZOO\x1b[K\r\n'
+  #
+  # The following splits the output at both \r, \n and \r\n
+  # boundaries, which gives:
+  #
+  #  [ '\r', 'FOO\x1b[K\r', 'BAR\x1b[K\r', 'ZOO\x1b[K\r\n' ]
+  #
+  decoded_lines = raw_output.decode('utf-8').splitlines(True)
+
+  # Remove any item that ends with a '\r' as this means its
+  # content will be overwritten by the next item in the list.
+  # For the previous example, this gives:
+  #
+  #  [ 'ZOO\x1b[K\r\n' ]
+  #
+  final_lines = [ l for l in decoded_lines if not l.endswith('\r') ]
+
+  # Return a single string that concatenates all filtered lines
+  # while removing any remaining \r in it. Needed to transform
+  # \r\n into \n.
+  #
+  #  "ZOO\x1b[K\n'
+  #
+  return ''.join(final_lines).replace('\r', '')
 
 class BuildDir:
     def __init__(self, build_ninja: str):
@@ -35,12 +70,18 @@ class BuildDir:
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.d.cleanup()
 
+    @property
+    def path(self) -> str:
+        return os.path.realpath(self.d.name)
+
+
     def run(
         self,
-        flags: str = '',
+        flags: T.Optional[str] = None,
         pipe: bool = False,
         raw_output: bool = False,
-        env: Dict[str, str] = default_env,
+        env: T.Dict[str, str] = default_env,
+        print_err_output = True,
     ) -> str:
         """Run Ninja command, and get filtered output.
 
@@ -56,13 +97,17 @@ class BuildDir:
 
           env: Optional environment dictionary to run the command in.
 
+          print_err_output: set to False if the test expects ninja to print
+            something to stderr. (Otherwise, an error message from Ninja
+            probably represents a failed test.)
+
         Returns:
           A UTF-8 string corresponding to the output (stdout only) of the
           Ninja command. By default, partial lines that were overwritten
           are removed according to the rules described in the comments
           below.
         """
-        ninja_cmd = '{} {}'.format(NINJA_PATH, flags)
+        ninja_cmd = '{} {}'.format(NINJA_PATH, flags if flags else '')
         try:
             if pipe:
                 output = subprocess.check_output(
@@ -74,57 +119,27 @@ class BuildDir:
                 output = subprocess.check_output(['script', '-qfec', ninja_cmd, '/dev/null'],
                                                  cwd=self.d.name, env=env)
         except subprocess.CalledProcessError as err:
-            sys.stdout.buffer.write(err.output)
+            if print_err_output:
+              sys.stdout.buffer.write(err.output)
+            err.cooked_output = remove_non_visible_lines(err.output)
             raise err
 
         if raw_output:
             return output.decode('utf-8')
-
-        # When running in a smart terminal, Ninja uses CR (\r) to
-        # return the cursor to the start of the current line, prints
-        # something, then uses `\x1b[K` to clear everything until
-        # the end of the line.
-        #
-        # Thus printing 'FOO', 'BAR', 'ZOO' on the same line, then
-        # jumping to the next one results in the following output
-        # on Posix:
-        #
-        # '\rFOO\x1b[K\rBAR\x1b[K\rZOO\x1b[K\r\n'
-        #
-        # The following splits the output at both \r, \n and \r\n
-        # boundaries, which gives:
-        #
-        #  [ '\r', 'FOO\x1b[K\r', 'BAR\x1b[K\r', 'ZOO\x1b[K\r\n' ]
-        #
-        decoded_lines = output.decode('utf-8').splitlines(True)
-
-        # Remove any item that ends with a '\r' as this means its
-        # content will be overwritten by the next item in the list.
-        # For the previous example, this gives:
-        #
-        #  [ 'ZOO\x1b[K\r\n' ]
-        #
-        final_lines = [ l for l in decoded_lines if not l.endswith('\r') ]
-
-        # Return a single string that concatenates all filtered lines
-        # while removing any remaining \r in it. Needed to transform
-        # \r\n into \n.
-        #
-        #  "ZOO\x1b[K\n'
-        #
-        return ''.join(final_lines).replace('\r', '')
+        return remove_non_visible_lines(output)
 
 def run(
     build_ninja: str,
-    flags: str = '',
+    flags: T.Optional[str] = None,
     pipe: bool = False,
     raw_output: bool = False,
-    env: Dict[str, str] = default_env,
+    env: T.Dict[str, str] = default_env,
+    print_err_output = True,
 ) -> str:
     """Run Ninja with a given build plan in a temporary directory.
     """
     with BuildDir(build_ninja) as b:
-        return b.run(flags, pipe, raw_output, env)
+        return b.run(flags, pipe, raw_output, env, print_err_output)
 
 @unittest.skipIf(platform.system() == 'Windows', 'These test methods do not work on Windows')
 class Output(unittest.TestCase):
@@ -136,6 +151,16 @@ class Output(unittest.TestCase):
         'build a: echo',
         '',
     ))
+
+    def _test_expected_error(self, plan: str, flags: T.Optional[str], expected: str):
+        """Run Ninja with a given plan and flags, and verify its cooked output against an expected content.
+        """
+        actual = ''
+        try:
+          actual = run(plan, flags, print_err_output=False)
+        except subprocess.CalledProcessError as err:
+          actual = err.cooked_output
+        self.assertEqual(expected, actual)
 
     def test_issue_1418(self) -> None:
         self.assertEqual(run(
@@ -369,6 +394,59 @@ out 2
           ),
           f'''in1\0out1\0out 2\0'''
         )
+
+
+    def test_tool_compdb_targets(self) -> None:
+        plan = '''
+rule cat
+  command = cat $in $out
+build out1 : cat in1
+build out2 : cat in2 out1
+build out3 : cat out2 out1
+build out4 : cat in4
+'''
+
+
+        self._test_expected_error(plan, '-t compdb-targets',
+'''ninja: error: compdb-targets expects the name of at least one target
+usage: ninja -t compdb [-hx] target [targets]
+
+options:
+  -h     display this help message
+  -x     expand @rspfile style response file invocations
+''')
+
+        self._test_expected_error(plan, '-t compdb-targets in1',
+            "ninja: fatal: 'in1' is not a target (i.e. it is not an output of any `build` statement)\n")
+
+        self._test_expected_error(plan, '-t compdb-targets nonexistent_target',
+            "ninja: fatal: unknown target 'nonexistent_target'\n")
+
+
+        with BuildDir(plan) as b:
+            actual = b.run(flags='-t compdb-targets out3')
+            expected = f'''[
+  {{
+    "directory": "{b.path}",
+    "command": "cat in1 out1",
+    "file": "in1",
+    "output": "out1"
+  }},
+  {{
+    "directory": "{b.path}",
+    "command": "cat in2 out1 out2",
+    "file": "in2",
+    "output": "out2"
+  }},
+  {{
+    "directory": "{b.path}",
+    "command": "cat out2 out1 out3",
+    "file": "out2",
+    "output": "out3"
+  }}
+]
+'''
+            self.assertEqual(expected, actual)
 
 
     def test_explain_output(self):
--- a/src/graph_test.cc
+++ b/src/graph_test.cc
@@ -15,6 +15,7 @@
 #include "graph.h"
 
 #include "build.h"
+#include "command_collector.h"
 #include "test.h"
 
 using namespace std;
@@ -308,6 +309,54 @@ TEST_F(GraphTest, InputsCollectorWithEscapes) {
 #endif
   EXPECT_EQ("implicit", inputs[3]);
   EXPECT_EQ("order_only", inputs[4]);
+}
+
+TEST_F(GraphTest, CommandCollector) {
+  ASSERT_NO_FATAL_FAILURE(AssertParse(&state_,
+                                      "build out1: cat in1\n"
+                                      "build mid1: cat in1\n"
+                                      "build out2: cat mid1\n"
+                                      "build out3 out4: cat mid1\n"
+                                      "build all: phony out1 out2 out3\n"));
+  {
+    CommandCollector collector;
+    auto& edges = collector.in_edges;
+
+    // Start visit from out2; this should add `build mid1` and `build out2` to
+    // the edge list.
+    collector.CollectFrom(GetNode("out2"));
+    ASSERT_EQ(2u, edges.size());
+    EXPECT_EQ("cat in1 > mid1", edges[0]->EvaluateCommand());
+    EXPECT_EQ("cat mid1 > out2", edges[1]->EvaluateCommand());
+
+    // Add a visit from out1, this should append `build out1`
+    collector.CollectFrom(GetNode("out1"));
+    ASSERT_EQ(3u, edges.size());
+    EXPECT_EQ("cat in1 > out1", edges[2]->EvaluateCommand());
+
+    // Another visit from all; this should add edges for out1, out2 and out3,
+    // but not all (because it's phony).
+    collector.CollectFrom(GetNode("all"));
+    ASSERT_EQ(4u, edges.size());
+    EXPECT_EQ("cat in1 > mid1", edges[0]->EvaluateCommand());
+    EXPECT_EQ("cat mid1 > out2", edges[1]->EvaluateCommand());
+    EXPECT_EQ("cat in1 > out1", edges[2]->EvaluateCommand());
+    EXPECT_EQ("cat mid1 > out3 out4", edges[3]->EvaluateCommand());
+  }
+
+  {
+    CommandCollector collector;
+    auto& edges = collector.in_edges;
+
+    // Starting directly from all, will add `build out1` before `build mid1`
+    // compared to the previous example above.
+    collector.CollectFrom(GetNode("all"));
+    ASSERT_EQ(4u, edges.size());
+    EXPECT_EQ("cat in1 > out1", edges[0]->EvaluateCommand());
+    EXPECT_EQ("cat in1 > mid1", edges[1]->EvaluateCommand());
+    EXPECT_EQ("cat mid1 > out2", edges[2]->EvaluateCommand());
+    EXPECT_EQ("cat mid1 > out3 out4", edges[3]->EvaluateCommand());
+  }
 }
 
 TEST_F(GraphTest, VarInOutPathEscaping) {
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -20,6 +20,8 @@
 
 #include <algorithm>
 #include <cstdlib>
+#include <cstring>
+#include <string>
 
 #ifdef _WIN32
 #include "getopt.h"
@@ -38,6 +40,7 @@
 #include "build_log.h"
 #include "deps_log.h"
 #include "clean.h"
+#include "command_collector.h"
 #include "debug_flags.h"
 #include "depfile_parser.h"
 #include "disk_interface.h"
@@ -130,6 +133,8 @@ struct NinjaMain : public BuildLogUser {
   int ToolClean(const Options* options, int argc, char* argv[]);
   int ToolCleanDead(const Options* options, int argc, char* argv[]);
   int ToolCompilationDatabase(const Options* options, int argc, char* argv[]);
+  int ToolCompilationDatabaseForTargets(const Options* options, int argc,
+                                        char* argv[]);
   int ToolRecompact(const Options* options, int argc, char* argv[]);
   int ToolRestat(const Options* options, int argc, char* argv[]);
   int ToolUrtle(const Options* options, int argc, char** argv);
@@ -932,8 +937,8 @@ std::string EvaluateCommandWithRspfile(const Edge* edge,
   return command;
 }
 
-void printCompdb(const char* const directory, const Edge* const edge,
-                 const EvaluateCommandMode eval_mode) {
+void PrintOneCompdbObject(std::string const& directory, const Edge* const edge,
+                          const EvaluateCommandMode eval_mode) {
   printf("\n  {\n    \"directory\": \"");
   PrintJSONString(directory);
   printf("\",\n    \"command\": \"");
@@ -977,37 +982,25 @@ int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
   argc -= optind;
 
   bool first = true;
-  vector<char> cwd;
-  char* success = NULL;
 
-  do {
-    cwd.resize(cwd.size() + 1024);
-    errno = 0;
-    success = getcwd(&cwd[0], cwd.size());
-  } while (!success && errno == ERANGE);
-  if (!success) {
-    Error("cannot determine working directory: %s", strerror(errno));
-    return 1;
-  }
-
+  std::string directory = GetWorkingDirectory();
   putchar('[');
-  for (vector<Edge*>::iterator e = state_.edges_.begin();
-       e != state_.edges_.end(); ++e) {
-    if ((*e)->inputs_.empty())
+  for (const Edge* edge : state_.edges_) {
+    if (edge->inputs_.empty())
       continue;
     if (argc == 0) {
       if (!first) {
         putchar(',');
       }
-      printCompdb(&cwd[0], *e, eval_mode);
+      PrintOneCompdbObject(directory, edge, eval_mode);
       first = false;
     } else {
       for (int i = 0; i != argc; ++i) {
-        if ((*e)->rule_->name() == argv[i]) {
+        if (edge->rule_->name() == argv[i]) {
           if (!first) {
             putchar(',');
           }
-          printCompdb(&cwd[0], *e, eval_mode);
+          PrintOneCompdbObject(directory, edge, eval_mode);
           first = false;
         }
       }
@@ -1087,6 +1080,118 @@ int NinjaMain::ToolRestat(const Options* options, int argc, char* argv[]) {
   return EXIT_SUCCESS;
 }
 
+struct CompdbTargets {
+  enum class Action { kDisplayHelpAndExit, kEmitCommands };
+
+  Action action;
+  EvaluateCommandMode eval_mode = ECM_NORMAL;
+
+  std::vector<std::string> targets;
+
+  static CompdbTargets CreateFromArgs(int argc, char* argv[]) {
+    //
+    // grammar:
+    //     ninja -t compdb-targets [-hx] target [targets]
+    //
+    CompdbTargets ret;
+
+    // getopt_long() expects argv[0] to contain the name of
+    // the tool, i.e. "compdb-targets".
+    argc++;
+    argv--;
+
+    // Phase 1: parse options:
+    optind = 1;  // see `man 3 getopt` for documentation on optind
+    int opt;
+    while ((opt = getopt(argc, argv, const_cast<char*>("hx"))) != -1) {
+      switch (opt) {
+      case 'x':
+        ret.eval_mode = ECM_EXPAND_RSPFILE;
+        break;
+      case 'h':
+      default:
+        ret.action = CompdbTargets::Action::kDisplayHelpAndExit;
+        return ret;
+      }
+    }
+
+    // Phase 2: parse operands:
+    int const targets_begin = optind;
+    int const targets_end = argc;
+
+    if (targets_begin == targets_end) {
+      Error("compdb-targets expects the name of at least one target");
+      ret.action = CompdbTargets::Action::kDisplayHelpAndExit;
+    } else {
+      ret.action = CompdbTargets::Action::kEmitCommands;
+      for (int i = targets_begin; i < targets_end; ++i) {
+        ret.targets.push_back(argv[i]);
+      }
+    }
+
+    return ret;
+  }
+};
+
+void PrintCompdb(std::string const& directory, std::vector<Edge*> const& edges,
+                 const EvaluateCommandMode eval_mode) {
+  putchar('[');
+
+  bool first = true;
+  for (const Edge* edge : edges) {
+    if (edge->is_phony() || edge->inputs_.empty())
+      continue;
+    if (!first)
+      putchar(',');
+    PrintOneCompdbObject(directory, edge, eval_mode);
+    first = false;
+  }
+
+  puts("\n]");
+}
+
+int NinjaMain::ToolCompilationDatabaseForTargets(const Options* options,
+                                                 int argc, char* argv[]) {
+  auto compdb = CompdbTargets::CreateFromArgs(argc, argv);
+
+  switch (compdb.action) {
+  case CompdbTargets::Action::kDisplayHelpAndExit: {
+    printf(
+        "usage: ninja -t compdb [-hx] target [targets]\n"
+        "\n"
+        "options:\n"
+        "  -h     display this help message\n"
+        "  -x     expand @rspfile style response file invocations\n");
+    return 1;
+  }
+
+  case CompdbTargets::Action::kEmitCommands: {
+    CommandCollector collector;
+
+    for (const std::string& target_arg : compdb.targets) {
+      std::string err;
+      Node* node = CollectTarget(target_arg.c_str(), &err);
+      if (!node) {
+        Fatal("%s", err.c_str());
+        return 1;
+      }
+      if (!node->in_edge()) {
+        Fatal(
+            "'%s' is not a target "
+            "(i.e. it is not an output of any `build` statement)",
+            node->path().c_str());
+      }
+      collector.CollectFrom(node);
+    }
+
+    std::string directory = GetWorkingDirectory();
+    PrintCompdb(directory, collector.in_edges, compdb.eval_mode);
+  } break;
+  }
+
+  return 0;
+}
+
 int NinjaMain::ToolUrtle(const Options* options, int argc, char** argv) {
   // RLE encoded.
   const char* urtle =
@@ -1141,6 +1246,9 @@ const Tool* ChooseTool(const string& tool_name) {
       Tool::RUN_AFTER_LOAD, &NinjaMain::ToolTargets },
     { "compdb",  "dump JSON compilation database to stdout",
       Tool::RUN_AFTER_LOAD, &NinjaMain::ToolCompilationDatabase },
+    { "compdb-targets",
+      "dump JSON compilation database for a given list of targets to stdout",
+      Tool::RUN_AFTER_LOAD, &NinjaMain::ToolCompilationDatabaseForTargets },
     { "recompact",  "recompacts ninja-internal data structures",
       Tool::RUN_AFTER_LOAD, &NinjaMain::ToolRecompact },
     { "restat",  "restats all outputs in the build log",
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -289,6 +289,11 @@ http://clang.llvm.org/docs/JSONCompilationDatabase.html[JSON format] expected
 by the Clang tooling interface.
 _Available since Ninja 1.2._
 
+`compdb-targets`:: like `compdb`, but takes a list of targets instead of rules,
+and expects at least one target. The resulting compilation database contains
+all commands required to build the indicated targets, and _only_ those
+commands.
+
 `deps`:: show all dependencies stored in the `.ninja_deps` file. When given a
 target, show just the target's dependencies. _Available since Ninja 1.4._
 
--- a/src/command_collector.h
+++ b/src/command_collector.h
@@ -1,0 +1,65 @@
+// Copyright 2024 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef NINJA_COMMAND_COLLECTOR_H_
+#define NINJA_COMMAND_COLLECTOR_H_
+
+#include <cassert>
+#include <unordered_set>
+#include <vector>
+
+#include "graph.h"
+
+/// Collects the transitive set of edges that lead into a given set
+/// of starting nodes. Used to implement the `compdb-targets` tool.
+///
+/// When collecting inputs, the outputs of phony edges are always ignored
+/// from the result, but are followed by the dependency walk.
+///
+/// Usage is:
+/// - Create instance.
+/// - Call CollectFrom() for each root node to collect edges from.
+/// - Call TakeResult() to retrieve the list of edges.
+///
+struct CommandCollector {
+  void CollectFrom(const Node* node) {
+    assert(node);
+
+    if (!visited_nodes_.insert(node).second)
+      return;
+
+    Edge* edge = node->in_edge();
+    if (!edge || !visited_edges_.insert(edge).second)
+      return;
+
+    for (Node* input_node : edge->inputs_)
+      CollectFrom(input_node);
+
+    if (!edge->is_phony())
+      in_edges.push_back(edge);
+  }
+
+ private:
+  std::unordered_set<const Node*> visited_nodes_;
+  std::unordered_set<Edge*> visited_edges_;
+
+  /// we use a vector to preserve order from requisites to their dependents.
+  /// This may help LSP server performance in languages that support modules,
+  /// but it also ensures that the output of `-t compdb-targets foo` is
+  /// consistent, which is useful in regression tests.
+ public:
+  std::vector<Edge*> in_edges;
+};
+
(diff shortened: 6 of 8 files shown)


mcprat:
squash to 1 commit to get rid of the merge commits and combine the original with your rename, add yourself to the tags at the end of the original commit

it would be nice if we can find a way to get rid of `stringstream` (for lower binary size?)

I still think the remaining arguments should not require a long option at all or otherwise `--` if its not possible to avoid long opts. Isn't it safe to assume everything after `-t compdb` will be a target? if not then whatever else can be after that should be the type of argument with it's own unique long option

JamesWidman:
@mcprat Agreed on squashing/rebasing, and i will look into the `stringstream` thing.

as for the command-line syntax: i wrote about this in #2319 (though i forgot to tag you there; sorry!), but it would be good to have it inline here too:

You wrote:
> Would it be better to shift out arguments so that no flag is required for the targets list?
> 
> so instead of
> 
> `-t compdb --target foo,bar`
> 
> you would have
> 
> `-t compdb foo bar`

Prior to the commits in this PR, the output of `ninja -t compdb -h` is:
```
usage: ninja -t compdb [options] [rules]

options:
```

so `ninja -t compdb foo bar` already means "emit every compile command that uses `rule foo` or `rule bar`", and we probably don't want to break that.

> or if a flag is really necessary to break out of flag-only parsing, use syntax like in `git`
> 
> `-t compdb -- foo bar`
> 
> so they can be space separated instead of comma separated

but that means the syntax would become:
```
ninja -t compdb [options] [rules] -- [targets]
```
...which means that users will need to remember that `-t compdb foo -- bar` means "*rule* foo and *target* bar", as opposed to "*target* foo and *rule* bar". 

How would you feel about a short option instead of a long option? (meaning: `-t TARGETS` instead of `--targets TARGETS`)

Also: if we go with an option, are you ok with making the target names comma-delimited? (i think the main reason to use commas is so that we stick to the format expected by `getopt()/getopt_long()`, where the set of all targets would be returned via `optarg`.)

Fwiw, in bash 4.2 and later, you can join an array with commas using the printf builtin like so:
```
echo $(printf "%s," ${targets[*]})
```

JamesWidman:
@mcprat oh wait, when you wrote:

> if not then whatever else can be after that should be the type of argument with it's own unique long option

are you proposing something like:
```
ninja -t compdb --target t1 --target t2 --target tN rule1 rule2 ruleN
```

JamesWidman:
@mcprat btw, i assumed that we don't want to break the pre-existing `[rules]` interface because i assume that someone, somewhere, depends on that in a script or something, but personally i can't see much of a use for it if we can specify targets.

Do we want to make the breaking change of removing the `[rules]` syntax (or replacing it with an option that allows something like `-r rule1 -r rule2 [...]`)?

JamesWidman:
@mcprat i just thought of another way we could do this:

up until now, i've been stuck thinking that `--targets` is an option that must have an argument, but we could alternatively tell `getopt()/getopt_long()` that `--targets` (or `-t`) *doesn't* take an argument.  Then we could change the syntactic form of the command, in a non-breaking way, to this:
```
ninja -t compdb [options] [items]
```
...where the `[items]` are interpreted as rules unless one of the `[options]` is `-t/--targets`, in which case the `[items]` are interpreted as targets. So in effect, we would have two alternative forms:
```
ninja -t compdb --targets [other_options]  target1 target2 [...] targetN
ninja -t compdb [other_options] rule1 rule2 [...] ruleN
```
...which, although it would add an option, it would at least mean that targets are separated by spaces as you requested.

(In this case, i would have a weak preference for the long form (`--targets` rather than `-t`)), just because it follows another `-t` option that means something completely different (the "tool" option). It's not a deal-breaker for me, though.)

mcprat:
@JamesWidman I'm aware of your replies in the other PR. I was reiterating, but also clarifying that neither a short or long option would be the ideal (closer to POSIX-like behavior), because it seems like what I said previously was interpreted as only preferring short option or `--` over long option.

All of this definitely clears things up, it is more complicated than I thought...

Here's an idea... what's stopping the parsing stage from checking both the list of targets and list of rules for each argument? Without having looked into it, there's probably different functions used depending on whether the argument is a rule or a target. Maybe each argument can be passed to both, a failure can be caught, and if only one has a problem that's expected, but both failing (or even if neither are failing) would indicate an actual problem.

mcprat:
or rather, since this is a printing-only tool, it can just ignore any errors...

JamesWidman:
> I was reiterating, but also clarifying that neither a short or long option would be the ideal (closer to POSIX-like behavior),
 
I am trying to understand: 

1. why is it desirable to avoid the introduction of a new option (like `--targets` or `-t`)?
2. how does POSIX enter into it?

mcprat:
> 1. why is it desirable to avoid the introduction of a new option (like `--targets` or `-t`)?

it's not really an option but rather more input, as it doesn't adjust existing outputs but rather just adds to it. consider the way the manual is written right now 

> given a list of rules, each of which is expected to be a C family language compiler rule whose first input is the name of the source file, prints on standard output a compilation database...

instead of now having to describe that there's different ways that one adds a rule to the list vs how one adds a target to the list, it would be nice if the manual can just say:

> given a list of rules and/or targets...


> 2. how does POSIX enter into it?

I consider the design of `make` as a general guide. even though most consider it a product of GNU, it is in fact defined by POSIX as well, and they try to mostly stay conformant. even though Ninja may not explicitly claim to be POSIX conforming we ought to follow it when possible.

POSIX has a section called "Utility Syntax Guidelines" that's used to form the standard syntax for `make`. there is mention there about the usage of `--` and how the order of arguments can affect the output, and that there should be only one argument to each option. I don't see much about how the definition of an option differs from the definition of a normal operand however...

JamesWidman:
@mcprat First: i think my confusion might be shrinking; thank you!

i didn't quite understand what you meant here:
> > 1. why is it desirable to avoid the introduction of a new option (like `--targets` or `-t`)?
> 
> it's not really an option but rather more input, as it doesn't adjust existing outputs but rather just adds to it. 

...but i think i understand this part:
> consider the way the manual is written right now
> 
> > given a list of rules, each of which is expected to be a C family language compiler rule whose first input is the name of the source file, prints on standard output a compilation database...
> 
> instead of now having to describe that there's different ways that one adds a rule to the list vs how one adds a target to the list, it would be nice if the manual can just say:
> 
> > given a list of rules and/or targets...

so... if i understand you correctly, it sounds like you're saying something like: 
"The user interface for producing a compilation database for a given set of ninja *rules* ought to be consistent with the user interface for producing a compilation database for a given set of ninja *targets*."

Is that about right?

(If so: i see that the consistency that you're asking for could make it easier for a user to transition from using rules to using targets or vice-versa, and that would end the confusion that prompted my first question above.)

Second: 
> POSIX has a section called "Utility Syntax Guidelines"

i started reading [Issue 8](https://pubs.opengroup.org/onlinepubs/9799919799/) (IEEE Std 1003.1™-2024 Edition) as well as [a draft from 1991](http://mirror.math.princeton.edu/pub/oldlinux/Linux.old/Ref-docs/POSIX/all.pdf), and i just realized that you might be targeting some other version in between (one that is old enough that it is implemented by all the platforms the ninja project wants to run on, but new enough that it specifies all the OS features that the project wants ninja to depend on).

i checked `CONTRIBUTING.md`, the wiki, and the ninja manual, but i didn't see any mention of POSIX. 

so: what version of POSIX should we be reading/targeting? 

Also: would it make sense to add a mention of that version of POSIX to the wiki and/or `CONTRIBUTING.md` to mention that version?  (I can see how POSIX compliance is another way to support consistency of user interfaces (in this case, consistency between ninja and POSIX ), and also helps us avoid breaking the rule-of-least-surprise. Thank you for the pointer!)

src/ninja.cc
```diff
@@ -945,6 +948,32 @@ void printCompdb(const char* const directory, const Edge* const edge,
   printf("\"\n  }");
 }
 
+struct DependentEdgeCollector {
```
digit-google, line 951: Please move this class definition to `src/graph.h` and provide proper documentation + unit-tests for it. See `InputsCollector` there for inspiration.

I also recommend to make `depend_edges` a member of the class, and add a method like `TakeResult()` to move the final result out of the instance one all root targets have been visited.

src/ninja.cc
```diff
@@ -956,19 +985,51 @@ int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
...
         eval_mode = ECM_EXPAND_RSPFILE;
         break;
 
+      case 't': {
+        std::istringstream iss(optarg);
+        std::string target;
+        std::vector<Edge*> user_interested_edges;
+        DependentEdgeCollector deps_edge_collector;
+
+        while (std::getline(iss, target, ',')) {
```
digit-google, line 1006: Use `SplitStringPiece()` from `string_piece_util.h` instead for simpler and smaller generated code. Use `std::` prefixes for all standard types, as we are trying to move away from `using namespace std`. E.g.:

  ```
  for (StringPiece target_piece : SplitStringPiece(target, ',')) {
    std::string err;
    std::string target = target_piece.AsString();
    Node* node = CollectTarget(target.c_str(), &err);
    ...
  }

src/ninja.cc
```diff
@@ -945,6 +948,32 @@ void printCompdb(const char* const directory, const Edge* const edge,
   printf("\"\n  }");
 }
 
+struct DependentEdgeCollector {
+  bool CollectFrom(Node* node, std::vector<Edge*>* depend_edges) {
```
digit-google, line 952: nit: use `const Node*` for the first argument :-)

digit-google:
We do really not care about following POSIX guidelines or imitating the GNU Make design.
On the other hand we do care about developer workflows _not_ _breaking_ when introducing changes.

As such, changing how non-option arguments to `-t compdb` are interpreted is a strict no-go.

It is ok to introduce new options, even if they change the tool's behavior, as long as everything is properly documented, and there is no place for ambiguity. If this is not possible, or makes the UI too cluncky, introducing a new tool might be the better option.

JamesWidman:
> We do really not care about following POSIX guidelines or imitating the GNU Make design. 

@digit-google Works for me, but is there consensus among maintainers on this point?  (I got the sense from @mcprat that some POSIX-ness and Make-ness was desired. I don't feel a strong preference either way, but it would help to have unambiguous direction.)

> On the other hand we do care about developer workflows _not_ _breaking_ when introducing changes.

[thumbs up]

> As such, changing how non-option arguments to `-t compdb` are interpreted is a strict no-go.

So for example (and please correct me if i'm wrong):
1. in `ninja -t compdb item1 item2`, we will continue to interpret the two items as rules *only* (and not as targets);
2. likewise,  we will NOT introduce a form like `ninja -t compdb --targets  item1 item2` where the two items are interpreted as targets

> It is ok to introduce new options, even if they change the tool's behavior, as long as everything is properly documented, and there is no place for ambiguity.

Sounds good to me!

>  If this is not possible, or makes the UI too cluncky, introducing a new tool might be the better option.

... and by "tool" you mean a new sub-command, right? For example, maybe like: 

```
ninja -t compdb-targets [options] [targets...]
```

JamesWidman:
@digit-google btw, what's our C++ version for compiling ninja? (i grepped for `CXX_STANDARD` in the cmake files but found nothing.) 

i see uses of `auto`, so that means we require at least C++11.

i ask because i wanted to use a [structured binding](https://en.cppreference.com/w/cpp/language/structured_binding) in a place where i thought it would help to add clarity, but that's a C++17 feature, which would require setting  [`CXX_STANDARD`](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html#prop_tgt:CXX_STANDARD) and [`CXX_STANDARD_REQUIRED`](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD_REQUIRED.html#prop_tgt:CXX_STANDARD_REQUIRED).

Note, according to the clang website:
> By default, Clang 16 or later builds C++ code according to the C++17 standard.

...and likewise, GCC 11 defaults to C++17.

So if we're *not* going to depend on C++14 or C++17 features, we may want to set `CXX_STANDARD` to 11 so that we get the desired compile errors in the event where people unwittingly uses more recent features.

digit-google:
IIRC, the goal is to have the core Ninja sources building with C++11 support only, in order to allow building the tool on older systems (in particular with `./configure.py --bootstrap` rather than CMake).

However, the tests now rely on GTest which mandates C++14 though (and soon C++17), which is why CXX_STANDARD is not applied to them in our `CMakeList.txt` file.

I think these requirements were in the top-level `CONTRIBUTING.md` file, but no longer are there.

I wouldn't mind switching to C++17 personally, but @jhasse is the authority on this specific issue (and this may require updating some of our CI recipes).

JamesWidman:
@digit-google @jhasse this is a little off-topic for this PR, but: what's our policy on GNU core-language  extensions)?    

some pros/cons (in case this hasn't come up before): 

pros: 
- in my personal projects i've found [statement-expressions](https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html) and [case ranges](https://gcc.gnu.org/onlinedocs/gcc/Case-Ranges.html) to be very handy. 
- clang supports these extensions even in MSVC-emulation mode, and [according to LLVM maintainers](https://hachyderm.io/@chandlerc/112692810247295476), this is by design. 

con:
- Allowing GNU extensions would rule out the use of MSVC to compile ninja (but i don't know whether that would be a deal-breaker for the ninja project).

digit-google:
I believe right now we absolutely want to build with the MSVC toolchain, and support non-GNU and non-Clang toolchains, so these are a deal breakers.

JamesWidman:
@digit-google 
> In this specific case, using `--targets` as a separator between rules and targets seems ok, and no one likes commas on the command-line. What I mean is something like `-t compdb [-x] [rule]... --targets [target]...`, but I am not sure that getopt supports this here.

so it turns out that this is doable, but it requires using `getopt_long()` in a way that is different from the normal usage pattern for `getopt()/getopt_long()`. To simplify things for `NinjaMain::ToolCompilationDatabase()`,  it seems best to move argument parsing for `compdb` into a separate routine that (1) parses all arguments and (2) returns a structure that's easier to act on.

at the moment i'm working on regression tests and documentation, but here's a preview of the argument parsing routine (for early feedback on the general shape of this approach):
 
(Before my next push, i will review the coding style guidelines, so hopefully that push will not contain e.g. any naming style violations that might appear below.)

```C++
Compdb parseCompdbArguments(int argc, char* argv[]) {
  Compdb ret;
  //
  // grammar:
  //     ninja -t compdb [-hx] [rules] [--targets targets]
  //
  // in phase 1 of argument parsing, we process [-hx] and "--targets". The
  // pseudo-option "--targets" is conceptually more like "--" than an option.
  // Traditionally, "--" means "every argument that follows is an operand [i.e.,
  // a non-option argument]". Here, "--targets" means "every argument that
  // follows is grammatically an operand, and semantically a target. Also, rule
  // operands may precede this argument."
  //
  // If getopt_long returns an instance of "--targets", we store its `optind`
  // (see `man getopt()`) as the starting index for the list of targets, and we
  // store the previous `optind` as the starting index for the list of rules.
  //
  // in phase 2 of argument parsing, we process the list of rules (if any) and
  // the list of targets (if any).

  /* [...] */

  return ret;
}
```
...where `Compdb` is:
```C++
struct Compdb {
  enum class Action {
    display_help_and_exit,
    emit_all_commands,
    emit_userSpecified_commands
  };

  Action action;
  EvaluateCommandMode eval_mode = ECM_NORMAL;

  typedef std::vector<StringPiece> ArgList;
  ArgList targets;
  ArgList rules;
};
```
...and the usage looks like this:
```C++
int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
                                       char* argv[]) {
  Compdb compdb = parseCompdbArguments(argc, argv);

  switch (compdb.action) {
  /* [...] */
  }

  return 0;
}

```

Feedback welcome!

src/ninja.cc
JamesWidman: > Please move this class definition to `src/graph.h`

is it ok if i move it to a new header file instead?  It's very `compdb`-specific, and only used by `ninja.cc` and `graph_test.cc`, so it would be nice if changing it didn't cause a lot of unrelated source files to be recompiled

mcprat:
> > POSIX has a section called "Utility Syntax Guidelines"
> 
> i started reading [Issue 8](https://pubs.opengroup.org/onlinepubs/9799919799/) (IEEE Std 1003.1™-2024 Edition)...

I did originally read the section from Issue 7, but I don't see anything different from Issue 8
It's base definitions section 12.2
https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap12.html



> > We do really not care about following POSIX guidelines or imitating the GNU Make design.
> 
> Works for me, but is there consensus among maintainers on this point? (I got the sense from @mcprat that some POSIX-ness and Make-ness was desired.


I'm not bringing up POSIX-ness or Make functionality to make things harder for us, it should actually make things *easier*. I understand Ninja does not have POSIX-compliance as a major goal, but I believe it should be done whenever it is easy to do so. POSIX likely recommends against an option having multiple arguments because it's simply not efficient, and doesn't match what most consider to be an "option" compared to what one would define as "input". I personally don't see why we need the secondary stage of `getopt` processing or similar alternatives at all...

imagine instead of 

`ninja -t compdb [-x] [RULES...] [--targets TARGETS...]`

or

`ninja -t compdb [-x] [--targets TARGETS...] [RULES...]`

we can just have

`ninja -t compdb [-x] [TARGETS...] [RULES...]`
or just writing it the other way around...
`ninja -t compdb [-x] [RULES...] [TARGETS...]`

where the order of operands affects the order of output, but not the actual total content (amount of JSON elements).

Then in the manual you don't have to explain a new "option", rather it can simply read:
"given a list of rules and/or targets ... prints on standard output a compilation database ..."

I'm going to try to get that method to work myself, to make sure I'm not preaching for something that isn't possible to begin with... I wanted to have it done already but I've been busy...

jhasse:
Sry misclicked.

Regarding C++ standard: I'm okay with everything that works relativly easy on the oldest supported RHEL version (8 currently).

I think it would be best to stick with a command line way that would also work easily with other command line parsers.

digit-google:
@mcprat , what you propose is ambiguous, and I assure you that someone, somewhere is going to have a rule and a target with the same name, and won´t understand why they cannot use the tool as expected. There needs to be a explicit separation in the parsing format. Or simply introduce a new tool name for targets, nobody will ever get confused.

digit-google:
Regarding C++, it looks like we should be able to switch to C++17 (to be performed in a separate CL of course).

[this page](https://access.redhat.com/solutions/19458#:~:text=RHEL8%20%3A%20gcc%208.,RHEL7%20%3A%20gcc%204.8.) says RHEL ships with GCC 8.x or 9.x, and [that page](https://gcc.gnu.org/projects/cxx-status.html#cxx17) says that all C++17 features are supported starting with GCC 8 (and even GCC 7 if you don't care about full support of template deductions).

Similarly, [this page](https://learn.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-170) shows that VS 2019, which we currently use in CI, supports C++17 (unlike VS 2017 which still lacks some features).

mcprat:
> a rule and a target with the same name

yes, this is a real problem. but options without arguments like `--targets` or `--rules` can block the printing of one type or the other.

if that's not acceptable, then I would vote for separate tools with separate names.

JamesWidman:
> POSIX likely recommends against an option having multiple arguments because it's simply not efficient,

it turns out that, as soon as `getopt_long()` returns 't' to indicate that it recognized the "--targets" option, we can stash the relevant `optind` values and break out of the loop over calls to `getopt_long()`.  So, in my current implementation, "--targets" is just like "--" in the sense that it immediately terminates the processing of options by `getopt_long()`. That seems sufficiently efficient...?

JamesWidman:
@mcprat One reason why i prefer having some explicit option (or separate sub-tool, as in `ninja -t compdb_targets`) is that i tend to use these tools in scripts, and i want those scripts to be as readable as possible so that it's as easy as possible for the reader (possibly my distant-future self) to (re-)build a good understanding of what a particular invocation does & why.

as a user reading a shell script, `compdb --targets foo` tells me unambiguously that "foo" is intended/expected to be a target. And i consider this desirable, because in general, i like to avoid being confused about the meaning of the code that i read (including shell commands).

But also, "--targets" makes it clear to *ninja* that subsequent arguments must be taken as names of targets, and that makes the diagnosis of errors very very simple & straightforward.

JamesWidman:
> Sry misclicked.

no worries (:

> I think it would be best to stick with a command line way that would also work easily with other command line parsers.

Sorry; i'm not sure i understand what you mean here. What do you mean by "other command line parsers", and how would they be affected by the suggested `[rules] [--targets targets]` syntax?

JamesWidman:
> @mcprat , what you propose is ambiguous, and I assure you that someone, somewhere is going to have a rule and a target with the same name, and won´t understand why they cannot use the tool as expected.

Agreed; Hyrum's Law definitely applies here.  

>  There needs to be a explicit separation in the parsing format. Or simply introduce a new tool name for targets, nobody will ever get confused.

I'm happy to implement `ninja -t compdb_targets [targets]` as an alternative to `ninja -t compdb --targets` if that's what people want.

@digit-google how should we/i proceed?

(Note, `-t compdb_targets` would seem to imply that you cannot mix rules & targets in one command and get a de-duplicated set of results (which is fine by me, at the moment, since i don't see a use-case for it).)

JamesWidman:
the commit in this force-push is obviously very much a work-in-progress, but it seems like people might benefit from experimenting with its proposed command syntax:
```zsh
ninja -t compdb [-hx] [rules] [--targets targets]
```
(note, if you don't use "--targets", then all non-option arguments after `compdb` are interpreted as rules, as is the status quo.)

At this point, it seems like the main open question is whether we want this `--targets` command-line interface as opposed to a new tool, whose grammar would probably look something like this:

```zsh
ninja -t compdb-targets [-hx] [targets]
```
(the `compdb-targets` tool is currently not implemented, but i can do that next if anyone wants to experiment with it.  At this point, it would be very easy for me to add.)

 i guess the main question is: does anyone actually have a use-case for merging the commands from a mix of both rules and targets (as implemented in the most recent push)? If not, then maybe it's better to proceed with `-t compdb-targets` instead of `-t compdb --targets`

@digit-google thoughts?

JamesWidman:
coding note: since this represents an overhaul of `-t compdb` rather than a minor change, i moved all the compdb stuff in ninja.cc to just after the definition of `NinjaMain::ToolRestat()` so as to reduce diff noise.

JamesWidman:
update: i was just playing with [jq](https://jqlang.github.io/jq/), and it turns out that it can be used to merge & de-duplicate objects from two different json files. So, if you want to have a single `compile_commands.json` that contains commands for both a specified set of rules *and* a specified targets (for some reason), with duplicates removed, you could do:
```zsh
ninja -t compdb rule1 rule2 [...] > a.json
ninja -t compdb --targets target1 target2 [...] > b.json
echo '[]' | jq --slurpfile a a.json --slurpfile b b.json '($a[] + $b[]) | unique_by(.output) > compile_commands.json
```

(the jq function `unique_by()` sorts the output, but [there are ways to prevent that](https://stackoverflow.com/questions/74678678/jq-remove-duplicates-from-arrays)).

Since this merging-and-deduplication can be done with jq, i propose that this PR should discontinue pursuit of the `--targets` option and instead proceed with the implementation of a new tool, which will look something like this:

> ```shell
> ninja -t compdb-targets [-hx] [targets]
> ```

@digit-google @jhasse What do you think?

eli-schwartz:
> I wouldn't mind switching to C++17 personally, but @jhasse is the authority on this specific issue (and this may require updating some of our CI recipes).

I continue to be regularly confused every time I look at ninja issues:

Are you, @digit-google, a maintainer authority on ninja or are you not? Sometimes you say "I am but a humble contributor", most of the time you use the possessive "our", sometimes you use wording such as "I require you to do X. Note that jhasse is the authority on Y".

JamesWidman:
@digit-google note, i moved your line-cooking stuff in `output_test.py` to its own function so that it can also be used to cook output that ninja sends to stderr.

doc/style.css
```diff
@@ -53,3 +53,7 @@ div.chapter {
 p {
     margin-top: 0;
 }
+
+code.literal {
+    white-space: nowrap;
```
jhasse, line 58: What is this for?

src/ninja.cc
```diff
@@ -185,6 +187,10 @@ struct NinjaMain : public BuildLogUser {
   }
 
   int64_t start_time_millis_;
+
+  static std::string GetWorkingDirectory();
+
+  std::string working_directory = GetWorkingDirectory();
```
jhasse, line 193: hm ... why have this here which means it runs at startup although it might not be needed? Am I missing something?

jhasse (commented):
I think this is a good approach, thanks for working on this :)

doc/style.css
JamesWidman: It keeps the label part of the `[horizontal]` labeled list from being broken into more than one line. Before adding `code.literal { white-space: nowrap;}`, to `style.css`, the new entry in the manual looks like this:

image

...and after adding that bit of CSS, it looks like this:

image

src/ninja.cc
JamesWidman: so, i was thinking that this would save a little complexity in the NinjaMain::ToolCompilationDatabase* functions, in exchange for a negligible added cost for non-compdb tools.

But i neglected to actually measure that cost until now. 

Based on some quick-&-dirty uses of SDL_GetCurrentTime() at the start & end of NinjaMain::GetWorkingDirectory(), it looks like the cost is somewhere in the ballpark of 13 to 15 microseconds. (I also stepped through the instructions of getcwd() on my system and found that the number of syscalls actually executed is somewhat larger than i expected.)

And while 15µs doesn't seem like much on its own, if we allow small unnecessary time expenditures like this as a _policy_, sooner or later they could add up to something noticeable.

So, I'll go ahead and move that call to GetWorkingDirectory(); out of the NinjaMain constructor and into the compdb functions.

misc/output_test.py
```diff
@@ -371,6 +395,62 @@ def test_tool_inputs(self) -> None:
...
+build out4 : cat in4
+'''
+
+        def test_expected_error(flags, expected):
+          actual = ''
+          try:
+            actual = run(plan, flags, print_err_output=False)
+          except subprocess.CalledProcessError as err:
+            actual = err.cooked_output
+          self.assertEqual(expected, actual)
```
digit-google, line 415: nit: I suggest you convert this into a standalone method, since `plan` is the only local variable that seems to be captured by this inner function, and that this functionaility could be reused elsewhere. Use a `_test_` prefix so that `unittest` doesn't think it is a test function, e.g.:

```
  def _test_expected_error(plan: str, flags : T.Union[str, None], expected: str):
     """Run Ninja with a given plan and flags, and verify its cooked output against an expected content.
     """
      ...
```
```

File: misc/output_test.py [hunk 7]
```diff
@@ -74,58 +115,41 @@ def run(
...
         return b.run(flags, pipe, raw_output, env)
 
+def run_and_generate_expected_output(
+    build_ninja: str,
+    flags: str = '',
+    expected: Callable[[str], str] = None,
+    pipe: bool = False,
+    raw_output: bool = False,
+    env: Dict[str, str] = default_env,
+) -> (str, str):
```
digit-google, line 147: Can you document the function, since it has a high number of non-trivial parameters. I also suggest renaming `expected` to `compute_expected_output` for clarity, documenting that it takes the build directory's real path as parameter.

Do we really want a function to do this here. It looks like doing this at call sites would be easier to understand (no lambda) but the PR is good nonetheless.

src/ninja.cc
```diff
@@ -1087,6 +1096,124 @@ int NinjaMain::ToolRestat(const Options* options, int argc, char* argv[]) {
   return EXIT_SUCCESS;
 }
 
+struct Compdb_Targets {
```
digit-google, line 1099: a few nit related to stylistic issues:

- Do not use underscores in type names.

- Use ALL_CAPS for constants in enum values, or as an alternative `kCamelCase` (the `k` prefix is an indication that this is a constant, not not a method).

- Use CamelCase for method and functions, except for simple getters and setters, and when trying to mimic standard library classes. E.g. `StringPiece::size()`.

src/ninja.cc
```diff
@@ -1087,6 +1096,124 @@ int NinjaMain::ToolRestat(const Options* options, int argc, char* argv[]) {
...
+    display_help_and_exit,
+    emit_commands
+  };
+
+  Action action;
+  EvaluateCommandMode eval_mode = ECM_NORMAL;
+
+  std::vector<StringPiece> targets;
+
+  static Compdb_Targets parse(int argc, char* argv[]) {
```
digit-google, line 1110: "parse" is a bit vague here, and you would usually use such a name for something that changes the state of the current instance, not return a new one. Can you call it something like "CreateFromArgs", "NewFromArgs" or something similar.

src/ninja.cc
```diff
@@ -1087,6 +1096,124 @@ int NinjaMain::ToolRestat(const Options* options, int argc, char* argv[]) {
...
+    //
+    Compdb_Targets ret;
+
+    // getopt_long() expects argv[0] to contain the name of
+    // the tool, i.e. "compdb".
+    argc++;
+    argv--;
+
+    // Phase 1: parse options:
+    {
```
digit-google, line 1123: nit drop the enclosing braces here and ion line 1140, they add very little. The comments and spacing are nice so keep them :)

src/ninja.cc
```diff
@@ -1087,6 +1096,124 @@ int NinjaMain::ToolRestat(const Options* options, int argc, char* argv[]) {
...
+  }
+};
+
+
+void printCompdb(std::string const &directory, std::vector<Edge *> const &edges,
+                 const EvaluateCommandMode eval_mode) {
+  putchar('[');
+
+  bool first = true;
+  for (Edge* edge : edges) {
```
digit-google, line 1164: nit: use `const Edge*`

src/ninja.cc
```diff
@@ -1087,6 +1096,124 @@ int NinjaMain::ToolRestat(const Options* options, int argc, char* argv[]) {
...
+        "usage: ninja -t compdb [-hx] target [targets]\n"
+        "\n"
+        "options:\n"
+        "  -h     display this help messsage\n"
+        "  -x     expand @rspfile style response file invocations\n");
+    return 1;
+  }
+
+  case Compdb_Targets::Action::emit_commands: {
+    InEdge_Collector inEdge_collector;
```
digit-google, line 1192: Avoid mixing CamelCase and snake_case whenever possible. Variable names use snake_case. So `in_edge_collector` or even just `edge_collector` or `collector` would be fine here.

src/ninja.cc
```diff
@@ -1087,6 +1096,124 @@ int NinjaMain::ToolRestat(const Options* options, int argc, char* argv[]) {
...
+
+    for (auto target_arg : compdb.targets) {
```
digit-google, line 1194: An `for (auto var : range)` statement usually leads to copying every item during the iteration, which can have surprising effects, and is thus best avoided unless the type of `var` is trivial to see in the context of the code. However this is not the case here.

I recommend using `for (StringPiece target_arg : compdb.targets)` here, since it shows that the copy is harmless. Another option, in general, is to use `for (const auto& var : range)` if the iteration doesn't mutate the range items.
 
Use `StringPiece target_arg` here, since the `auto` is hard to resolve without looking way higher in the source code.

src/ninja.cc
```diff
@@ -372,6 +378,21 @@ bool NinjaMain::CollectTargetsFromArgs(int argc, char* argv[],
...
+    errno = 0;
+    success = getcwd(&ret[0], ret.size());
+  } while (!success && errno == ERANGE);
+  if (!success) {
+    Fatal("cannot determine working directory: %s", strerror(errno));
+  }
+  ret.resize(strlen(&ret[0]));
+  return ret;
+}
+
```
digit-google, line 395: This is nice. Can you move this to src/util.cc.

src/ninja.cc
```diff
@@ -1087,6 +1096,124 @@ int NinjaMain::ToolRestat(const Options* options, int argc, char* argv[]) {
...
+        "  -x     expand @rspfile style response file invocations\n");
+    return 1;
+  }
+
+  case Compdb_Targets::Action::emit_commands: {
+    InEdge_Collector inEdge_collector;
+
+    for (auto target_arg : compdb.targets) {
+      std::string err;
+      Node* node = CollectTarget(target_arg.begin(), &err);
```
digit-google, line 1196: There is no guarantee that `StringPiece::begin()` is zero-terminated, so I think there may be a bug here.
Use `target_arg.AsString().c_str()` instead to fix it.

src/graph.h
```diff
@@ -459,4 +461,48 @@ struct InputsCollector {
...
+///
+/// When collecting inputs, the outputs of phony edges are always ignored
+/// from the result, but are followed by the dependency walk.
+///
+/// Usage is:
+/// - Create instance.
+/// - Call CollectFrom() for each root node to collect edges from.
+/// - Call TakeResult() to retrieve the list of edges.
+///
+struct InEdge_Collector {
```
digit-google, line 475: nit: Do not use _ in type names, juse plain CamelCase.

src/graph_test.cc
```diff
@@ -282,6 +282,54 @@ TEST_F(GraphTest, InputsCollector) {
   EXPECT_EQ("out3", inputs[4]);
 }
 
+TEST_F(GraphTest, InEdgeCollector) {
```
digit-google, line 285: You added this test between InputsCollector and InputsCollectorWithEscapes, can you move it past the latter instead?

src/graph.h
```diff
@@ -459,4 +461,48 @@ struct InputsCollector {
...
+    for (Node* input_node : edge->inputs_)
+      CollectFrom(input_node);
+
+    if (!edge->is_phony())
+      in_edges.push_back(edge);
+
+  }
+
+  std::vector<Edge*> TakeResult() { return std::move(in_edges); }
+  bool Empty() { return in_edges.empty(); }
```
digit-google, line 496: It looks like you're never using these methods. Feel free to remove them.

src/graph.h
```diff
@@ -459,4 +461,48 @@ struct InputsCollector {
...
+
+  std::unordered_set<const Node*> visited_nodes_;
+  std::unordered_set<Edge*> visited_edges_;
```
digit-google, line 499: Move these to a `private` section.

src/ninja.cc
```diff
@@ -932,7 +953,7 @@ std::string EvaluateCommandWithRspfile(const Edge* edge,
   return command;
 }
 
-void printCompdb(const char* const directory, const Edge* const edge,
+void printOneCompdbObject(std::string const &directory, const Edge* const edge,
```
digit-google, line 956: nit: `PrintOneCompdbObject`.

Use `git clang-format` to reformat your commits, the `&` goes on the side of the type. See `CONTRIBUTING.md` for the link to the style guide.

src/ninja.cc
```diff
@@ -977,18 +998,6 @@ int NinjaMain::ToolCompilationDatabase(const Options* options, int argc,
...
-    errno = 0;
-    success = getcwd(&cwd[0], cwd.size());
-  } while (!success && errno == ERANGE);
-  if (!success) {
-    Error("cannot determine working directory: %s", strerror(errno));
-    return 1;
-  }
 
   putchar('[');
   for (vector<Edge*>::iterator e = state_.edges_.begin();
```
digit-google, line 1003: Since we're touching this part of the code, can I ask you to change this to use a for-range based loop:

```
for (const Edge* edge : state_.edges_) {
  ...
}
```

src/ninja.cc
```diff
@@ -1087,6 +1096,124 @@ int NinjaMain::ToolRestat(const Options* options, int argc, char* argv[]) {
...
+      int opt;
+      while ((opt = getopt(argc, argv, "hx")) != -1) {
+        switch (opt) {
+        case 'x':
+          ret.eval_mode = ECM_EXPAND_RSPFILE;
+          break;
+        case 'h':
+        default:
+          ret.action = Compdb_Targets::Action::display_help_and_exit;
+          goto fin;
```
digit-google, line 1134: nit: Just `return ret` here.

misc/output_test.py
JamesWidman: > Do we really want a function to do this here. It looks like doing this at call sites would be easier to understand (no lambda) but the PR is good nonetheless.

I was prompted to introduce the `expected`/`compute_expected_output` parameter because of this member of `class BuildDir`:

```
    def __enter__(self):
        self.d = tempfile.TemporaryDirectory()
        #  [...]
```
...which is implicitly called upon entering this `with` statement:

```
    with BuildDir(build_ninja) as b:
        return b.run(flags, pipe, raw_output, env)
```
So the only time when it's possible to get the name of the temporary directory (which we need for the "directory" part of each compilation database entry in the expected output) is during an execution of that `with` statement. 

That said... if i invoke `with` directly from within `test_tool_compdb_targets()`, i won't need a lambda.

Alternatively, BuildDir.run() could return a tuple that contains both the output and the name of the temporary directory...?

src/ninja.cc
JamesWidman: fwiw, i deliberately mix camel case and snake case whenever the equivalent english phrase would be partially hyphenated; e.g. an "in-edge collector" is a collector that collects the (direct and indirect) in-edges of the given node (as opposed to out-edges of said node), so called because it uses `node->in_edge()` (as opposed to `node->out_edges()`)

since hyphens are not allowed in C++ identifiers, i squish the hyphenated part into camel case, and replace the space with an underscore; thus "in-edge collector" is transformed into "InEdge_Collector".

in the case of the variable, i'm fine with renaming it `collector`, but i guess the class would have to be renamed to `In_Edge_Collector`, which... feels unfortunate, because it loses the grouping, and on first read, you could mistakenly read it as "in edge-collector" (i.e. "to be inside an edge collector"...?)

But i'm not going to die on this particular hill. (:

src/ninja.cc
JamesWidman: thank you for catching this!

misc/output_test.py
digit-google: I suggest adding a `path` property to the `BuildDir` class then and use that, it will be simpler :)

JamesWidman: Ah, like:
```python
class BuildDir:
    path: str = ''

    def __enter__(self):
        self.d = tempfile.TemporaryDirectory()
        BuildDir.path=os.path.realpath(self.d.name)
```
i didn't think of that, probably because, by the time we use BuilDir.path, that directory will no longer exist. But it doesn't need to exist; we only need that absolute path to compute the expected output.

src/ninja.cc
JamesWidman: agreed!
[happily retreats from hill]

src/ninja.cc
JamesWidman: wait; on second thought: in this case, `target_arg.begin()` (a.k.a. `target_arg.str_`) _is_ guaranteed to be null-terminated, because it's the same pointer value as one of the elements of `argv` from `main()`; e.g. if there are no options, then the first `target_arg.str_` ==  `argv[3]` (and if that's not null-terminated, then something weirdly bad must have happened much earlier).

also, every correct use of `StringPiece` depends on knowing the provenance of the string pointed-to. Since we need to know the provenance anyway, and in this case knowing the provenance also comes with the knowledge that the string is null-terminated, it seems ok to depend on that...?

misc/output_test.py
digit-google: BuildDir.path would be a class-level member,  which is very unusual, and technically wrong in the case several instances exist at the same time. Use a proper property instead, e.g.:

```
@property
def path(self) -> str:
    return self.d.name
```

Also just with `with Builder(...) as b: ...` in the test itself to access `b.path`, there is nothing wrong with that, it adds very little code and everything is clear.

src/ninja.cc
digit-google: In _general_, StringPiece::begin() is not zero-terminated. In this specific line, it happens to be the case though, because of the way compdb.targets is _currently_ built, but this is not visible from the current code's context.

It is easy to imagine that someone else later changes how combd.targets is populated (for example by expanding the list from a response file, or to support the `target^` syntax), and then you will have an unexpected bug.

Maybe StringPiece is not the right type for the targets vector, and that a `const char*` type would be better because it exposes the fact that the strings are zero-terminated? Or just use a vector of strings?

misc/output_test.py
JamesWidman: i pretty much just copied & pasted this (though i changed the type of `flags` to `Optional[str]`); would you like me to add you as a co-author? 

(full disclosure: i have not touched python in years, so i've depended heavily on your suggestions here)

JamesWidman:
(note, `CommandCollector` (formerly known as `InEdge_Collector`) moved from src/graph.h to src/command_collector.h)

misc/output_test.py
```diff
@@ -12,14 +12,50 @@
 import tempfile
 import unittest
 from textwrap import dedent
-from typing import Dict
+from typing import Dict, Optional
+from collections.abc import Callable
```
JamesWidman, line 16: oops; i forgot to remove the import of `Callable`; this will be gone in the next push.

JamesWidman:
is there a recommended way to use `clang-tidy`?  (Apparently it ships with a configuration that matches google's style guidelines.)

misc/output_test.py
eli-schwartz: Superior style for avoiding this:

```python
import os
import sys
import tempfile
import foobar
import typing as T

def func(args: T.Optional[T.Dict[str, str]]) -> T.Callable[........]:
    pass
```

A commonly used module where the symbols in use are regularly in flux is a strange time to spend most of your time micromanaging which symbols are currently in scope. But everyone does it. I'm always baffled.

The semantic difference between the two is trivial:
- `from x import y` fills up the local namespace and in return accessing the "y" symbol avoids having to do a name lookup inside "x"

The effects on code readability are considerable:
- with `from x import y` you lose the ability to clearly reason at a glance, that many symbols are part of the same logical context

JamesWidman: done and pushed; thank you! 

(as i mentioned upthread, i have not touched python in years, so i'm grateful for any python help!)

misc/output_test.py
```diff
@@ -12,14 +12,49 @@
...
-from typing import Dict
+import typing as T
 
 default_env = dict(os.environ)
 default_env.pop('NINJA_STATUS', None)
 default_env.pop('CLICOLOR_FORCE', None)
 default_env['TERM'] = ''
 NINJA_PATH = os.path.abspath('./ninja')
 
+def cook(raw_output: bytes) -> str:
```
digit-google, line 23: may I suggest a more meaningful name, e.g. `remove_non_visible_lines` here?

misc/output_test.py
eli-schwartz: Well, like I said an awful lot of people seem to do it so you're not exactly in bad company. :P 

Probably the rationale is that `from` imports contribute less to line max width. That is the usual reason to use `from` imports for meaningful runtime code.

And if the alternative was truly 
```python
import typing 

def func(args: typing.Optional[typing.Dict[str, str]]) -> typing.Callable[........]:
    pass
```

I'd perhaps concede. That's a 7 char cost per symbol and typing symbols always appear on a line where something else is already present. But `T` is quite distinctive and makes sense to reserve for a language level feature in a way that `frobnicator.core.utils` as `U` doesn't really make sense. :shrug: I wish they'd named it "T" all along.

digit-google:
Thank you, the last PR looks good apart from minor nits. @jhasse, could you mark this PR as ready for review so that CI builders start building them. I compiled it locally for Linux and Windows only (cross-compiled with mingw64 only).

JamesWidman:
@digit-google fwiw, before today's pushes, one of the linux-hosted automated tests was failing, but i forgot to look into it before the force-push.

JamesWidman:
i finally looked at the job logs (:
(fixed errors from codespell about misspelled instances of "message", and a missing #include of `<direct.h>` for the MSVC build.)

doc/style.css
JamesWidman: update: for the next push, i've changed this bit of CSS to:
```CSS
table.horizontal tbody tr td:nth-child(1) {
    white-space: nowrap;
    padding-right: 1.5em;
}
```

this makes it so that:
- the `nowrap` property value is now applied only to the left column of a `[horizontal]` labeled list (which fixes an issue where other instances of `code.literal` were prevented from wrapping in the middle of a paragraph, which led to some awkwardly-short lines leading up to the then-unbreakable code), and 
- there is now some padding on the right (which is missing in the 2nd screenshot above).

JamesWidman:
> Thank you, the last PR looks good apart from minor nits. @jhasse, could you mark this PR as ready for review so that CI builders start building them. I compiled it locally for Linux and Windows only (cross-compiled with mingw64 only).

@digit-google @jhasse yeah; in the workflow for windows, i see that it successfully builds with MSVC, and that the unit tests completed successfully, but it looks like we don't run `output_test.py` there, so it would be good to know whether an MSVC-built ninja actually produces the desired output for `-t compdb-targets`.

JamesWidman:
> it would be good to know whether an MSVC-built ninja actually produces the desired output for `-t compdb-targets`.

on the other hand, `compdb-targets` isn't doing anything OS-specific that `compdb` doesn't do, and `compdb` presumably works on windows, so...

JamesWidman:
side note: when i try to build the documentation on macOS, using the version of `xsltproc` that ships with macOS (`/usr/bin/xsltproc`), i get:

```zsh
% which xsltproc
/usr/bin/xsltproc
% ./configure.py
wrote build.ninja.
% ninja manual
[1/1] XSLTPROC doc/manual.html
FAILED: doc/manual.html
xsltproc --nonet doc/docbook.xsl build/manual.xml > doc/manual.html
I/O error : Attempt to load network entity http://docbook.sourceforge.net/release/xsl/current/html/docbook.xsl
warning: failed to load external entity "http://docbook.sourceforge.net/release/xsl/current/html/docbook.xsl"
compilation error: file doc/docbook.xsl line 8 element import
xsl:import : unable to load http://docbook.sourceforge.net/release/xsl/current/html/docbook.xsl
%
```
...but if i use the version of libxslt provided by [Homebrew](https://brew.sh/) (after installing XSLT 1.0 Stylesheets for DocBook with `brew install docbook-xsl`), everything works: 
```zsh
% export PATH="$(brew --prefix libxslt)/bin:$PATH"
% ninja manual
[1/1] XSLTPROC doc/manual.html
%
```
i'm guessing that the homebrew version is checking homebrew-standard paths to find the docbook stuff, whereas the macOS version doesn't...

should we mention something about this in `README.md`? E.g. words to the effect of: 

> to build documentation on macOS, first run:
```sh
brew install libxslt docbook-xsl
export PATH="$(brew --prefix libxslt)/bin:$PATH"
```

digit-google:
Mentioning it in the README.md would be nice. Detecting the version of xslt at configure time and warning about it would be nicer, imho, but I don't know enough about this library and its dependencies to know if this is simple enough to implement.

doc/style.css
mcprat: might want to consider making this a separate commit

JamesWidman: done!

JamesWidman:
is it normal for workflows to be re-executed when the content of the most recent push is identical to the previous push?

mcprat:
> is it normal for workflows to be re-executed when the content is identical to the previous push?

unfortunately yes, the CI is not that smart, it operates on whether the commit hash has changed not the content. however, GitHub remembers which hashes passed or not so it can display that when a bad commit is placed on top

eli-schwartz:
> is it normal for workflows to be re-executed when the content of the most recent push is identical to the previous push?

https://github.com/orgs/community/discussions/64405

I fear you will be waiting quite long for a resolution but fortunately the CI here doesn't take 2 hours on each run.

JamesWidman:
@jhasse is there anything else i need to do for this PR? (I thought I fixed or addressed everything, but please let me know if I missed anything!)

JamesWidman:
thank you both!

PR #2503, opened 2024-09-26
Fix #2499: Status update regression.

### LINKED ISSUE #2499: Ninja lost the ability to print compiled file names in the same line
Commit [1ba5570807e3ded62c4939ca87f20d8c8696fa46](https://github.com/ninja-build/ninja/commit/1ba5570807e3ded62c4939ca87f20d8c8696fa46) kills ninja's ability to avoid flooding with the names of compiled files.
Old behavior:
good

New behavior (after the commit):
bad

digit-google:
PR #2487 introduced a regression, where a completed command without an output would force a newline, preventing the next status update to appear on the same line in smart terminals.

This fixes the issue by adding the missing `!outputs.empty()` condition + adding a proper regression test to catch future breaks.

Fixed: #2499

### DIFF
--- a/misc/output_test.py
+++ b/misc/output_test.py
@@ -39,8 +39,29 @@ class BuildDir:
         self,
         flags: str = '',
         pipe: bool = False,
+        raw_output: bool = False,
         env: Dict[str, str] = default_env,
     ) -> str:
+        """Run Ninja command, and get filtered output.
+
+        Args:
+          flags: Extra arguments passed to Ninja.
+
+          pipe: set to True to run Ninja in a non-interactive terminal.
+            If False (the default), this runs Ninja in a pty to simulate
+            a smart terminal (this feature cannot work on Windows!).
+
+          raw_output: set to True to return the raw, unfiltered command
+            output.
+
+          env: Optional environment dictionary to run the command in.
+
+        Returns:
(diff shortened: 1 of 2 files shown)


misc/output_test.py
```diff
@@ -39,8 +39,29 @@ def run(
...
+        """Run Ninja command, and get filtered output.
+
+        Args:
+          flags: Extra arguments passed to Ninja.
+
+          pipe: set to True to run Ninja in a non-interactive terminal.
+            If False (the default), this runs Ninja in a pty to simulate
+            a smart terminal (this feature cannot work on Windows!).
+
+          raw_output: set to True to return the raw, unfilted command
```
jhasse, line 54: unfiltered

misc/output_test.py
```diff
@@ -39,8 +39,29 @@ def run(
...
+            output.
+
+          env: Optional environment dictionary to run the command in.
+
+        Returns:
+          An UTF-8 corresponding to the output (stdout only) of the
```
jhasse, line 60: missing "string"?

misc/output_test.py
digit-google: Fixed. Thanks.

misc/output_test.py
digit-google: Indeed :-)

Issue #2515, opened 2024-10-29
Ninja does not update elapsed time in status for long build commands


digit-google:
`NINJA_STATUS` now supports new formatters, such as `%e` or `%w`, which print the elapsed time in seconds.

Their usefulness is however severely limited, because Ninja, still *only updates its status when a command completes*.
This results in poor user experience in smart terminals when long commands dominate a build (e.g. Rust link operations). Consider this example build plan:

```
rule sleep_then_print
  command = sleep 10 && echo done

build all: sleep_then_print
```

Running the command `NINJA_STATUS="[%w] " ninja`  in a terminal results in the line:

```
[00:00] sleep 10 && echo done
```

Being printed for 10 seconds, *without the time being updated*. Then on completion, it is overwritten into:

```
[00:10] sleep 10 && echo done
done
```

The root of the problem is that `SubprocessSet::DoWork()` is used to wait for events during the build, and only returns when it detects that a process completed, or in the case of a user interruption.

To fix this issue the following changes are proposed:

- Modify `SubprocessSet::DoWork()` to accept a timeout parameter in milliseconds, and return an enum to distinguish between timeout / process completion / user interruption.

- Add a `Status::Refresh(int64_t)` method, and implement it in `StatusPrinter::Refresh()` to update the status when in interactive terminals.

- Modify the `RealCommandRunner` function that calls `DoWork()` to handle timeouts and call `status->Refresh()` in this case.
  Current prototype at https://github.com/ninja-build/ninja/compare/master...digit-google:ninja:periodic-status-updates

Note that doing this opens the door to other improvements that could appear later, such as:

- Printing a [table of running commands](https://fuchsia.dev/fuchsia-src/development/build/ninja_fuchsia_improvements?hl=en#feature_status_of_running_commands) under the status line, to better understand what's going on.

- Polling for new jobserver slots appearing in the pool to start new tasks as soon as possible, as explained [in this comment](https://github.com/ninja-build/ninja/issues/1139#issuecomment-2389078694), though ideally, it would be nicer if this could be handled as a new type of event for `DoWork()`.

jhasse:
Also see #111.

Issue #1328, opened 2017-09-13
option to search in parent directories for build.ninja


vvuk:
It would be handy to have an option to `ninja` that would cause it to keep searching up the directory tree to find a `build.ninja` (or `-f` whatever) file if one isn't present in the current directory.  Right now if you're in a subdirectory of your build tree you have to either change directories first or explicitly `-C ../..` (or an absolute path).

PR #2516, opened 2024-10-29
Update status periodically in smart terminals

digit-google:
Ensure that time-related `NINJA_STATUS` formatters such as `%w` as properly updated when running in a smart terminal.

This fixes issue #2515 by ensuring that Ninja can refresh its status even when only long commands are running.

- Modify `Subprocess::DoWork()` to accept a timeout, and return an enum describing the exit condition.
- Add `Status::Refresh()` and `StatusPrinter::Refresh()` to update the status on smart terminals.
- Modify `RealCommandRunner` to call `DoWork()` with a timeout, and call `status->Refresh()` on expiration to trigger the update.

The end result is that the status is properly updated in smart terminals when a single long command is invoked (see the example build plan in #2515).

NOTE: This adds a new regression test in `misc/output_test.py`, however, the test is flaky since it is time-dependent, i.e. it runs successfully on a lightly loaded laptop, but fails frequently on Github CI. Hence it is disabled by default but can be run manually during development.

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -190,7 +190,8 @@ you don't need to pass `-j`.)
 Environment variables
 ~~~~~~~~~~~~~~~~~~~~~
 
-Ninja supports one environment variable to control its behavior:
+Ninja supports a few environment variables to control its behavior:
+
 `NINJA_STATUS`, the progress status printed before the rule being run.
 
 Several placeholders are available:
@@ -214,6 +215,12 @@ specified by `-j` or its default)
 The default progress status is `"[%f/%t] "` (note the trailing space
 to separate from the build rule). Another example of possible progress status
 could be `"[%u/%r/%f] "`.
+
+`NINJA_STATUS_REFRESH_MILLIS`, the refresh timeout in milliseconds
+for status updates in interactive terminals. The default value is 1000,
+to allow time-sensitive formatters like `%w` to be updated during
+long build runs (e.g. when one or more build commands run for a long
+time).
 
 Extra tools
 ~~~~~~~~~~~
--- a/misc/output_test.py
+++ b/misc/output_test.py
@@ -16,6 +16,7 @@ from typing import Dict
 
 default_env = dict(os.environ)
 default_env.pop('NINJA_STATUS', None)
+default_env.pop('NINJA_STATUS_REFRESH_MILLIS', None)
 default_env.pop('CLICOLOR_FORCE', None)
 default_env['TERM'] = ''
 NINJA_PATH = os.path.abspath('./ninja')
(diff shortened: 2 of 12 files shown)


digit-google:
Interesting, would you care sharing your plans about this rewrite? This might have a very important impact on the Fuchsia fork, as we try very hard to stay close to upstream (i.e. I routinely rebase our stack of patchs on top of upstream Ninja to make them manageable without any merges).

Since we are sharing our intentions:

I didn't mention here is that for the Fuchsia fork, which also needs to implement a client/server mode, we refactored Ninja considerably to introduce an [AsyncLoop abstraction](https://fuchsia.googlesource.com/third_party/github.com/ninja-build/ninja/+/refs/heads/main/src/async_loop.h) which allows for waiting for different types of I/O events, and made `SubprocessSet` use it for both Posix and Win32. This also supports timers, which are used in our implementation of [`StatusTable`](https://fuchsia.googlesource.com/third_party/github.com/ninja-build/ninja/+/refs/heads/main/src/status_table.h), which is how we [print running commands in smart terminals](https://fuchsia.dev/fuchsia-src/development/build/ninja_fuchsia_improvements?hl=en#feature_status_of_running_commands).

I didn't want to introduce such level of complexity here, and I don't think upstream needs the `AsyncLoop` interface at this point.

I have also [another branch](https://github.com/digit-google/ninja/tree/status-table) to implement `StatusTable` in upstream Ninja that uses a simpler version of the class (i.e. one which doesn't depend on `AsyncLoop` at all). It hasn't been uploaded as a PR yet, but works on top of this PR properly.

For the record, this has been used by my team for over a year, and has helped _tremendously_ in identifying build commands that were bogging the build for no good reason, which led to _several_ build performance improvements that would not have been possible with Ninja's current output. Which is why I think having such a feature in v1 would benefit

Finally, we are working to introduce in our own fork a new Status sub-class to generate a [Build Event Protocol](https://bazel.build/remote/bep) stream. The BEP originated in Bazel but has become a format supported by many other build / CI analytics tools, it is a binary protobuf with a well-known fixed schema that provides rich information about what's happening during the build, is easy to convert to other formats (in particular into perfetto traces), and doesn't have the problems of JSON (slow encoding + bad non-UTF8 support). We do not expect upstream to accept such a feature though, but any change to how the frontend works is important for us. Surprisingly it doesn't add much code to the source tree.

FWIW, the Android Ninja fork already has such a feature, except that they use their own custom protobuf schema, which includes far more metrics in the event stream, but we prefer to stick to BEP for compatibility. They have been using it for years and are very happy with it.

Finally, I am a little wary about delivery. A major rewrite like the one you describe would take a large amount of work, and I didn't have the feeling that you have a lot of time to work on this. How many months / years are you forecasting for such work? I am really curious to know.

jhasse:
Interesting!

I think some kind of async event loop is a good idea. I've used Boost.Asio since I'm not good enough to implement that from scratch, but I guess Boost is banned at Google because NIH? I guess a hand-crafted solution is also much faster to compile, results in a smaller binary and can be tuned for performance.

Those extensions you have in your Fuchsia branch are very interesting! That's why I dig the idea of a second v2 binary / folder: I could basically merge PRs that only touch v2 right away without much considerations, because it would be marked as "beta" and I don't have to worry about breaking something for existing users.

These are the breaking changes I have in mind:

* After all the nice frontend improvements were added, make them the default instead of just options.
* Make description part of the status format string in order to allow to change its position, see #1045
* Introduce a `--status` command line flag, see #1180, and remove support for `NINJA_STATUS` env var.
* Add support for `NINJA_FLAGS` #797. That's the replacement for `NINJA_STATUS` then, too.
* Always realpath everything in order to fix #1251.
* Only use hashes instead of timestamps, see #1459.
* To mitigate the performance impact of the previous two points: Add client/server mode.
* Look at a lot more places for build manifest, e.g. in `build/` or parent directories, see #1328.
* Automatically configure CMake, etc. projects. I dream of a README.md of a C++ which just lists `ninja` as the build instructions :D
* Replace getopt with a good C++ arg parsing library. getopt has caused problems in the past, the recent case being your `--jobserver` flag. When doing that, also improve the argument names / format / etc.
* Completely drop support for building ninja on old systems. This holds us back and if someone wants to compile ninja on their toaster they should just use samurai instead (or v1).

> How many months / years are you forecasting for such work? I am really curious to know.

As that would be in parallel I would guess years? Because it's a second binary and the source would be in a `v2` directory, I could quickly merge PRs that only touch on that code. My hope is that this encourages people to help out.

For the past years there were many good PRs that added useful functionality but weren't merged because there's always the risk of breaking something. I want to change that while keeping the stable ninja v1.

btw: My changes don't touch the existing code that much, nearly nothing at all. So you shouldn't have any problem rebasing the Fuchsia fork on top of them :)
I really should find the time to open the PR and show you my prototype ...

PR #2519, opened 2024-11-01
Speed up parsing

sesse:
This patch series speeds up ninja parsing (as measured by a no-op Chromium build) by about 40–50%. The main win is reducing allocation rate by punting StringPiece allocation to a arena/bump allocator, but we also switch out the hash table and hash functions in use.

The series might seem large, but a) most of it is vendoring EmHash8 and rapidhash, and b) most of the rest is just piping an arena pointer through to the various functions and tests.

### DIFF
--- a/src/deps_log.cc
+++ b/src/deps_log.cc
@@ -186,15 +186,13 @@ LoadStatus DepsLog::Load(const string& path, State* state, string* err) {
     return LOAD_SUCCESS;
   }
 
-  long offset;
+  long offset = ftell(f);
   bool read_failed = false;
   int unique_dep_record_count = 0;
   int total_dep_record_count = 0;
   for (;;) {
-    offset = ftell(f);
-
     unsigned size;
-    if (fread(&size, 4, 1, f) < 1) {
+    if (fread(&size, sizeof(size), 1, f) < 1) {
       if (!feof(f))
         read_failed = true;
       break;
@@ -206,6 +204,7 @@ LoadStatus DepsLog::Load(const string& path, State* state, string* err) {
       read_failed = true;
       break;
     }
+    offset += size + sizeof(size);
 
     if (is_deps) {
       if ((size % 4) != 0) {
--- a/src/eval_env.cc
+++ b/src/eval_env.cc
@@ -99,6 +99,10 @@ string BindingEnv::LookupWithFallback(const string& var,
 }
 
 string EvalString::Evaluate(Env* env) const {
+  if (parsed_.empty()) {
+    return single_token_;
+  }
+
   string result;
   for (TokenList::const_iterator i = parsed_.begin(); i != parsed_.end(); ++i) {
     if (i->second == RAW)
@@ -110,40 +114,57 @@ string EvalString::Evaluate(Env* env) const {
 }
 
 void EvalString::AddText(StringPiece text) {
-  // Add it to the end of an existing RAW token if possible.
-  if (!parsed_.empty() && parsed_.back().second == RAW) {
-    parsed_.back().first.append(text.str_, text.len_);
+  if (parsed_.empty()) {
+    single_token_.append(text.begin(), text.end());
+  } else if (!parsed_.empty() && parsed_.back().second == RAW) {
+    parsed_.back().first.append(text.begin(), text.end());
   } else {
-    parsed_.push_back(make_pair(text.AsString(), RAW));
+    parsed_.push_back(std::make_pair(text.AsString(), RAW));
   }
 }
+
 void EvalString::AddSpecial(StringPiece text) {
-  parsed_.push_back(make_pair(text.AsString(), SPECIAL));
+  if (parsed_.empty() && !single_token_.empty()) {
+    // Going from one to two tokens, so we can no longer apply
+    // our single_token_ optimization and need to push everything
+    // onto the vector.
+    parsed_.push_back(std::make_pair(std::move(single_token_), RAW));
+  }
+  parsed_.push_back(std::make_pair(text.AsString(), SPECIAL));
 }
 
 string EvalString::Serialize() const {
   string result;
-  for (TokenList::const_iterator i = parsed_.begin();
-       i != parsed_.end(); ++i) {
+  if (parsed_.empty() && !single_token_.empty()) {
     result.append("[");
-    if (i->second == SPECIAL)
-      result.append("$");
-    result.append(i->first);
+    result.append(single_token_);
     result.append("]");
+  } else {
+    for (const auto& pair : parsed_) {
+      result.append("[");
+      if (pair.second == SPECIAL)
+        result.append("$");
+      result.append(pair.first.begin(), pair.first.end());
+      result.append("]");
+    }
   }
   return result;
 }
 
 string EvalString::Unparse() const {
   string result;
-  for (TokenList::const_iterator i = parsed_.begin();
-       i != parsed_.end(); ++i) {
-    bool special = (i->second == SPECIAL);
-    if (special)
-      result.append("${");
-    result.append(i->first);
-    if (special)
-      result.append("}");
+  if (parsed_.empty() && !single_token_.empty()) {
+    result.append(single_token_.begin(), single_token_.end());
+  } else {
+    for (TokenList::const_iterator i = parsed_.begin();
+         i != parsed_.end(); ++i) {
+      bool special = (i->second == SPECIAL);
+      if (special)
+        result.append("${");
+      result.append(i->first.begin(), i->first.end());
+      if (special)
+        result.append("}");
+    }
   }
   return result;
 }
--- a/src/third_party/rapidhash/README.ninja
+++ b/src/third_party/rapidhash/README.ninja
@@ -1,0 +1,7 @@
+Description: Very fast, high quality, platform-independent hashing algorithm.
+Version: commit 4a6b2570e868536be84800353efd92c699f37d2c
+URL: https://github.com/Nicoshev/rapidhash
+Copyright: Copyright (C) 2024 Nicolas De Carli, Based on 'wyhash', by Wang Yi <godspeed_china@yeah.net>
+SPDX-License-Identifier: BSD-2-Clause
+Local changes:
+ - Changed to UNIX line endings
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -53,63 +53,14 @@ using namespace std;
 namespace {
 
 const char kFileSignature[] = "# ninja log v%d\n";
-const int kOldestSupportedVersion = 6;
-const int kCurrentVersion = 6;
-
-// 64bit MurmurHash2, by Austin Appleby
-#if defined(_MSC_VER)
-#define BIG_CONSTANT(x) (x)
-#else   // defined(_MSC_VER)
-#define BIG_CONSTANT(x) (x##LLU)
-#endif // !defined(_MSC_VER)
-inline
-uint64_t MurmurHash64A(const void* key, size_t len) {
-  static const uint64_t seed = 0xDECAFBADDECAFBADull;
-  const uint64_t m = BIG_CONSTANT(0xc6a4a7935bd1e995);
-  const int r = 47;
-  uint64_t h = seed ^ (len * m);
-  const unsigned char* data = static_cast<const unsigned char*>(key);
-  while (len >= 8) {
-    uint64_t k;
-    memcpy(&k, data, sizeof k);
-    k *= m;
-    k ^= k >> r;
-    k *= m;
-    h ^= k;
-    h *= m;
-    data += 8;
-    len -= 8;
-  }
-  switch (len & 7)
-  {
-  case 7: h ^= uint64_t(data[6]) << 48;
-          NINJA_FALLTHROUGH;
-  case 6: h ^= uint64_t(data[5]) << 40;
-          NINJA_FALLTHROUGH;
-  case 5: h ^= uint64_t(data[4]) << 32;
-          NINJA_FALLTHROUGH;
-  case 4: h ^= uint64_t(data[3]) << 24;
-          NINJA_FALLTHROUGH;
-  case 3: h ^= uint64_t(data[2]) << 16;
-          NINJA_FALLTHROUGH;
-  case 2: h ^= uint64_t(data[1]) << 8;
-          NINJA_FALLTHROUGH;
-  case 1: h ^= uint64_t(data[0]);
-          h *= m;
-  };
-  h ^= h >> r;
-  h *= m;
-  h ^= h >> r;
-  return h;
-}
-#undef BIG_CONSTANT
-
+const int kOldestSupportedVersion = 7;
+const int kCurrentVersion = 7;
 
 }  // namespace
 
 // static
 uint64_t BuildLog::LogEntry::HashCommand(StringPiece command) {
-  return MurmurHash64A(command.str_, command.len_);
+  return rapidhash(command.str_, command.len_);
 }
 
 BuildLog::LogEntry::LogEntry(const string& output)
--- a/src/build_log_test.cc
+++ b/src/build_log_test.cc
@@ -104,7 +104,7 @@ TEST_F(BuildLogTest, FirstWriteAddsSignature) {
 
 TEST_F(BuildLogTest, DoubleEntry) {
   FILE* f = fopen(kTestFilename, "wb");
-  fprintf(f, "# ninja log v6\n");
+  fprintf(f, "# ninja log v7\n");
   fprintf(f, "0\t1\t2\tout\t%" PRIx64 "\n",
       BuildLog::LogEntry::HashCommand("command abc"));
   fprintf(f, "0\t1\t2\tout\t%" PRIx64 "\n",
@@ -177,7 +177,7 @@ TEST_F(BuildLogTest, ObsoleteOldVersion) {
 
 TEST_F(BuildLogTest, SpacesInOutput) {
   FILE* f = fopen(kTestFilename, "wb");
-  fprintf(f, "# ninja log v6\n");
+  fprintf(f, "# ninja log v7\n");
   fprintf(f, "123\t456\t456\tout with space\t%" PRIx64 "\n",
       BuildLog::LogEntry::HashCommand("command"));
   fclose(f);
@@ -200,10 +200,10 @@ TEST_F(BuildLogTest, DuplicateVersionHeader) {
   // build log on Windows. This shouldn't crash, and the second version header
   // should be ignored.
   FILE* f = fopen(kTestFilename, "wb");
-  fprintf(f, "# ninja log v6\n");
+  fprintf(f, "# ninja log v7\n");
   fprintf(f, "123\t456\t456\tout\t%" PRIx64 "\n",
       BuildLog::LogEntry::HashCommand("command"));
-  fprintf(f, "# ninja log v6\n");
+  fprintf(f, "# ninja log v7\n");
   fprintf(f, "456\t789\t789\tout2\t%" PRIx64 "\n",
       BuildLog::LogEntry::HashCommand("command2"));
   fclose(f);
@@ -252,7 +252,7 @@ struct TestDiskInterface : public DiskInterface {
 
 TEST_F(BuildLogTest, Restat) {
   FILE* f = fopen(kTestFilename, "wb");
-  fprintf(f, "# ninja log v6\n"
+  fprintf(f, "# ninja log v7\n"
              "1\t2\t3\tout\tcommand\n");
   fclose(f);
   std::string err;
@@ -280,7 +280,7 @@ TEST_F(BuildLogTest, VeryLongInputLine) {
   // Ninja's build log buffer is currently 256kB. Lines longer than that are
   // silently ignored, but don't affect parsing of other lines.
   FILE* f = fopen(kTestFilename, "wb");
-  fprintf(f, "# ninja log v6\n");
+  fprintf(f, "# ninja log v7\n");
   fprintf(f, "123\t456\t456\tout\tcommand start");
   for (size_t i = 0; i < (512 << 10) / strlen(" more_command"); ++i)
     fputs(" more_command", f);
--- a/src/eval_env.h
+++ b/src/eval_env.h
@@ -39,8 +39,8 @@ struct EvalString {
   /// @return The string with variables not expanded.
   std::string Unparse() const;
 
-  void Clear() { parsed_.clear(); }
-  bool empty() const { return parsed_.empty(); }
+  void Clear() { parsed_.clear(); single_token_.clear(); }
+  bool empty() const { return parsed_.empty() && single_token_.empty(); }
 
   void AddText(StringPiece text);
   void AddSpecial(StringPiece text);
@@ -53,6 +53,12 @@ private:
   enum TokenType { RAW, SPECIAL };
   typedef std::vector<std::pair<std::string, TokenType> > TokenList;
   TokenList parsed_;
+
+  // If we hold only a single RAW token, then we keep it here instead of
+  // pushing it on TokenList. This saves a bunch of allocations for
+  // what is a common case. If parsed_ is nonempty, then this value
+  // must be ignored.
+  std::string single_token_;
 };
 
 /// An invocable build command and associated metadata (description, etc.).
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -740,12 +740,13 @@ bool ImplicitDepLoader::LoadDepsFromLog(Edge* edge, string* err) {
     return false;
   }
 
-  vector<Node*>::iterator implicit_dep =
-      PreallocateSpace(edge, deps->node_count);
-  for (int i = 0; i < deps->node_count; ++i, ++implicit_dep) {
-    Node* node = deps->nodes[i];
-    *implicit_dep = node;
-    node->AddOutEdge(edge);
+  Node** nodes = deps->nodes;
+  size_t node_count = deps->node_count;
+  edge->inputs_.insert(edge->inputs_.end() - edge->order_only_deps_,
+                       nodes, nodes + node_count);
+  edge->implicit_deps_ += node_count;
+  for (size_t i = 0; i < node_count; ++i) {
+    nodes[i]->AddOutEdge(edge);
   }
   return true;
 }
--- a/src/hash_map.h
+++ b/src/hash_map.h
@@ -19,6 +19,7 @@
 #include <string.h>
 #include "string_piece.h"
 #include "util.h"
+#include "third_party/emhash/hash_table8.hpp"
 
 // MurmurHash2, by Austin Appleby
 static inline
@@ -53,8 +54,6 @@ unsigned int MurmurHash2(const void* key, size_t len) {
   return h;
 }
 
-#include <unordered_map>
-
 namespace std {
 template<>
 struct hash<StringPiece> {
@@ -73,7 +72,7 @@ struct hash<StringPiece> {
 /// mapping StringPiece => Foo*.
 template<typename V>
 struct ExternalStringHashMap {
-  typedef std::unordered_map<StringPiece, V> Type;
+  typedef emhash8::HashMap<StringPiece, V> Type;
 };
 
 #endif // NINJA_MAP_H_
--- a/src/hash_map.h
+++ b/src/hash_map.h
@@ -19,40 +19,9 @@
 #include <string.h>
 #include "string_piece.h"
 #include "util.h"
-#include "third_party/emhash/hash_table8.hpp"
 
-// MurmurHash2, by Austin Appleby
-static inline
-unsigned int MurmurHash2(const void* key, size_t len) {
-  static const unsigned int seed = 0xDECAFBAD;
-  const unsigned int m = 0x5bd1e995;
-  const int r = 24;
-  unsigned int h = seed ^ len;
-  const unsigned char* data = static_cast<const unsigned char*>(key);
-  while (len >= 4) {
-    unsigned int k;
-    memcpy(&k, data, sizeof k);
-    k *= m;
-    k ^= k >> r;
-    k *= m;
-    h *= m;
-    h ^= k;
-    data += 4;
-    len -= 4;
-  }
-  switch (len) {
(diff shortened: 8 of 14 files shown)


digit-google:
Thanks, cool ideas in there. While I didn't look into it for very long, a few remarks:

- When adding new classes, or new metthods, provide unit-tests for them.

- Please use a "_" suffix when adding new class members (e.g. `ins_`, `outs_`, `validations_` instead of `ins`, `outs`, `validations`).

- Put third-party headers under `src/third_party/xxx` to make their origin clearer. Add a `README.ninja` file as well indicating the original URL, SPDX License tag, etc.. If patches were applied, there must be some way to show that to make future updates easier (e.g. Chromium usually provides a patches/ sub-directory that contains the patches applied to the local copy).

- Separating these into multiple PRs will help them get accepted quicker.

- `StringPiece::Persist()` should be `Arena::PersistStringPiece()`. The `StringPiece` class should not know or depend on `Arena` at all.

- I don't think the log version needs to be changed if the hash is computed differently. @jhasse, what do you think?

- Make the arena a member of the `State` class, which will remove 90% of the code you added to send its pointer to various layers of the code, and make everything much easier to understand. It looks like there are no benefit for using an `Arena` whose lifecycle exceeds that of `State`, but let me know if I'm wrong.

- Would appreciate measurements related to peak RAM usage before/after the patch is applied for large build graphs. I'll try to get these numbers myself for the Fuchsia build (which is about 10x larger than the Chromium one) next week.

nico:
I was about to hit merge when I saw digit's comments. This is all small self-contained commits; I thought the PR is in pretty good shape. (In particular, I don't think it necessarily needs to be split. Splitting it wouldn't hurt either ofc, but it's more work, and I'd rather merge this in this form that not merging it due to us not merging it over review back-and-forth.)

The third_party suggestion is good.

I think a hash change does need a manifest version bump.

If the arena `State` suggestion works out, that's also a good suggestion.

Sending the hash map mingw fix upstream would be nice (but not a blocker).

Thanks for the PR!

digit-google:
Oh, and this fails to build because I didn't add `arena.cc` to either `configure.py` or `CMakeLists.txt` !!

src/arena.h
```diff
@@ -0,0 +1,60 @@
...
+//
+// The arena starts by allocating a single 4 kB block, and then increases by
+// 50% every time it needs a new block. This gives O(1) calls to malloc.
+
+struct Arena {
+ public:
+  static constexpr size_t kAlignTarget = sizeof(uint64_t);
+
+  char* Alloc(size_t num_bytes) {
+    num_bytes = (num_bytes + kAlignTarget) & ~kAlignTarget;
```
digit-google, line 39: nit: use `(num_bytes  + kAlignTarget - 1) & ~kAlignTarget)` to avoid over-allocating memory.

src/arena.h
```diff
@@ -0,0 +1,60 @@
...
+// All pointers returned by Alloc() are valid until the arena is destroyed,
+// at which point everything is deallocated all at once. No destructors
+// are run.
+//
+// The arena starts by allocating a single 4 kB block, and then increases by
+// 50% every time it needs a new block. This gives O(1) calls to malloc.
+
+struct Arena {
+ public:
+  static constexpr size_t kAlignTarget = sizeof(uint64_t);
```
digit-google, line 36: Regretabbly a `static constexpr size_t` member will need an empty declaration in the corresponding .cc file, or some compilers will complain in debug mode. But this definition seems to only be used inside of Alloc(), do you really it, i.e. could you turn that into a simple function?

src/arena.cc
```diff
@@ -0,0 +1,35 @@
...
+#include "arena.h"
+
+#include <algorithm>
+
+char* Arena::AllocSlowPath(size_t num_bytes)
+{
+  size_t to_allocate = std::max(next_size_, num_bytes);
+
+  Block new_block;
+  new_block.mem.reset(new char[to_allocate]);
```
digit-google, line 24: nit: Used an aligned new here since there is technically no guarantee that this will happen (yes, I know most allocators would use size_t as a minimum). For now the arena is only used to store character sequences so this doesn't matter, but this could become problematic if it is used later to allocate other things.

src/arena.cc
```diff
@@ -0,0 +1,35 @@
...
+
+  char *ret = new_block.mem.get();
+  cur_ptr_ = new_block.mem.get() + num_bytes;
+  cur_end_ = new_block.mem.get() + to_allocate;
```
digit-google, line 28: I am not sure the `Block` type serves any useful purpose. What would you think about using a simple `std::vector<std::unique_ptr<char[]>>` instead? The code can also be simplified with something like:

```
blocks_.emplace_back(new char [to_allocate]);
char* last_block = blocks_.back().get();
cur_ptr_ = last_block + num_bytes;
cur_end_ = last_block + to_allocate;
```

src/arena.h
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+ private:
+  struct Block {
+    std::unique_ptr<char[]> mem;
+  };
+
+  char* AllocSlowPath(size_t num_bytes);
+
+  std::vector<Block> blocks;
```
digit-google, line 56: nit: Call this `blocks_` and add a unit-test.

src/eval_env.cc
```diff
@@ -102,23 +102,20 @@ string EvalString::Evaluate(Env* env) const {
...
 }
 
 void EvalString::AddText(StringPiece text) {
-  // Add it to the end of an existing RAW token if possible.
-  if (!parsed_.empty() && parsed_.back().second == RAW) {
-    parsed_.back().first.append(text.str_, text.len_);
-  } else {
-    parsed_.push_back(make_pair(text.AsString(), RAW));
-  }
+  parsed_.push_back(make_pair(text, RAW));
```
digit-google, line 132: nit: can you use `std::make_pair()`, we'd like to get away from `using std`.

src/eval_env.cc
```diff
@@ -112,35 +116,65 @@ string EvalString::Evaluate(Env* env) const {
...
-    bool special = (i->second == SPECIAL);
-    if (special)
-      result.append("${");
-    result.append(i->first.begin(), i->first.end());
-    if (special)
-      result.append("}");
+  if (parsed_.empty() && !single_token_.empty()) {
+    result.append(single_token_.begin(), single_token_.end());
+  } else {
+    for (TokenList::const_iterator i = parsed_.begin();
```
digit-google, line 169: can you convert this to a C++11 for-range loop? (e.g. `for (const auto& pair : parsed_) { ... }`)

src/eval_env.cc
```diff
@@ -112,35 +116,65 @@ string EvalString::Evaluate(Env* env) const {
...
+      // our single_token_ optimization and need to push everything
+      // onto the vector.
+      parsed_.push_back(make_pair(single_token_, RAW));
+    } else {
+      // This is the first (nonempty) token, so we don't need to
+      // allocate anything on the vector (yet).
+      single_token_ = text;
+      return;
+    }
+  }
```
digit-google, line 131: This  is very neat, thank you so much :-)

sesse:
I think everything should be taken care of now, short of the alignment issue.

For the memory usage, I wonder if we should consider making a short-string optimization to StringPiece()? It should be possible to have a 15-byte string inline, without allocating anything on the arena. We could then also possibly bring back the concatenation optimization in EvalString, so that if you do AddString("foo"); AddString("bar"); you get a single short-string StringPiece with 6 bytes in them.

src/arena.cc
digit-google: Good point. In theory you should be able to use a dedicated type with `alignas()` specifier, and use this in a `new AlignedType[(to_allocate + sizeof(AlignedType) - 1) & ~sizeof(AlignedType)])`, but this is getting ugly, and frankly there is no need for this feature here. What do you think about dropping the alignment requirement entirely, and just call this `StringPieceArena` for clarity?

digit-google:
I think `StringPiece` is really designed to be a view over content (just like `std::string_view`). What you describe would be a completely different type, but that's a interesting idea. Maybe for another PR?

Since we're talking about StringPiece / string lookup performance, the Android Ninja fork has introduced a HashedStringView type that embeds a (pointer + 32-bit size + 32-bit hash) to speed up hash table lookups considerably (with a corresponding HashedString type, which is an std::string + 32-bit hash).

I had ported the idea on top of my fork some time ago, and usint this improved things noticeably (between 5% and 11% for no-op Ninja builds, depending on the build graph). This could be another way to improve speed, at the cost of *extra complexity* though, and historically Ninja maintainers have been very reluctant of such changes.

It would be nice to know if @nico and @jhasse would be ok with this before trying to implement these schemes.

sesse:
I looked a bit more at this; the issue isn't that we have a lot of short stirngs (we don't). it is simply that EvalString doesn't need to live past the end of ManifestParser::ParseDefault(). So we could simply have a small arena that lives only in ManifestParser, and is cleared (with memory available for reuse) at the end of ParseEdge(). It would be a little more complex, but we would probably get rid of all of the memory bloat. What do you think?

sesse:
I see some autobuilders are failing, too:

 - Linux / Fedora complains that rapidhash.h has trailing spaces. Should I fix this?
 - macOS seems to build in C++03 mode, so e.g. constexpr is not allowed. This seems to conflict with the desire to use C++11 range-based for loops in this review?
 - Linux / build complains about spelling errors in third-party code. What about this one?

digit-google:
Thanks a ton @sesse. I have tried your latest patch with a moderate Fuchsia build plan (e.g. `fx set core.x64 --with //bundles/tests`) and there is a great improvement in no-op time (e.g. from 11.6s to 10.8s, i.e. **-7%**). On the other hand, peal RAM usage (measured with `/usr/bin/time`) goes from 1.7 GiB to 2.4 GiB which really considerable (+40%).

I wanted to see how much the non-arena related improvements impacted performance, and after a small rebase, I got a newer version of your patchset (see attached patch) that actually runs slightly faster, without increasing memory at all. So it looks like the arena is not helping after all.

[0001-Remove-arena-from-pr-2519-commit-790f571.patch.zip](https://github.com/user-attachments/files/17618292/0001-Remove-arena-from-pr-2519-commit-790f571.patch.zip)

digit-google:
The MacOS compilation issue seems to be a bug, this should definitely compile as C++11. I think this is entirely unrelated to this Cl. This is bad :-(

For the spelling issue, the codespell invocation in `.github/workflows/linux.yml` needs to be updated to add a command-line option, probably something like `--skip="src/third_party/*"` or similar.

For linting, `misc/ci.py` needs to be modified to skip `src/third_party` as well.

sesse:
I definitely see differences between the arena and non-arena; there's a measurement there right at the first patch. But like I said, maybe the arena can do with a smaller scope/lifetime.

We can review the non-arena parts first, and then come back to it after the other stuff is in? I'm not _that_ keen on all the re-measuring to get the commit messages right, though.

digit-google:
Looking at the MacOS issue, it looks like the `CMakeLists.txt` only forces C++11 when compiling `libninja` sources. For anything else, e.g. tests and binaries, the compiler's defaults are being used. I guess adding `-DCMAKE_CXX_STANDARD=11` in the CMake configuration command might help).

sesse:
I took out the arena (I intend to re-measure it after we have all of the other stuff in). It was my own rebase, but I looked to yours for confirmation in a couple of places, so that was useful; thanks.

digit-google:
Arrgh, modifying `misc/ci.py` is quite easy, but it seems there is no way to tell codespell to skip a given directory, only filename globs are supported in the `skip:` section (or the `--skip` command-line argument).

sesse:
I can always include a separate patch to just fix the spelling mistakes…

digit-google:
Whatever gets accepted by @jhasse and @nico is ok with me :-)

I have uploaded PR #2521 and PR #2522 for the other issues. I am ok with them being integrated in this PR if the upstream reviewers are ok with that (I just prefer to have a series of small commits to help rebase my fork, and others, on top of upstream).

digit-google:
For the spelling mistakes, what about updating `.github/workflows/linux.yml` to add a few extra words to the ignore list instead? Patching third-party code for something so trivial is ... really annoying imho.

sesse:
I made a prototype of using the arena in a more limited lifetime (most values would be cleared in ParseEdge(), a few have to live past that). On top of this patch series, it took time from 4.300 to 3.859 seconds (~11% win), and memory from 876 to 899 MB, which should hopefully be bearable.

I would assume it's more on platforms with weak malloc (e.g. macOS), but I haven't tested.

src/arena.cc
sesse: The arena is taken out now, so we'll deal with it in a separate PR if we get there.

digit-google:
Can you share this prototype? I'd like to benchmark it against my build plans.

sesse:
Sure, tip of https://github.com/sesse/ninja/tree/arena has it.

sesse:
> For the spelling errors / trailing whitespace I would create a PR on the upstream project.

I doubt you can fix everything anyway; one of them was Windows line endings, and I doubt rapidhash would take in such a change.

digit-google:
Thanks a ton @sesse, this is great, here are my benchmarking results for the following versions:

- `ninja-upstream`: Unmodified upstream Ninja sources
- `ninja-sesse1`: Your original patchset with the initial arena implementation
- `ninja-sesse1-no-arena`: The same without the arena (my version)
- `ninja-sesse3`: The version you just uploaded as commit a0f10b9, which is currently the `master` commit in your repo.
- `ninja-sesse3-arena`: The wip version that is the `arena` commit in you repo right now (777574b5d36d366c3a37f57b90b052ed51289916):

First, regarding performance:

```shell
$ ARGS="-C out/default --quiet -n"
$ hyperfine --runs=5 "/tmp/ninja-upstream $ARGS" "/tmp/ninja-sesse1 $ARGS" "/tmp/ninja-sesse1-no-arena $ARGS" "/tmp/ninja-sesse3 $ARGS" "/tmp/ninja-sesse3-arena $ARGS"
Benchmark 1: /tmp/ninja-upstream -C out/default -n --quiet
  Time (mean ± σ):     11.448 s ±  0.062 s    [User: 6.813 s, System: 4.612 s]
  Range (min … max):   11.397 s … 11.550 s    5 runs
 
Benchmark 2: /tmp/ninja-sesse1 -C out/default -n --quiet
  Time (mean ± σ):     10.999 s ±  0.054 s    [User: 6.099 s, System: 4.877 s]
  Range (min … max):   10.921 s … 11.061 s    5 runs
 
Benchmark 3: /tmp/ninja-sesse1-no-arena -C out/default -n --quiet
  Time (mean ± σ):     10.663 s ±  0.025 s    [User: 6.101 s, System: 4.537 s]
  Range (min … max):   10.619 s … 10.682 s    5 runs
 
Benchmark 4: /tmp/ninja-sesse3 -C out/default -n --quiet
  Time (mean ± σ):     10.586 s ±  0.032 s    [User: 6.076 s, System: 4.485 s]
  Range (min … max):   10.561 s … 10.642 s    5 runs
 
  Warning: The first benchmarking run for this command was significantly slower than the rest (10.642 s). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.
 
Benchmark 5: /tmp/ninja-sesse3-arena -C out/default -n --quiet
  Time (mean ± σ):     10.820 s ±  0.040 s    [User: 6.160 s, System: 4.636 s]
  Range (min … max):   10.779 s … 10.869 s    5 runs
 
Summary
  /tmp/ninja-sesse3 -C out/default -n --quiet ran
    1.01 ± 0.00 times faster than /tmp/ninja-sesse1-no-arena -C out/default -n --quiet
    1.02 ± 0.00 times faster than /tmp/ninja-sesse3-arena -C out/default -n --quiet
    1.04 ± 0.01 times faster than /tmp/ninja-sesse1 -C out/default -n --quiet
    1.08 ± 0.01 times faster than /tmp/ninja-upstream -C out/default -n --quiet
```

Second, regarding peak RAM usage:

```shell
$ /usr/bin/time -f%M /tmp/ninja-upstream $ARGS
1808180

$ /usr/bin/time -f%M /tmp/ninja-sesse3 $ARGS
1803832

$ /usr/bin/time -f%M /tmp/ninja-sesse3-arena $ARGS
1918924
```

Conclusion: for this specific build plan, the current patchset at `a0f10b9` is the best. Surprisingly the arena is still slightly slower here, but the memory increase has been drastically reduced. I am now curious about numbers for your build plan :)

digit-google:
@tsniatowski : You make an excellent point. The current patchset contains the log version change to 7 and seems to be a great improvement, I say let's ship it!

sesse:
I'm surprised you don't win more than 8%, but I guess perhaps your time is dominated by something else than parsing? It's impossible to say for sure without seeing a profile, although 4.6 seconds system time hints at some heavy OS involvement (stat-ing, perhaps?). (Also, how are you running these against the same build directory without having the problem of the hash changing and getting a full rebuild?)

In any case, the arena is now not part of the PR, so we can discuss that separately, I believe.

jhasse:
> one of them was Windows line endings, and I doubt rapidhash would take in such a change.

You don't know until you try :)

Where do the changes in lexer.cc come from?

Technically (unlike with e.g. the zlib license) the MIT license is infectious, meaning that we would have to distribute the copyright and the license terms with binary distributions of ninja after this PR. 99% of people aren't aware of this and break this rule all the time though, not sure if it really matters.
Also not sure about BSD-2-Clause.

sesse:
I don't know what changed lexer.cc, I guess something in the build system does? I haven't modified it by hand. :-)

src/deps_log.cc
```diff
@@ -206,6 +204,7 @@ LoadStatus DepsLog::Load(const string& path, State* state, string* err) {
       read_failed = true;
       break;
     }
+    offset += size + 4;
```
jhasse, line 207: why 4?

jhasse:
> I don't know what changed lexer.cc, I guess something in the build system does? I haven't modified it by hand. :-)

Ah yes, I think building via Python changes it in-place using re2c. Can you remove that change from your commit? Only needed when modifying lexer.in.cc.

sesse:
I don't really trust hyperfine's statistics, but since you wanted comparative measurements:

* `ninja-old` is current master (089af0065cf00d3b3137b7e666a949379051ddf4)
* `ninja` is this PR (a0f10b93c20a0987708ddb0abd936bf152c085d5)

```
Benchmark 1: ./ninja-old -C ~/chromium/src/out/Default -n --quiet
  Time (mean ± σ):      3.097 s ±  0.010 s    [User: 2.745 s, System: 0.351 s]
  Range (min … max):    3.089 s …  3.113 s    5 runs
 
Benchmark 2: ./ninja -C ~/chromium/src/out/Default -n --quiet
  Time (mean ± σ):      2.122 s ±  0.009 s    [User: 1.730 s, System: 0.391 s]
  Range (min … max):    2.115 s …  2.136 s    5 runs
 
Summary
  ./ninja -C ~/chromium/src/out/Default -n --quiet ran
    1.46 ± 0.01 times faster than ./ninja-old -C ~/chromium/src/out/Default -n --quiet
```

This is on a 5950X (Zen 3), with a fairly normal NVMe SSD and Debian unstable.

sesse:
I took out the lexer.cc changes.

src/deps_log.cc
sesse: fread() reads first 4 bytes (line 196), then later `size` bytes.

jhasse: Ah right. Can you change this to `offset += sizeof(size) + size;` and also replace the 4 above to make this clearer? Maybe even change `unsigned size;` to `uint32_t size;`.

sesse: Done.

jhasse:
Can you benchmark only the hasmap change against MSVC's std::unordered_map and libc++ (e.g. on macOS)?

sesse:
I haven't really had a Windows machine since 2001 or so, so that's a bit tricky :-) I have a Mac at work, so I can make the test there next week. Or I can probably make a test with libc++ on Linux (with Clang) if that works?

jhasse:
Yes, would be even more interesting to have a direct comparison on the same system between libstdc++ and libc++ :)

sesse:
Linux, still 5950X:

```
Benchmark 1: ./ninja-libstdc++ -C ~/chromium/src/out/Default -n --quiet
  Time (mean ± σ):      5.262 s ±  0.095 s    [User: 4.153 s, System: 1.108 s]
  Range (min … max):    5.094 s …  5.317 s    5 runs
 
Benchmark 2: ./ninja-libc++ -C ~/chromium/src/out/Default -n --quiet
  Time (mean ± σ):      5.192 s ±  0.101 s    [User: 4.097 s, System: 1.093 s]
  Range (min … max):    5.080 s …  5.275 s    5 runs
 
Benchmark 3: ./ninja-libstdc++-emhash8 -C ~/chromium/src/out/Default -n --quiet
  Time (mean ± σ):      4.913 s ±  0.066 s    [User: 3.808 s, System: 1.104 s]
  Range (min … max):    4.800 s …  4.973 s    5 runs
 
Benchmark 4: ./ninja-libc++-emhash8 -C ~/chromium/src/out/Default -n --quiet
  Time (mean ± σ):      4.815 s ±  0.101 s    [User: 3.720 s, System: 1.094 s]
  Range (min … max):    4.752 s …  4.990 s    5 runs
 
Summary
  ./ninja-libc++-emhash8 -C ~/chromium/src/out/Default -n --quiet ran
    1.02 ± 0.03 times faster than ./ninja-libstdc++-emhash8 -C ~/chromium/src/out/Default -n --quiet
    1.08 ± 0.03 times faster than ./ninja-libc++ -C ~/chromium/src/out/Default -n --quiet
    1.09 ± 0.03 times faster than ./ninja-libstdc++ -C ~/chromium/src/out/Default -n --quiet
```

sesse:
@jhasse Is there anything missing here for this to be merged?

jhasse:
Not from my side, I just remembered that I hate Windows anyway :D
I'm also not using ninja on such big projects, so I'm personally completely unaffected by this ...

Maybe @nico and @digit-google have some comments left though :) Let's wait a bit, then I'll merge.

sesse:
Great :-) I still have the arena PR, but we'll deal with that after this.

digit-google:
This is a great PR, please ship it asap :-)

For the record, I have added below my own Linux benchmarks for three versions of the Ninja sources, and several compiler + c++ runtime and memory allocator combos, run against a small Fuchsia build plan, where:

- `ninja-upstream-*` is for current Ninja sources.
- `ninja-sesse-*` are the changes in the PR *without* the emhash8 changes
- `ninja-emhash8-*` are everything in this PR.

This shows that `ninja-sesse-*` is consistently better than `ninja-upstream-*`, and `ninja-emhash8-*` is consistently better than `ninja-sesse-*`, even when using custom allocators.

```
Benchmark 1: /tmp/ninja-upstream-clang16-glibc-stdlibc++ -C out/default -n --quiet
  Time (mean ± σ):      4.862 s ±  0.022 s    [User: 2.727 s, System: 2.124 s]
  Range (min … max):    4.834 s …  4.895 s    10 runs
 
Benchmark 2: /tmp/ninja-sesse-clang16-glibc-stdlibc++ -C out/default -n --quiet
  Time (mean ± σ):      4.503 s ±  0.018 s    [User: 2.391 s, System: 2.100 s]
  Range (min … max):    4.465 s …  4.524 s    10 runs
 
Benchmark 3: /tmp/ninja-emhash8-clang16-glibc-libstdc++ -C out/default -n --quiet
  Time (mean ± σ):      4.407 s ±  0.021 s    [User: 2.267 s, System: 2.129 s]
  Range (min … max):    4.358 s …  4.437 s    10 runs
 
Benchmark 4: /tmp/ninja-upstream-clang20-glibc-stdlibc++ -C out/default -n --quiet
  Time (mean ± σ):      4.786 s ±  0.026 s    [User: 2.665 s, System: 2.109 s]
  Range (min … max):    4.725 s …  4.808 s    10 runs
 
Benchmark 5: /tmp/ninja-sesse-clang20-glibc-stdlibc++ -C out/default -n --quiet
  Time (mean ± σ):      4.439 s ±  0.029 s    [User: 2.320 s, System: 2.107 s]
  Range (min … max):    4.398 s …  4.485 s    10 runs
 
Benchmark 6: /tmp/ninja-emhash8-clang20-glibc-libstdc++ -C out/default -n --quiet
  Time (mean ± σ):      4.307 s ±  0.018 s    [User: 2.208 s, System: 2.088 s]
  Range (min … max):    4.277 s …  4.330 s    10 runs
 
Benchmark 7: /tmp/ninja-emhash8-clang20-glibc-libc++ -C out/default -n --quiet
  Time (mean ± σ):      4.319 s ±  0.015 s    [User: 2.200 s, System: 2.107 s]
  Range (min … max):    4.287 s …  4.336 s    10 runs
 
Benchmark 8: /tmp/ninja-upstream-clang20-jemalloc-libc++ -C out/default -n --quiet
  Time (mean ± σ):      4.021 s ±  0.020 s    [User: 2.071 s, System: 1.940 s]
  Range (min … max):    3.988 s …  4.046 s    10 runs
 
Benchmark 9: /tmp/ninja-sesse-clang20-jemalloc-libc++ -C out/default -n --quiet
  Time (mean ± σ):      3.801 s ±  0.019 s    [User: 1.859 s, System: 1.931 s]
  Range (min … max):    3.772 s …  3.837 s    10 runs
 
Benchmark 10: /tmp/ninja-emhash8-clang20-jemalloc-libc++ -C out/default -n --quiet
  Time (mean ± σ):      3.733 s ±  0.015 s    [User: 1.796 s, System: 1.927 s]
  Range (min … max):    3.706 s …  3.754 s    10 runs
 
Benchmark 11: /tmp/ninja-upstream-clang20-rpmalloc-libc++ -C out/default -n --quiet
  Time (mean ± σ):      4.068 s ±  0.022 s    [User: 2.120 s, System: 1.937 s]
  Range (min … max):    4.033 s …  4.111 s    10 runs
 
Benchmark 12: /tmp/ninja-sesse-clang20-rpmalloc-libc++ -C out/default -n --quiet
  Time (mean ± σ):      3.827 s ±  0.011 s    [User: 1.882 s, System: 1.932 s]
  Range (min … max):    3.802 s …  3.843 s    10 runs
 
Benchmark 13: /tmp/ninja-emhash8-clang20-rpmalloc-libc++ -C out/default -n --quiet
  Time (mean ± σ):      3.781 s ±  0.015 s    [User: 1.820 s, System: 1.948 s]
  Range (min … max):    3.759 s …  3.799 s    10 runs
 
Benchmark 14: /tmp/ninja-upstream-gcc13-glibc-stdlibc++ -C out/default -n --quiet
  Time (mean ± σ):      4.823 s ±  0.035 s    [User: 2.666 s, System: 2.145 s]
  Range (min … max):    4.759 s …  4.865 s    10 runs
 
Benchmark 15: /tmp/ninja-sesse-gcc13-glibc-stdlibc++ -C out/default -n --quiet
  Time (mean ± σ):      4.488 s ±  0.021 s    [User: 2.305 s, System: 2.172 s]
  Range (min … max):    4.450 s …  4.513 s    10 runs
 
Benchmark 16: /tmp/ninja-emhash8-gcc13-glibc-libstdc++ -C out/default -n --quiet
  Time (mean ± σ):      4.383 s ±  0.030 s    [User: 2.203 s, System: 2.168 s]
  Range (min … max):    4.341 s …  4.424 s    10 runs
 
Summary
  /tmp/ninja-emhash8-clang20-jemalloc-libc++ -C out/default -n --quiet ran
    1.01 ± 0.01 times faster than /tmp/ninja-emhash8-clang20-rpmalloc-libc++ -C out/default -n --quiet
    1.02 ± 0.01 times faster than /tmp/ninja-sesse-clang20-jemalloc-libc++ -C out/default -n --quiet
    1.03 ± 0.01 times faster than /tmp/ninja-sesse-clang20-rpmalloc-libc++ -C out/default -n --quiet
    1.08 ± 0.01 times faster than /tmp/ninja-upstream-clang20-jemalloc-libc++ -C out/default -n --quiet
    1.09 ± 0.01 times faster than /tmp/ninja-upstream-clang20-rpmalloc-libc++ -C out/default -n --quiet
    1.15 ± 0.01 times faster than /tmp/ninja-emhash8-clang20-glibc-libstdc++ -C out/default -n --quiet
    1.16 ± 0.01 times faster than /tmp/ninja-emhash8-clang20-glibc-libc++ -C out/default -n --quiet
    1.17 ± 0.01 times faster than /tmp/ninja-emhash8-gcc13-glibc-libstdc++ -C out/default -n --quiet
    1.18 ± 0.01 times faster than /tmp/ninja-emhash8-clang16-glibc-libstdc++ -C out/default -n --quiet
    1.19 ± 0.01 times faster than /tmp/ninja-sesse-clang20-glibc-stdlibc++ -C out/default -n --quiet
    1.20 ± 0.01 times faster than /tmp/ninja-sesse-gcc13-glibc-stdlibc++ -C out/default -n --quiet
    1.21 ± 0.01 times faster than /tmp/ninja-sesse-clang16-glibc-stdlibc++ -C out/default -n --quiet
    1.28 ± 0.01 times faster than /tmp/ninja-upstream-clang20-glibc-stdlibc++ -C out/default -n --quiet
    1.29 ± 0.01 times faster than /tmp/ninja-upstream-gcc13-glibc-stdlibc++ -C out/default -n --quiet
    1.30 ± 0.01 times faster than /tmp/ninja-upstream-clang16-glibc-stdlibc++ -C out/default -n --quiet
```

digit-google:
friendly ping for @jhasse .

src/third_party/rapidhash/README.ninja
```diff
@@ -0,0 +1,7 @@
+Description: Very fast, high quality, platform-independent hashing algorithm.
+Version: commit 4a6b2570e868536be84800353efd92c699f37d2c
+URL: https://github.com/Nicoshev/rapidhash
+Copyright: Copyright (C) 2024 Nicolas De Carli, Based on 'wyhash', by Wang Yi <godspeed_china@yeah.net>
+SPDX-License-Identifier: BSD-2-Clause
```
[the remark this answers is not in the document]
sesse: Done.

Issue #2520, opened 2024-11-04
MacOS CI builder uses C++03 by default.


digit-google:
This was found by @sesse in PR #2519, the `CMakeLists.txt` only enforces C++11 for `libninja`. Anything else, i.e. the `ninja` binary and tests are compiled with the compiler's default setting.

It looks like this is still C++03 for the MacOS builder, resulting in surprising compilation failures.
The fix is to enforce C++11 as the minimum standard in the CMake configuration step of the corresponding github workflow file.

PR #2521, opened 2024-11-04
Fix for #2520: Force C++11 for MacOS CI builder.

digit-google:
MacOS CI builders use a compiler which defaults to C++03. Unfortunately, CMakeLists.txt only forces C++11 for libninja, try to force it through CMAKE_CXX_STANDARD.

Note a few interesting points:

- Tests use GoogleTest which officially *requires* C++14. However, it seems that compiling them with C++03 or C++11 doesn´t seem to be an issue on this platform for now.

- To avoid future problems, it might be useful to enforce a separate default of C++14 for test programs.

- GoogleTest plans to switch to C++17 in December 2024 [1],

[1] https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -21,7 +21,7 @@ jobs:
       env:
         MACOSX_DEPLOYMENT_TARGET: 10.15
       run: |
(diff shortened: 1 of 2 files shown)


digit-google:
Oh I see, I'll upload a different PR then, thanks.

PR #2522, opened 2024-11-04
misc/ci.py: Ignore src/third_party directory.

digit-google:
This is necessary to be able to use third-party code that contains trailing spaces and other oddities.

See PR #2519 for context.


jhasse:
Use the `ignores` array above instead :)

I also think this should be part of the PR that adds the directory, on its own it doesn't make any sense.

PR #2523, opened 2024-11-07
Fix memory leak of Rules, use bool phony_ member to identify phony edges

jhasse:
Not sure if this is worth it and what the performance impact is.

### DIFF
--- a/src/eval_env.cc
+++ b/src/eval_env.cc
@@ -31,22 +31,22 @@ void BindingEnv::AddBinding(const string& key, const string& val) {
   bindings_[key] = val;
 }
 
-void BindingEnv::AddRule(const Rule* rule) {
+void BindingEnv::AddRule(std::unique_ptr<const Rule> rule) {
   assert(LookupRuleCurrentScope(rule->name()) == NULL);
-  rules_[rule->name()] = rule;
+  rules_[rule->name()] = std::move(rule);
 }
 
 const Rule* BindingEnv::LookupRuleCurrentScope(const string& rule_name) {
-  map<string, const Rule*>::iterator i = rules_.find(rule_name);
+  auto i = rules_.find(rule_name);
   if (i == rules_.end())
     return NULL;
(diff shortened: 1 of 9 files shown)


src/eval_env.cc
```diff
@@ -63,6 +63,16 @@ const EvalString* Rule::GetBinding(const string& key) const {
...
 }
 
+std::unique_ptr<Rule> Rule::Phony() {
+  auto rule = std::unique_ptr<Rule>(new Rule("phony"));
+  rule->phony_ = true;
+  return rule;
+}
+
+bool Rule::IsPhony() const {
+  return phony_;
```
digit-google, line 73: nit: Can we just store the address of this Rule in `State::phony_rule_` and implement `State::IsPhony(const Rule* rule) { return rule == phony_rule_; }` instead? This would avoid adding 8 bytes to each Rule instance due to padding.

digit-google:
Apart from my nit, this LGTM.

src/eval_env.cc
jhasse: you mean like it was before? That doesn't work anymore since then the Rule would be deleted twice in `std::map<std::string, std::unique_ptr<const Rule>> rules_;`, wouldn't it?

digit-google: Note necessarily with something like:

```
// state.h
struct State {
  ...
  const Rule* phony_rule() const { return phony_rule_; }
  
  Rule* phony_rule_ = nullptr;
  ...
};

// state.cc
State::State() {
  ...
  // Add the phony rule to it.
  phony_rule_ = new Rule("phony");
  bindings_.AddRule(phony_rule_);  // passes ownership to bindings_.
}
```

The caveat is that you need a State pointer whenever your want to implement `rule->IsPhony()`, but this is a minor modification to the rest of the code.

Frankly, this could go in a different PR. The memory leak fix is more important, so feel free to submit if you want.

PR #2524, opened 2024-11-08
Add include target option to inputs tool

freand76:
This feature can be used if one would like to to a reverse search on what 
targets that will be built by ninja (for example from the file listing of a git commit)

Instead of just listing all the inputs (that ninja -t inputs do) it will also list the target that use that specific input.

ninja -t inputs -i targetA targetB targetC

Ninja will then output:
targetA:file1.c
targetB:file1.c
targetC:file1.c
targetB:file2.c
targetC:file2.c

Using the list above:
If we have a git commit that changes file1 we know that we want to **test** everything related to targetA, targetB, targetC.

on the other hand if we have a git commit that changes file2 we only have to **test** everything related to targetB and targetC.

We use this feature a lot to get the correct (reduce the) test scope of a certain git commit.

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -770,16 +770,18 @@ int NinjaMain::ToolInputs(const Options* options, int argc, char* argv[]) {
   bool print0 = false;
   bool shell_escape = true;
   bool dependency_order = false;
+  bool include_target = false;
 
   optind = 1;
   int opt;
   const option kLongOptions[] = { { "help", no_argument, NULL, 'h' },
                                   { "no-shell-escape", no_argument, NULL, 'E' },
                                   { "print0", no_argument, NULL, '0' },
+                                  { "include-target", no_argument, NULL, 'i' },
                                   { "dependency-order", no_argument, NULL,
                                     'd' },
                                   { NULL, 0, NULL, 0 } };
-  while ((opt = getopt_long(argc, argv, "h0Ed", kLongOptions, NULL)) != -1) {
+  while ((opt = getopt_long(argc, argv, "h0Edi", kLongOptions, NULL)) != -1) {
     switch (opt) {
     case 'd':
       dependency_order = true;
@@ -789,6 +791,9 @@ int NinjaMain::ToolInputs(const Options* options, int argc, char* argv[]) {
       break;
     case '0':
       print0 = true;
+      break;
+    case 'i':
+      include_target = true;
       break;
     case 'h':
     default:
@@ -804,6 +809,7 @@ int NinjaMain::ToolInputs(const Options* options, int argc, char* argv[]) {
 "  -0, --print0            Use \\0, instead of \\n as a line terminator.\n"
 "  -E, --no-shell-escape   Do not shell escape the result.\n"
 "  -d, --dependency-order  Sort results by dependency order.\n"
+"  -i, --include-target    Add target before result in every line, separated by the : character.\n"
       );
       // clang-format on
       return 1;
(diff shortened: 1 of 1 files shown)


freand76:
Please feel free to ask if you need more information about why this feature might be useful by ninja users.

src/ninja.cc
```diff
@@ -819,23 +825,45 @@ int NinjaMain::ToolInputs(const Options* options, int argc, char* argv[]) {
...
+    if (print0) {
+      for (const std::string& input : inputs) {
+        fwrite(input.c_str(), input.size(), 1, stdout);
+        fputc('\0', stdout);
+      }
+      fflush(stdout);
+    } else {
+      for (const std::string& input : inputs)
+        puts(input.c_str());
+    }
```
digit-google, line 866: Avoid code duplication by defining a local lambda that captures `shell_escape` and `dependency_order`, takes an `InputsCollector&` and ` const char* prefix` as input and does the printing. You'll be able to call it from the two branches of your `if (include_target)` condition.

src/ninja.cc
```diff
@@ -804,6 +809,7 @@ int NinjaMain::ToolInputs(const Options* options, int argc, char* argv[]) {
 "  -0, --print0            Use \\0, instead of \\n as a line terminator.\n"
 "  -E, --no-shell-escape   Do not shell escape the result.\n"
 "  -d, --dependency-order  Sort results by dependency order.\n"
+"  -i, --include-target    Add target before result in every line, separated by the : character.\n"
```
digit-google, line 812: If I understand correctly, this is a convenience feature to avoid calling `ninja -t inputs <target>` in a loop, by reusing the same build plan loaded from memory.

However, it *drastically* changes the behavior of this tool as it now performs one graph traversal per target, instead of a single one, and changes the output considerably: in addition to the prefix, actual input paths will be duplicated when they are used by more than one target, which does not happen when not using this flag.

I think the documentation should reflect that *much* more clearly.

digit-google: I thus suggest finding a more descriptive name than `-include-target`. for example `--separate-target-inputs`, wdyt?

digit-google: Finally, this needs a regression test in `misc/output_test.py`. Please add the correspond test cases to `test_tool_inputs()` in that file.

digit-google: Also, are we sure that target names are never going to include a `':` in their path? Maybe use a TAB character as the separator instead, at least these are forbidden from Ninja build plans at the moment. Or allow the user to provide an alternative separator here? Just thinking out loud, let me know what you think.

digit-google: Yeah, a new tool seems to make sense. What about `multi-inputs` to clarify that it provides different sets of inputs?

Regarding flags/features:

- Make `--dependency-order` the default behavior. Historically, this flag was added to `inputs` to make the tool faster (this is important for me too :-)). The initial implementation (before `InputsCollector`) was actually slightly broken and would collect duplicate targets, so `std::sort()` + `std::unique()` calls were needed to get rid of them. I don't see why keeping the output sorted in the new tool would be useful, though you could add a `--sort` option if you really like it.

- Do not implement escaping, but do support `--print0` or `--format=print0` for scripts. It is trivial to implement, and it's a better alternative imho.

Regarding the delimiter, I think TAB is a good default choice, with the ability to use a different one with a flag. Another alternative is to provide a `--format=json` flag to output a JSON object that maps target paths to list of inputs instead (but the result will be significantly slower to generate, and to consume).

freand76: New PR created https://github.com/ninja-build/ninja/pull/2527

freand76:
Dropping this in favor of #2527

PR #2527, opened 2024-11-12
Add multi-inputs tool

freand76:
The 'multi-inputs' option will list all <target> + <inputs> for the given targets.

Run:
ninja -t multi-inputs target1 target2 target3

Ninja will then output:
target1 input_x
target1 input_y
target2 input_x
target2 input_z
target3 input_y

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -265,6 +265,47 @@ output files are out of date.
 rebuild those targets.
 _Available since Ninja 1.11._
 
+`multi-inputs`:: print one or more sets of inputs required to build targets.
+Each line will consist of a target, a delimiter, an input and a terminator character.
+The list produced by the tool can be helpful if one would like to know which targets
+that are affected by a certain input.
++
+The output will be a series of lines with the following elements:
++
+----
+<target> <delimiter> <input> <terminator>
+----
++
+The default `<delimiter>` is a single TAB character.
+The delimiter can be modified to any string using the `--delimiter` argument.
++
+The default `<terminator>` is a line terminator (i.e. `\n` on Posix and `\r\n` on Windows).
+The terminator can be changed to `\0` by using the `--print0` argument.
++
+----
+----
++
+Example usage of the `multi-inputs` tool:
++
+----
+ninja -t multi-inputs target1 target2 target3
+----
++
+Example of produced output from the `multi-inputs` tool:
++
+----
+target1 file1.c
+target2 file1.c
+target2 file2.c
+target3 file1.c
+target3 file2.c
+target3 file3.c
+----
++
+_Note that a given input may appear for several targets if it is used by more
+than one targets._
+_Available since Ninja 1.13._
+
 `clean`:: remove built files. By default, it removes all built files
 except for those created by the generator.  Adding the `-g` flag also
 removes built files created by the generator (see <<ref_rule,the rule
--- a/misc/output_test.py
+++ b/misc/output_test.py
@@ -449,6 +449,40 @@ options:
             self.assertEqual(expected, actual)
 
 
+    def test_tool_multi_inputs(self) -> None:
+        plan = '''
+rule cat
+  command = cat $in $out
+build out1 : cat in1
+build out2 : cat in1 in2
+build out3 : cat in1 in2 in3
+'''
+        self.assertEqual(run(plan, flags='-t multi-inputs out1'),
+'''out1<TAB>in1
+'''.replace("<TAB>", "\t"))
+
+        self.assertEqual(run(plan, flags='-t multi-inputs out1 out2 out3'),
+'''out1<TAB>in1
+out2<TAB>in1
+out2<TAB>in2
+out3<TAB>in1
+out3<TAB>in2
+out3<TAB>in3
+'''.replace("<TAB>", "\t"))
+
+        self.assertEqual(run(plan, flags='-t multi-inputs -d: out1'),
+'''out1:in1
+''')
+
+        self.assertEqual(
+          run(
+            plan,
+            flags='-t multi-inputs -d, --print0 out1 out2'
+          ),
+          '''out1,in1\0out2,in1\0out2,in2\0'''
+        )
+
+
     def test_explain_output(self):
         b = BuildDir('''\
             build .FORCE: phony
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -130,6 +130,7 @@ struct NinjaMain : public BuildLogUser {
   int ToolTargets(const Options* options, int argc, char* argv[]);
   int ToolCommands(const Options* options, int argc, char* argv[]);
   int ToolInputs(const Options* options, int argc, char* argv[]);
+  int ToolMultiInputs(const Options* options, int argc, char* argv[]);
   int ToolClean(const Options* options, int argc, char* argv[]);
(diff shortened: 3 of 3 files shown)


freand76:
Replaces PR https://github.com/ninja-build/ninja/pull/2524

doc/manual.asciidoc
```diff
@@ -265,6 +265,11 @@ output files are out of date.
 rebuild those targets.
 _Available since Ninja 1.11._
 
+`multi-inputs`:: given a list of targets, print a list of all inputs used
+to rebuild those targets. Each line will constst of a target, a field separator
+and a file path. The list produced by the tool can be helpful if one
+would like to know which targets that are affected by a certain file.
```
digit-google, line 271: nit: Please clarify why this is different from `inputs` here. In particular explain that this prints different sets of inputs, one per target, and that a given input may appear several times in the output if it is used by more than one target. Maybe add one example as you did in your commit message for clarity?

Also s/constst/consist/ :-)

src/ninja.cc
```diff
@@ -845,6 +846,76 @@ int NinjaMain::ToolInputs(const Options* options, int argc, char* argv[]) {
...
+    case '0':
+      print0 = true;
+      break;
+    case 'h':
+    default:
+      // clang-format off
+      printf(
+"Usage '-t multiinputs [options] [targets]\n"
+"\n"
+"List all target+inputs used for a set of targets, sorted in dependency order.\n"
```
digit-google, line 880: nit: clarify a bit what the expected output is here too. In particular desscribe the output format as a series of lines, each one like `<target> <delimiter> <input> <terminator>`

src/ninja.cc
```diff
@@ -845,6 +846,76 @@ int NinjaMain::ToolInputs(const Options* options, int argc, char* argv[]) {
...
+
+    collector.VisitNode(node);
+    std::vector<std::string> inputs = collector.GetInputsAsStrings();
+
+    for (const std::string& input : inputs) {
+      printf("%s%s%s", node->path().c_str(), fieldseparator, input.c_str());
+      if (print0) {
+        fputc('\0', stdout);
+      } else {
+        fputc('\n', stdout);
```
digit-google, line 911: nit: I recommend you use a `char terminator = '\n'` local variable, change its value to '\0' in the option processor  for `--print0` and just use `fputc(terminator, stdout)` here for simplicity.

src/ninja.cc
```diff
@@ -845,6 +846,76 @@ int NinjaMain::ToolInputs(const Options* options, int argc, char* argv[]) {
...
+  // the tool, i.e. "inputs".
+  argc++;
+  argv--;
+
+  bool print0 = false;
+
+  optind = 1;
+  int opt;
+  const char* defaultfieldseparator = "\t";
+  const char* fieldseparator = defaultfieldseparator;
```
digit-google, line 860: the `defaultfieldseparator` variable doesn't seem to be useful here. Also please use snake_case for naming variables, they'll be much more readable :-)

misc/output_test.py
```diff
@@ -449,6 +449,40 @@ def test_tool_compdb_targets(self) -> None:
...
+    def test_tool_multi_inputs(self) -> None:
+        plan = '''
+rule cat
+  command = cat $in $out
+build out1 : cat in1
+build out2 : cat in1 in2
+build out3 : cat in1 in2 in3
+'''
+        self.assertEqual(run(plan, flags='-t multi-inputs out1'),
+'''out1	in1
```
digit-google, line 461: Can you avoid using embedded tabs for readability? For example, replace them in with "<TAB>" in the result before the comparison, as in:

```
self.assertEqual(run(...).replace("\t", "<TAB>"), ''out1<TAB>in1
''')
```

misc/output_test.py
```diff
@@ -449,6 +449,40 @@ def test_tool_compdb_targets(self) -> None:
...
+        self.assertEqual(run(plan, flags='-t multi-inputs -F: out1'),
+'''out1:in1
+''')
+
+        self.assertEqual(
+          run(
+            plan,
+            flags='-t multi-inputs -F, --print0 out1 out2'
+          ),
+          f'''out1,in1\0out2,in1\0out2,in2\0'''
```
digit-google, line 482: just curious, but why a formatted string here? Does a simple `"out1,in1\0out2,in1\0out2,in2\0"` work?

src/ninja.cc
```diff
@@ -1234,6 +1305,8 @@ const Tool* ChooseTool(const string& tool_name) {
       Tool::RUN_AFTER_LOAD, &NinjaMain::ToolCommands },
     { "inputs", "list all inputs required to rebuild given targets",
       Tool::RUN_AFTER_LOAD, &NinjaMain::ToolInputs},
+    { "multi-inputs", "list all target+inputs required to rebuild given targets",
```
digit-google, line 1308: suggestion: "print one or more sets of inputs required to build targets."

misc/output_test.py
freand76: changed to **replace("<TAB>", "\t")** instead ;-)

misc/output_test.py
freand76: Copy paste error :) same thing is available in test_input_tool ;-)

src/ninja.cc
freand76: Updated... hopefully for the better.

doc/manual.asciidoc
```diff
@@ -265,6 +265,37 @@ output files are out of date.
...
+`multi-inputs`:: print one or more sets of inputs required to build targets.
+Each line will consist of a target, a delimiter, an input and a terminator character.
+The list produced by the tool can be helpful if one would like to know which targets
+that are affected by a certain input.
++
+The output will be a series of lines with the following elements:
++
+----
+<target> <delimiter> <input> <terminator>
+----
```
digit-google, line 277: nit: Add a line stating that by default `<delimiter>` is a single TAB character, and `<terminator>` is a line terminator (i.e. `\n` on Posix, and `\r\n` on Windows), but that flags can be used to change their values.

misc/output_test.py
```diff
@@ -449,6 +449,40 @@ def test_tool_compdb_targets(self) -> None:
...
+        plan = '''
+rule cat
+  command = cat $in $out
+build out1 : cat in1
+build out2 : cat in1 in2
+build out3 : cat in1 in2 in3
+'''
+        self.assertEqual(run(plan, flags='-t multi-inputs out1'),
+'''out1<TAB>in1
+'''.replace("<TAB>", "\t"))
```
digit-google, line 462: Oh, this is actually much nicer than my suggestion, thank you!

digit-google:
Thanks a lot, apart from my last documentation nit, this looks good to me. @jhasse, what do you think?

doc/manual.asciidoc
freand76: Updated

freand76:
> Thanks a lot, apart from my last documentation nit, this looks good to me. @jhasse, what do you think?

Any progress on this? (looking forward to having this in the next release ;-)

jhasse:
doc should mention _Available since Ninja 1.13._

freand76:
> doc should mention _Available since Ninja 1.13._

Of course :-) Fixed and pushed.

PR #2532, opened 2024-11-25
Add an arena for allocating strings.

sesse:
This is the patch we took out during the review of PR #2519, now rebased and re-measured. It's much less memory-bad, but still wins ~12%.

```
Benchmark 1: ./ninja.old -C ~/chromium/src/out/Default content_shell
  Time (mean ± σ):      4.608 s ±  0.014 s    [User: 3.073 s, System: 1.531 s]
  Range (min … max):    4.590 s …  4.634 s    10 runs
 
Benchmark 2: ./ninja -C ~/chromium/src/out/Default content_shell
  Time (mean ± σ):      4.084 s ±  0.009 s    [User: 2.558 s, System: 1.523 s]
  Range (min … max):    4.068 s …  4.097 s    10 runs
 
Summary
  ./ninja -C ~/chromium/src/out/Default content_shell ran
    1.13 ± 0.00 times faster than ./ninja.old -C ~/chromium/src/out/Default content_shell
```

What do you think? Should we still pursue this?


src/manifest_parser.cc
```diff
@@ -152,6 +152,10 @@ bool ManifestParser::ParseRule(string* err) {
       return false;
 
     if (Rule::IsReservedBinding(key)) {
+      for (auto it = value.parsed_.begin(); it != value.parsed_.end(); ++it) {
+        it->first = it->first.PersistOutsideArena();
+      }
+      value.single_token_ = value.single_token_.PersistOutsideArena();
```
digit-google, line 158: Correct me if I'm wrong, but this looks like a memory  leak, since StringPiece is designed to be a non-owning view over string content, the result of PersistOutsideArena() will never be released. This method is probably confusing: returning an _owning_ StringPiece value is surprising and error-prone, using StringPiece::AsString() is much better because the lifetime of an std::string is easy to understand.

sesse: I'm not entirely sure if I understand. Everything allocated on an arena is, in a sense, a memory leak; it's not freed until the arena goes out of scope. But at that point, it really is.

Are you saying that the lifetime of the arena here is unreasonably long?

sesse: Wait, I'm probably confused. I'll need to reacquaint myself with my own code, it has been so long now. :-)

sesse:
I moved so that the EvalStrings used in Rule are now owned by a smaller arena on the BindingEnv instead of being allocated on the heap and leaked.

elliotgoodrich:
I spent a bit of time removing allocations in [trimja](https://github.com/elliotgoodrich/trimja) and the main changes related to `EvalString` were to:

  1. Change `EvalString` so that its puts all segments in a single `std::string` so its full capacity is retained when cleared ([trimja/evalstring.h](https://github.com/elliotgoodrich/trimja/blob/main/src/evalstring.h))
  2. Structure the `ManifestParser` to reuse the same `EvalString` throughout parsing ([`m_storage`](https://github.com/elliotgoodrich/trimja/blob/65c9056d4323b8133dad0bf71a0999722398474b/src/manifestparser.h#L249))

Then the only copies you need to make of `EvalString` is in [`Rule::AddBinding`](https://github.com/ninja-build/ninja/blob/048a68dffbca4055d19b8573908b09b6ec168122/src/eval_env.h#L70), which would be faster as the copy constructor of trimja's `EvalString` only needs 1 allocation and copy of contiguous data. Otherwise, during parsing the capacity is reused in the single `EvalString` and eventually we'll stop any allocations.

Maybe changing the entire `ManifestParser` is too much work, but it looks possible to add an `EvalString` member variable and reuse this?

elliotgoodrich:
I've submitted 1. here https://github.com/ninja-build/ninja/pull/2696

The second part would need a `ReusableStack<EvalString>` that doesn't actually destruct elements when clearing and will reuse them later on.  This will replace the `vector<EvalString>` inside `ManifestParser` to remove the remaining allocations in this area.

PR #2534, opened 2024-11-25
Fix build log leak

digit-google:
This patch fixes a memory leak in BuildLog::Recompact(), when calling `entries_.erase()` for dead outputs would not free the pointed-to LogEntry.

Rather than doing a local fix with an expliciti delete, this changes the `Entries` type to use `std::unique_ptr<LogEntry>` values instead of raw pointers, as this will prevent similar bugs to appear in the future. Other call sites are updated appropriately.

This also contains minor cleanups in separate commits:

- Use default member initialization in `build_log.h`.
- Remove `using namespace std` from `build_log.cc` and `build_log_test.cc`
- Remove one unnecessary string copy per entry when reading the build log.

### DIFF
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -41,8 +41,6 @@
 #define strtoll _strtoi64
 #endif
 
-using namespace std;
-
 // Implementation details:
 // Each run's log appends to the log file.
 // To load, we run through all log entries in series, throwing away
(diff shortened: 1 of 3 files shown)


digit-google:
I understand :-) That's one of the reasons why I decomposed this into a series of small commits, apart from making reviews easier, it minimizes the chances of conflicts, or make them much easier to resolve in practice. Thanks again.

PR #2540, opened 2024-12-03
Propagate supbrocess' exit codes to the ninja exit code

### LINKED ISSUE #1123: Return codes to distinguish when ninja fails to execute the flow or to startup the execution
Currently, on linux, the ninja process exits with status 1 if any kind of error exists.

I propose two different exit statuses, for the ninja process, to determine whether:
1. the flow did not started the execution at all, i.e. ninja file has a syntax error
2. or the flow execution has failed 

The main goal is to simplify the implementation of automated workflows based on ninja files generators.

Adriano

### LINKED ISSUE #1507: Ninja hides the exit status of failing subprocesses

Felixoid:
Now, ninjas could exit with codes 0, 1, and 2.

With these codes, it's nearly impossible to get what went wrong in the subcommands.

This patch propagates the subcommand's exit code to the ninja's exit code.

It fixes #1123, fixes #1507, and technically substitutes the https://github.com/ninja-build/ninja/pull/1805

### DIFF
--- a/misc/output_test.py
+++ b/misc/output_test.py
@@ -152,14 +152,18 @@ class Output(unittest.TestCase):
         '',
     ))
 
-    def _test_expected_error(self, plan: str, flags: T.Optional[str], expected: str):
+    def _test_expected_error(self, plan: str, flags: T.Optional[str],expected: str,
+                             *args, exit_code: T.Optional[int]=None, **kwargs)->None:
         """Run Ninja with a given plan and flags, and verify its cooked output against an expected content.
+        All *args and **kwargs are passed to the `run` function
         """
         actual = ''
-        try:
-          actual = run(plan, flags, print_err_output=False)
-        except subprocess.CalledProcessError as err:
-          actual = err.cooked_output
+        kwargs['print_err_output'] = False
+        with self.assertRaises(subprocess.CalledProcessError) as cm:
+            run(plan, flags, *args,  **kwargs)
+        actual = cm.exception.cooked_output
+        if exit_code is not None:
+            self.assertEqual(cm.exception.returncode, exit_code)
         self.assertEqual(expected, actual)
 
     def test_issue_1418(self) -> None:
@@ -280,6 +284,75 @@ build zoo: touch bar
                 )
             except subprocess.CalledProcessError as err:
                 self.fail("non-zero exit code with: " + err.output)
+
+    def test_pr_2540(self)->None:
+        py = sys.executable
+        plan = f'''\
+rule CUSTOM_COMMAND
+  command = $COMMAND
+
+build 124: CUSTOM_COMMAND
+  COMMAND = {py} -c 'exit(124)'
+
+build 127: CUSTOM_COMMAND
+  COMMAND = {py} -c 'exit(127)'
+
+build 130: CUSTOM_COMMAND
+  COMMAND = {py} -c 'exit(130)'
+
+build 137: CUSTOM_COMMAND
+  COMMAND = {py} -c 'exit(137)'
+
+build success: CUSTOM_COMMAND
+  COMMAND = sleep 0.3; echo success
+'''
+        # Disable colors
+        env = default_env.copy()
+        env['TERM'] = 'dumb'
+        self._test_expected_error(
+            plan, '124',
+            f'''[1/1] {py} -c 'exit(124)'
+FAILED: [code=124] 124 \n{py} -c 'exit(124)'
+ninja: build stopped: subcommand failed.
+''',
+            exit_code=124, env=env,
+        )
+        self._test_expected_error(
+            plan, '127',
+            f'''[1/1] {py} -c 'exit(127)'
+FAILED: [code=127] 127 \n{py} -c 'exit(127)'
+ninja: build stopped: subcommand failed.
+''',
+            exit_code=127, env=env,
+        )
+        self._test_expected_error(
+            plan, '130',
+            'ninja: build stopped: interrupted by user.\n',
+            exit_code=130, env=env,
+        )
+        self._test_expected_error(
+            plan, '137',
+            f'''[1/1] {py} -c 'exit(137)'
+FAILED: [code=137] 137 \n{py} -c 'exit(137)'
+ninja: build stopped: subcommand failed.
+''',
+            exit_code=137, env=env,
+        )
+        self._test_expected_error(
+            plan, 'non-existent-target',
+            "ninja: error: unknown target 'non-existent-target'\n",
+            exit_code=1, env=env,
+        )
+        self._test_expected_error(
+            plan, '-j2 success 127',
+            f'''[1/2] {py} -c 'exit(127)'
+FAILED: [code=127] 127 \n{py} -c 'exit(127)'
+[2/2] sleep 0.3; echo success
+success
+ninja: build stopped: subcommand failed.
+''',
+            exit_code=127, env=env,
+        )
 
     def test_depfile_directory_creation(self) -> None:
         b = BuildDir('''\
--- a/src/build.cc
+++ b/src/build.cc
@@ -34,12 +34,12 @@
 #include "depfile_parser.h"
 #include "deps_log.h"
 #include "disk_interface.h"
+#include "exit_status.h"
 #include "explanations.h"
 #include "graph.h"
 #include "metrics.h"
 #include "state.h"
 #include "status.h"
-#include "subprocess.h"
 #include "util.h"
 
 using namespace std;
@@ -687,7 +687,7 @@ bool Builder::AlreadyUpToDate() const {
   return !plan_.more_to_do();
 }
 
-bool Builder::Build(string* err) {
+ExitStatus Builder::Build(string* err) {
   assert(!AlreadyUpToDate());
   plan_.PrepareQueue();
 
@@ -726,14 +726,14 @@ bool Builder::Build(string* err) {
         if (!StartEdge(edge, err)) {
           Cleanup();
           status_->BuildFinished();
-          return false;
+          return ExitFailure;
         }
 
         if (edge->is_phony()) {
           if (!plan_.EdgeFinished(edge, Plan::kEdgeSucceeded, err)) {
             Cleanup();
             status_->BuildFinished();
-            return false;
+            return ExitFailure;
           }
         } else {
           ++pending_commands;
@@ -760,14 +760,16 @@ bool Builder::Build(string* err) {
         Cleanup();
         status_->BuildFinished();
         *err = "interrupted by user";
-        return false;
+        return result.status;
       }
 
       --pending_commands;
-      if (!FinishCommand(&result, err)) {
+      bool command_finished = FinishCommand(&result, err);
+      SetFailureCode(result.status);
+      if (!command_finished) {
         Cleanup();
         status_->BuildFinished();
-        return false;
+        return result.status;
       }
 
       if (!result.success()) {
@@ -791,11 +793,11 @@ bool Builder::Build(string* err) {
     else
       *err = "stuck [this is a bug]";
 
-    return false;
+    return GetExitCode();
   }
 
   status_->BuildFinished();
-  return true;
+  return ExitSuccess;
 }
 
 bool Builder::StartEdge(Edge* edge, string* err) {
@@ -1035,4 +1037,11 @@ bool Builder::LoadDyndeps(Node* node, string* err) {
     return false;
 
   return true;
+}
+
+void Builder::SetFailureCode(ExitStatus code) {
+  // ExitSuccess should not overwrite any error
+  if (code != ExitSuccess) {
+    exit_code_ = code;
+  }
 }
--- a/src/build.cc
+++ b/src/build.cc
@@ -885,7 +885,7 @@ bool Builder::FinishCommand(CommandRunner::Result* result, string* err) {
   running_edges_.erase(it);
 
   status_->BuildEdgeFinished(edge, start_time_millis, end_time_millis,
-                             result->success(), result->output);
+                             result->status, result->output);
 
   // The rest of this function only applies to successful commands.
   if (!result->success()) {
--- a/src/build.h
+++ b/src/build.h
@@ -209,9 +209,9 @@ struct Builder {
   /// Returns true if the build targets are already up to date.
   bool AlreadyUpToDate() const;
 
-  /// Run the build.  Returns false on error.
+  /// Run the build.  Returns ExitStatus or the exit code of the last failed job.
   /// It is an error to call this function when AlreadyUpToDate() is true.
-  bool Build(std::string* err);
+  ExitStatus Build(std::string* err);
 
   bool StartEdge(Edge* edge, std::string* err);
 
@@ -233,6 +233,10 @@ struct Builder {
   std::unique_ptr<CommandRunner> command_runner_;
   Status* status_;
 
+  /// Returns ExitStatus or the exit code of the last failed job
+  /// (doesn't need to be an enum value of ExitStatus)
+  ExitStatus GetExitCode() const { return exit_code_; }
+
  private:
   bool ExtractDeps(CommandRunner::Result* result, const std::string& deps_type,
                    const std::string& deps_prefix,
@@ -252,6 +256,10 @@ struct Builder {
   std::unique_ptr<Explanations> explanations_;
 
   DependencyScan scan_;
+
+  /// Keep the global exit code for the build
+  ExitStatus exit_code_ = ExitSuccess;
+  void SetFailureCode(ExitStatus code);
 
   // Unimplemented copy ctor and operator= ensure we don't copy the auto_ptr.
   Builder(const Builder &other);        // DO NOT IMPLEMENT
--- a/src/build_test.cc
+++ b/src/build_test.cc
@@ -20,6 +20,7 @@
 
 #include "build_log.h"
 #include "deps_log.h"
+#include "exit_status.h"
 #include "graph.h"
 #include "status_printer.h"
 #include "test.h"
@@ -616,8 +617,8 @@ void BuildTest::RebuildTarget(const string& target, const char* manifest,
   command_runner_.commands_ran_.clear();
   builder.command_runner_.reset(&command_runner_);
   if (!builder.AlreadyUpToDate()) {
-    bool build_res = builder.Build(&err);
-    EXPECT_TRUE(build_res);
+    ExitStatus build_res = builder.Build(&err);
+    EXPECT_EQ(build_res, ExitSuccess);
   }
   builder.command_runner_.release();
 }
@@ -823,7 +824,7 @@ TEST_F(BuildTest, OneStep) {
   string err;
   EXPECT_TRUE(builder_.AddTarget("cat1", &err));
   ASSERT_EQ("", err);
-  EXPECT_TRUE(builder_.Build(&err));
+  EXPECT_EQ(builder_.Build(&err), ExitSuccess);
   ASSERT_EQ("", err);
 
   ASSERT_EQ(1u, command_runner_.commands_ran_.size());
@@ -837,7 +838,7 @@ TEST_F(BuildTest, OneStep2) {
   string err;
   EXPECT_TRUE(builder_.AddTarget("cat1", &err));
   ASSERT_EQ("", err);
-  EXPECT_TRUE(builder_.Build(&err));
+  EXPECT_EQ(builder_.Build(&err), ExitSuccess);
   EXPECT_EQ("", err);
 
   ASSERT_EQ(1u, command_runner_.commands_ran_.size());
@@ -848,7 +849,7 @@ TEST_F(BuildTest, TwoStep) {
   string err;
   EXPECT_TRUE(builder_.AddTarget("cat12", &err));
   ASSERT_EQ("", err);
-  EXPECT_TRUE(builder_.Build(&err));
+  EXPECT_EQ(builder_.Build(&err), ExitSuccess);
   EXPECT_EQ("", err);
   ASSERT_EQ(3u, command_runner_.commands_ran_.size());
   // Depending on how the pointers work out, we could've ran
@@ -868,7 +869,7 @@ TEST_F(BuildTest, TwoStep) {
   state_.Reset();
   EXPECT_TRUE(builder_.AddTarget("cat12", &err));
   ASSERT_EQ("", err);
-  EXPECT_TRUE(builder_.Build(&err));
+  EXPECT_EQ(builder_.Build(&err), ExitSuccess);
   ASSERT_EQ("", err);
   ASSERT_EQ(5u, command_runner_.commands_ran_.size());
   EXPECT_EQ("cat in1 in2 > cat2", command_runner_.commands_ran_[3]);
@@ -886,7 +887,7 @@ TEST_F(BuildTest, TwoOutputs) {
   string err;
   EXPECT_TRUE(builder_.AddTarget("out1", &err));
   ASSERT_EQ("", err);
-  EXPECT_TRUE(builder_.Build(&err));
+  EXPECT_EQ(builder_.Build(&err), ExitSuccess);
   EXPECT_EQ("", err);
   ASSERT_EQ(1u, command_runner_.commands_ran_.size());
   EXPECT_EQ("touch out1 out2", command_runner_.commands_ran_[0]);
@@ -902,7 +903,7 @@ TEST_F(BuildTest, ImplicitOutput) {
   string err;
   EXPECT_TRUE(builder_.AddTarget("out.imp", &err));
   ASSERT_EQ("", err);
-  EXPECT_TRUE(builder_.Build(&err));
+  EXPECT_EQ(builder_.Build(&err), ExitSuccess);
   EXPECT_EQ("", err);
   ASSERT_EQ(1u, command_runner_.commands_ran_.size());
   EXPECT_EQ("touch out out.imp", command_runner_.commands_ran_[0]);
@@ -924,7 +925,7 @@ TEST_F(BuildTest, MultiOutIn) {
   string err;
   EXPECT_TRUE(builder_.AddTarget("out", &err));
   ASSERT_EQ("", err);
-  EXPECT_TRUE(builder_.Build(&err));
+  EXPECT_EQ(builder_.Build(&err), ExitSuccess);
   EXPECT_EQ("", err);
 }
 
@@ -940,7 +941,7 @@ TEST_F(BuildTest, Chain) {
   string err;
   EXPECT_TRUE(builder_.AddTarget("c5", &err));
   ASSERT_EQ("", err);
-  EXPECT_TRUE(builder_.Build(&err));
+  EXPECT_EQ(builder_.Build(&err), ExitSuccess);
   EXPECT_EQ("", err);
   ASSERT_EQ(4u, command_runner_.commands_ran_.size());
 
@@ -960,7 +961,7 @@ TEST_F(BuildTest, Chain) {
   EXPECT_TRUE(builder_.AddTarget("c5", &err));
   ASSERT_EQ("", err);
   EXPECT_FALSE(builder_.AlreadyUpToDate());
-  EXPECT_TRUE(builder_.Build(&err));
+  EXPECT_EQ(builder_.Build(&err), ExitSuccess);
   ASSERT_EQ(2u, command_runner_.commands_ran_.size());  // 3->4, 4->5
 }
 
@@ -1001,7 +1002,7 @@ TEST_F(BuildTest, MakeDirs) {
   EXPECT_TRUE(builder_.AddTarget("subdir/dir2/file", &err));
 
   EXPECT_EQ("", err);
-  EXPECT_TRUE(builder_.Build(&err));
+  EXPECT_EQ(builder_.Build(&err), ExitSuccess);
   ASSERT_EQ("", err);
   ASSERT_EQ(2u, fs_.directories_made_.size());
   EXPECT_EQ("subdir", fs_.directories_made_[0]);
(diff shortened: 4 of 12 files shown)


Felixoid:
The reason why do we need it:

In ClickHouse, we use [prlimit](https://github.com/ClickHouse/ClickHouse/blob/854b72ade26a8ab551a9600050645198ea93bc32/CMakeLists.txt#L78) to identify when some binary becomes too big, but it fails in the entirely same way as a standard failing build.

Felixoid:
After the last push, it works like this:

```
> ./ninja test-exit; echo $?
[1/1] exit 155
FAILED: test-exit
exit 155
155
> ./ninja test-exit; echo $?
[1/1] exit 1
FAILED: test-exit
exit 1
1
> ./ninja test-exit; echo $?
[0/1] exit 2
130
```

Felixoid:
The latest commit adds the exit code to the `FAILED` log line:

```
> ./ninja prlimit-kill prlimit-allocate prlimit-sigv ; echo $?
[1/3] prlimit --data=1 clickhouse-local -q 'SELECT * FROM (SELECT * FROM system.numbers GROUP BY number LIMIT 100000000000) GROUP BY number'
FAILED[code 139] prlimit-sigv
prlimit --data=1 clickhouse-local -q 'SELECT * FROM (SELECT * FROM system.numbers GROUP BY number LIMIT 100000000000) GROUP BY number'
[2/3] prlimit --data=9999999 clickhouse-local -q 'SELECT * FROM (SELECT * FROM system.numbers GROUP BY number LIMIT 100000000000) GROUP BY number'
FAILED[code 139] prlimit-allocate
prlimit --data=9999999 clickhouse-local -q 'SELECT * FROM (SELECT * FROM system.numbers GROUP BY number LIMIT 100000000000) GROUP BY number'
[3/3] prlimit --cpu=1 clickhouse-local -q 'SELECT * FROM (SELECT * FROM system.numbers GROUP BY number LIMIT 100000000000) GROUP BY number'
FAILED[code 137] prlimit-kill
prlimit --cpu=1 clickhouse-local -q 'SELECT * FROM (SELECT * FROM system.numbers GROUP BY number LIMIT 100000000000) GROUP BY number'
137
```

Felixoid:
Dear @jhasse and @digit-google, could you please leave a comment on whether it makes sense or if there is something left to improve before the merge?

digit-google:
Thank you for your contribution. I am very busy this week, but I'll take a look the next one. Sorry about that.

digit-google:
Thank you very much for you PR. This feature will be useful.

I think that keeping a dedicated type for the exit status will help keeping the code easier to read and understand. Besides, native exit codes on Win32 are unsigned long, even though the `exit()` function invoked from Ninja still takes an `int`. I suggest to achieve this by:

1) Using C++11 enum sizing with platform-specific types, as in:

```
// from exit_status.h

// The underlying type of the ExitStatus enum, used to represent a platform-specific
// process exit code.
#ifdef _WIN32
#define EXIT_STATUS_TYPE unsigned long
#else  // !_WIN32
#define EXIT_STATUS_TYPE int
#endif  // !_WIN32

enum ExitStatus : EXIT_STATUS_TYPE {
  ExitSuccess = 0,
  ExitFailure,
  ExitInterrupted,
};
```

2) Modify `Builder::Build()` to return an `ExitStatus` value instead of a boolean.

Also, on Posix, ensure that an interrupted build returns `static_cast<ExitStatus>(130)` directly, to get rid of the hack on line 1580 of `ninja.cc` to do that.

3) Only cast the ExitStatus to an `int` when needed, i.e. when calling `exit()` from `ninja.cc`. This means NinjaMain::RunBuild() would also return an `ExitStatus` value instead of an `int`.

Apart from that, can I ask you to rebase your commits to remove temporary fixes?  Nobody really needs this information but you, and it makes reviews, and later looking at the git history, difficult. Ideally tests should be introduced with the modifications they are related to. In this specific case, I would recommend one commit to propagate the status (+tests), then another commit to change the printed status line in case of failure to include the code.

Finally add a regression test in `output_test.py` for each commit's changes. I.e. one that verifies the value returned by Ninja when invoking a command that fails voluntarily, then another one that checks the text printed by Ninja in case of failure to verify it includes the right code.

There is a tiny chance that the status line change is going to break existing parsing scripts. I recommend you try to keep the original `FAILED:` sub-string in the output in case of failure to minimize that. I.e. print `FAILED: [code=<value>]` instead of `FAILED[code=<value>]:`. @jhasse what do you think about this specific problem?

Felixoid:
Hi @digit-google, thanks for the review!

I think the only unclear thing to me is how to use ExitStatus for other codes than are defined there. I think, you mean using EXIT_STATUS_TYPE in all the places? Than, looks good to me.

I'll apply all the fixes

digit-google:
Can you clarify what you mean. It should be straightforward to just use the `ExitStatus` type for return values and local/member variables types. Use a static cast when you need to convert an `int` or `DWORD` into an `ExitStatus` in the subprocess `Finish()` functions, and that should be enough. Wdyt?

Felixoid:
Just in case I am not a real C++ dev 👉👈, I could miss the point. 
The idea is to exit Ninja with one of the exit codes of the subprocesses. So, `return static_cast<ExitStatus>(42);` should still work, right?

digit-google:
It's the other way around, lines 1804-1807 of ninja.cc look like this:

```
    int result = ninja.RunBuild(argc, argv, status);
    if (g_metrics)
      ninja.DumpMetrics();
    exit(result);
```

What I am suggesting is having `NinjaMain::RunBuild()` return an ExitStatus value, then do:

```
    ExitStatus status = ninja.RunBuild(argc, argv, status);
    if (g_metrics)
      ninja.DumpMetrics();
    exit(static_cast<int>(status));
```

Felixoid:
but [Finish()](https://github.com/ninja-build/ninja/pull/2540/files#diff-e70adfab80664f71832232d71b3055c78e054dd75fbeec69d019cb0e9dce6b2dR169) should as well return the same code it received from the subprocess, so it will be passed from there to `ninja.RunBuild`?

```diff
-int Subprocess::Finish() {
+ExitStatus Subprocess::Finish() {
   assert(pid_ != -1);
   int status;
   if (waitpid(pid_, &status, 0) < 0)
@@ -166,7 +168,7 @@ int Subprocess::Finish() {

   if (WIFEXITED(status)) {
     // propagate the status transparently
-    return WEXITSTATUS(status);
+    return static_cast<ExitStatus>(WEXITSTATUS(status));
   }
   if (WIFSIGNALED(status)) {
     // Overwrite interrupts to exit code 2
@@ -175,7 +177,7 @@ int Subprocess::Finish() {
       return ExitInterrupted;
   }
   // At this point, we exit with any other signal+128
-  return status | 0x80;
+  return static_cast<ExitStatus>(status | 0x80);
 }
```

Felixoid:
> There is a tiny chance that the status line change is going to break existing parsing scripts. I recommend you try to keep the original `FAILED:` sub-string in the output in case of failure to minimize that. I.e. print `FAILED: [code=<value>]` instead of `FAILED[code=<value>]:`.

One note, there's no `FAILED:` now. Only `FAILED executed_command`.

I think using `FAILED [code=<value]:` should be good. It won't break the parsers

Felixoid:
I think all points are addressed:

- `ExitStatus` is used everywhere to pass the exit codes
- `output_test.py` has tests for both logging and exiting

Besides, I updated all the tests depending on `Builder::Build`, so they are also aware of the `ExitStatus` return.

src/build.h
```diff
@@ -233,6 +233,11 @@ struct Builder {
   std::unique_ptr<CommandRunner> command_runner_;
   Status* status_;
 
+  /// Keep the global exit code for the run
+  ExitStatus exit_code_;
```
digit-google, line 237: nit: move this variable and the `SetExitCode()` function to the `private` section to avoid direct access. Maybe also rename `ExitCode()` to `GetExitCode()` for clarity.

src/exit_status.h
```diff
@@ -15,10 +15,18 @@
...
+#define EXIT_STATUS_TYPE unsigned long
+#else  // !_WIN32
+#define EXIT_STATUS_TYPE int
+#endif  // !_WIN32
+
+enum ExitStatus : EXIT_STATUS_TYPE {
+  ExitSuccess=0,
   ExitFailure,
-  ExitInterrupted
+  ExitInterrupted=130
```
digit-google, line 29: The value 130 is Posix-specific. On Windows, the convention is that the C runtime will return 3 in case of Ctrl-C, though some programs also return 1. I suggest you define the value with another platform-specific macro, e.g.:

```
// The underlying type of the ExitStatus enum, used to represent a platform-specific
// process exit code.
#ifdef _WIN32
#define EXIT_STATUS_TYPE unsigned long
#else  // !_WIN32
#define EXIT_STATUS_TYPE int
#endif  // !_WIN32

// The platform-specific value of ExitInterrupted returned by Ninja in case of use interruption (e.g. Ctrl-C).
// On Windows, the convention used by the C runtime library is to return 3, even
// though the system error `ERROR_CONTROL_C_EXIT` is 572. Some applications also return 1
// which is ExitFailure. The value 2 is chosen to preserve the existing Ninja behavior on this
// platform. On Posix, this is simply 128 + SIGINT. Note that Ninja will map SIGHUP
// and SIGTERM interrupted to ExitInterrupted as well.
#ifdef _WIN32
#  define EXIT_INTERRUPTED_VALUE  2
#else  // !_WIN32
#  define EXIT_INTERRUPTED_VALUE  130
#endif  // !_WIN32

enum ExitStatus : EXIT_STATUS_TYPE {
  ExitSuccess=0,
  ExitFailure,
  ExitInterrupted = EXIT_INTERRUPTED_VALUED,
};
```

Also, document in the commit message that this changes the value returned by Ninja on Posix for user interruptions. This is important for the future release notes.

misc/output_test.py
```diff
@@ -152,14 +153,18 @@ class Output(unittest.TestCase):
         '',
     ))
 
-    def _test_expected_error(self, plan: str, flags: T.Optional[str], expected: str):
+    def _test_expected_error(self,  expected: str, *args, **kwargs):
```
digit-google, line 156: Please avoid changing the order of arguments if possible, as this reduces the number of potential merge / rebase conflicts for all other unsubmitted PRs. In this specific case, I believe it should be possible to add the extra args at the end of the current signature without requiring changing other call sites.

misc/output_test.py
```diff
@@ -281,6 +286,70 @@ def test_issue_2048(self) -> None:
             except subprocess.CalledProcessError as err:
                 self.fail("non-zero exit code with: " + err.output)
 
+    def test_pr_2540(self)->None:
+        plan = '''\
+rule CUSTOM_COMMAND
+  command = $COMMAND
+
+build 124: CUSTOM_COMMAND
+  COMMAND = python3 -c 'exit(124)'
```
digit-google, line 295: Please do not assume that `python3` is available when running the test (as a general rule, `python3` is never available on Windows, nor on OS X iirc). Instead define a local variable such as `python = sys.executable` and use it in formatted strings for the build plan and the expected results below.

digit-google:
Thanks very much for the latest commits. I added a few review comments, please address them, but this is looking very good so far. Now it's up to @jhasse to add his feedback and approve the PR.

misc/output_test.py
Felixoid: About keeping as few changes in signatures as possible to reduce conflicts, this specific case limits the usage of calling the following `run` function.

Passing every possible argument into `run` in the wrong order, as I've done with [env](https://github.com/ninja-build/ninja/pull/2540/files#diff-32108e2fd2aaf591ed4fae7d0955d688432825bc286bdd99a4962a7db3a8d6a2R317) would be a mess. 

Keeping the current order would mean something like `(self, plan: str, flags: T.Optional[str], expected: str, exit_code: Optional[int] = 0, env: Dict[str, str] = {})` or `(self, plan: str, flags: T.Optional[str], expected: str, env: Dict[str, str] = {}, exit_code: Optional[int] = 0)`, and it means we split run-related args with the method-related. IMHO, it doesn't look nice, and adding more run-related args in the future will be even more problematic.

Is it worth changing it now until the method is not spread widely?

src/exit_status.h
Felixoid: Thanks. I'll take a look on how to comment the commit properly

misc/output_test.py
digit-google: I think there is no point in making all arguments explicit, i.e. you should be able to just add `*args, **kwargs` at the end of the current definition, and check `kwargs["exit_code"]` manually.

misc/output_test.py
```diff
@@ -174,11 +174,11 @@ def test_issue_1418(self) -> None:
...
 
 build a: echo
-  delay = 3
+  delay = 0.9
 build b: echo
-  delay = 2
+  delay = 0.6
 build c: echo
-  delay = 1
+  delay = 0.3
```
digit-google, line 181: I suggest leaving this out of this PR, it's unrelated to what you're trying to do. Unfortunately, we had issues in the past because CI tests are run on VMs that often are very contended, meaning that tests that expect exact timing sequences can fail randomly in Github actions :-/ I am wary that such a change might trigger this condition more often.

misc/output_test.py
Felixoid: The issue, `flags: T.Optional[str],` doesn't have the default value, although it has it in both [run](https://github.com/ninja-build/ninja/blob/d3444792771cfef3d47174e3535bab63e4c97197/misc/output_test.py#L80) and [run](https://github.com/ninja-build/ninja/blob/d3444792771cfef3d47174e3535bab63e4c97197/misc/output_test.py#L133).

It is worth having a default for `None`, but the following `expected` without defaults must stay in front of `flags` then.

To be honest, the current signature is very questionable. If it's a blocker, I could leave the first arguments as they are and pass all the `**kwargs`.

Felixoid:
```diff
diff --git a/misc/output_test.py b/misc/output_test.py
index 641a444..b3e2a15 100755
--- a/misc/output_test.py
+++ b/misc/output_test.py
@@ -153,7 +153,7 @@ class Output(unittest.TestCase):
         '',
     ))
 
-    def _test_expected_error(self,  expected: str, *args, **kwargs):
+    def _test_expected_error(self, plan: str, flags: T.Optional[str], expected: str, *args, **kwargs):
         """Run Ninja with a given plan and flags, and verify its cooked output against an expected content.
         All *args and **kwargs are passed to the `run`, except `exit_code`
         """
@@ -161,7 +161,7 @@ class Output(unittest.TestCase):
         kwargs['print_err_output'] = False
         exit_code = kwargs.pop("exit_code", None)
         with self.assertRaises(subprocess.CalledProcessError) as cm:
-            run(*args,  **kwargs)
+            run(plan, flags, *args,  **kwargs)
         actual = cm.exception.cooked_output
         if exit_code is not None:
             self.assertEqual(cm.exception.returncode, exit_code)
@@ -311,42 +311,48 @@ build success: CUSTOM_COMMAND
         env = default_env.copy()
         env['TERM'] = 'dumb'
         self._test_expected_error(
+            plan, '124',
             f'''[1/1] {py} -c 'exit(124)'
 FAILED: [code=124] 124 \n{py} -c 'exit(124)'
 ninja: build stopped: subcommand failed.
 ''',
-            plan, '124', exit_code=124, env=env,
+            exit_code=124, env=env,
         )
         self._test_expected_error(
+            plan, '127',
             f'''[1/1] {py} -c 'exit(127)'
 FAILED: [code=127] 127 \n{py} -c 'exit(127)'
 ninja: build stopped: subcommand failed.
 ''',
-            plan, '127', exit_code=127, env=env,
+            exit_code=127, env=env,
         )
         self._test_expected_error(
+            plan, '130',
             'ninja: build stopped: interrupted by user.\n',
-            plan, '130', exit_code=130, env=env,
+            exit_code=130, env=env,
         )
         self._test_expected_error(
+            plan, '137',
             f'''[1/1] {py} -c 'exit(137)'
 FAILED: [code=137] 137 \n{py} -c 'exit(137)'
 ninja: build stopped: subcommand failed.
 ''',
-            plan, '137', exit_code=137, env=env,
+            exit_code=137, env=env,
         )
         self._test_expected_error(
+            plan, 'non-existent-target',
             "ninja: error: unknown target 'non-existent-target'\n",
-            plan, 'non-existent-target', exit_code=1, env=env,
+            exit_code=1, env=env,
         )
         self._test_expected_error(
+            plan, '-j2 success 127',
             f'''[1/2] {py} -c 'exit(127)'
 FAILED: [code=127] 127 \n{py} -c 'exit(127)'
 [2/2] sleep 0.3; echo success
 success
 ninja: build stopped: subcommand failed.
 ''',
-            plan, '-j2 success 127', exit_code=127, env=env,
+            exit_code=127, env=env,
         )
 
 
@@ -478,6 +484,7 @@ build out4 : cat in4
 
 
         self._test_expected_error(
+            plan, '-t compdb-targets',
             '''ninja: error: compdb-targets expects the name of at least one target
 usage: ninja -t compdb [-hx] target [targets]
 
@@ -485,17 +492,16 @@ options:
   -h     display this help message
   -x     expand @rspfile style response file invocations
 ''',
-            plan, '-t compdb-targets',
         )
 
         self._test_expected_error(
-            "ninja: fatal: 'in1' is not a target (i.e. it is not an output of any `build` statement)\n",
             plan, '-t compdb-targets in1',
+            "ninja: fatal: 'in1' is not a target (i.e. it is not an output of any `build` statement)\n",
         )
 
         self._test_expected_error(
-            "ninja: fatal: unknown target 'nonexistent_target'\n",
             plan, '-t compdb-targets nonexistent_target',
+            "ninja: fatal: unknown target 'nonexistent_target'\n",
         )
```

Like this?

misc/output_test.py
Felixoid: Sure, I will drop it on the following rebase.

misc/output_test.py
Felixoid: the patch is applied in the latest rebase

digit-google:
Thank you very much, latest version seems perfect to me.

src/status_printer.cc
```diff
@@ -202,16 +204,17 @@ void StatusPrinter::BuildEdgeFinished(Edge* edge, int64_t start_time_millis,
   --running_edges_;
 
   // Print the command that is spewing before printing its output.
-  if (!success) {
+  if (exit_code != 0) {
```
jhasse, line 207: `!= ExitSuccess`

src/subprocess-posix.cc
```diff
@@ -23,6 +24,7 @@
 #include <string.h>
 #include <sys/wait.h>
 #include <spawn.h>
+#include <cmath>
```
jhasse, line 27: the same as above

src/subprocess-posix.cc
```diff
@@ -165,15 +167,17 @@ ExitStatus Subprocess::Finish() {
...
   if (WIFEXITED(status)) {
-    int exit = WEXITSTATUS(status);
-    if (exit == 0)
-      return ExitSuccess;
-  } else if (WIFSIGNALED(status)) {
+    // propagate the status transparently
+    return static_cast<ExitStatus>(WEXITSTATUS(status));
+  }
+  if (WIFSIGNALED(status)) {
+    // Overwrite interrupts to exit code 2
```
jhasse, line 174: comment is wrong, I would just remove it.

src/subprocess-posix.cc
```diff
@@ -165,15 +167,17 @@ ExitStatus Subprocess::Finish() {
...
     if (WTERMSIG(status) == SIGINT || WTERMSIG(status) == SIGTERM
         || WTERMSIG(status) == SIGHUP)
       return ExitInterrupted;
   }
-  return ExitFailure;
+  // At this point, we exit with any other signal+128
+  return static_cast<ExitStatus>(status | 0x80);
```
jhasse, line 180: why not `+ 128`?

src/subprocess_test.cc
```diff
@@ -50,7 +51,13 @@ TEST_F(SubprocessTest, BadCommandStderr) {
     subprocs_.DoWork();
   }
 
-  EXPECT_EQ(ExitFailure, subproc->Finish());
+  ExitStatus exit = subproc->Finish();
+#ifdef _POSIX_VERSION
```
jhasse, line 55: Use `#else` on the _WIN32 branch instead.

src/subprocess_test.cc
```diff
@@ -64,7 +71,13 @@ TEST_F(SubprocessTest, NoSuchCommand) {
...
```
jhasse, line 75: same as above

jhasse (commented):
Thanks for the PR!

From my experience you're entering a world of pain as soon as you want to do anything with exit codes outside of the range 0 to 254. Therefore I'm not sure if Ninja should support it, even on Windows.

I will think about it.

src/build_test.cc
```diff
@@ -16,10 +16,12 @@
 
 #include <assert.h>
 #include <climits>
+#include <gtest/gtest.h>
```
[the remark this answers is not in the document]
Felixoid: It was added automatically during autocomplete. It looks like https://github.com/ycm-core/YouCompleteMe does it.

src/build.cc
```diff
@@ -1036,3 +1038,8 @@ bool Builder::LoadDyndeps(Node* node, string* err) {
 
   return true;
 }
+
+void Builder::SetExitCode(ExitStatus code) {
+  // Set code to the most recent error
+  if (code != 0) exit_code_ = code;
```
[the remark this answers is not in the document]
Felixoid: Then, somebody could accidentally call it to overwrite the error code by 0.

Does it make sense to rename the method, maybe?

```
void Builder::SetFailureCode(ExitStatus code) {
  // The ExitSuccess should not overwrite any error
  if (code != ExitSuccess) exit_code_ = code;
}
```

misc/output_test.py
```diff
@@ -10,16 +10,17 @@
 import subprocess
 import sys
 import tempfile
+import typing as T
 import unittest
 from textwrap import dedent
-import typing as T
```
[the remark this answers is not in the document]
Felixoid: It was done by https://github.com/PyCQA/isort/, yes.

misc/output_test.py
```diff
@@ -10,16 +10,17 @@
...
 
 default_env = dict(os.environ)
 default_env.pop('NINJA_STATUS', None)
 default_env.pop('CLICOLOR_FORCE', None)
 default_env['TERM'] = ''
 NINJA_PATH = os.path.abspath('./ninja')
 
+
```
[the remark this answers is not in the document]
Felixoid: This is the pep-8 requirement "two empty lines before and after function/class definitions"

misc/output_test.py
```diff
@@ -120,7 +121,7 @@ def run(
                                                  cwd=self.d.name, env=env)
         except subprocess.CalledProcessError as err:
             if print_err_output:
-              sys.stdout.buffer.write(err.output)
+                sys.stdout.buffer.write(err.output)
```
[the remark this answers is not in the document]
Felixoid: Having mixed alignments in the same file confuses very much. Should I revert it?

src/ninja.cc
```diff
@@ -1572,18 +1572,18 @@ int NinjaMain::RunBuild(int argc, char** argv, Status* status) {
...
-    return 0;
+    return ExitSuccess;
   }
 
-  if (!builder.Build(&err)) {
+  ExitStatus exit_status = builder.Build(&err);
+  if (exit_status != ExitSuccess) {
     status->Info("build stopped: %s.", err.c_str());
     if (err.find("interrupted by user") != string::npos) {
-      return 130;
```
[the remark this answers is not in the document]
Felixoid: I think it did indeed. So, should we make the same behavior and use `ExitInterrupted=130`?

src/subprocess-posix.cc
Felixoid: It's hard to say; I've seen it in some sources during the first research.

misc/output_test.py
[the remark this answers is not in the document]
Felixoid: Is it ok to keep it sorted? The same question regarding the other pep8-related changes

src/exit_status.h
```diff
@@ -15,10 +15,30 @@
...
+#  define EXIT_INTERRUPTED_VALUE  2
+#else  // !_WIN32
+#  define EXIT_INTERRUPTED_VALUE  130
+#endif  // !_WIN32
+
+enum ExitStatus : EXIT_STATUS_TYPE {
+  ExitSuccess=0,
   ExitFailure,
-  ExitInterrupted
+  ExitInterrupted=EXIT_INTERRUPTED_VALUE,
```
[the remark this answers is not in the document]
Felixoid: Since the exit code used to be [130](https://github.com/ninja-build/ninja/blob/b78335e0f960bb547c208254e65c9a20fcd4da29/src/ninja.cc#L1578-L1583) for all the OS, should we make it 130 here as well, maybe?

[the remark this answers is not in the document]
Felixoid: Pushed `ExitInterrupted=130`. Please let me know if you have a better idea regarding it.

src/ninja.cc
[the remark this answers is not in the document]
Felixoid: Let's continue this discussion in https://github.com/ninja-build/ninja/pull/2540#discussion_r1878808943 branch

Felixoid:
Addressed most of the questions and made review-related changes as fixups to track the progress.

src/exit_status.h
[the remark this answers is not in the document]
digit-google: Oh, my mistake for suggesting otherwise. Yes, I would be in favor of keeping returning 130 on Windows if that's what we already do (I guess nobody cares though).

src/status_printer.cc
```diff
@@ -25,6 +26,7 @@
 
 #include <stdarg.h>
 #include <stdlib.h>
+#include <string>
```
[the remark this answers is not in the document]
Felixoid: Thanks for noticing it. I disabled it now, ref https://github.com/ycm-core/YouCompleteMe/issues/3788

Felixoid:
I've rebased all the changes, let me know if something here is still blocker to proceed :pray:

src/build.cc
```diff
@@ -1036,3 +1038,8 @@ bool Builder::LoadDyndeps(Node* node, string* err) {
 
   return true;
 }
+
+void Builder::SetFailureCode(ExitStatus code) {
+  // The ExitSuccess should not overwrite any error
```
jhasse, line 1043: Remove `The `

src/build.cc
```diff
@@ -1036,3 +1038,8 @@ bool Builder::LoadDyndeps(Node* node, string* err) {
...
+  if (code != ExitSuccess) exit_code_ = code;
```
jhasse, line 1044: Please add {}

src/subprocess_test.cc
jhasse: still missing

jhasse (changes requested):
Few nitpicks, besides that it LGTM :)

misc/output_test.py
```diff
@@ -152,14 +153,18 @@ class Output(unittest.TestCase):
         '',
     ))
 
-    def _test_expected_error(self, plan: str, flags: T.Optional[str], expected: str):
+    def _test_expected_error(self, plan: str, flags: T.Optional[str],expected: str,
+                             *args, exit_code: T.Optional[int]=None, **kwargs)->None:
```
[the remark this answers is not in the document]
Felixoid: It's a proper typing for functions return when nothing is returned. Here's an example https://docs.python.org/3/library/typing.html#type-aliases

The typing is non-complete currently. And `mypy` complains about it

src/subprocess_test.cc
Felixoid: I don't get what's missing. There's no `_WIN32` for `exit` anymore [here](https://github.com/ninja-build/ninja/blob/02a19024a9cefbbf1dfa7583e8dc0235895d1925/src/subprocess_test.cc#L55-L59) and [here](https://github.com/ninja-build/ninja/blob/02a19024a9cefbbf1dfa7583e8dc0235895d1925/src/subprocess_test.cc#L74-L78)

Felixoid:
I reverted all things in output_test.py unrelated to the test itself, and will implement automatic checks for them separately.

The rest of the comments are addressed, including [_WIN32](https://github.com/ninja-build/ninja/pull/2540/commits/a3532131962a054bb9f5da990c7421e977672d5d#diff-fd66706b3c2066e7d93bab190af8ba03ebf3c3a4683541a6f7175a05a8ad43baR55)

src/subprocess_test.cc
jhasse: Ah I was confused. But still wondering: Why is there a difference between `_POSIX_VERSION` and the rest?

Felixoid: I don't know, but I would assume it's the `command not found` exit code for the win32. And POSIX returns 127 for it:

```
> command-not-found ; echo $?
zsh: command not found: command-not-found
127
```

Previously `subprocess-posix.cc` didn't propagate it, so any error was 1.

jhasse: I think we should just adjust the test that it checks != ExitSuccess then.

Felixoid: No problem, will do in a moment

jhasse:
Thanks! No more objections from my side, will wait a few days to see if anyone else has comments.

digit-google:
Thanks and happy new year :) The latest commits look fine to me as well. Great work.

Felixoid:
Thanks! Happy New Year, too!

So, does it look like good to go?

jhasse (approved):

PR #2541, opened 2024-12-05
Add default virtual destructor for `Parser` class.

zufuliu:
This fixed clang `[-Wdelete-non-abstract-non-virtual-dtor]` warning:
    delete called on non-final 'ManifestParser' that has virtual
    functions but non-virtual destructor.
for `subparser_.reset(new ManifestParser(state_, file_reader_, options_));`.

### DIFF
--- a/src/parser.h
+++ b/src/parser.h
@@ -26,6 +26,7 @@ struct State;
 struct Parser {
   Parser(State* state, FileReader* file_reader)
       : state_(state), file_reader_(file_reader) {}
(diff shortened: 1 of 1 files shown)


digit-google:
Thanks, I think it would make more sense to add a virtual (default) destructor to the Parser base class instead. Can you do that?

digit-google:
Thanks, this looks good to me. @jhasse , wdyt?

jhasse:
thx :)

PR #2547, opened 2024-12-28
Fix rspfile name length exceeding MAX_FILENAME_SIZE (140) characters

timfleck:
### Description

This Pull Request addresses an issue where `rspfile` names exceeding 144 characters cause build failures due to encrypted filesystem limitations. The proposed solution truncates overly long `rspfile` names and appends a unique identifier to ensure they remain within the acceptable length.

### Changes Made

- **ManifestParser::ParseRule:** Introduced a length check for `rspfile` names and applied truncation when necessary.
- **checkFileName:** Checks the lengeth of the filename strings. If necessary it shortens it by cut the first 'n' characters and add a checksum to keep the files unique. 
- **simpleHash:** Generates a checksum to keep the shortened files unique.

### Benefits

- Prevents build failures caused by excessively long `rspfile` names.
- Enhances compatibility with filesystems that impose strict filename length restrictions.
- Maintains uniqueness of `rspfile` names to avoid conflicts.

### Testing

- Created test cases with `rspfile` names exceeding 140 characters to ensure they are correctly truncated.
- Verified that build processes succeed without `rspfile` name length issues.

### DIFF
--- a/src/manifest_parser.cc
+++ b/src/manifest_parser.cc
@@ -152,6 +152,15 @@ bool ManifestParser::ParseRule(string* err) {
       return false;
 
     if (Rule::IsReservedBinding(key)) {
+      if (key == "rspfile") {
+        //Info("*** ManifestParser::ParseRule() Key:'%s' Value:'%s'****\n",key.c_str(),value.Unparse().c_str());
+        char safeFileName[MAX_FILENAME_SIZE];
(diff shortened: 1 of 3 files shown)


digit-google:
Please open an issue describing your problem so we can discuss the real issue. For example, I fail to see why this would be needed for rspfiles only for example.

Issue #2426, opened 2024-04-19
status_printer.h:53:38: error: ‘int64_t’ has not been declared


userdocs:
A recent change post v1.12.0 release has caused this build error:

```cpp
Run docker exec -w /root/ninja multiarch cmake --build build
[  1%] Generating lexer.cc
[  2%] Generating depfile_parser.cc
[  4%] Building CXX object CMakeFiles/libninja-re2c.dir/depfile_parser.cc.o
[  5%] Building CXX object CMakeFiles/libninja-re2c.dir/lexer.cc.o
[  5%] Built target libninja-re2c
[  6%] Building CXX object CMakeFiles/libninja.dir/src/build_log.cc.o
[  8%] Building CXX object CMakeFiles/libninja.dir/src/build.cc.o
[  9%] Building CXX object CMakeFiles/libninja.dir/src/clean.cc.o
[ 10%] Building CXX object CMakeFiles/libninja.dir/src/clparser.cc.o
[ 12%] Building CXX object CMakeFiles/libninja.dir/src/dyndep.cc.o
[ 13%] Building CXX object CMakeFiles/libninja.dir/src/dyndep_parser.cc.o
[ 14%] Building CXX object CMakeFiles/libninja.dir/src/debug_flags.cc.o
[ 16%] Building CXX object CMakeFiles/libninja.dir/src/deps_log.cc.o
[ 17%] Building CXX object CMakeFiles/libninja.dir/src/disk_interface.cc.o
[ 18%] Building CXX object CMakeFiles/libninja.dir/src/edit_distance.cc.o
[ 20%] Building CXX object CMakeFiles/libninja.dir/src/eval_env.cc.o
[ 21%] Building CXX object CMakeFiles/libninja.dir/src/graph.cc.o
[ 22%] Building CXX object CMakeFiles/libninja.dir/src/graphviz.cc.o
[ 24%] Building CXX object CMakeFiles/libninja.dir/src/json.cc.o
[ 25%] Building CXX object CMakeFiles/libninja.dir/src/line_printer.cc.o
[ 27%] Building CXX object CMakeFiles/libninja.dir/src/manifest_parser.cc.o
[ 28%] Building CXX object CMakeFiles/libninja.dir/src/metrics.cc.o
[ 29%] Building CXX object CMakeFiles/libninja.dir/src/missing_deps.cc.o
[ 31%] Building CXX object CMakeFiles/libninja.dir/src/parser.cc.o
[...]

PR #2548, opened 2025-01-01
Create linux-musl.yml

userdocs:
A static musl crossbuild binaries workflow to test building against musl cc across a range of arch targets,

A continuation of this; https://github.com/ninja-build/ninja/issues/2426 

It uses a variation of this https://github.com/richfelker/musl-cross-make via https://github.com/userdocs/qbt-musl-cross-make which provides the crossbuild toolchains on demand, here using the docker images.

They are using alpine linux settings from here https://git.alpinelinux.org/aports/tree/main/gcc/APKBUILD#n287

You could also release the static binaries using this.

Take about 3 minutes per build https://github.com/userdocs/ninja/actions/runs/12575349270

### DIFF
--- a/.github/workflows/linux-musl.yml
+++ b/.github/workflows/linux-musl.yml
@@ -33,6 +33,9 @@ jobs:
 
     name: ${{ matrix.build_method}} ${{ matrix.binary_name }}
 
+    env:
+      container_name: crossbuild
+
     steps:
       - name: Host - checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


jhasse:
Why though?

jhasse:
Ah I remember now. Can you simplify this? I think one job is more than enough. As Ninja is C++ I would suspect musl fans are already crying their eyes out anyway and will prefer https://github.com/michaelforney/samurai.

jhasse:
Actually I don't want any artifacts for musl.

jhasse:
But for x86 it works?

jhasse:
Thanks! Sorry for the back and forth :)

Issue #2444, opened 2024-05-07
Stuck at "Cleaning... X files" for long periods of time


user062:
Ninja version:1.12.0
OS: Archlinux

Whenever I run `ninja` to build a project, it gets stuck at `Cleaning... X files` for a long period of time (sometimes up to 5 minutes) even when there are no files to clean.
While it is stuck, there is 0 CPU usage, so it's not doing anything.

htop shows that a spawned process is in a zombie state waiting to be reaped:
![ninja_meson_problem](https://github.com/ninja-build/ninja/assets/48716714/0b56072d-1dac-46da-b547-388537156989)

Here is a related meson report: https://github.com/mesonbuild/meson/discussions/10645

I appreciate any help debugging this further.

Thanks a lot.

user062:
@scivision, This happens when building [GIMP](https://gitlab.gnome.org/GNOME/gimp/).

I investigated a bit. I found that it's waiting on this ppoll call:
[subprocess-posix.cc:267](https://github.com/ninja-build/ninja/blob/master/src/subprocess-posix.cc#L267)

ntrrgc:
This is occasionally also hitting GStreamer (also using meson).

Attaching a debugger I could confirm ninja was waiting inside ppoll() for a events on a pipe (the only pipe open in that process, arguably the one given to meson, fd 4, listening to `POLLIN | POLLPRI`).

The weird thing is that doing `ps -n -P |grep <pipe inode>` only returned the ninja process, meaning that the other end of the pipe had already been closed. meson was a zombie (`<defunt>`) process at that point.

When I detached the debugger ninja unstuck itself.

There is still the possiblity that meson shared that pipe with another process I didn't catch and that process terminated somewhere between the time I attached the debugger and inspected for processes that had that same pipe open. Otherwise this would hint to a kernel bug, because you would expect ppoll() to return and notify `POLLHUP` if the other end of the pipe was closed and all data has been read (and if not all data has been read, `ppoll()` should still return with a `POLLIN` event while data is available).

ntrrgc:
I managed to reproduce the bug a second time, and this time I caught who was keeping the write-side of the pipe open.

```
$ ls /proc/801351/fd
total 0
lrwx------ 1 ntrrgc ntrrgc 64 Jan  8 14:58 0 -> /dev/pts/1
lrwx------ 1 ntrrgc ntrrgc 64 Jan  8 14:58 1 -> /dev/pts/1
[...]

PR #2550, opened 2025-01-10
Replace covert pipe with SIGCHLD handler

ntrrgc:
For background, see https://github.com/ninja-build/ninja/issues/2444#issuecomment-2577999973.

In short, when running subprocesses that share the terminal, ninja intentionally leaves a pipe open before exec() so that it can use EOF from that pipe to detect when the subprocess has exited.

That mechanism is problematic: If the subprocess ends up spawning background processes (e.g. sccache), those would also inherit the pipe by default. In that case, ninja may not detect process termination until all background processes have quitted.

This patch makes it so that, instead of propagating the pipe file descriptor to the subprocess without its knowledge, ninja keeps both ends of the pipe to itself, and uses a SIGCHLD handler to close the write end of the pipe when the subprocess has truly exited.

During testing I found Subprocess::Finish() lacked EINTR retrying, which made ninja crash prematurely. This patch also fixes that.

Fixes https://github.com/ninja-build/ninja/issues/2444

### DIFF
--- a/src/subprocess-posix.cc
+++ b/src/subprocess-posix.cc
@@ -36,6 +36,10 @@ extern char** environ;
 
 using namespace std;
 
+namespace {
+  ExitStatus ParseExitStatus(int status);
+}
+
 Subprocess::Subprocess(bool use_console) : fd_(-1), pid_(-1),
                                            use_console_(use_console) {
 }
@@ -49,26 +53,34 @@ Subprocess::~Subprocess() {
 }
 
 bool Subprocess::Start(SubprocessSet* set, const string& command) {
-  int output_pipe[2];
-  if (pipe(output_pipe) < 0)
-    Fatal("pipe: %s", strerror(errno));
-  fd_ = output_pipe[0];
+  int subproc_stdout_fd = -1;
+  if (use_console_) {
+    fd_ = -1;
+  } else {
+    int output_pipe[2];
+    if (pipe(output_pipe) < 0)
+      Fatal("pipe: %s", strerror(errno));
+    fd_ = output_pipe[0];
+    subproc_stdout_fd = output_pipe[1];
 #if !defined(USE_PPOLL)
-  // If available, we use ppoll in DoWork(); otherwise we use pselect
-  // and so must avoid overly-large FDs.
-  if (fd_ >= static_cast<int>(FD_SETSIZE))
-    Fatal("pipe: %s", strerror(EMFILE));
+    // If available, we use ppoll in DoWork(); otherwise we use pselect
+    // and so must avoid overly-large FDs.
+    if (fd_ >= static_cast<int>(FD_SETSIZE))
+      Fatal("pipe: %s", strerror(EMFILE));
 #endif  // !USE_PPOLL
-  SetCloseOnExec(fd_);
+    SetCloseOnExec(fd_);
+  }
 
   posix_spawn_file_actions_t action;
   int err = posix_spawn_file_actions_init(&action);
   if (err != 0)
     Fatal("posix_spawn_file_actions_init: %s", strerror(err));
 
-  err = posix_spawn_file_actions_addclose(&action, output_pipe[0]);
-  if (err != 0)
-    Fatal("posix_spawn_file_actions_addclose: %s", strerror(err));
+  if (!use_console_) {
+    err = posix_spawn_file_actions_addclose(&action, fd_);
+    if (err != 0)
+      Fatal("posix_spawn_file_actions_addclose: %s", strerror(err));
+  }
 
   posix_spawnattr_t attr;
   err = posix_spawnattr_init(&attr);
@@ -97,18 +109,17 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
       Fatal("posix_spawn_file_actions_addopen: %s", strerror(err));
     }
 
-    err = posix_spawn_file_actions_adddup2(&action, output_pipe[1], 1);
+    err = posix_spawn_file_actions_adddup2(&action, subproc_stdout_fd, 1);
     if (err != 0)
       Fatal("posix_spawn_file_actions_adddup2: %s", strerror(err));
-    err = posix_spawn_file_actions_adddup2(&action, output_pipe[1], 2);
+    err = posix_spawn_file_actions_adddup2(&action, subproc_stdout_fd, 2);
     if (err != 0)
       Fatal("posix_spawn_file_actions_adddup2: %s", strerror(err));
-    err = posix_spawn_file_actions_addclose(&action, output_pipe[1]);
+    err = posix_spawn_file_actions_addclose(&action, subproc_stdout_fd);
     if (err != 0)
       Fatal("posix_spawn_file_actions_addclose: %s", strerror(err));
-    // In the console case, output_pipe is still inherited by the child and
-    // closed when the subprocess finishes, which then notifies ninja.
   }
+
 #ifdef POSIX_SPAWN_USEVFORK
   flags |= POSIX_SPAWN_USEVFORK;
 #endif
@@ -130,7 +141,8 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
   if (err != 0)
     Fatal("posix_spawn_file_actions_destroy: %s", strerror(err));
 
-  close(output_pipe[1]);
+  if (!use_console_)
+    close(subproc_stdout_fd);
   return true;
 }
 
@@ -147,13 +159,32 @@ void Subprocess::OnPipeReady() {
   }
 }
 
-ExitStatus Subprocess::Finish() {
-  assert(pid_ != -1);
-  int status;
-  if (waitpid(pid_, &status, 0) < 0)
-    Fatal("waitpid(%d): %s", pid_, strerror(errno));
-  pid_ = -1;
 
+bool Subprocess::TryFinish(int waitpid_options) {
+  assert(pid_ != -1);
+  int status, ret;
+  while ((ret = waitpid(pid_, &status, waitpid_options)) < 0) {
+    if (errno != EINTR)
+      Fatal("waitpid(%d): %s", pid_, strerror(errno));
+  }
+  if (ret == 0)
+    return false; // Subprocess is alive (WNOHANG-only).
+  pid_ = -1;
+  exit_status_ = ParseExitStatus(status);
+  return true; // Subprocess has terminated.
+}
+
+ExitStatus Subprocess::Finish() {
+  if (pid_ != -1) {
+    TryFinish(0);
+    assert(pid_ == -1);
+  }
+  return exit_status_;
+}
+
+namespace {
+
+ExitStatus ParseExitStatus(int status) {
 #ifdef _AIX
   if (WIFEXITED(status) && WEXITSTATUS(status) & 0x80) {
     // Map the shell's exit code used for signal failure (128 + signal) to the
@@ -176,18 +207,29 @@ ExitStatus Subprocess::Finish() {
   return ExitFailure;
 }
 
+} // anonymous namespace
+
 bool Subprocess::Done() const {
-  return fd_ == -1;
+  // Console subprocesses share console with ninja, and we consider them done
+  // when they exit.
+  // For other processes, we consider them done when we have consumed all their
+  // output and closed their associated pipe.
+  return (use_console_ && pid_ == -1) || (!use_console_ && fd_ == -1);
 }
 
 const string& Subprocess::GetOutput() const {
   return buf_;
 }
 
-int SubprocessSet::interrupted_;
+volatile sig_atomic_t SubprocessSet::interrupted_;
+volatile sig_atomic_t SubprocessSet::s_sigchld_received;
 
 void SubprocessSet::SetInterruptedFlag(int signum) {
   interrupted_ = signum;
+}
+
+void SubprocessSet::SigChldHandler(int signo, siginfo_t* info, void* context) {
+  s_sigchld_received = 1;
 }
 
 void SubprocessSet::HandlePendingInterruption() {
@@ -206,11 +248,14 @@ void SubprocessSet::HandlePendingInterruption() {
 }
 
 SubprocessSet::SubprocessSet() {
+  // Block all these signals.
+  // Their handlers will only be enabled during ppoll/pselect().
   sigset_t set;
   sigemptyset(&set);
   sigaddset(&set, SIGINT);
   sigaddset(&set, SIGTERM);
   sigaddset(&set, SIGHUP);
+  sigaddset(&set, SIGCHLD);
   if (sigprocmask(SIG_BLOCK, &set, &old_mask_) < 0)
     Fatal("sigprocmask: %s", strerror(errno));
 
@@ -223,6 +268,25 @@ SubprocessSet::SubprocessSet() {
     Fatal("sigaction: %s", strerror(errno));
   if (sigaction(SIGHUP, &act, &old_hup_act_) < 0)
     Fatal("sigaction: %s", strerror(errno));
+
+  memset(&act, 0, sizeof(act));
+  act.sa_flags = SA_SIGINFO | SA_NOCLDSTOP;
+  act.sa_sigaction = SigChldHandler;
+  if (sigaction(SIGCHLD, &act, &old_chld_act_) < 0)
+    Fatal("sigaction: %s", strerror(errno));
+}
+
+void SubprocessSet::CheckConsoleProcessTerminated() {
+  if (!s_sigchld_received)
+    return;
+  for (auto i = running_.begin(); i != running_.end(); ) {
+    if ((*i)->use_console_ && (*i)->TryFinish(WNOHANG)) {
+      finished_.push(*i);
+      i = running_.erase(i);
+    } else {
+      ++i;
+    }
+  }
 }
 
 SubprocessSet::~SubprocessSet() {
@@ -233,6 +297,8 @@ SubprocessSet::~SubprocessSet() {
   if (sigaction(SIGTERM, &old_term_act_, 0) < 0)
     Fatal("sigaction: %s", strerror(errno));
   if (sigaction(SIGHUP, &old_hup_act_, 0) < 0)
+    Fatal("sigaction: %s", strerror(errno));
+  if (sigaction(SIGCHLD, &old_chld_act_, 0) < 0)
     Fatal("sigaction: %s", strerror(errno));
   if (sigprocmask(SIG_SETMASK, &old_mask_, 0) < 0)
     Fatal("sigprocmask: %s", strerror(errno));
@@ -262,9 +328,18 @@ bool SubprocessSet::DoWork() {
     fds.push_back(pfd);
     ++nfds;
   }
+  if (nfds == 0) {
+    // Add a dummy entry to prevent using an empty pollfd vector.
+    // ppoll() allows to do this by setting fd < 0.
+    pollfd pfd = { -1, 0, 0 };
+    fds.push_back(pfd);
+    ++nfds;
+  }
 
   interrupted_ = 0;
+  s_sigchld_received = 0;
   int ret = ppoll(&fds.front(), nfds, NULL, &old_mask_);
+  CheckConsoleProcessTerminated();
   if (ret == -1) {
     if (errno != EINTR) {
       perror("ninja: ppoll");
@@ -273,6 +348,9 @@ bool SubprocessSet::DoWork() {
     return IsInterrupted();
   }
 
+  // ppoll/pselect prioritizes file descriptor events over a signal delivery.
+  // However, if the user is trying to quit ninja, we should react as fast as
+  // possible.
   HandlePendingInterruption();
   if (IsInterrupted())
     return true;
@@ -315,7 +393,9 @@ bool SubprocessSet::DoWork() {
   }
 
   interrupted_ = 0;
-  int ret = pselect(nfds, &set, 0, 0, 0, &old_mask_);
+  s_sigchld_received = 0;
+  int ret = pselect(nfds, (nfds > 0 ? &set : nullptr), 0, 0, 0, &old_mask_);
+  CheckConsoleProcessTerminated();
   if (ret == -1) {
     if (errno != EINTR) {
       perror("ninja: pselect");
@@ -324,6 +404,9 @@ bool SubprocessSet::DoWork() {
     return IsInterrupted();
   }
 
+  // ppoll/pselect prioritizes file descriptor events over a signal delivery.
+  // However, if the user is trying to quit ninja, we should react as fast as
+  // possible.
   HandlePendingInterruption();
   if (IsInterrupted())
     return true;
--- a/src/subprocess.h
+++ b/src/subprocess.h
@@ -68,8 +68,27 @@ struct Subprocess {
   char overlapped_buf_[4 << 10];
   bool is_reading_;
 #else
+  /// The file descriptor that will be used in ppoll/pselect() for this process,
+  /// if any. Otherwise -1.
+  /// In non-console mode, this is the read-side of a pipe that was created
+  /// specifically for this subprocess. The write-side of the pipe is given to
+  /// the subprocess as combined stdout and stderr.
+  /// In console mode no pipe is created: fd_ is -1, and process termination is
+  /// detected using the SIGCHLD signal and waitpid(WNOHANG).
   int fd_;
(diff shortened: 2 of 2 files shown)


src/subprocess-posix.cc
```diff
@@ -117,20 +120,42 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
   if (err != 0)
     Fatal("posix_spawnattr_setflags: %s", strerror(err));
 
+  if (use_console_) {
+    // Block SIGCHLD to prevent a race where the new subprocess exits before we
+    // register its PID.
```
digit-google, line 125: This should not be necessary as SIGCHLD should be blocked when this function is called. See comment below for more details.

Apart from that, a proper version of this code should save the old procmask and restore it on line 146 instead.

src/subprocess-posix.cc
```diff
@@ -223,6 +337,12 @@ SubprocessSet::SubprocessSet() {
     Fatal("sigaction: %s", strerror(errno));
   if (sigaction(SIGHUP, &act, &old_hup_act_) < 0)
     Fatal("sigaction: %s", strerror(errno));
+
+  memset(&act, 0, sizeof(act));
+  act.sa_flags = SA_SIGINFO | SA_NOCLDSTOP;
+  act.sa_sigaction = SigChldHandler;
+  if (sigaction(SIGCHLD, &act, &old_chld_act_) < 0)
+    Fatal("sigaction: %s", strerror(errno));
```
digit-google, line 345: The signal should first be blocked before changing its signal handler. Add a `sigaddset(&set, SIGCHLD)` after line 327 above.

Note that the loop waiting in DoWork() uses `pselect() / ppoll()` specifically to unblock the signals atomically when waiting for events. This is why there is no reason to block it temporarily when spawning a new process.

src/subprocess-posix.cc
```diff
@@ -205,6 +232,93 @@ void SubprocessSet::HandlePendingInterruption() {
     interrupted_ = SIGHUP;
 }
 
+// SIGCHLD handling:
+
+struct PidFdEntry;
+typedef unsigned int IxEntry;
+// PidFdList is used to store the PID and file descriptor pairs of the
```
digit-google, line 239: Avoid complicated logic in the signal handler. Instead write the child pid to a self-pipe that can be waited on in DoWork() (ensure the pipe descriptors are not leaked to spawned commands too).  This avoids leaking the file descriptor to a global table, which makes reasoning about lifecycle difficult (e.g. there are code paths where this descriptor will never be closed properly in your code).

Using a linked list or any kind of map to find the fd from the pid is probably not needed. Just scan the array of Subprocess instances linearly, since command termination is not in the critical performance path (even when 1000+ of commands are launched in parallel).

ntrrgc: > Just scan the array of Subprocess instances linearly
I considered this originally, but I don't know how many processes could end up in that table if someone really runs a lot of jobs share terminal, or if that is ever an use case.

> Instead write the child pid to a self-pipe that can be waited on in DoWork()
I was trying to avoid modifications on `DoWork()`. I considered writing to pipes, but that has the additional risk that a write can potentially deadlock if the pipe buffer is full.
However, if we can rely on ppoll/pselect() returning EINTR after the first SIGCHLD signal handler execution, we could instead use a simple `int` field to communicate between the two, similar to how it's done for SIGINT:

Before ppoll(), set the "terminated PID" field to -1. Call ppoll(), if you get EINTR, check whether that field got a value other than -1. If it did, that's a process that is done. At that point we wouldn't even need the pipes, although they may still be useful to keep things orthogonal between the console and non-console use cases.

ntrrgc:
Just realized I also forgot to remove this comment:

```
    // In the console case, output_pipe is still inherited by the child and
    // closed when the subprocess finishes, which then notifies ninja.
```

src/subprocess-posix.cc
digit-google: > I considered this originally, but I don't know how many processes could end up in that table if someone really runs a lot of jobs share terminal, or if that is ever an use case.

Ah, it turns out that Ninja, in its current design, only allows to run a single "console" sub-process at a time (this is implemented elsewhere and is not visible in this file). I think this can be leveraged to avoid using a self-pipe entirely.

> I was trying to avoid modifications on DoWork(). I considered writing to pipes, but that has the additional risk that a write can potentially deadlock if the pipe buffer is full.

Technically, this is extremely unlikely. In this code, the signal handler can only run during the `pselect()` / `ppoll()` call. It would require thousands of processes to *all* terminate during that exact syscall to block the pipe buffer (which are very large these days, see https://github.com/afborchert/pipebuf?tab=readme-ov-file#some-results for some not-so-recent practical results).

But we can avoid pipes nonetheless.

> However, if we can rely on ppoll/pselect() returning EINTR after the first SIGCHLD signal handler execution, we could instead use a simple int field to communicate between the two, similar to how it's done for SIGINT:

There is no actual guarantee that the system call would return after only a single SIGCHLD signal was handled.

On the other hand, because there is only one console subprocess, it should be possible to write its pid value to a global variable that the signal handler compares to. In case of success, it would set an atomic flag to true that can be trivially checked in DoWork(). More specifically:

- Add `SubprocessSet::console_subproc_` as a `Subprocess*` pointer to the current console process if any.
  Ensure that starting a new subprocess updates the pointer if needed (and assert that only one can exist).
  
- Add two global sig_atomic_t values. One `s_console_pid`, will contain the pid of the console subprocess after it is started, or a special value (e.g. -1) to indicate there is no console process currently (which would be written in Subprocess::Finish). The second `s_console_exited` will be used as a boolean flag.

- The SIGCHLD handler simply compares the signal's pid to the value of `s_console_pid`. If they match, it sets `s_console_exited` to 1.

- In DoWork(), set `s_console_exited` to 0 before calling `pselect()` or `ppoll()`, and look at its value after the call.

Wdyt?

ntrrgc:
What is the point of `HandlePendingInterruption()`?

```c++
interrupted_ = 0;
int ret = ppoll(&fds.front(), nfds, NULL, &old_mask_);
if (ret == -1) {
  if (errno != EINTR) {
    perror("ninja: ppoll");
    return false;
  }
  CheckForProcessTermination();
  return IsInterrupted();
}

HandlePendingInterruption();
if (IsInterrupted())
  return true;
```

If there is a SIGINT pending by the time `HandlePendingInterruption()` is called, wouldn't ppoll() handle it and return EINTR on the next DoWork() cycle?

src/subprocess-posix.cc
digit-google: Ah, I am trying to wrap my head around signals again, this is all so subtle, some details coming back:

`SIGCHLD` is a standard signals, not a real one, which means that it is not queued. When several processes terminate outside of the `pselect()` call, they are collapsed into a single signal handler call during the syscall (probably passing the pid of the last process). See 
https://stackoverflow.com/questions/48750382/can-not-receive-all-the-sigchld

In other words, you we can only treat `SIGCHLD` as a boolean flag that says "some child has stopped", then have to use `waitpid(..., WNOHANG)` to scan the state of all processes of interest. Luckily for Ninja, that would be looking at the state of the single console process.

ntrrgc: Wow, that makes `si_pid` in SIGCHLD near-useless. If that is true, sigaction(2) should be updated to point it out.

ntrrgc: > The SIGCHLD handler simply compares the signal's pid to the value of s_console_pid. If they match, it sets s_console_exited to 1.

Something irks me about this: Using `si_pid` in SIGCHLD is only safe if we only have one children and therefore one potential sender of SIGCHLD (assuming none is trying to send SIGCHLD maliciously to ninja too). But if we only have one children, we already know its PID.

And yet, it feels so fragile to assume this! An alternative would be to use waitpid with WNOHANG but that also requires tweaking the Finish() code.

digit-google: I agree, it looks like the only thing that a SIGCHLD reliably indicates is that "at least one child process terminated", hence not much can be done in the signal handler, and instead the main loop should, when this even happens, call waitpid() with WNOHANG on all known child processes to verify their status.

digit-google:
Regarding `HandlePendingInterruption()`, from looking at the git log for `src/subprocess-posix.cc`, it looks like this call was introduced  by https://github.com/ninja-build/ninja/commit/f2bfbda312ac92de3656ff3ae284a7ec82e59e74 , which commit message states that:

```
ppoll/pselect prioritizes file descriptor events over
a signal delivery. So a flood of events prevents ninja
from reacting keyboard interruption by the user.

This CL adds a check for pending keyboard interruptions after
file descriptor events.
```

That would explain it.

src/subprocess-posix.cc
ntrrgc: Yesterday I retook this a little bit: The main disadvantage of `waitpid(WNOHANG)` for figuring out the terminated children is that it's a mutable operation, so it requires refactoring code so that the waitpid() is not done accidentally twice on the same PID (also, being very strict, the PID stops existing exactly from the time the waitpid() is called and can be reused).

Fortunately, I think this is perfectly doable within ninja's Subprocess and SubprocessSet design.

I got to implement it, but I must have gotten something wrong as I get `ninja: fatal: waitpid(930444): No child processes` on the first call to waitpid when running under gdb. I'll keep investigating.

---

Oh, regarding removing the self-pipe: I'm also doing that; however, be aware this also adds some complexity.

One of the parts where it adds complexity is that you have to support doing ppoll() and pselect() with zero file descriptors, and do it in a portable way. The man pages are a bit less helpful than I'd like:

* The Linux man page for pselect() and select() talks about how in the olden days people used select() with nfds=0 and the three sets, and a timeout, as «a fairly portable way to sleep with subsecond precision».

  I do hope that is indeed portable enough and it extends to pselect(), as nfds == 0 is not mentioned elsewhere in that page.

* The Linux man page for ppoll() doesn't specify _Nullable for fds. It also makes no mention of whether nfds == 0 is valid or not.

  However, you're allowed to use a negative fd inside a pollfd entry to have "unused" entries. I hope nfds == 1 with a single entry with fd < 0 is supported across operating systems; but again, the man pages don't mention this explicitly.

ntrrgc:
> ppoll/pselect prioritizes file descriptor events over
> a signal delivery. So a flood of events prevents ninja
> from reacting keyboard interruption by the user.

I see, that explains the mystery, thanks for looking it up. I will try to remember to add a comment so that it's not a mystery for the next person as well.

src/subprocess-posix.cc
ntrrgc: > I must have gotten something wrong as I get `ninja: fatal: waitpid(930444): No child processes` on the first call to waitpid when running under gdb.

What I had gotten wrong was that I had written this:

```
while ((ret = waitpid(pid_, &status, options) < 0)) {
```

instead of this:

```
while ((ret = waitpid(pid_, &status, options)) < 0) {
```

After that I did some more debugging and cleaning and I'm now sending the updated patch for review.

ntrrgc:
I should also update the commit and PR title now that this version doesn't use a self-pipe trick at all.

ntrrgc:
@digit-google Do you have any thoughts on the patch?

src/subprocess-posix.cc
```diff
@@ -262,17 +322,30 @@ bool SubprocessSet::DoWork() {
...
+    // Add a dummy entry to prevent using an empty pollfd vector.
+    // ppoll() allows to do this by setting fd < 0.
+    pollfd pfd = { -1, 0, 0 };
+    fds.push_back(pfd);
+    ++nfds;
+  }
 
   interrupted_ = 0;
+  s_sigchld_received = 0;
+  std::atomic_signal_fence(std::memory_order_release);
```
digit-google, line 335: nit: This is not necessary, this only tells the compiler to avoid re-ordering writes to the `s_sigchld_received`  variable after the fence. However:

- This is already enforced at each read or write of this variable due to the `volatile` qualifier when defining this variable.

- The next function call to `ppoll()` will act as a memory barrier as well.

src/subprocess-posix.cc
```diff
@@ -223,6 +261,26 @@ SubprocessSet::SubprocessSet() {
...
+  act.sa_flags = SA_SIGINFO | SA_NOCLDSTOP;
+  act.sa_sigaction = SigChldHandler;
+  if (sigaction(SIGCHLD, &act, &old_chld_act_) < 0)
+    Fatal("sigaction: %s", strerror(errno));
+}
+
+void SubprocessSet::CheckConsoleProcessTerminated() {
+  if (!s_sigchld_received)
+    return;
+  for (vector<Subprocess*>::iterator i = running_.begin();
```
digit-google, line 275: we are now using C++14, so I suggest you use range-based loops with `auto` here as in:

```
for (auto it = running_.begin(); it != running_.end(); ++it) {
  ...
}
```

src/subprocess-posix.cc
```diff
@@ -16,6 +16,7 @@
 
 #include <sys/select.h>
 #include <assert.h>
+#include <atomic>
```
digit-google, line 19: nit: try separating C++ and C system headers into different blocks. I recommend using `git clang-format` do that automatically. Otherwise, push the `<atomic`>  include in a separate line below 22.

src/subprocess-posix.cc
```diff
@@ -49,26 +52,32 @@ Subprocess::~Subprocess() {
 }
 
 bool Subprocess::Start(SubprocessSet* set, const string& command) {
-  int output_pipe[2];
-  if (pipe(output_pipe) < 0)
-    Fatal("pipe: %s", strerror(errno));
-  fd_ = output_pipe[0];
+  int output_pipe[2] = {-1, -1};
```
digit-google, line 55: nit: for clarity, can I ask you only define a single `subproc_stdout_fd` local variable, and only use `fd_` and `subproc_output_fd` in the rest of the code if possible, e.g.:

```
// The stdout/stderr write descriptor for non-console subprocesses.
int subproc_stdout_fd = -1;
if (use_console_) {
  fd_ = -1;
} else {
  int output_pipe[2];
  if (pipe(output_pipe) < 0) ...;
  fd_ = output_pipe[0];
  subproc_stdout_fd = output_pipe[1];
  ...
}

...

if (subproc_stdout_fd >= 0) {
  // Do not leak this descriptor to child processes. See https://github.com/ninja-build/ninja/issues/2444
  err = posix_spawn_file_actions_addclose(&action, subproc_stdout_fd);
  ...
}

....

if (subproc_stdout_fd >= 0)
  close(subproc_stdout_fd);
}
```

src/subprocess.h
```diff
@@ -68,8 +68,27 @@ struct Subprocess {
...
+  /// a certain subprocess has finished. This is done for terminal subprocesses.
+  /// However, this also causes the subprocess to be reaped before Finish() is
+  /// called, so we need to store the ExitStatus so that a later Finish()
+  /// invocation can return it.
+  ExitStatus exit_status_;
+
+  /// Call waitpid() on the subprocess with the provided options and update the
+  /// pid_ and exit_status_ fields.
+  /// Return a boolean indicating whether the subprocess has indeed terminated.
+  bool TryFinish(int options);
```
digit-google, line 91: nit: Call argument `waitpid_options` for clarity.

src/subprocess-posix.cc
ntrrgc: I'm not a big fan of `auto`, as it tends to obscure types, but I can use it.

I will not place `++i` in the loop however because this loop erases elements and therefore `i` must only be increased when an erase didn't happen.

src/subprocess-posix.cc
```diff
@@ -315,15 +387,20 @@ bool SubprocessSet::DoWork() {
...
   interrupted_ = 0;
-  int ret = pselect(nfds, &set, 0, 0, 0, &old_mask_);
+  s_sigchld_received = 0;
+  int ret = pselect(nfds, (nfds > 0 ? &set : nullptr), 0, 0, 0, &old_mask_);
   if (ret == -1) {
     if (errno != EINTR) {
       perror("ninja: pselect");
       return false;
     }
+    CheckConsoleProcessTerminated();
```
digit-google, line 397: Thanks a lot, this is definitely cleaner!

However, shouldn't this function also be called in the non-error case as well? E.g. if there is a lot of file descriptor activity + a console process termination at the same time (similar to the situation explained in the comment on lines 401-403), pselect() would not necessarily return an error, and the `s_sigchld_received` flag change would be missed.  Same goes for ppoll() below. What do you think?

ntrrgc: You're right! I'm changing that.

src/subprocess.h
```diff
@@ -96,16 +115,24 @@ struct SubprocessSet {
...
 #else
   static void SetInterruptedFlag(int signum);
-  static void HandlePendingInterruption();
+  static void SigChldHandler(int signo, siginfo_t* info, void* context);
+
   /// Store the signal number that causes the interruption.
   /// 0 if not interruption.
-  static int interrupted_;
-
+  static volatile int interrupted_;
```
jhasse, line 122: Why did you make this volatile?

src/subprocess-posix.cc
```diff
@@ -36,6 +36,8 @@ extern char** environ;
 
 using namespace std;
 
+static ExitStatus ParseExitStatus(int status);
```
jhasse, line 39: use anonymous namespace instead, see https://clang.llvm.org/extra/clang-tidy/checks/misc/use-anonymous-namespace.html

jhasse (commented):
Thanks! I had a very brief look for now.

src/subprocess.h
ntrrgc: I will make it your call whether `volatile` here stays or goes away.

digit-google: The best option is probably to use [sig_atomic_t](https://en.cppreference.com/w/c/program/sig_atomic_t) which was designed specially for this exact use case.

src/subprocess-posix.cc
```diff
@@ -176,20 +207,31 @@ ExitStatus Subprocess::Finish() {
...
+  // output and closed their associated pipe.
+  return (use_console_ && pid_ == -1) || (!use_console_ && fd_ == -1);
 }
 
 const string& Subprocess::GetOutput() const {
   return buf_;
 }
 
-int SubprocessSet::interrupted_;
+volatile int SubprocessSet::interrupted_;
```
jhasse, line 224: sig_atomic_t here

jhasse:
Thank you for this :) Subprocess management is horrible on POSIX, glad that you took the time to look into it.
Some kind of test would be great, but we can add that later. Let's get this some testing on master.

jhasse:
Unfortunately had to revert due to #2586. Please have a look.

ntrrgc:
I'll have a look.

ntrrgc:
Some initial observations: the test file hangs because it gets in an infinite loop due to a missing iterator increment in the ppoll-based implementation `SubprocessSet::DoWork()`:

```cpp
nfds_t cur_nfd = 0;
for (vector<Subprocess*>::iterator i = running_.begin();
      i != running_.end(); ) {
  int fd = (*i)->fd_;
  if (fd < 0)
    continue;  // <--- BUG: Endless loop here, the iterator is not advanced.
  assert(fd == fds[cur_nfd].fd);
  if (fds[cur_nfd++].revents) {
    (*i)->OnPipeReady();
    if ((*i)->Done()) {
      finished_.push(*i);
      i = running_.erase(i);
      continue;
    }
  }
  ++i;
}
```

Likely this codepath wasn't being reached before, so the bug remained hidden.

The pselect-based implementation of `SubprocessSet::DoWork()` doesn't miss the iterator increment and it doesn't get stuck.

I'll study this code a bit more just in case I missed something else too and send a patch.

src/subprocess-posix.cc
```diff
@@ -262,9 +328,18 @@ bool SubprocessSet::DoWork() {
     fds.push_back(pfd);
     ++nfds;
   }
+  if (nfds == 0) {
```
digit-google, line 331: @ntrrgc , can you remind me what was the rationale for this specific change? Can we instead change the ppoll() call to do something like that instead:

```
int ret = ppoll(nfds > 0 ? &fds.front() : nullptr, nfds, ...);
```

As that would match the pselect() call below.

ntrrgc: Fortunately Me of the Past left a comment underneath that explains it. I can add some context that may not have been immediately obvious:

If you look at `man 2 ppoll` in Linux you'll notice that the first argument is not marked as nullable (and further reinforcing this is the fact that other arguments are in that same function are). Since we can't safely pass a NULL, we do the next best thing, which is using a tiny list whose only entry is a dummy fd.

ppoll() will then block forever... or until a signal not masked by `old_mask_` is received! And that's how we are supposed to detect either SIGCHLD (if the subprocess exits) or SIGINT (if the user presses ^C).

Additional context:

`old_mask_` is initialized in `SubprocessSet::SubprocessSet()` with the signal mask the process ninja had before spawning subprocess. Notice however that signal masks are inheritable. If ninja's parent process masked SIGINT while spawning ninja, `old_mask_` will contain SIGINT and therefore ninja will actually not react to ^C, *intentionally*. The rationale being that if your parent process spawned you with SIGINT blocked it was for a reason and you should respect its wishes. (This is not new in my changes, that code is 13 years old according to git blame; but it's helpful to keep in mind when debugging signal-related bugs.)

digit-google: Thanks for the context, this is pretty subtle and would likely benefit from a comment to make maintenance of this code path easier in the future :)

I believe it is pretty standard to use something like pselect() with no file descriptors and a timeout to wait for a specific amount of time and/or until a signal is raised, I don't see a reason why this wouldn't work for ppoll().

Did you experience an actual failure?

I just looked at the source code for the Linux kernel and the FreeBSD one, and I don't see any reason to think a null pointer would be a problem.

I also looked at the sources for libev, libuv and libevent, which are all quite portable and heavily tested, and they essentially use calloc() and realloc() to allocate the array. Whether this returns null or not when the size is 0 is implementation-dependent according to Posix.

The Linux manpage for malloc() states that GLibc will never return NULL for a size of 0, but the one for FreeBSD or MacOS do not specify any behavior in that case. It is likely that they all return not-NULL in this case then.

Another way to address the issue would be to use `realloc()` too to allocate the array, instead of an std::vector<> instance, in which case we would have the same level of "support" as other event libraries, and a slightly simpler wait loop.

Regarding `old_mask_`, maybe we should ensure that it does not block SIGCHLD then, because I don't see the code working if the parent blocked it (though this would be highly unusual, normally SIGCHLD's handler is set to SIG_IGN instead, which technically does not block signal delivery). What do you think?

ntrrgc: > Did you experience an actual failure?

I did not test actually passing NULL, simply because the documentation implies not to do that. I did consider it for a second, but then I realized that even if it worked in Linux, it would be very hard for me to make sure it works in all operating systems ninja supports, or that it working is intentional and not an implementation detail that could change in future versions of some OS. I simply decided to play it safe by keeping my expectations low.

Notice that the POSIX man page also doesn't specify anywhere that `fds=NULL` is safe: https://pubs.opengroup.org/onlinepubs/9799919799/functions/poll.html

However, if you have a reasonable guarantee that passing {fds=NULL, nfds=0} is safe in all relevant systems, I would be perfectly okay with the change, since it would indeed make the code a bit simpler. In the absence of that guarantee, the current approach feels safer and more explicit to me than switching std::vector to realloc() or assuming non-zero return for malloc(0). The latter approaches may allow to remove a few lines of code, but the complexity would deceptively remain there.

> Regarding old_mask_, maybe we should ensure that it does not block SIGCHLD

Good point. Yes, we should. Copying old_mask_ into a new variable and using `sigdelset(&sigset, SIGCHLD)` should do it.

digit-google: > I did not test actually passing NULL, simply because the documentation implies not to do that. I did consider it for a second, but then I realized that even if it worked in Linux, it would be very hard for me to make sure it works in all operating systems ninja supports, or that it working is intentional and not an implementation detail that could change in future versions of some OS. I simply decided to play it safe by keeping my expectations low.
> 
Oh I agree with you, it's just that it is very surprising as I am not aware of any other code that does the same thing when calling `ppoll()`, hence my question for additional context. We can keep the code as-is, but I'll upload a PR that adds relevant comments.

> > Regarding old_mask_, maybe we should ensure that it does not block SIGCHLD
> 
> Good point. Yes, we should. Copying old_mask_ into a new variable and using `sigdelset(&sigset, SIGCHLD)` should do it.

Will add that too. Thanks.

Issue #2531, opened 2024-11-24
MSVC 2022 17.12.1 `configure.py --bootstrap` fails on `emhash/hash_table8.hpp`


zufuliu:
`warning C4456: declaration of 'slot' hides previous local declaration`:
https://github.com/ninja-build/ninja/blob/b40052cd6f3d7d801e035cd768b2462c06ff4ffd/src/third_party/emhash/hash_table8.hpp#L1329-L1340
https://github.com/ninja-build/ninja/blob/b40052cd6f3d7d801e035cd768b2462c06ff4ffd/src/third_party/emhash/hash_table8.hpp#L1364-L1375

full `configure.py --bootstrap --platform msvc` output:
<details>

```console
H:\llvm\ninja>configure.py --bootstrap --platform msvc
bootstrapping ninja...
depfile_parser.cc
lexer.cc
build.cc
build_log.cc
H:\llvm\ninja\src\third_party/emhash/hash_table8.hpp(1340): error C2220: the following warning is treated as an error
H:\llvm\ninja\src\third_party/emhash/hash_table8.hpp(1340): warning C4456: declaration of 'slot' hides previous local declaration
H:\llvm\ninja\src\third_party/emhash/hash_table8.hpp(1329): note: see declaration of 'slot'
H:\llvm\ninja\src\third_party/emhash/hash_table8.hpp(1340): note: the template instantiation context (the oldest one first) is
H:\llvm\ninja\src\build_log.h(101): note: see reference to class template instantiation 'emhash8::HashMap<StringPiece,V,std::hash<StringPiece>,std::equal_to<KeyT>,std::allocator<std::pair<KeyT,ValueT>>,emhash8::DefaultPolicy>' being compiled
        with
        [
            V=BuildLog::LogEntry *,
            KeyT=StringPiece,
            ValueT=BuildLog::LogEntry *
        ]
H:\llvm\ninja\src\third_party/emhash/hash_table8.hpp(1322): note: while compiling class template member function 'emhash8::HashMap<StringPiece,V,std::hash<StringPiece>,std::equal_to<KeyT>,std::allocator<std::pair<KeyT,ValueT>>,emhash8::DefaultPolicy>::size_type emhash8::HashMap<KeyT,ValueT,std::hash<StringPiece>,std::equal_to<KeyT>,std::allocator<std::pair<KeyT,ValueT>>,emhash8::DefaultPolicy>::find_filled_bucket(const KeyT &,uint64_t) noexcept const'
        with
[...]

PR #2551, opened 2025-01-11
Issue 2531 update hash-table8.hpp to latest

atamagaii:
Based on Issue #2531 I brought the latest version of hash-table8.hpp into the project. Additionally while attempting to build ninja_test I found that the compiler was getting caught up on a couple warnings so I added those to configure.py

These changes were tested against 
Microsoft (R) C/C++ Optimizing Compiler Version 19.43.34618 for x86

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -341,6 +341,8 @@ if platform.is_msvc():
               '/wd4355',
               # Disable warnings about ignored typedef in DbgHelp.h
               '/wd4091',
+              # Disable warning about signed/unsigned mismatch and _unlink in gtest
+              '/wd4389', '/wd4996',
               '/GR-',  # Disable RTTI.
               '/Zc:__cplusplus',
               # Disable size_t -> int truncation warning.
--- a/configure.py
+++ b/configure.py
@@ -364,7 +364,7 @@ else:
               '-Wno-unused-parameter',
               '-fno-rtti',
               '-fno-exceptions',
-              '-std=c++11',
+              '-std=c++14',
               '-fvisibility=hidden', '-pipe',
               '-DNINJA_PYTHON="%s"' % options.with_python]
     if options.debug:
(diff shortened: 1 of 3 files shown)


atamagaii:
The recent changes to hash_table8.hpp introduced `constexpr` functionality that exceeds the capabilities of `C++11`. As a result, I moved the main build process to use `C++14`, which provides the necessary support for these features.

However, I’m not familiar enough with the broader project requirements to determine if maintaining compatibility with `C++11` was intentional—for example, to ensure Ninja can be built by a wider range of toolchains or environments. If there are specific reasons for sticking to `C++11`, thoughts?

configure.py
```diff
@@ -341,6 +341,8 @@ def search_system_path(file_name: str) -> Optional[str]:  # type: ignore # Missi
               '/wd4355',
               # Disable warnings about ignored typedef in DbgHelp.h
               '/wd4091',
+              # Disable warning about signed/unsigned mismatch and _unlink in gtest
```
digit-google, line 344: Is it possible to fix the underlying source issues instead of disabling warnings (which has the potential of hiding bugs)?

configure.py
```diff
@@ -364,7 +364,7 @@ def search_system_path(file_name: str) -> Optional[str]:  # type: ignore # Missi
               '-Wno-unused-parameter',
               '-fno-rtti',
               '-fno-exceptions',
-              '-std=c++11',
+              '-std=c++14',
```
digit-google, line 367: I am not opposed to upgrading to C++14 (would love being able to use `std::make_unique<>` for example), but that would probably be better done in a separate PR, with an update to `CONTRIBUTING.md` that specifies our current requirements. @jhasse what do you think?

jhasse: My only requirement is: CI shouldn't fail (so it has to compile on RHEL 8). So no opposition on my side, too :)

atamagaii:
Sounds good. I'll make another branch for the C++14 changes this weekend
and do it as a separate PR. Once that's happy we can revisit this PR to fix
the MSVC issue.

PR #2552, opened 2025-01-13
Fix includes normalize win32

digit-google:
Remove the use of fixed-size buffers in `src/includes_normalize-win32.cc`  in order to support long file paths.

This also requires using GetFullPathNameW() as, surprisingly, GetFullPathNameA() would fail with long paths even when long paths are enabled on the host system running the tests.

To achieve this, introduce ConvertWin32UnicodeToUTF8() and ConvertUTF8ToWin32Unicode() functions to perform conversions properly in `src/util.cc`.

### DIFF
--- a/src/includes_normalize-win32.cc
+++ b/src/includes_normalize-win32.cc
@@ -24,12 +24,10 @@
 
 #include <windows.h>
 
-using namespace std;
-
 namespace {
 
 bool InternalGetFullPathName(const StringPiece& file_name, char* buffer,
-                             size_t buffer_length, string *err) {
+                             size_t buffer_length, std::string* err) {
   DWORD result_size = GetFullPathNameA(file_name.AsString().c_str(),
                                        buffer_length, buffer, NULL);
   if (result_size == 0) {
@@ -71,7 +69,7 @@ bool SameDriveFast(StringPiece a, StringPiece b) {
 }
 
 // Return true if paths a and b are on the same Windows drive.
-bool SameDrive(StringPiece a, StringPiece b, string* err)  {
+bool SameDrive(StringPiece a, StringPiece b, std::string* err) {
   if (SameDriveFast(a, b)) {
     return true;
   }
@@ -126,8 +124,8 @@ bool IsFullPathName(StringPiece s) {
 
 }  // anonymous namespace
 
-IncludesNormalize::IncludesNormalize(const string& relative_to) {
-  string err;
+IncludesNormalize::IncludesNormalize(const std::string& relative_to) {
+  std::string err;
   relative_to_ = AbsPath(relative_to, &err);
   if (!err.empty()) {
     Fatal("Initializing IncludesNormalize(): %s", err.c_str());
@@ -135,9 +133,9 @@ IncludesNormalize::IncludesNormalize(const string& relative_to) {
   split_relative_to_ = SplitStringPiece(relative_to_, '/');
 }
 
-string IncludesNormalize::AbsPath(StringPiece s, string* err) {
+std::string IncludesNormalize::AbsPath(StringPiece s, std::string* err) {
   if (IsFullPathName(s)) {
-    string result = s.AsString();
+    std::string result = s.AsString();
     for (size_t i = 0; i < result.size(); ++i) {
       if (result[i] == '\\') {
         result[i] = '/';
@@ -156,21 +154,23 @@ string IncludesNormalize::AbsPath(StringPiece s, string* err) {
   return result;
 }
 
-string IncludesNormalize::Relativize(
-    StringPiece path, const vector<StringPiece>& start_list, string* err) {
-  string abs_path = AbsPath(path, err);
+std::string IncludesNormalize::Relativize(
+    StringPiece path, const std::vector<StringPiece>& start_list,
+    std::string* err) {
+  std::string abs_path = AbsPath(path, err);
   if (!err->empty())
     return "";
-  vector<StringPiece> path_list = SplitStringPiece(abs_path, '/');
+  std::vector<StringPiece> path_list = SplitStringPiece(abs_path, '/');
   int i;
-  for (i = 0; i < static_cast<int>(min(start_list.size(), path_list.size()));
+  for (i = 0;
+       i < static_cast<int>(std::min(start_list.size(), path_list.size()));
        ++i) {
     if (!EqualsCaseInsensitiveASCII(start_list[i], path_list[i])) {
       break;
     }
   }
 
-  vector<StringPiece> rel_list;
+  std::vector<StringPiece> rel_list;
   rel_list.reserve(start_list.size() - i + path_list.size() - i);
   for (int j = 0; j < static_cast<int>(start_list.size() - i); ++j)
     rel_list.push_back("..");
@@ -181,8 +181,8 @@ string IncludesNormalize::Relativize(
   return JoinStringPiece(rel_list, '/');
 }
 
-bool IncludesNormalize::Normalize(const string& input,
-                                  string* result, string* err) const {
+bool IncludesNormalize::Normalize(const std::string& input, std::string* result,
+                                  std::string* err) const {
   char copy[_MAX_PATH + 1];
   size_t len = input.size();
   if (len > _MAX_PATH) {
@@ -193,7 +193,7 @@ bool IncludesNormalize::Normalize(const string& input,
   uint64_t slash_bits;
   CanonicalizePath(copy, &len, &slash_bits);
   StringPiece partially_fixed(copy, len);
-  string abs_input = AbsPath(partially_fixed, err);
+  std::string abs_input = AbsPath(partially_fixed, err);
   if (!err->empty())
     return false;
 
--- a/src/includes_normalize-win32.cc
+++ b/src/includes_normalize-win32.cc
@@ -12,32 +12,99 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "includes_normalize.h"
-
-#include "string_piece.h"
-#include "string_piece_util.h"
-#include "util.h"
+#include <string.h>
+#include <windows.h>
 
 #include <algorithm>
 #include <iterator>
 #include <sstream>
 
-#include <windows.h>
+#include "includes_normalize.h"
+#include "string_piece.h"
+#include "string_piece_util.h"
+#include "util.h"
 
 namespace {
 
-bool InternalGetFullPathName(const StringPiece& file_name, char* buffer,
-                             size_t buffer_length, std::string* err) {
-  DWORD result_size = GetFullPathNameA(file_name.AsString().c_str(),
-                                       buffer_length, buffer, NULL);
-  if (result_size == 0) {
-    *err = "GetFullPathNameA(" + file_name.AsString() + "): " +
-        GetLastErrorString();
+// Get the full path of a given filename. On success set |*path| and return
+// true. On failure, clear |path|, set |*err| then result false.
+bool InternalGetFullPathName(const StringPiece& file_name, std::string* path,
+                             std::string* err) {
+  // IMPORTANT: Using GetFullPathNameA() with a long paths will fail with
+  // "The filename or extension is too long" even if long path supported is
+  // enabled. GetFullPathNameW() must be used for this function to work!
+#if 1
+  path->clear();
+  // Convert to wide filename first.
+  std::string filename_str = file_name.AsString();
+  std::wstring wide_filename;
+  if (!ConvertUTF8ToWin32Unicode(filename_str, &wide_filename, err))
     return false;
-  } else if (result_size > buffer_length) {
-    *err = "path too long";
+
+  // Call GetFullPathNameW()
+  DWORD wide_full_size = GetFullPathNameW(wide_filename.c_str(), 0, NULL, NULL);
+  if (wide_full_size == 0) {
+    *err = "GetFullPathNameW(" +
+           std::string(wide_filename.begin(), wide_filename.end()) +
+           "): " + GetLastErrorString();
     return false;
   }
+
+  // NOTE: wide_full_size includes the null-terminating character.
+  std::wstring wide_path;
+  wide_path.resize(static_cast<size_t>(wide_full_size - 1));
+  DWORD wide_full_size2 =
+      GetFullPathNameW(wide_filename.c_str(), wide_full_size,
+                       const_cast<wchar_t*>(wide_path.data()), NULL);
+  if (wide_full_size2 == 0) {
+    *err = "GetFullPathNameA(" + filename_str + "): " + GetLastErrorString();
+    path->clear();
+    return false;
+  }
+
+  // Convert wide_path to Unicode.
+  return ConvertWin32UnicodeToUTF8(wide_path, path, err);
+#else
+  path->clear();
+  std::string filename_str = file_name.AsString();
+  DWORD full_size = GetFullPathNameA(filename_str.c_str(), 0, NULL, NULL);
+  if (full_size == 0) {
+    *err = "GetFullPathNameA(" + filename_str + "): " + GetLastErrorString();
+    return false;
+  }
+
+  // NOTE: full_size includes the null-terminating character.
+  path->resize(static_cast<size_t>(full_size - 1));
+  DWORD result2 = GetFullPathNameA(filename_str.c_str(), full_size,
+                                   const_cast<char*>(path->data()), NULL);
+  if (result2 == 0) {
+    *err = "GetFullPathNameA(" + filename_str + "): " + GetLastErrorString();
+    return false;
+  }
+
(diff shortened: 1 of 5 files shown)


jhasse:
Have you created a bug report for Microsoft? I think we should try to avoid working around bugs in Windows as much as possible, so that the underlying issue can be fixed and everyone benefits.

digit-google:
So I actually tried to file a bug, which requires registering a Microsoft account (don't have one, don't want one), and from what I can gather on various Internet comments, essentially goes into a /dev/null bucket most of the time. Even if they fixed this, existing Vista / Windows 10 / whatever installs would never get the fix, so I think we'll have to just work-around this here.

maxhmv:
@jhasse  After reading the [Maximum File Path Limitation](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation) documentation again from my understanding this [section](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#functions-without-max_path-restrictions) states that only the *W functions don't have any MAX_PATH limitations, which is in line with @digit-google [observation](https://github.com/ninja-build/ninja/pull/2552/files#diff-df243def54c33e3772ef3ea49c4d1820ab34b5681032ca56369fec091020a5d4R33) that GetFullPathNameA() fails for long paths even when the system option is set in the registry and during build.

Therefore I don't understand how working with GetFullPathNameA() can fix the problem.

Thanks to the both of you for looking into this.

maxhmv:
@jhasse While I agree with you about this being a bug on Microsoft's side (I would say the whole existence of a MAX_PATH in 2025 is a bug), i strongly doubt that this is something they will likely fix. The implications of such a change could be quite big from Microsoft's perspective, because potentially people will depend on the current (documented) behavior.

Just to clarify: 

- You will not merge the PR as is but only the #else part with the *A version, and ninja wont support long path names under Windows for the time being, right? If this is the case, as a compromise would you accept a cmake option to switch the behavior? 
- Would you merge the *W part if there is a bug report at Microsoft?

Thank you.

jhasse:
> You will not merge the PR as is but only the #else part with the *A version, and ninja wont support long path names under Windows for the time being, right?

Well, not for the normalization of long paths at least. There is a subset of features where we support long paths already.

> If this is the case, as a compromise would you accept a cmake option to switch the behavior?

No, if someone builds from source anyway, they can checkout one of the PRs that use the *W function.

> Would you merge the *W part if there is a bug report at Microsoft?

That depends on Microsoft's answer.

> There is no guarantee that this won't break in the future (though Microsoft is very good at maintaining backwards compatibility).

The best thing to make them not break it, is to rely on it ;)

> Ninja relying on A functions for long paths support could be considered a bug in Ninja itself.

We're using A functions for UTF-8, not for long path support.

jhasse:
The fact that Microsoft documented their bug does not make it not a bug.

pkasting:
This behavior is intentional on Microsoft's part, for backward compat reasons (the same reason why the A/W split was introduced alongside macros mapping to one or the other to begin with). Their assumption is that the only apps using the A versions are older apps, for whom the APIs must not change behavior, even for arguable "bugfix" reasons. So this will never be implemented in the A versions.

From Chromium's perspective, we no longer care, since we no longer build with ninja. However, as a neutral third party, this is something where ninja should change its behavior (to use the W versions, not because it wants UCS-2 _or_ because it wants this bug fix per se, but because that is what "apps which support more modern conventions such as Unicode and long paths" are assumed to use).

src/includes_normalize-win32.cc
```diff
@@ -12,32 +12,99 @@
...
+  }
+
+  // NOTE: wide_full_size includes the null-terminating character.
+  std::wstring wide_path;
+  wide_path.resize(static_cast<size_t>(wide_full_size - 1));
+  DWORD wide_full_size2 =
+      GetFullPathNameW(wide_filename.c_str(), wide_full_size,
+                       const_cast<wchar_t*>(wide_path.data()), NULL);
+  if (wide_full_size2 == 0) {
+    *err = "GetFullPathNameA(" + filename_str + "): " + GetLastErrorString();
```
pkasting: I believe this was supposed to be a W (no functional effect, but misleading).

jhasse:
> Their assumption is that the only apps using the A versions are older apps, for whom the APIs must not change behavior, even for arguable "bugfix" reasons. So this will never be implemented in the A versions.

Even the A versions absolutely change behavior, that's what manifests are for.

UCS-2 was a mistake and that's why the A API is the future (with the introduction of UTF-8). It's true that there was a time where the programming community thought that wide charsets are more modern but that has definitely shifted in recent years.

pkasting:
In theory, yes, Microsoft could have made these changes to the A APIs under the manifest gating. But in practice that's not how these specific APIs work. The A ones do not change at all, and the W ones change under manifest gates.

I consider this overly cautious, and I understand where you're coming from in principle. It is certainly how some of their APIs work, after all. But it is what it is. I could be missing something about how AppCompat and future planned registry tweaks affect this that makes Microsoft's behavior sane; either way, i think the future is not the A APIs (which are often not so much UTF-8 as ASCII or 1252, but it varies by function), but something entirely different. Until then, the change here seems like the best course to me.

src/includes_normalize-win32.cc
digit-google: Indeed, thanks. Fixed.

pkasting:
(See also https://learn.microsoft.com/en-us/windows/win32/intl/conventions-for-function-prototypes, especially the highlighted note: "New Windows applications should use Unicode...They should be written with generic functions, and should define UNICODE to compile the functions into Unicode functions." The "A" versions are for ANSI, i.e. the old Windows code pages, and may have arbitrary functional gaps; they most definitely do not guarantee UTF-8 input/output will always work.)

jhasse:
Of course Microsoft wants to shift the blame on the developers instead of fixing their shit. As I have said here and in other issues/PRs: It's a bug on Microsoft's side which I don't want to work-around.

pkasting:
Shrug; you can believe what you want. They have stated reasons, they don't consider it buggy but a principled compromise, they believe changing this is worse than not changing it. Regardless of the truth of their position, the only practical impact refusing to accept the PR here has at this point is that ninja is unusable for certain users and projects, punishing users for something they didn't cause and can't control.

As I said, I don't personally care. Chromium has ditched ninja and now Valve has too, so it doesn't affect me.

Cheers,
PK

mummynobbit:
How come long paths issue is not fixed? I remember seeing a manifest embedded in the executable in one of the latest releases of ninja (1.12.1).
For which users doesn't it work exactly?

sztomi:
@mummynobbit you are commenting on the exact issue that needs fixing. I personally encountered this when building a large-ish Qt project. `moc` generates deeply nested include paths, ninja tries (and fails) to normalize them on Windows.

@jhasse I would argue that the question of this being a bug on Microsoft's end or not is completely orthogonal to the issue. Currently this bug is present on this platform and there is a valid, documented fix for it. ninja is broken without the fix. I can see in the repo history that you are willing to merge workarounds for compatibility with other external software, so why not Windows? (even though I disagree with calling this a workaround, but whatever, let's call it that). It seems like you are letting whatever feelings you may have towards this platform or Microsoft cloud your judgement. And I'm saying this as someone who has strong feelings and opinions regarding Microsoft, Windows and their role in our society. That doesn't change the fact that they exist and widely used, so we need to live with them.

panther7:
@jhasse Is some other fix for Windows builder without this PR fix?

jhasse:
> How come long paths issue is not fixed? I remember seeing a manifest embedded in the executable in one of the latest releases of ninja (1.12.1).

The flag in the manifest only results in some of Microsoft's APIs to support long paths. This results in some cases being fixed, but other still not working.
Also there were (are?) some Ninja internal path limitations.

> For which users doesn't it work exactly?

We don't know, the bug reports haven't been a lot more detailed than "works" or "doesn't work".
It seems to depend on the Windows Version, the compiler being used, the CRT being used, Registry flags and maybe even some other factors.

> @jhasse Is some other fix for Windows builder without this PR fix?

Sorry, what do you mean?

digit-google:
I agree with @pkasting and others here that.

Independently from what you may think about Microsoft and the decisions they've made, they have been consistent about documenting that only the W functions are compatible with long path support, and that applications should be modified to drop usage of the A functions for this.

The fact that *some*, but not all, A functions do work when the manifest includes the right attribute cannot even be relied upon. In theory, though unlikely, it could break in any future Windows update. Plus the A functions that do not work, as shown in this PR, and which already create issues.

In other words, Ninja should use the documented and supported way to support long paths.
Not doing it is just punishing Ninja users for no good technical reason.

Moreover, the current, incomplete, state of long path support leads to situations where failures are confusing to developers, and not easily reproducible between different setups.

@jhasse, can you reconsider your position for the benefit of Ninja users on Windows and, to be frank, the sanity of Ninja maintainers (you included) since this issue pops up frequently in the issue tracker?

jhasse:
Yes. The discussion drifted a bit away from this PR on to the general issue, let's focus again.

Would it be possible to add a test for this? Also if you would create a PR with the `#else` version I could merge that right away as mentioned above.

(I also think that putting <> includes over "" includes is completely wrong, did clang-format do this?)

digit-google:
> Would it be possible to add a test for this?

Actually, the tests were changed to check the new behavior (but I failed to update the comments in them, so this wasn't obvious, I'll fix it).

> Also if you would create a PR with the `#else` version I could merge that right away as mentioned above.

I assume you mean just removing the use of fixed-size buffers, right? Yes I can, but the function will still fail due to the GetFullPathNameA() failure. Please confirm if this is what you really want (I see little benefit in this).

> (I also think that putting <> includes over "" includes is completely wrong, did clang-format do this?)

Indeed, this is what `clang-format` does with the current `.clang-format` settings. I try to use `git clang-format` as much as possible for my CLs so I don't have to deal with style issues myself. If you feel strongly about that style, I suggest we discuss this in a separate PR where we update `.clang-format` first.

digit-google:
I have refactored the PR into four commits for clarity:

- The first two remove some bounded buffer usage and removes the `using namespacec std`, but only fix one partial issue related to long paths.

- The last two introduce UTF-8 <> Win32 Unicode conversion functions, and fix the problem properly using GetFullPathNameW().

Both adjust the unit-tests accordingly.

panther7:
@digit-google hi, i tried your ninja (1.14.0.git) and my build always end with full rebuild "warning: premature end of file; recovering"

digit-google:
> @digit-google hi, i tried your ninja (1.14.0.git) and my build always end with full rebuild "warning: premature end of file; recovering"

That means your `.ninja_deps` is corrupted, and the truncation that is applied by Ninja in this case doesn't fix it.
I have seen similar issue crop up rarely in my project's build, and diagnosed this to an invalid checksum value (but don't have a good repro case or explanation for this).

Maybe what you are seeing is incidental, what happens if you remove the `.ninja_deps` file and rebuild in your case?

panther7:
@digit-google It's clear build, `.ninja_deps` is fresh file.

builds commands are:
```sh
# fresh repo/dir
# ...
ninja -C out/x64 electron
# ... full build
# ... done
ninja -C out/x64 electron:electron_dist_zip
# ninja: warning: premature end of file; recovering
# ... full build again
```

digit-google:
That doesn't seem related to this PR at all, why are you asking this here?

Also what do you mean exactly by "your ninja" here?

If you are talking about the tip-of-tree of github.com/ninja-build/ninja, then please file a different Github issue with reproduction steps to carry the conversation there.

If you are talking about the [Fuchsia Ninja program](https://fuchsia.dev/fuchsia-src/development/build/ninja_fuchsia_improvements?hl=en) instead, report that through the [Fuchsia public tracker](https://issues.fuchsia.dev/issues) instead. Please do not pollute upstream issues with unrelated content.

panther7:
I only say, that ninja v1.31.1 works properly, but "1.14.0.git" from `fix-includes_normalize-win32` has bug with full rebuild.

digit-google:
Thanks for the clarification. In that case what about 1.14.0.git tip-of-tree (i.e. without the `fix-includes_normalize-win32` commits)?

panther7:
I tried [1.14.0 from master](https://github.com/ninja-build/ninja/actions/runs/17777971838), and I think, that works properly.
Build [1.14.0 from `fix-includes_normalize-win32`](https://github.com/ninja-build/ninja/actions/runs/18034077615) repeating build like [this](https://github.com/ninja-build/ninja/pull/2552#issuecomment-3353999181).
@digit-google

bebuch:
I can confirm that a Ninja compiled from this branch was able to build our CMake code base on Windows 11, which consists of over 6,000 steps, including various third-party tools such as Qt, protobuf/gRPC, ActiveX, etc.

@jhasse Are more changes required? We would be very happy to see this in an offical release.

bebuch:
@jhasse ping

jhasse:
Do I understand it correctly that @panther7 found a regression?

digit-google:
Sorry for the long answer.

> Do I understand it correctly that @panther7 found a regression?

This is unclear, @panther7 , can you clarify exactly what you are seeing? As I wrote previously, it looks like you are trigerring another Ninja bug that may be unrelated to this PR. Exact repro steps would thus be very welcomed here.

lygstate:
I am also facing this issue, don't know if its another Ninja bug
```
[0/245] Running ninja for QtWebEngineCore in D:/work/xtal/xtal-wasm/build-msvc/qtwebengine/src/core/Release/AMD64ninja: warning: premature end of file; recovering
ninja: Entering directory `D:/work/xtal/xtal-wasm/build-msvc/qtwebengine/src/core/Release/AMD64'
[215/18340] CXX obj/base/allocator/partition_allocator/src/partition_alloc/allocator_base/process_handle_win.obj
C:\Program Files (x86)\Windows Kits\10\\include\10.0.26100.0\\um\handleapi.h(27): warning C4005: 'INVALID_HANDLE_VALUE': macro redefinition
../../../../../../qt-everywhere-src-6.10.2/qtwebengine/src/3rdparty/chromium/base/allocator/partition_allocator/src\partition_alloc/partition_alloc_base/win/windows_types.h(69): note: see previous definition of 'INVALID_HANDLE_VALUE'
[295/18340] CC obj/third_party/nasm/nasm/outmacho.obj
```

lygstate:
> Sorry for the long answer.
> 
> > Do I understand it correctly that @panther7 found a regression?
> 
> This is unclear, @panther7 , can you clarify exactly what you are seeing? As I wrote previously, it looks like you are trigerring another Ninja bug that may be unrelated to this PR. Exact repro steps would thus be very welcomed here.

```c
      unsigned checksum = *reinterpret_cast<unsigned*>(buf + size - 4);
      int expected_id = ~checksum;
      int id = static_cast<int>(nodes_.size());
      if (id != expected_id || node->id() >= 0) {
        read_failed = true;
        break;
      }
```
anything wrong with this?
`id == expected_id  but node->id() >=0`

I found `../../../../../../qt-everywhere-src-6.10.2/qtwebengine/src/3rdparty/chromium/third_party/protobuf/src/google/protobuf/arenaz_sampler.h` appeared twice
in `.ninja_deps`
so what's the cause?

lygstate:
> Sorry for the long answer.
> 
> > Do I understand it correctly that @panther7 found a regression?
> 
> This is unclear, @panther7 , can you clarify exactly what you are seeing? As I wrote previously, it looks like you are trigerring another Ninja bug that may be unrelated to this PR. Exact repro steps would thus be very welcomed here.

I am verfied the issue is triggered this MR, I redo this mr at https://github.com/ninja-build/ninja/pull/2723, and now it's fine

itavero:
Our team is again facing issues on Windows 11 due to this bug. It's a bit unclear to me who's in the lead now and what needs to happen to get either this PR or #2723 to be merged.

Is there anything we (as users. potential first-time contributors) can do to help?

jhasse:
Yes, complain to Microsoft.

itavero:
> Yes, complain to Microsoft.

I thought people tried that already?
I also fail to see how that would help to get one of these PRs merged.

jhasse:
In my opinion it wasn't done enough. Merging this PR takes the pressure off, so I'm very reluctant to help the billion dollar company in my free time.

The more I think about it ... fixing this also improves Windows as a development platform in general. Merging this might actually be unethical to me :O

lygstate:
> Yes, complain to Microsoft.

The complain didn't have any good things, can we just deal with it first? as many people suffer from it and cost so many time to deal with it, can we just just keep complain Microsoft and still get it working

lygstate:
> Our team is again facing issues on Windows 11 due to this bug. It's a bit unclear to me who's in the lead now and what needs to happen to get either this PR or #2723 to be merged.
> 
> Is there anything we (as users. potential first-time contributors) can do to help?

https://github.com/ninja-build/ninja/pull/2723 Solved this issue:
https://github.com/ninja-build/ninja/pull/2552#issuecomment-3864819332

PR #2554, opened 2025-01-17
Update linux.yml

userdocs:
use ubuntu-24.04-arm instead of emulation. 10 minutes down to 1:40 from 10:40

image

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -12,140 +12,140 @@ jobs:
     container:
       image: fedora:40
     steps:
-    - uses: actions/checkout@v4
-    - name: Install dependencies
-      run: dnf install -y ninja-build cmake gtest-devel re2c clang util-linux clang-tools-extra
-    - name: Linting
-      run: misc/ci.py
-    - name: Configure with CMake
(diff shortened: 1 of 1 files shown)


jhasse:
Thanks! Can you do the cmake change, too?

jhasse:
The ARM build crashes now. Any idea or should I revert his PR for now?

digit-google:
I don't see the ARM build crashes, on the other hand, the Linux one fails on the spell-check step.
Link to the current failure from the latest pushed PR:

https://github.com/ninja-build/ninja/actions/runs/13019410459/job/36316411606

Either fix it locally, or revert?
Ideally we should probably disable spell checking in third_party/, there is no benefit, only pain, in doing so.
I don't know how to do that though.

jhasse:
@digit-google You can see it here for example: #2561

I think it's easier to fix the typo than to exclude third_party/. At least for now.

digit-google:
Thanks for the link. Wow, a segmentation fault at link time. No idea what's happening here for such a trivial change.
That looks like a flaky failure. Is it possible to restart it to verify?

I agree on fixing the typo immediately, but the problem will keep re-appearing if we don't find a way to exclude third_party/ :-)

PR #2556, opened 2025-01-21
Issue 2531 fix vs2022 build

atamagaii:
Okay my plan here is to take a multi stage approach to getting the latest hash_table8 moved into the project.

1. [This PR] Fix all VS2022 build errors including the original shadowing error mentioned in #2531 
2. Target c++14
3. Pull in the latest version of hash_table8

### Changes

1. Fixed the variable shadowing in hash_table8. The justification for fixing here instead of taking version 1.7 directly from the emhash project is that we are targeting c++11 and can't support the latest version.
2. We had defined the macro ```#define unlink _unlink``` but the compiler still viewed 'unlink' as a POSIX function. I created a function that injects the correct version of unlink depending on the platform.
3. Lots of sign mismatch warnings. Logically something like ```EXPECT_EQ(list.size(), 1);``` makes sense, but the compiler doesn't like it. Ideally we would be matching all of our assertions to the exact type. But realistically ```EXPECT_EQ(list.size(), size_t(1));``` and ```EXPECT_EQ(list.size(), 1u);``` are going to be equivalent for small values.

### DIFF
--- a/src/build_log.cc
+++ b/src/build_log.cc
@@ -237,7 +237,7 @@ LoadStatus BuildLog::Load(const std::string& path, std::string* err) {
       }
       if (invalid_log_version) {
         fclose(file);
-        unlink(path.c_str());
+        _unlink(path.c_str());
         // Don't report this as a failure. A missing build log will cause
         // us to rebuild the outputs anyway.
         return LOAD_NOT_FOUND;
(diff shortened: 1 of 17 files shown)


jhasse:
Thanks :)

PR #2560, opened 2025-01-28
Migrate to c++14

atamagaii:
This is related to #2531 

Before we can bring in the latest version of hash_table8 we need to migrate to c++14. This should also bring some quality of life to future development as developers will have access to ```std::make_unique```. There weren't any major deprecations in c++14 so we shouldn't see any warnings there if we ever happen to remove ```-Wno-deprecated```. 

### What changed

Updated configure.py to...
1. set cflags to target c++14.
2. remove a function that replaced c++11 with c++14 when building ninja_test as we are now targeting the required version.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -623,7 +623,7 @@ if gtest_src_dir:
     n.comment('Tests all build into ninja_test executable.')
 
     # Test-specific version of cflags, must include the GoogleTest
-    # include directory. 
+    # include directory.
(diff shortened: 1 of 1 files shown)


atamagaii:
@jhasse I noticed that we are getting CI failures on ARM64. The seg fault makes me wonder if it is a problem with the image... has this been a problem in the past?

jhasse:
See #2554

PR #2561, opened 2025-01-28
Update hash_table8.hpp

userdocs:
fixes workflow job failure: https://github.com/ninja-build/ninja/actions/runs/13009149905/job/36282726008


jhasse:
Should also be fixed here: https://github.com/ktprime/emhash/blob/master/hash_table8.hpp#L1467

jhasse:
Thanks :)

PR #2574, opened 2025-02-19
build_test: ensure dry runs with dyndep files work

mathstuf:
Released versions of `ninja` would fail to perform `ninja -n` when a `dyndep`-containing rule existed. `ninja` would fail to load the `dyndep` file that the (unexecuted) command did not run.

The behavior was fixed in 311bf93 (Propagate jobs' exit codes to the ninja's exit code, 2024-12-03). Add a test to ensure it keeps working.

---
Cc: @Felixoid @bradking @jhasse

### DIFF
--- a/src/build_test.cc
+++ b/src/build_test.cc
@@ -2025,6 +2025,29 @@ TEST_F(BuildDryRun, AllCommandsShown) {
   ASSERT_EQ(3u, command_runner_.commands_ran_.size());
 }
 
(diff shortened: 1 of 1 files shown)


mathstuf:
The added test fails before commit 311bf9341687db433c86731894443faa810b87b8 (from #2540) and passes afterwards.

Felixoid:
So, I fixed something more unintentionally? It's a nice bonus! :tada:

Issue #1261, opened 2017-03-28
Long build line can hit MAX_ARG_STRLEN on Linux


AndreasKempfNEC:
Dear Ninja team,

while trying to port a large codebase (> 10000 files) to the Meson build system on Linux, I noticed that really long build lines can fail without providing an error message explaining the problem. This can easily be replicated by, e.g., the following (useless) Ninja build file with files main.c and test.c:

```
cc = gcc 

rule build
  command = $cc $in -o $out

build prog: build main.c test.c test.c test.c ... [replicate thousands of times]
```

```
$ ninja
... test.c [replicate thousands of times] ... -o prog
ninja: build stopped: subcommand failed.
```

More specifically, it seems to fail with line lengths longer than around 128 kiB. Having done some digging, I suspect that Ninja hits the MAX_ARG_STRLEN limit of the Linux kernel [1, 2, 3],

> Number of arguments and maximum length of one argument
> At least on Linux 2.6, there's also a limit on the maximum number of arguments in argv[].
> On Linux 2.6.14 the function do_execve() in fs/exec.c tests if the number exceeds
>
>  PAGE_SIZE*MAX_ARG_PAGES-sizeof(void *) / sizeof(void *)
>
>On a 32-bit Linux, this is ARGMAX/4-1 (32767). This becomes relevant if the average length of arguments is smaller than 4.
>
>Since Linux 2.6.23, this function tests if the number exceeds MAX_ARG_STRINGS in <linux/binfmts.h> (2^32-1 = 4294967296-1).
>
>And as additional limit since 2.6.23, one argument must not be longer than MAX_ARG_STRLEN (131072).
>This might become relevant if you generate a long call like "sh -c 'automatically generated with many arguments'".
>(pointed out by Xan Lopez and Ralf Wildenhues) 

Without really knowing too much about the Ninja source, I would suspect the following lines to cause this,
```
const char* spawned_args[] = { "/bin/sh", "-c", command.c_str(), NULL };
  if (posix_spawn(&pid_, "/bin/sh", &action, &attr,
                  const_cast<char**>(spawned_args), environ) != 0)
Fatal("posix_spawn: %s", strerror(errno));
```
https://github.com/ninja-build/ninja/blob/9e71431e6f8323be8ced8997409cfe7a389c6583/src/subprocess-posix.cc#L103

It seems that the build line is passed to /bin/sh as a single command line argument. As explained above, this might hit a limit at some point.

[...]

PR #2580, opened 2025-03-28
Use pipe instead of 'sh -c' to spawn subprocesses

ecnelises:
Filling the whole command into 'sh -c' may hit length limit of single argument. For non-console subprocesses, use pipe to invoke sh instead.

Fixes https://github.com/ninja-build/ninja/issues/1261

### DIFF
--- a/src/subprocess-posix.cc
+++ b/src/subprocess-posix.cc
@@ -51,6 +51,7 @@ Subprocess::~Subprocess() {
 
 bool Subprocess::Start(SubprocessSet* set, const string& command) {
   int output_pipe[2];
+  int input_pipe[2] = { -1, -1 };
   if (pipe(output_pipe) < 0)
     Fatal("pipe: %s", strerror(errno));
   fd_ = output_pipe[0];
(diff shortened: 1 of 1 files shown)


src/subprocess-posix.cc
```diff
@@ -119,10 +120,32 @@ bool Subprocess::Start(SubprocessSet* set, const string& command) {
...
+      Fatal("posix_spawn_file_actions_addclose: %s", strerror(err));
+    spawned_args[1] = NULL;
+  }
   err = posix_spawn(&pid_, "/bin/sh", &action, &attr,
         const_cast<char**>(spawned_args), environ);
   if (err != 0)
     Fatal("posix_spawn: %s", strerror(err));
+  
+  if (!use_console_) {
+    ssize_t written = write(input_pipe[1], command.c_str(), command.size());
```
digit-google, line 144: Does that work for large commands?

Most systems use a default pipe buffer size of 64 kiB, some even smaller (see https://github.com/afborchert/pipebuf) so this write() call will block infinitely if `command.size()` is larger than that (and the bug referenced by this PR talks about command sizes of 128 kiB).

Linux has `fcntl(F_SETPIPE_SZ)` to change that the pipe buffer size, but I am not aware of similar features on other systems. Even the Linux pipe manpage tsates that applications should not rely on a specific capacity.

It might be more portable and simple to write long commands to a file that is passed as the first argument to the shell instead. And that would also work for console processes too.

digit-google: friendly ping?

jhasse:
Closing until comment addressed.

src/subprocess-posix.cc
ecnelises: Would a temp file-based approach require us to parse the command string just to check the argument limit, to avoid generating a command file for every single task? The alternative pessimistic scan for space separators seems simpler but could lead to various edge cases.

digit-google: It's only a size limit, not an argument count one, so I don't see why parsing would be needed here. Can you give an example of the edge cases you're thinking of?

PR #2584, opened 2025-04-09
[Doc] add cmake -DBUILD_TESTING option hint for users building with CMake

kenchung285:
Building with CMake currently defaults to building unit tests, which pulls in the full GoogleTest source. This is unnecessary for users who only need the ninja binary, and it increases build time and disk usage — especially problematic in minimal environments like Docker.

This update adds a note to the README showing how to disable test builds using: `-DBUILD_TESTING=OFF`  
This makes the build process faster and lighter for users who don't need testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,15 +49,17 @@ version of the library.
 
 ### CMake
 
+To build the ninja binary without building the unit tests, disable test building by setting `BUILD_TESTING` to `false`:
+
 ```
-cmake -Bbuild-cmake
+cmake -Bbuild-cmake -DBUILD_TESTING=false
 cmake --build build-cmake
 ```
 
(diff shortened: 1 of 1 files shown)


kenchung285:
Hi there, I recommend this description in readme because I encounter the problem during building `ninja` in docker env.  
I need to inspect CMakeLists.txt to figure out the usage of `-DBUILD_TESTING=false` option

jhasse:
Thanks, good suggestion!

Can you use 0 or OFF instead of false though?

kenchung285:
@jhasse Modified in https://github.com/ninja-build/ninja/pull/2584/commits/1c5443084c5539b4cf7e487e59e326b831f5441b

Issue #2568, opened 2025-02-04
windows device paths are handled incorrectly in depfiles


chrislimpach:
I'm using cmake with corrosion and cbindgen -- cbindgen creates gcc style dependency files, for which cmake sets "deps = gcc".  It does this since cmake version 3.29 (or at least 3.28 does not do this).

The generated depfiles use windows device paths, i.e. \\?\C:\path\to\file.

cmake's cmake_transform_depfile changes these to use forward slashes, i.e. //?/C:/path/to/file.

When these are added to .ninja_deps, the //? prefix is stripped and /C:/path/to/file can be seen in the .ninja_deps file.

Subsequent ninja runs fail with "ninja: error: FindFirstFileExA(/C:/path/to): ..."

The build.ninja file below is a standalone example showing the failure.  It needs a template.txt file in the same directory with some text.

The 1st COMMAND line shows the handling of a depfile matching what is produced by cmake's cmake_transform_depfile (i.e. containing paths with forward slashes).

The 2nd COMMAND line (commented out) shows the handling if the cmake step what skipped and the depfile contains paths with backslashes.  Note that ninja adds these paths as //?/C:/path/to/file, but doesn't strip a prefix.

\\?\C:\path\to\file is a valid pathname, and cmake arguably turns this into a less valid pathname, but since ninja ultimately stores the path with /, it would be good if it handled depfiles with slashes correctly.

```
rule CUSTOM_COMMAND
  command = $COMMAND

build output.txt: CUSTOM_COMMAND
  COMMAND = cmd.exe /c "copy template.txt output.txt& echo output.txt: //?/%CD:\=/%/template.txt >test.d"
  #COMMAND = cmd.exe /c "copy template.txt output.txt& echo output.txt: \\?\%CD%\template.txt >test.d"
  depfile = test.d
  deps = gcc
  restat = 1
```

digit-google:
Thanks for reporting this, I can reproduce something like this on Linux with the following `build.ninja` file:

```ninja
rule output
  command = touch $out && echo "$out: //?/C:/path/to/input" > $out.d
  depfile = $out.d
  deps = gcc
  restat = 1

build output.txt: output

default output.txt
```

Using the following commands:

```bash
$ ninja
[1/1] touch output.txt && echo "output.txt: //?/C:/path/to/input" > output.txt.d

$ ninja -t deps
output.txt: #deps 2, deps mtime 1738673321279399520 (VALID)
    /
    /C:/path/to/input
```

[...]

PR #2590, opened 2025-04-23
Add support for filenames with ampersands/quotes in depfile parser

Balletie:
Similar to #2569 (and issue #2568), I needed support for depfiles that have paths with ampersands or quotes in them. This patch solves it in the same way as in #2569 by just adding those characters to the same regex. I also added a test to cover these situations.

I constructed a similar minimal example to show the current behavior:

```ninja
rule output
  command = touch $out && echo "$out: test&input" > $out.d
  depfile = $out.d
  deps = gcc
  restat = 1

build output.txt: output

default output.txt
```

Inspecting the deps for that build config gives the following result with the current version:

```bash
$ ninja
[1/1] touch output.txt && echo "output.txt: test&input" > output.txt.d

$ ninja -t deps
output.txt: #deps 2, deps mtime 1745375193596114292 (VALID)
    test
    input
```

After my change, it looks like this:

```bash
$ ninja/ninja
ninja: warning: build log version is too old; starting over
[1/1] touch output.txt && echo "output.txt: test&input" > output.txt.d

$ ninja/ninja -t deps
output.txt: #deps 1, deps mtime 1745375261451409696 (VALID)
    test&input
```

My installed version of re2c (for generating the parser file) is 3.0.

Please let me know what you think!


pbaksa:
Why not also allow *;<>`
<> is not valid on windows, but maybe someone doesn't want to target windows?
cmake documentation says any character should be accepted: https://cmake.org/cmake/help/latest/command/add_custom_command.html

jhasse:
Thanks!

@pbaksa I think it would be better to avoid these characters in filenames - while it might be okay now, it will nearly always cause trouble somewhere down the line. Just not worth the trouble.

PR #2591, opened 2025-04-27
Add support for cgroup2 in CPU quota parsing

### LINKED ISSUE #2166: ParseCPUFromCGroup not supporting cgroup2
With cgroup2 there is no longer `cpu.cfs_quota_us` and `cpu.cfs_period_us` present which make the following logic https://github.com/ninja-build/ninja/blob/master/src/util.cc#L637-L651 no longer working in case you try to limit for example within Docker container context the number of CPUs with `--cpus`. Instead there is a `/sys/fs/cgroup/cpu.max` file containing the quota and period values to calculate the number of CPUs in case cgroup2 is active. The latter could for example be detected by checking the presence of `/sys/fs/cgroup/cgroup.controllers`.

SR4ven:
On systems using cgroup2, CPU quota auto detection would fail.
Now, cgroup v2 cpu.max will be parsed with a fallback to cgroup v1.

Tested with Docker on Ubuntu 24.04 with cgroup v2 and Ubuntu 20.04 with cgroup v1.

```
docker run -it --rm --cpus 2.5 ubuntu:24.04
ninja -h

...
-j N     run N jobs in parallel (0 means infinity) [default=3 on this system]
```

Fixes #2166

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -39,6 +39,7 @@
 #include <sys/time.h>
 #endif
 
+#include <algorithm>
 #include <vector>
 
 #if defined(__APPLE__) || defined(__FreeBSD__)
@@ -689,16 +690,31 @@ map<string, string> ParseMountInfo(map<string, CGroupSubSys>& subsystems) {
     MountPoint mp;
     if (!mp.parse(line))
       continue;
-    if (mp.fsType != "cgroup")
-      continue;
-    for (size_t i = 0; i < mp.superOptions.size(); i++) {
-      string opt = mp.superOptions[i].AsString();
-      map<string, CGroupSubSys>::iterator subsys = subsystems.find(opt);
-      if (subsys == subsystems.end())
+    if (mp.fsType == "cgroup") {
+      for (size_t i = 0; i < mp.superOptions.size(); i++) {
+        string opt = mp.superOptions[i].AsString();
+        map<string, CGroupSubSys>::iterator subsys = subsystems.find(opt);
+        if (subsys == subsystems.end())
+          continue;
+        string newPath = mp.translate(subsys->second.name);
+        if (!newPath.empty())
+          cgroups.insert(make_pair(opt, newPath));
+      }
+    } else if (mp.fsType == "cgroup2") {
+      // Find cgroup2 entry in format "0::/path/to/cgroup"
+      auto subsys = std::find_if(subsystems.begin(), subsystems.end(),
+                                 [](const auto& sys) {
+                                   return sys.first == "" && sys.second.id == 0;
+                                 });
+      if (subsys == subsystems.end()) {
         continue;
-      string newPath = mp.translate(subsys->second.name);
-      if (!newPath.empty())
-        cgroups.insert(make_pair(opt, newPath));
+      }
+      std::string path = mp.mountPoint.AsString();
+      if (subsys->second.name != "/") {
+        // Append the relative path for the cgroup to the mount point
+        path.append(subsys->second.name);
+      }
+      cgroups.insert(std::make_pair("cgroup2", path));
     }
   }
   return cgroups;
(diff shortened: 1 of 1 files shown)


src/util.cc
```diff
@@ -722,23 +738,74 @@ map<string, CGroupSubSys> ParseSelfCGroup() {
...
+    return -1;
+  }
+  return quota / period;
+}
+
+int ParseCPUFromCGroup() {
+  map<string, CGroupSubSys> subsystems = ParseSelfCGroup();
+  map<string, string> cgroups = ParseMountInfo(subsystems);
+  const auto cgroup2 = cgroups.find("cgroup2");
+  const auto cpu = cgroups.find("cpu");
```
jhasse, line 798: I would move this down after the cgroup2 check because it might not be needed.

src/util.cc
```diff
@@ -722,23 +738,74 @@ map<string, CGroupSubSys> ParseSelfCGroup() {
...
+  // Check for conversion errors
+  if (errno == ERANGE || period_end == period_str.c_str() ||
+      *period_end != '\0' || period <= 0) {
+    return -1;
+  }
+  return quota / period;
+}
+
+int ParseCPUFromCGroup() {
+  map<string, CGroupSubSys> subsystems = ParseSelfCGroup();
```
jhasse, line 795: Please add std::

src/util.cc
```diff
@@ -722,23 +738,74 @@ map<string, CGroupSubSys> ParseSelfCGroup() {
...
+  map<string, string> cgroups = ParseMountInfo(subsystems);
```
jhasse, line 796: Please add std::

src/util.cc
```diff
@@ -689,16 +690,31 @@ map<string, string> ParseMountInfo(map<string, CGroupSubSys>& subsystems) {
...
-    if (mp.fsType != "cgroup")
-      continue;
-    for (size_t i = 0; i < mp.superOptions.size(); i++) {
-      string opt = mp.superOptions[i].AsString();
-      map<string, CGroupSubSys>::iterator subsys = subsystems.find(opt);
-      if (subsys == subsystems.end())
+    if (mp.fsType == "cgroup") {
+      for (size_t i = 0; i < mp.superOptions.size(); i++) {
+        string opt = mp.superOptions[i].AsString();
+        map<string, CGroupSubSys>::iterator subsys = subsystems.find(opt);
```
jhasse, line 696: std:: (or auto)

src/util.cc
```diff
@@ -689,16 +690,31 @@ map<string, string> ParseMountInfo(map<string, CGroupSubSys>& subsystems) {
...
+        if (subsys == subsystems.end())
```
jhasse, line 697: Please add {}

src/util.cc
```diff
@@ -689,16 +690,31 @@ map<string, string> ParseMountInfo(map<string, CGroupSubSys>& subsystems) {
...
+          continue;
+        string newPath = mp.translate(subsys->second.name);
```
jhasse, line 699: std::

src/util.cc
```diff
@@ -689,16 +690,31 @@ map<string, string> ParseMountInfo(map<string, CGroupSubSys>& subsystems) {
...
+        if (!newPath.empty())
```
jhasse, line 700: {}

src/util.cc
```diff
@@ -689,16 +690,31 @@ map<string, string> ParseMountInfo(map<string, CGroupSubSys>& subsystems) {
...
+          cgroups.insert(make_pair(opt, newPath));
```
jhasse, line 701: emplace instead

jhasse (changes requested):
thanks!

nkltz:
Hi, this doesn't fully support cgroups2 as it doesn't take parents of nested cgroups into account:
```
$ cat /proc/self/cgroup
0::/my-own-cgroup
$ cat /sys/fs/cgroup/my-own-cgroup/cpu.max # this PR jumps here and stops if cpu.max is absent or it states max
max 100000
$ cat /sys/fs/cgroup/cpu.max # this is where the information is, we should be using 5 CPUs
500000 100000
```
This is common in kubernetes containers and I have documented this in more details in #2806. @jhasse if you get the chance to review the PR, it will be much appreciated!

PR #2594, opened 2025-05-04
add option to subtool graph

moritzx22:
the command
`ninja -t graph mytarget`
will generate a .dot file to visualize the dependency graph. This is a good working feature to understand the build.
But for large (cmake) projects, the generated .dot file is getting large as well. Generating a graphic file from those large files is practically not possible.

For the case you only want to see a subset of the graph of a dedicated target, I propose to add the following option:
`ninja -t graph -d 1 mytarget`
`ninja -t graph -d [integer] mytarget`
The -d option gives a possibility to specify the maximum dependency distance of the target and their inputs.
Without the -d flag, the behavior is as before, same as for negative values.

In the following, some examples using the ninja as an target

`ninja -t graph -d 0 ninja`
depth0
`ninja -t graph -d 1 ninja`
depth1
`ninja -t graph ninja`
original

### DIFF
--- a/src/graphviz.cc
+++ b/src/graphviz.cc
@@ -22,13 +22,20 @@
 
 using namespace std;
 
-void GraphViz::AddTarget(Node* node) {
+void GraphViz::printNodeLabel(Node* node){
+   if (labeled_nodes_.find(node) == labeled_nodes_.end()) {
+    string pathstr = node->path();
+    replace(pathstr.begin(), pathstr.end(), '\\', '/');
+    printf("\"%p\" [label=\"%s\"]\n", node, pathstr.c_str());
+    labeled_nodes_.insert(node);
+  }
+}
+
+void GraphViz::AddTarget(Node* node, const int depth) {
   if (visited_nodes_.find(node) != visited_nodes_.end())
     return;
 
-  string pathstr = node->path();
-  replace(pathstr.begin(), pathstr.end(), '\\', '/');
-  printf("\"%p\" [label=\"%s\"]\n", node, pathstr.c_str());
+  printNodeLabel(node);
   visited_nodes_.insert(node);
 
   Edge* edge = node->in_edge();
@@ -54,11 +61,15 @@ void GraphViz::AddTarget(Node* node) {
     // Can draw simply.
     // Note extra space before label text -- this is cosmetic and feels
     // like a graphviz bug.
-    printf("\"%p\" -> \"%p\" [label=\" %s\"]\n",
-           edge->inputs_[0], edge->outputs_[0], edge->rule_->name().c_str());
+    printNodeLabel(edge->inputs_[0]);
+    printf("\"%p\" -> \"%p\" [label=\" %s\"]\n", edge->inputs_[0],
+           edge->outputs_[0], edge->rule_->name().c_str());    
   } else {
-    printf("\"%p\" [label=\"%s\", shape=ellipse]\n",
-           edge, edge->rule_->name().c_str());
+    if (labeled_edges_.find(edge) == labeled_edges_.end()) {
+      printf("\"%p\" [label=\"%s\", shape=ellipse]\n", edge,
+             edge->rule_->name().c_str());
+      labeled_edges_.insert(edge);
+    }
     for (vector<Node*>::iterator out = edge->outputs_.begin();
          out != edge->outputs_.end(); ++out) {
       printf("\"%p\" -> \"%p\"\n", edge, *out);
@@ -68,13 +79,16 @@ void GraphViz::AddTarget(Node* node) {
       const char* order_only = "";
       if (edge->is_order_only(in - edge->inputs_.begin()))
         order_only = " style=dotted";
+      printNodeLabel(*in);
       printf("\"%p\" -> \"%p\" [arrowhead=none%s]\n", (*in), edge, order_only);
     }
   }
 
-  for (vector<Node*>::iterator in = edge->inputs_.begin();
-       in != edge->inputs_.end(); ++in) {
-    AddTarget(*in);
+  if (depth != 0) {
+    for (vector<Node*>::iterator in = edge->inputs_.begin();
+         in != edge->inputs_.end(); ++in) {
+      AddTarget(*in, depth < 0 ? -1 : depth - 1);
+    }
   }
 }
 
--- a/src/graphviz.h
+++ b/src/graphviz.h
@@ -30,12 +30,16 @@ struct GraphViz {
   GraphViz(State* state, DiskInterface* disk_interface)
       : dyndep_loader_(state, disk_interface) {}
   void Start();
-  void AddTarget(Node* node);
+  void AddTarget(Node* node, int depth);
   void Finish();
 
+  void printNodeLabel(Node* node);
+
   DyndepLoader dyndep_loader_;
-  std::set<Node*> visited_nodes_;
+  std::set<Node*> visited_nodes_;  
   EdgeSet visited_edges_;
+  std::set<Node*> labeled_nodes_;
+  EdgeSet labeled_edges_;
 };
 
 #endif  // NINJA_GRAPHVIZ_H_
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -379,6 +379,45 @@ bool NinjaMain::CollectTargetsFromArgs(int argc, char* argv[],
 }
 
 int NinjaMain::ToolGraph(const Options* options, int argc, char* argv[]) {
+
+  // The graph tool uses getopt, and expects argv[0] to contain the name of
+  // the tool, i.e. "graph".
+  argc++;
+  argv--;
+
+  int depth = -1; 
+
+  optind = 1;
+  int opt;
+  const option kLongOptions[] = { { "help", no_argument, NULL, 'h' },
+                                  { "depth", required_argument, NULL, 'd' },
+                                  { NULL, 0, NULL, 0 } };
+  while ((opt = getopt_long(argc, argv, "d:h0", kLongOptions, NULL)) != -1) {
+    switch (opt) {
+    case 'd':
+      depth = atoi(optarg);
+      break;   
+    case 'h':
+    default:
+      // clang-format off
+      printf(
+"Usage '-t graph [options] [targets]\n"
+"\n"
+"output graphviz dot file for targets.\n"
+"For smaller outputs try the --depth option.\n"
+"  example:\n"
+"   ninja -t graph mytarget | dot -Tsvg -ograph.svg\n"
+"Options:\n"
+"  -h, --help   Print this message.\n"
+"  -d, --depth  limit the dependency depth to inputs.\n"
+      );
+      // clang-format on
+      return 1;
+    }
+  }
+  argv += optind;
+  argc -= optind;
+
   vector<Node*> nodes;
   string err;
   if (!CollectTargetsFromArgs(argc, argv, &nodes, &err)) {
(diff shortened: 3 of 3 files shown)


moritzx22:
the command line
`ninja -t graph --help`
prints the following output
```
Usage '-t graph [options] [targets]

output graphviz dot file for targets.
For smaller outputs try the --depth option.
  example:
   ninja -t graph mytarget | dot -Tsvg -ograph.svg
Options:
  -h, --help   Print this message.
  -d, --depth  limit the dependency depth to inputs.
```

src/graphviz.cc
```diff
@@ -54,11 +61,15 @@ void GraphViz::AddTarget(Node* node) {
...
+    printf("\"%p\" -> \"%p\" [label=\" %s\"]\n", edge->inputs_[0],
+           edge->outputs_[0], edge->rule_->name().c_str());    
   } else {
-    printf("\"%p\" [label=\"%s\", shape=ellipse]\n",
-           edge, edge->rule_->name().c_str());
+    if (labeled_edges_.find(edge) == labeled_edges_.end()) {
+      printf("\"%p\" [label=\"%s\", shape=ellipse]\n", edge,
+             edge->rule_->name().c_str());
+      labeled_edges_.insert(edge);
+    }
```
digit-google, line 560: nit: Avoid double look-ups by using the result of `insert()` directly as in:

```
if (labeled_edges_.insert(edge).second) {
    printf(...)
}
```

src/graphviz.cc
```diff
@@ -22,13 +22,20 @@
...
+}
+
+void GraphViz::AddTarget(Node* node, const int depth) {
   if (visited_nodes_.find(node) != visited_nodes_.end())
     return;
 
-  string pathstr = node->path();
-  replace(pathstr.begin(), pathstr.end(), '\\', '/');
-  printf("\"%p\" [label=\"%s\"]\n", node, pathstr.c_str());
+  printNodeLabel(node);
```
digit-google: nit: Put this after the insert() call and use its result to avoid double lookups:

```
if (!visited_nodes_.insert(node).second)
  return;
  
printNodeLabel(node);
  
Edge* edge = node->in_edge();
...
```

src/graphviz.cc
```diff
@@ -68,13 +79,16 @@ void GraphViz::AddTarget(Node* node) {
...
   }
 
-  for (vector<Node*>::iterator in = edge->inputs_.begin();
-       in != edge->inputs_.end(); ++in) {
-    AddTarget(*in);
+  if (depth != 0) {
+    for (vector<Node*>::iterator in = edge->inputs_.begin();
+         in != edge->inputs_.end(); ++in) {
+      AddTarget(*in, depth < 0 ? -1 : depth - 1);
+    }
```
digit-google, line 646: For clarity, I suggest using `if (depth > 0) { ... }` here and just call `AddTarget(..., depth - 1)`.

Also, we are now supporting C++14 so feel free to use range-based loops as in:

```
if (depth > 0) {
  for (const Node* input : edge->inputs_)
    AddTarget(input, depth - 1);
}
```

src/graphviz.h
```diff
@@ -30,12 +30,16 @@ struct GraphViz {
...
   void Start();
-  void AddTarget(Node* node);
+  void AddTarget(Node* node, int depth);
   void Finish();
 
+  void printNodeLabel(Node* node);
+
   DyndepLoader dyndep_loader_;
-  std::set<Node*> visited_nodes_;
+  std::set<Node*> visited_nodes_;
```
digit-google: Please remove all trailing spaces from your commit (i.e. squash your second one into the first).

src/graphviz.cc
digit-google: Just to clarify, this also means setting `depth = INT_MAX` by default.

src/ninja.cc
```diff
@@ -379,6 +379,45 @@ bool NinjaMain::CollectTargetsFromArgs(int argc, char* argv[],
...
+      printf(
+"Usage '-t graph [options] [targets]\n"
+"\n"
+"output graphviz dot file for targets.\n"
+"For smaller outputs try the --depth option.\n"
+"  example:\n"
+"   ninja -t graph mytarget | dot -Tsvg -ograph.svg\n"
+"Options:\n"
+"  -h, --help   Print this message.\n"
+"  -d, --depth  limit the dependency depth to inputs.\n"
```
digit-google: nit: please document argument here, and clarify its meaning e.g.:

```
  -d, --depth=DEPTH  limit depth of each target dependencies sub-tree.
```

src/ninja.cc
```diff
@@ -379,6 +379,45 @@ bool NinjaMain::CollectTargetsFromArgs(int argc, char* argv[],
...
+    case 'd':
+      depth = atoi(optarg);
+      break;   
+    case 'h':
+    default:
+      // clang-format off
+      printf(
+"Usage '-t graph [options] [targets]\n"
+"\n"
+"output graphviz dot file for targets.\n"
```
digit-google: nit: Start the phrase with a capital letter (consistent to other tool help messages), e.g. "Output graphviz ..."

src/ninja.cc
```diff
@@ -389,7 +428,7 @@ int NinjaMain::ToolGraph(const Options* options, int argc, char* argv[]) {
   GraphViz graph(&state_, &disk_interface_);
   graph.Start();
   for (vector<Node*>::const_iterator n = nodes.begin(); n != nodes.end(); ++n)
-    graph.AddTarget(*n);
+    graph.AddTarget(*n, depth);
```
digit-google: Can you change this to use a ranged-based for loop for clarity, i.e.:

```
for (const Node* node : nodes)
  graph.AddTarget(node, depth);
```

src/graphviz.cc
```diff
@@ -22,13 +22,20 @@
 
 using namespace std;
 
-void GraphViz::AddTarget(Node* node) {
+void GraphViz::printNodeLabel(Node* node){
+   if (labeled_nodes_.find(node) == labeled_nodes_.end()) {
+    string pathstr = node->path();
+    replace(pathstr.begin(), pathstr.end(), '\\', '/');
+    printf("\"%p\" [label=\"%s\"]\n", node, pathstr.c_str());
+    labeled_nodes_.insert(node);
```
digit-google: Can you explain why is this using `labeled_nodes_` here while the original code that now calls this did not? It looks like this would change the output considerably if `-d` is not used.

src/graphviz.cc
moritzx22: ### Clarification cli

- `ninja -t graph ninja`
  - depth is unlimited by default
- `ninja -t graph -d 0 ninja`
  - depth is limited
  - function GraphViz::AddTarget is executed one time
  - recursion depth is set to 0
- `ninja -t graph -d 1 ninja`
  - depth is limited
  - function GraphViz::AddTarget is executed several times
  - recursion depth is set to 1

see for an example: https://github.com/ninja-build/ninja/pull/2594#issue-3038090730

moritzx22: clarification for variable depth

- 0:
  - no recursion anymore
- +1:
  - recursion depth of one is left
- +2:
  - recursion depth of two is left, and so on...
- -1:
  - recursion depth is unlimited
  -  this is the default value if no option d is chosen by the user
  - valid for any negative number

src/graphviz.cc
moritzx22: Excellent question. My first approach was not using the `labeled_nodes_`, the effect was that some of the nodes were displayed with their hashes, not with names.

This happened due to early interruption of the recursion, some of the `"print labeling"` did miss. This never happened in the original algorithm, as there was no filtering.

`labeled_nodes_` shall make sure the label is only printed once. The new algorithm is doing the labeling immediately a node/linkage is printed. The original algorithm could make the labeling in a next recursion for nodes that has already been printed. This is fine as long as the recursion does not end early...

if `-d` is not used, there is (or shall) no semantic change in the output. Just the order of the single dot commands might  change...

moritzx22: ### Note:
I think that the current algorithm could be split in two part to be more clean and future proof:
- Filtering
  - result -> `std::set<const Node*>`
- generating the output
  - print `std::set<const Node*>`

Right now the algorithm does that all in one step. I did this because code changes are much smaller and it is easier to track changes for the reviewer. The current filter option is easy to implement with current design of the algorithm.

I could think of other filter options, here some ideas:

- exclude targets
   - `ninja -t graph -x NegativeRegexPattern myTarget`
   - would print my target excluding target matching NegativeRegexPattern
 - exclude link types
   - `ninja -t graph -xOrderOnly myTarget`
   -  do not print order only links
 - `ninja -t graph -dependency MyTarget1..myTarget2..myTarget3 
   - would print all target that are in between the linkage path of all target specified
   - this would cover the case if the user wants to see the relationship of 2 or more targets, but is not interested in the remaining targets
   - this would be a more complicated filter and most likely breaking the current design of the algorithm
  
There might be more useful filter options I do not have in mind yet. But before making bigger steps and invest more effort, there should be a decision if the feature "filter option for graph" should be part of the ninja-build.

src/graphviz.cc
moritzx22: done

src/graphviz.cc
moritzx22: done

src/graphviz.h
moritzx22: done

src/ninja.cc
moritzx22: done

src/ninja.cc
moritzx22: done

src/ninja.cc
moritzx22: done. Additionally added "const correctness".

src/graphviz.cc
moritzx22: changed depth to be positive only. [590c905](https://github.com/ninja-build/ninja/commit/590c9057280092ca7f3d8fdbb82a217871f38f84)

moritzx22:
- graphviz algorithm has been redesigned(commit history is obsolete)
  1. read and modify the graph (dyndep and depsfile)
  2. apply separate independent filters to the graph and dump each 'single dependencies' to one set
  3. export the set to dot
- additional filters can be added in future
- overall goal is still to make the subtool `ninja -t graph` usable to real world builds(no matter of size)

for functionality see the help message:
```
$ ./ninja -t graph -h
Usage: ninja -t graph [options] [targets] [options] [targets] ...

Generate a Graphviz DOT file representing the build graph for the given targets.
The graph is constructed by iterating over each target and following its inputs
You can filter or group targets to reduce output size and focus on specific parts.
Grouped Options apply independently to each group of targets.

Examples:
  ninja -t graph mytarget | dot -Tsvg -ograph.svg
  ninja -t graph -x '.*substring.*' mytarget
  ninja -t graph -r myTarget1 myTarget2 myTarget3
  ninja -t graph -r myTarget1 myTarget2 -d 0 -o myTarget3 @default

Global Options:
  -h, --help              Show this help message.
  -E, --elide=LEN         Limit label length to LEN characters (default: unlimited).
  -p, --disable-dyndep    Do not load dyndep files.
  -P, --enable-dep        Force loading of dep files.
  -S, --enable-scan       Load deps and dyndeps using the standard ninja algorithm.
  -T, --statistic         Print statistics about the exported graph.

Grouped Options (apply per target group):
  -d, --depth=DEPTH       Limit depth of dependency sub-tree.
  -o, --no-order-only     Exclude order-only dependencies.
  -i, --no-implicit       Exclude implicit dependencies.
  -e, --no-explicit       Exclude explicit dependencies.
  -D, --no-gen-depload    Exclude nodes generated by deps-loader.
                          (effective only if deps/dyndeps are loaded;
                           recommended when deps are loaded)
  -x, --exclude-target=RE Exclude targets matching regex RE.
  -s, --no-siblings       Exclude sibling targets.

Traverse Modes (choose maximum one per group):
  -R, --reverse           Traverse graph in reverse order.
  -r, --relations         Show all relations between targets.
                          (ignores -s and -d)
  -Q, --query             Behave similar to 'ninja -t query'.

Special Targets:
  @root                   All targets with no outputs.
  @default                Default targets.
                          (equivalent to 'ninja -t graph' without args)

```

PR #2604, opened 2025-05-27
Update to GTest 1.16, latest version to support C++14

userdocs:
https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html

> Changed in version 4.0: Compatibility with versions of CMake older than 3.5 is removed. Calls to [cmake_minimum_required(VERSION)](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#command:cmake_minimum_required) or [cmake_policy(VERSION)](https://cmake.org/cmake/help/latest/command/cmake_policy.html#version) that do not specify at least 3.5 as their policy version (optionally via ...<max>) will produce an error in CMake 4.0 and above.

Current release used googletest v1.10.0 - https://github.com/ninja-build/ninja/blob/v1.12.1/CMakeLists.txt#L235 

This version does not meet the requirement of 3.5 [cmake_minimum_required(VERSION 2.8.8)](https://github.com/google/googletest/blob/release-1.10.0/CMakeLists.txt#L4)

Latest master here has [v1.12.1](https://github.com/ninja-build/ninja/blob/master/CMakeLists.txt#L265 ) which meets the minimum requirement https://github.com/google/googletest/blob/release-1.12.1/CMakeLists.txt#L4

There is a warning 3.10 is depreciated support will be removed next so this is the oldest version of google test to support higher than 3.10

googletest v1.14.x - https://github.com/google/googletest/blob/v1.14.x/CMakeLists.txt

issue is a new release is needed but also maybe you can bump google test version to avoid cmake 3.10 support being dropped before next release.

Tested to work with googletest latest release https://github.com/google/googletest/releases/tag/v1.17.0


userdocs:
Just noticed, google test [1.17.0](https://github.com/google/googletest/blob/v1.17.0/CMakeLists.txt#L4) is cmake 3.16, higher than ninja min of [3.15](https://github.com/ninja-build/ninja/blob/master/CMakeLists.txt#L1)

v1.16.0 is 3.13 https://github.com/google/googletest/blob/v1.16.0/CMakeLists.txt#L4 - maybe this version is the better choice?

jhasse:
Yes, we don't want to bump the CMake version for now.

@digit-google Wasn't there a reason for the version we chose?

gruenich:
1) Limiting to gtest 1.16 has the advantage, that it uses C++14 and not C++17 as gtest 1.17 does.
2) Changing the title of this merge request to "Increasing Google Test to 1.16" would help people to guess what this pull request is about.

CMakeLists.txt
```diff
@@ -262,8 +262,8 @@ if(BUILD_TESTING)
     include(FetchContent)
     FetchContent_Declare(
       googletest
-      URL https://github.com/google/googletest/archive/refs/tags/release-1.12.1.tar.gz
-      URL_HASH SHA256=81964fe578e9bd7c94dfdb09c8e4d6e6759e19967e397dbea48d1c10e45d0df2
+      URL https://github.com/google/googletest/releases/download/v1.16.0/googletest-1.16.0.tar.gz
+      URL_HASH SHA256=78c676fc63881529bf97bf9d45948d905a66833fbfa5318ea2cd7478cb98f399
```
digit-google, line 268: Very nice, I recommend adding a comment there explaining we are limited to 1.16 for now because 1.17 and above require C++17 which is not required yet by Ninja.

digit-google:
Indeed, it would be nice if the commit message was more explicit (e.g. "Update to GTest 1.16, latest version to support C++14").

CMakeLists.txt
userdocs: done

userdocs:
How does it look now?

PR #2605, opened 2025-06-05
Prevent ninja starting new job when physical memory become low.

Marc-Pierre-Barbier:
Inspired by https://github.com/ninja-build/ninja/issues/1571

this adds a new option: --keep-free-memory AMOUNT
and the amount can be specified with G/M/m/K/k/B units (defaulting to byte)

If the amount specified is greater than the available memory then we stop starting new build jobs.

Most importantly, this doesn't solve all memory issues you will need to swap if you don't put a great enough value, too big of a value will result in having very few core being used. This is purely here to stop the problem from getting worse and not fix it.

### DIFF
--- a/src/real_command_runner.cc
+++ b/src/real_command_runner.cc
@@ -81,7 +81,7 @@ size_t RealCommandRunner::CanRunMore() const {
     // Ensure that we make progress.
     capacity = 1;
 
-  //this just pause the creation of new processes when we run out of ran, there is no way to estimate the number of process we can run.
+  //this just pause the creation of new processes when we run out of ram, there is no way to estimate the number of process we can run.
   if (config_.desired_free_ram > 0 && GetFreeMemory() < config_.desired_free_ram ) {
     capacity = 0;
   }
--- a/src/real_command_runner.cc
+++ b/src/real_command_runner.cc
@@ -81,6 +81,11 @@ size_t RealCommandRunner::CanRunMore() const {
     // Ensure that we make progress.
     capacity = 1;
 
+  //this just pause the creation of new processes when we run out of ran, there is no way to estimate the number of process we can run.
+  if (config_.desired_free_ram > 0 && GetFreeMemory() < config_.desired_free_ram ) {
+    capacity = 0;
+  }
+
   return capacity;
 }
 
--- a/src/real_command_runner.cc
+++ b/src/real_command_runner.cc
@@ -77,7 +77,6 @@ size_t RealCommandRunner::CanRunMore() const {
   //this just pause the creation of new processes when we run out of ram
   if (config_.desired_free_ram > 0) {
     long memory_capacity = GetFreeMemory() / config_.desired_free_ram;
-    fprintf(stderr, "Mem cap=%ld\n", memory_capacity);
     if (memory_capacity < capacity)
       capacity = memory_capacity;
   }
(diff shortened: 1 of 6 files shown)


Marc-Pierre-Barbier:
Marked as draft until I can test on macOS so I can use sysconf.

src/util.cc
```diff
@@ -1026,3 +1029,63 @@ int platformAwareUnlink(const char* filename) {
...
+#elif defined(__linux__)
+long GetFreeMemory() {
+  static long swapped_idle = LONG_MAX;
+  struct sysinfo infos;
+  sysinfo(&infos);
+  const long swapped = (infos.totalswap - infos.freeswap);
+  //since system use commited memory normaly, we store the smallest amount we have seen to guess how much
+  // paging is non-ninja related
+  swapped_idle = std::min(swapped_idle, swapped);
+  return infos.freeram - (swapped - swapped_idle);
```
talshadow, line 1085: I guess it should be 
return (infos.totalram - infos.bufferram - infos.sharedram) * infos.mem_unit;

src/real_command_runner.cc
```diff
@@ -74,6 +74,13 @@ size_t RealCommandRunner::CanRunMore() const {
       capacity = load_capacity;
   }
 
+  //this just pause the creation of new processes when we run out of ram
+  if (config_.desired_free_ram > 0) {
+    long memory_capacity = GetFreeMemory() / config_.desired_free_ram;
+    if (memory_capacity < capacity)
+      capacity = memory_capacity;
+  }
```
talshadow, line 82: if (config_.desired_free_ram > 0) {
    long memory_capacity = GetFreeMemory() - config_.desired_free_ram;
    if (memory_capacity <= 0)
      capacity = 0;
  }

What about this approach?

src/util.cc
talshadow: I'm not sure about  "- infos.bufferram", maybe we need to omit it too

src/real_command_runner.cc
```diff
@@ -74,6 +74,13 @@ size_t RealCommandRunner::CanRunMore() const {
       capacity = load_capacity;
   }
 
+  //this just pause the creation of new processes when we run out of ram
+  if (config_.desired_free_ram > 0) {
+    long memory_capacity = GetFreeMemory() / config_.desired_free_ram;
+    if (memory_capacity < capacity)
+      capacity = memory_capacity;
```
mcprat, line 81: just do an early return instead of redefining capacity

return memory_capacity;

Marc-Pierre-Barbier: In this function, there is a check later on that does the following:

```c++
if (capacity == 0 && subprocs_.running_.empty())
    // Ensure that we make progress.
    capacity = 1;
```
That's why there is no other usage of early returns in this function, as this mechanism prevent lockups even if we have to ignore our ram condition.

kevin-robb (approved):
Random person here, I've been using a build of this branch (built via `./configure.py --bootstrap`) on Ubuntu 22 for a large CMake project, and this has worked wonderfully. Using `--keep-free-memory 4G` does the trick, making ninja usable and still faster than default Unix make. Thanks for implementing this! I'd love to see this get merged into the release version of ninja

notable-equivalent:
This is a very useful feature. What further work needs to be done to get this merged?

vulder:
The proposed functionality here would save me and a group of projects that I'm supporting a lot of trouble in our local development and CI setup. Is there anything that I can do to help move this change forward?

PR #2606, opened 2025-06-06
Support $^ escape sequence for newlines.

ildus:
Replace '$^' with the actual newlines.

Straight to use case:

```
rule one
    command = cmd /C file.bat
    rspfile = file.bat
    rspfile_content = echo "1" > $out $^echo "2"

build one.txt : one
```

This is required when the build system generates scripts, but trying to make one liners from every script is too bug prone. For example 'set ' in Windows cmd.exe requires delayed expansion when the variable (which is just set) was used in the same line. But writer of the building script could not know such details.

Another case is platforms where scripts can not be converted to oneliners in any way (I have a such use case for OpenVMS dcl).

'$^' currently gives an error (bad escape), so this change should not break the current usage of ninja in any way.

### DIFF
--- a/src/lexer.in.cc
+++ b/src/lexer.in.cc
@@ -22,8 +22,8 @@
 using namespace std;
 
 // $^ supported starting this version
-const int minNewlineEscapeVersionMajor = 1;
-const int minNewlineEscapeVersionMinor = 14;
+const int kMinNewlineEscapeVersionMajor = 1;
+const int kMinNewlineEscapeVersionMinor = 14;
 
 bool Lexer::Error(const string& message, string* err) {
   // Compute line/column.
@@ -266,9 +266,9 @@ bool Lexer::ReadEvalString(EvalString* eval, bool path, string* err) {
     "$^" {
       if (!newline_version_checked_)
       {
-        if ((manifest_version_major < minNewlineEscapeVersionMajor) ||
-            (manifest_version_major = minNewlineEscapeVersionMajor &&
-             manifest_version_minor < minNewlineEscapeVersionMinor))
+        if ((manifest_version_major < kMinNewlineEscapeVersionMajor) ||
+            (manifest_version_major = kMinNewlineEscapeVersionMajor &&
+             manifest_version_minor < kMinNewlineEscapeVersionMinor))
         {
           return Error("using $^ escape requires specifying 'ninja_required_version' with version greater or equal 1.14", err);
         }
--- a/src/lexer.cc
+++ b/src/lexer.cc
@@ -23,8 +23,8 @@
 using namespace std;
 
 // $^ supported starting this version
-const int minNewlineEscapeVersionMajor = 1;
-const int minNewlineEscapeVersionMinor = 14;
+const int kMinNewlineEscapeVersionMajor = 1;
+const int kMinNewlineEscapeVersionMinor = 14;
 
 bool Lexer::Error(const string& message, string* err) {
   // Compute line/column.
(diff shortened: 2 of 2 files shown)


doc/manual.asciidoc
```diff
@@ -917,6 +917,8 @@ paths, where a space would otherwise separate filenames.  See below.)
 `$:` :: a colon.  (This is only necessary in `build` lines, where a colon
 would otherwise terminate the list of outputs.)
 
+`$^` :: a newline. Inserts '\n' into the resulting string.
```
AJIOB: Why it named not a `$n`, similar to the `$:`?

ildus: Not so visible when surrounded with an another text, and ^ usually means the beginning like in regular expressions.

jhasse:
Thanks! For this not to cause a lot of confusion in the future I think we should print a warning (or error?) if some uses it without `ninja_minimum_version = 1.13` in the first line. Because then the error message for someone with older ninja will be more helpful. Not sure if that is easily doable with the current parser.

digit-google:
Following the regression described in #2616, should these newlines be `\r\n` on Windows instead?

ildus:
@digit-google That's a good question. I think it depends on the use case. For the scripts like in example `cmd` works just fine with both line endings, but I remember `bash` from cygwin could fail on windows endings. Although I could not reproduce it today, on my cygwin today it worked ok with '\r\n', maybe they fixed that.

ildus:
Changed behavior on Windows for consistency, will add '\r\n' instead just '\n'.

src/version.h
```diff
@@ -26,6 +26,7 @@ void ParseVersion(const std::string& version, int* major, int* minor);
 
 /// Check whether \a version is compatible with the current Ninja version,
 /// aborting if not.
-void CheckNinjaVersion(const std::string& required_version);
+void CheckNinjaVersion(const std::string& required_version, int &file_major,
+    int &file_minor);
```
digit-google: nit: Prefer using int pointers here, as it makes it more obvious that these are output parameters.

doc/manual.asciidoc
```diff
@@ -917,6 +917,8 @@ paths, where a space would otherwise separate filenames.  See below.)
 `$:` :: a colon.  (This is only necessary in `build` lines, where a colon
 would otherwise terminate the list of outputs.)
 
+`$^` :: a newline. Inserts '\n' ('\r\n' on Windows) into the resulting string.
+
```
digit-google, line 928: Please document that this is only available since Ninja 1.14 *and* that this requires `ninja_required_version` to be at least "1.14" too. It would be nice to add a phrase describing its usefulness (e.g. "useful to write build commands that require several shell lines as a single `command`  value in the build plan").

doc/manual.asciidoc
```diff
@@ -917,7 +917,13 @@ paths, where a space would otherwise separate filenames.  See below.)
...
 would otherwise terminate the list of outputs.)
 
-`$^` :: a newline. Inserts '\n' ('\r\n' on Windows) into the resulting string.
+`$^` :: a newline _(available since Ninja 1.14)_.
+
+Inserts '\n' ('\r\n' on Windows) into the resulting string. This is useful to
+write build commands that require several shell lines as a single `command`
+value in the build plan.
+
+Requires `ninja_required_version` to be specified in the build file.
```
digit-google: nit: specified and greater or equal to 1.14

src/version.h
```diff
@@ -26,7 +26,7 @@ void ParseVersion(const std::string& version, int* major, int* minor);
 
 /// Check whether \a version is compatible with the current Ninja version,
 /// aborting if not.
-void CheckNinjaVersion(const std::string& required_version, int &file_major,
-    int &file_minor);
+void CheckNinjaVersion(const std::string& required_version, int *file_major,
+    int *file_minor);
```
digit-google: nit: star goes near the type, for consistency, use `git clang-format` :)

digit-google:
Thank you very much, apart from the latest two nits, this looks good to me. @jhasse , wdyt?

ildus:
@digit-google Thanks for review!

ildus:
Hi @jhasse , any chance this PR could be merged?

src/lexer.in.cc
```diff
@@ -21,6 +21,10 @@
 
 using namespace std;
 
+// $^ supported starting this version
+const int minNewlineEscapeVersionMajor = 1;
```
jhasse: rename to kMinNewline... as per Google C++ Style

ildus: Fixed!

ildus:
Thanks! Reverted '\r\n' for Windows part.

PR #2622, opened 2025-06-27
Remove 1.13 warnings and enable -Werror for release builds

digit-google:
Remove a number of compiler warnings triggered by the 1.13.0 sources and ensure warnings are turned into errors in our CI builders, to catch similar issues early in the future.

- Add a new `--warnings-as-errors` option to `configure.py`
- Use the new option in github workflow files that invoke `configure.py`
- Use `-DCMAKE_COMPILE_WARNING_AS_ERROR=1` when configuring CMake for non-debug builds. Note that enabling it for debug builds triggers far too many obnoxious warnings on Ubuntu 24:04, so this has been left out here, including in re2-generated code, or third_party one.
- Fix minor issues found in the sources, revealed by our CI builders with these new flags enabled.

### DIFF
--- a/src/jobserver_test.cc
+++ b/src/jobserver_test.cc
@@ -331,7 +331,8 @@ TEST(Jobserver, PosixFifoClient) {
   ASSERT_TRUE(write_fd.IsValid()) << "Cannot open FIFO at: " << strerror(errno);
   for (size_t n = 0; n < kSlotCount; ++n) {
     uint8_t slot_byte = static_cast<uint8_t>('0' + n);
-    ::write(write_fd.fd_, &slot_byte, 1);
+    ssize_t ret = ::write(write_fd.fd_, &slot_byte, 1);
+    (void)ret;  // make compiler happy
   }
   // Keep the file descriptor opened to ensure the fifo's content
   // persists in kernel memory.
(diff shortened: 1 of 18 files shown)


src/jobserver_test.cc
```diff
@@ -331,7 +331,8 @@ TEST(Jobserver, PosixFifoClient) {
   ASSERT_TRUE(write_fd.IsValid()) << "Cannot open FIFO at: " << strerror(errno);
   for (size_t n = 0; n < kSlotCount; ++n) {
     uint8_t slot_byte = static_cast<uint8_t>('0' + n);
-    ::write(write_fd.fd_, &slot_byte, 1);
+    ssize_t ret = ::write(write_fd.fd_, &slot_byte, 1);
```
jhasse, line 334: I would just do `(void)::write(write_fd.fd_, &slot_byte, 1);`, but not important

jhasse:
Thanks! It's quite a lot of little changes though. Shouldn't cause any issues, but I'm always a bit scared about these.

src/jobserver_test.cc
digit-google: That's what I tried initially, but the compiler was still complaining, hence the current code.

Issue #2616, opened 2025-06-20
Ninja 1.13.0 with MSVC `link.exe` fails, apparently due to RSP line limit of 16,383 characters


Grimrukh:
When using Ninja 1.13.0 and MSVC 17 2022 to build a configured CMake project (whether via CMake with CMAKE_MAKE_PROGRAM set or directly via `ninja -C build`), it seems to run into issues when generated RSP files contain space-separated entries producing lines that are longer than 16,383 characters. When this happens, the first prefix of the first path in the long line is stripped, appearing to `link.exe` as an invalid absolute path. This does not happen with Ninja 1.12.1; it links and builds fine.

Here's the output with some names stripped:

```
[636/692] Linking CXX executable bin\some_test.exe
FAILED: [code=4294967295] bin/some_test.exe
C:\WINDOWS\system32\cmd.exe /C "cd . && "C:\path\to\cmake.exe" -E vs_link_exe --msvc-ver=1943 --intdir=dev\package\tests\CMakeFiles\some_test.dir --rc=C:\PROGRA~2\WI3CF2~1\10\bin\100226~1.0\x64\rc.exe --mt=C:\PROGRA~2\WI3CF2~1\10\bin\100226~1.0\x64\mt.exe --manifests  -- C:\PROGRA~1\MICROS~4\2022\PROFES~1\VC\Tools\MSVC\1443~1.348\bin\Hostx64\x64\link.exe /nologo @CMakeFiles\some_test.rsp  /out:bin\some_test.exe /implib:lib\some_test.lib /pdb:bin\some_test.pdb /version:0.0 /machine:x64 /debug /INCREMENTAL /subsystem:console  /GUARD:CF /CETCOMPAT /DYNAMICBASE && cd ."

LINK Pass 1: command "C:\PROGRA~1\MICROS~4\2022\PROFES~1\VC\Tools\MSVC\1443~1.348\bin\Hostx64\x64\link.exe /nologo @CMakeFiles\some_test.rsp /out:bin\some_test.exe /implib:lib\some_test.lib /pdb:bin\some_test.pdb /version:0.0 /machine:x64 /debug /INCREMENTAL /subsystem:console /GUARD:CF /CETCOMPAT /DYNAMICBASE /MANIFEST /MANIFESTFILE:dev\package\tests\CMakeFiles\some_test.dir/intermediate.manifest dev\package\tests\CMakeFiles\some_test.dir/manifest.res" failed (exit code 1104) with the following output:
[...]

PR #2623, opened 2025-06-27
Use crlf on windows for in newline

digit-google:
This is to fix the Ninja 1.13 regression described issue #2616 

This doesn't change anything on Posix, only on Windows.

Note that while the regression itself only impacts the content of the `rspfile` written by Ninja, this change also modifies the output of `$in_newline` when used on a command-line (as in performed in a manifest_parser unit-test).

It is however unlikely that a build plan relied on `\n` instead of `\r\n` being present on the command-line when invoking a Windows tool.


jhasse:
Hm ... I think \r\n should be avoided whenever possible. What's the problem with only using it for rspfiles?

jhasse:
I went with my version because I wanted to get a fix out ASAP and I have some minor issues with your approach.

The `BuildTest.RspFileNewlines` checks if `VirtualFileSystem::WriteTextFile` (a test only method) does the right thing, which is a bit weird. You could also just check if WriteTextFile gets called with the RSP file content. Then the other test checks if WriteTextFile converts \r\n to \n - okay. But let's say we later change the code to do the \r\n -> \n conversion before calling Write(Text)File: The unit test could then be removed, but that won't be obvious.
I think with the two tests and the new method this increases the complexity too much. My variant doesn't touch/move the WriteText body and IMHO makes it a bit clearer where the regression comes from ("w" was changed to "wb") and what the issue is.

I think a high level integration test would be nice for this. We can't use output_test.py because it doesn't work on Windows but maybe we could add a stripped down Windows version (output isn't important in this case).

The BuildConfig change is good, I would merge it for 1.14 (but having it in 1.13.1 was a bit too risky, too, another reason I went with my PR).

digit-google:
No problem, I understand, and I was away for a few days anyway. Thank you :-)

PR #2624, opened 2025-06-29
workflows - linux.yml - add aarch64 matrix builds

userdocs:
merge build-aarch64 job into build job and remove former. Add aarch64 matrix to other jobs.

All jobs will run amd64 and arm64 version of that job.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -3,16 +3,32 @@ name: Linux
 on:
   pull_request:
   push:
+    branches: ['**']
+    tags-ignore: ['**']  # Don't trigger on tag pushes to stop two runs on release publish
   release:
     types: published
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   fedora:
-    runs-on: [ubuntu-latest]
-    container:
-      image: fedora:40
+    permissions:
+      contents: read
+    strategy:
+      matrix:
+        host-os: ["ubuntu-latest", "ubuntu-24.04-arm"]
+      fail-fast: false
+    defaults:
+      run:
+        shell: bash
+    runs-on: ${{ matrix.host-os }}
+    container: fedora:40
     steps:
       - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
       - name: Install dependencies
         run: dnf install -y ninja-build cmake gtest-devel re2c clang util-linux clang-tools-extra
       - name: Linting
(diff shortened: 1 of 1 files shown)


userdocs:
I have a question about this jobs, since i don't fully understand what it's doing.

https://github.com/ninja-build/ninja/pull/2624/files#diff-21364b2e6fae1f2875cee1ab3daefb0685403687eaf8bc32b5c6eacda351c9d3R123-R130

1: ubuntu `20.04` is EOL, so should this become `22.04`?
2: `pip3 install cmake==3.17.*` had to become  `pip3 install cmake==3.20.3` as this is the [first release to have arm64 support.](https://pypi.org/project/cmake/3.20.3/) -I don't know why this is version locked. does it matter? should we upgrade/change it?

As for the rest, the matrix additions should be pretty simple. Just an aarch64 version of existing job

userdocs:
I updated `test` to 22.04 and added `fail-fast: false` to all matrix jobs.

I applied [zizmor audit stuff](https://github.com/zizmorcore/zizmor)

https://docs.zizmor.sh/audits/#artipacked
https://docs.zizmor.sh/audits/#excessive-permissions

Does `build-with-python` job need [ubuntu:20.04](https://github.com/ninja-build/ninja/pull/2624/files#diff-21364b2e6fae1f2875cee1ab3daefb0685403687eaf8bc32b5c6eacda351c9d3R189) or can that be removed?

jhasse:
Thanks! I'm a bit overwhelmed though. Can you split this PR in two,
1. one that does the bare minimum change so that the `build` job runs on ARM, too, 
    _(for 1.13.1)_
2. and another one with the rest of your changes?

> Does `build-with-python` job need [ubuntu:20.04](https://github.com/ninja-build/ninja/pull/2624/files#diff-21364b2e6fae1f2875cee1ab3daefb0685403687eaf8bc32b5c6eacda351c9d3R189) or can that be removed?

No, I think we can switch that to 22.04 now.

> `pip3 install cmake==3.17.*` had to become `pip3 install cmake==3.20.3` as this is the [first release to have arm64 support.](https://pypi.org/project/cmake/3.20.3/) -I don't know why this is version locked. does it matter? should we upgrade/change it?

It's version-locked so that CMake doesn't do a breaking change and our CI fails. 3.20 should be fine now, but use * for the patch, because that really doesn't matter.

userdocs:
@jhasse ok, i'll do that.

userdocs:
one question, i know it still works but this was archived in 2021

[uses: actions/upload-release-asset@v1](https://github.com/ninja-build/ninja/pull/2624/files#diff-21364b2e6fae1f2875cee1ab3daefb0685403687eaf8bc32b5c6eacda351c9d3L94)

Shall I also include the change that swaps this to a maintained for to the build job or do that separately?

jhasse:
Your call :) Might be a good idea to test it with the next release and might also be a good idea not to touch a working system.

userdocs:
I'm breaking the remaining changes up into mini pull requests for each related change. then you can review them that way.

I've already tested them to know they don't immediately break anything. I do think a a maintained one is less hassle than an archived one, if it breaks you have to swap. if the maintained one breaks the maintainer will fix it (hopefully)

PR #2626, opened 2025-06-30
Add --depfile flag to `inputs` and `multi-inputs` tools.

### LINKED ISSUE #2618: Feature Request: Include inputs from depfile in `ninja -t inputs`
## Background

The inputs from depfile is missing from stdout of `ninja -t inputs`

I propose to add `--depfile` option to `ninja -t inputs`,
in order to add inputs contained in depfile

## How is this useful?

With this feature, I can generate a depfile for any target of ninja sub project, 
that includes all inputs of the target, using the stdout of `ninja -t inputs`.

This new generated depfile can be used by ninja of an external project,
to relative precisely determine if the target of ninja sub project needs to be rebuilt.

This is useful for CMake project that uses CMake API `ExternalProject_Add`

## Error handling

This feature should be used after the target has been built, so depfile is up-to-date.
But I am not sure what to do if depfile is missing. Should it be ignore or raise an error,
or add other options to differ two behaviors.

## Implementation

The implementation is relatively simple that I have implemented a demo in my local machine.
Construct a `ImplicitDepLoader` and call `ImplicitDepLoader::LoadDeps` in `InputsCollector::VisitNode`

I can submit a Pull Request later, if this idea of feature request is accepted by the maintainer.

## Remaining issues

I am not sure how to handle "Dynamic Dependencies" feature of ninja build,
because I have never used this feature.
My currently thought is that this feature only focus on "depfile"
"Dynamic Dependencies" is not considered.

## Alternatives

Instead of providing argument `--depfile`, 
`ninja -t inputs` always contains inputs from depfile, if depfile exists

digit-google:
This forces the load of depfile entries from either the Ninja log or explicit depfiles to be performed when visiting the graph, thus making them visible in the result.

- graph.h, graph.cc: Add optional ImplicitDepLoader* argument to InputsCollector constructor, and if provided, use it to load deps during the visit if needed.

- ninja.cc: Modify `inputs` and `multi-inputs` implementation to support a new `--depfile` or `-D` flag, and create an ImplicitDepLoader instance when it is used to initialize the InputsCollector instances.

  NOTE: Using std::make_unique<> fails on our MacOS CI builder, which still uses -std=gnu+11 for some unknown reason, so this uses `std::unique_ptr<T>::reset(new T(...))` instead :-/

- output_test.py: Add regression tests + fix minor typing issues.

Fixes #2618

### DIFF
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -762,10 +762,25 @@ vector<Node*>::iterator ImplicitDepLoader::PreallocateSpace(Edge* edge,
 }
 
 void InputsCollector::VisitNode(const Node* node) {
-  const Edge* edge = node->in_edge();
+  Edge* edge = node->in_edge();
 
   if (!edge)  // A source file.
     return;
+
+  if (implicit_dep_loader_ && !edge->deps_loaded_) {
+    // Record that the deps were loaded in |deps_loaded_| as
+    // multiple visits to the same edge can be performed by
+    // repeated InputsCollector uses, as for the multi-inputs tool.
+    edge->deps_loaded_ = true;
+
+    // Ignore errors when loading depfile entries.
+    std::string err;
+    if (!implicit_dep_loader_->LoadDeps(edge, &err) && !err.empty()) {
+      // Print the error as a warning on stderr when an error occurred during
+      // the load.
+      Warning("%s", err.c_str());
+    }
+  }
 
   // Add inputs of the producing edge to the result,
   // except if they are themselves produced by a phony
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -784,14 +784,20 @@ int NinjaMain::ToolInputs(const Options* options, int argc, char* argv[]) {
 
   optind = 1;
   int opt;
+  bool include_depfile_inputs = false;
+  enum { OPT_DEPFILE = 1 };
   const option kLongOptions[] = { { "help", no_argument, NULL, 'h' },
                                   { "no-shell-escape", no_argument, NULL, 'E' },
                                   { "print0", no_argument, NULL, '0' },
+                                  { "depfile", no_argument, NULL, OPT_DEPFILE },
                                   { "dependency-order", no_argument, NULL,
                                     'd' },
                                   { NULL, 0, NULL, 0 } };
   while ((opt = getopt_long(argc, argv, "h0Ed", kLongOptions, NULL)) != -1) {
     switch (opt) {
+    case OPT_DEPFILE:
+      include_depfile_inputs = true;
+      break;
     case 'd':
       dependency_order = true;
       break;
@@ -815,6 +821,7 @@ int NinjaMain::ToolInputs(const Options* options, int argc, char* argv[]) {
 "  -0, --print0            Use \\0, instead of \\n as a line terminator.\n"
 "  -E, --no-shell-escape   Do not shell escape the result.\n"
 "  -d, --dependency-order  Sort results by dependency order.\n"
+"  --depfile               Include depfile inputs in the result.\n"
       );
       // clang-format on
       return 1;
@@ -830,7 +837,13 @@ int NinjaMain::ToolInputs(const Options* options, int argc, char* argv[]) {
     return 1;
   }
 
-  InputsCollector collector;
+  std::unique_ptr<ImplicitDepLoader> implicit_dep_loader;
+  if (include_depfile_inputs) {
+    implicit_dep_loader = std::make_unique<ImplicitDepLoader>(
+        &state_, &deps_log_, &disk_interface_, nullptr, nullptr);
+  }
+
+  InputsCollector collector(implicit_dep_loader.get());
   for (const Node* node : nodes)
     collector.VisitNode(node);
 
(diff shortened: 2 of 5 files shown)


hehaoqian:
Hi, I wonder if this loads dependencies from the DynDep file?
If it does, the flag of `--depfile` may be inappropriate

src/graph.cc
```diff
@@ -767,6 +767,19 @@ void InputsCollector::VisitNode(const Node* node) {
   if (!edge)  // A source file.
     return;
```
hehaoqian, line 768: Remove `const` in `const Edge* edge = node->in_edge();` ,
instead of using `const_cast`?

src/graph.cc
```diff
@@ -767,6 +767,19 @@ void InputsCollector::VisitNode(const Node* node) {
...
+
+    // Record that the deps were loaded in |deps_loaded_| as
+    // multiple visits to the same edge can be performed by
+    // repeated InputsCollector uses, as for the multi-inputs tool.
+    mutable_edge->deps_loaded_ = true;
+
+    // Ignore errors when loading depfile entries.
+    std::string err;
+    (void)implicit_dep_loader_->LoadDeps(mutable_edge, &err);
+  }
```
hehaoqian, line 783: I think the default behavior is to handle the error.
Because the most common usage of this feature is to
`ninja -t inputs TARGET` immediately after `ninja TARGET`,
and dependency loading should always succeed.

To ignore error, I think a separate flag should be added

src/ninja.cc
```diff
@@ -861,13 +873,17 @@ int NinjaMain::ToolMultiInputs(const Options* options, int argc, char* argv[]) {
...
+  bool include_depfile_inputs = false;
   const option kLongOptions[] = { { "help", no_argument, NULL, 'h' },
-                                  { "delimiter", required_argument, NULL,
-                                    'd' },
+                                  { "delimiter", required_argument, NULL, 'd' },
+                                  { "depfile", no_argument, NULL, 'D' },
                                   { "print0", no_argument, NULL, '0' },
                                   { NULL, 0, NULL, 0 } };
-  while ((opt = getopt_long(argc, argv, "d:h0", kLongOptions, NULL)) != -1) {
+  while ((opt = getopt_long(argc, argv, "Dd:h0", kLongOptions, NULL)) != -1) {
```
hehaoqian: Confusing to have both big `-D` and small `-d` option.
I suggest to change another letter, such as `-f`

src/graph.cc
digit-google: Thanks, the code was changed to print the error to stderr. Note that this can only happen when using an explicit `depfile` without `deps = gcc` or `deps = msvc`, since the log is already loaded when this function is invoked, so should be pretty rare in practice.

src/ninja.cc
digit-google: I find `-f` even more confusing. I'd rather not have a short version then. What do you think?

src/graph.cc
digit-google: Good idea! Done.

digit-google:
> Hi, I wonder if this loads dependencies from the DynDep file? If it does, the flag of `--depfile` may be inappropriate

No, the code only loads inputs from the deps log, so `--depfile` is an appropriate name (also this is what the original feature request asked for).

For dyndep files, the changes to the graph that they involve are a bit more aggressive, and I admit I am not sure at this point that the current collection algorithm would be correct when they are loaded. So probably better left for another PR.

src/ninja.cc
hehaoqian: Ok. No short argument is better

digit-google: Done, `-D` is now gone, only `--depfile` remains.

src/graph.cc
```diff
@@ -762,11 +762,26 @@ vector<Node*>::iterator ImplicitDepLoader::PreallocateSpace(Edge* edge,
...
+    // multiple visits to the same edge can be performed by
+    // repeated InputsCollector uses, as for the multi-inputs tool.
+    edge->deps_loaded_ = true;
+
+    // Ignore errors when loading depfile entries.
+    std::string err;
+    if (!implicit_dep_loader_->LoadDeps(edge, &err)) {
+      // Print the error as a warning on stderr when an error occurred during
+      // the load.
+      Warning("%s", err.c_str());
```
jyn514, line 781: FYI I tried out this branch just now and got the following output on ninja itself:
```
(-zsh@hephaestus.local) ~/src/ninja (detached HEAD)
; ./ninja -t inputs --depfile
ninja: warning:
...
ninja: warning:
ninja: warning:
ninja: warning:
ninja: warning:
build/browse.o
```
Probably there is a bug here somewhere? I would at least expect `err` to be non-empty.

jyn514: I also don't see it including any of the depfile inputs, which may or may not be related.

digit-google:
Getting back to this PR after some time. I just pushed a fix for the `--depfile` flag.

As @jyn514 tested, it was not working correctly because the tools now need the `Tool::RUN_AFTER_LOGS` flag (instead of `Tool::RUN_AFTER_LOAD`) to see the content of the deps log.

The regression tests were not catching these because they used `depfile = ...` without `deps = gcc` in the rule definition, which triggers a different code path in Ninja (where the deps are loaded directly from the .d files that are not removed). So I also modified these to handle both cases.

jyn514:
I tested your fixes and they work like a charm :)

Issue #2619, opened 2025-06-25
Version 1.13.0 not working on Ubuntu 22.04 LTS ARM


agarny:
I have just downloaded ninja 1.13.0 and tried it on Ubuntu 22.04 LTS ARM and got the following:
```
$ ninja --version
ninja: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by ninja)
ninja: /lib/aarch64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.32' not found (required by ninja)
```
Was dropping support for GLIBC_2.35 in ninja planned on ARM? Ubuntu 22.04 LTS is officially supported until April 2027, so I am surprised that support for it was dropped just like this. (To drop Ubuntu 20.04 LTS is fine since official support for it ended in May 2025.)

I explicitly mentioned Ubuntu 22.04 LTS ARM because everything is indeed fine with Ubuntu 22.04 LTS Intel.

jhasse:
I think at the time there were no Ubuntu 22.04 ARM runners. The old method using emulation was very very slow, see: https://github.com/ninja-build/ninja/commit/9a87d2c7c24934293a2f43cbc8e8b3f6659f812b

But I found https://github.com/actions/partner-runner-images/blob/main/images/arm-ubuntu-22-image.md so I guess we could just change the runner from 24.04-arm to 22.04-arm?

digit-google:
Using 22.04-arm should fix this indeed.

Another option is to use the arm64 version of rocky linux 8 in a container, as is done for the linux/x64 builder, but that would be a much larger change.

jhasse:
@userdocs Do you remember why you used 24.04 instead of 22.04?

> Another option is to use the arm64 version of rocky linux 8 in a container, as is done for the linux/x64 builder, but that would be a much larger change.

I would like this, because I think Ubuntu sucks. But my hate isn't big enough to put in the work :D

userdocs:
@jhasse at the time of the commit it was because [ubuntu-latest was transitioning to `24.04`](https://github.com/actions/runner-images/issues/10636) and when [I changed from emulation using `ubuntu-latest`](https://github.com/ninja-build/ninja/commit/9a87d2c7c24934293a2f43cbc8e8b3f6659f812b#diff-21364b2e6fae1f2875cee1ab3daefb0685403687eaf8bc32b5c6eacda351c9d3L172-L186) to the native arm runners, there is no `latest` option.

[...]

PR #2627, opened 2025-06-30
add aarch64 matrix to linux.yml build job using rockylinux 8 container

userdocs:
instead of a separate amd64 and arm64 builds jobs, turn the `build` job into a matrix using amd64 and arm64 ubuntu runners and remove the `build-aarch64` job.


userdocs:
@jhasse this should be the minimum required to fix https://github.com/ninja-build/ninja/issues/2619

jhasse:
Awesome! Thanks :)

PR #2628, opened 2025-06-30
prevent duplicate runs being triggered when a release tag is published via webui

userdocs:
creating a release through the webui would trigger a non release run alongside the release run. It's not needed.

The release run fulfils alone is enough in this scenario.

before 

image

after

image

### DIFF
--- a/.github/workflows/linux-musl.yml
+++ b/.github/workflows/linux-musl.yml
@@ -4,6 +4,8 @@ on:
   workflow_dispatch:
   pull_request:
   push:
+    branches: ['**']
+    tags-ignore: ['**']  # Don't trigger on tag pushes
   release:
     types: [published]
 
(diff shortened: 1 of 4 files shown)


userdocs:
We only want this the `published` jobs to run here right and not a merge/pr one?

jhasse:
> We only want this the `published` jobs to run here right and not a merge/pr one?

What do you mean by that?

userdocs:
When you publish a release it seems to trigger two events, in this order

A push event where the ref is the tag just created

image

Followed by a release event.

image

This prevents the push/tag event from being trigged. In this case it's the same workflows being run but with no release.

I noticed the [concurrency](https://github.com/ninja-build/ninja/pull/2632) would have done something similar by cancelling the push/tag event and only running the release event

image

userdocs:
but it does not prevent runs from pull requests. Just tags.

jhasse:
I see. Thanks for the PR :)

PR #2629, opened 2025-06-30
replace archived in 2021 release action for currently maintained one

userdocs:
actions/upload-release-asset is archived in 2021 but ncipollo/release-action is actively maintained a suitable drop in replacement with feature parity.

It can also be adapted to work the same when triggering aa release via a pushed  tag, should the need arise.

linux.yml only - tidied up dependency install commands.

note: As I understand the release process it goes:
- you create the release via the webui, 
- you create the tag select which branch, 
- create the body and add the changelog link
- you publish the release
- it triggers the workflows to build and upload the assets.

Here is a demonstration to show it should mimic that process.

https://github.com/userdocs/ninja/releases/tag/v1.13.19

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -58,10 +58,9 @@ jobs:
           ignore_words_list: fo,wee,addin,notin
       - name: Install dependencies
         run: |
-          dnf update -y
-          dnf install -y dnf-plugins-core epel-release
+          dnf install -y make gcc-c++ libasan clang-analyzer cmake dnf-plugins-core epel-release
           dnf config-manager --set-enabled powertools
-          dnf install -y clang-analyzer cmake gcc-c++ gtest-devel libasan make ninja-build p7zip p7zip-plugins re2c
+          dnf install -y gtest-devel p7zip p7zip-plugins ninja-build
 
       - name: Build debug ninja
         env:
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -58,9 +58,10 @@ jobs:
           ignore_words_list: fo,wee,addin,notin
       - name: Install dependencies
         run: |
-          dnf install -y make gcc-c++ libasan clang-analyzer cmake dnf-plugins-core epel-release
+          dnf update -y
+          dnf install -y dnf-plugins-core epel-release
           dnf config-manager --set-enabled powertools
-          dnf install -y gtest-devel p7zip p7zip-plugins ninja-build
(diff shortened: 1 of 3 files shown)


.github/workflows/linux.yml
```diff
@@ -58,9 +58,10 @@ jobs:
           ignore_words_list: fo,wee,addin,notin
       - name: Install dependencies
         run: |
-          dnf install -y make gcc-c++ libasan clang-analyzer cmake dnf-plugins-core epel-release
+          dnf update -y
```
jhasse: Only bad package managers require this, DNF isn't one of them.

.github/workflows/linux.yml
```diff
@@ -58,9 +58,10 @@ jobs:
...
+          dnf install -y dnf-plugins-core epel-release
           dnf config-manager --set-enabled powertools
-          dnf install -y gtest-devel p7zip p7zip-plugins ninja-build
+          dnf install -y clang-analyzer cmake gcc-c++ gtest-devel libasan make ninja-build p7zip p7zip-plugins re2c
```
jhasse: Hm ... I don't see the advantage of this change? Why is it cleaner?

I would say it's cleaner to have all the packages here that are part of EPEL and powertools.

jhasse (changes requested):
Why doesn't this action require the token secret anymore?

userdocs:
> Why doesn't this action require the token secret anymore?

I believe it defaults to `github.token` which is stated in the readme

https://github.com/ncipollo/release-action?tab=readme-ov-file#action-inputs

image

.github/workflows/linux.yml
userdocs: I have to check this, I think I needed it on rocky 8 to install something successfully.

.github/workflows/linux.yml
userdocs: So like `dnf install -y dnf-plugins-core ... related apps` on the same lines?

.github/workflows/linux.yml
userdocs: rocky 9 didn't need it.

I'll confirm and update.

jhasse: Hm ... that should only happen on amateur distros like Debian (SCNR). DNF notices if there's no index (like in a fresh Docker images) and downloads it. The same after adding repositories. It also updates the index periodically.

jhasse:
I see, thanks :) Can you remove the Linux change from the commit? Then I will merge.

userdocs:
Done and tests passed.

PR #2630, opened 2025-06-30
linux.yml - test - add matrix to job

userdocs:
add matrix to this job to test on amd64/arm64 on jammy and noble LTS releases instead of only amd64 focal which is EOL.

Applied two zizmor audits commented in the changes.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -122,15 +122,10 @@ jobs:
     container: ${{ matrix.image }}
     steps:
       - uses: actions/checkout@v4
-        with:
-          persist-credentials: false #https://docs.zizmor.sh/audits/#artipacked
       - name: Install dependencies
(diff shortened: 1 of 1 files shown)


jhasse:
Regarding `persist-credentials: false`: If it's better 99% of the time, why isn't it the default? I don't really understand what it implies so I would rather have a shorter yml file.

That re2c wasn't installed is actually intended: We want to test both with and without re2c.

For CMake: Can we use cmake from apt now that we have 22.04 as a minimum?

userdocs:
> persist-credentials: false

I don't think I meant to leave that in, it's a zizmor recommendation. It's default is actually contested as you can see here. I guess it's just a convenience thing.

https://github.com/actions/checkout/issues/485

I can just remove it.

The other points i'll apply and check.

PR #2631, opened 2025-06-30
linux.yml - build-with-python - add matrix to job

userdocs:
add matrix to this job to test on amd64/arm64 on jammy and noble LTS releases instead of only amd64 focal which is EOL.

Applied two zizmor audits commented in the changes.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -157,19 +157,27 @@ jobs:
         working-directory: build-clang/Release
 
   build-with-python:
-    runs-on: [ubuntu-latest]
-    container:
-      image: ${{ matrix.image }}
+    permissions: # https://docs.zizmor.sh/audits/#excessive-permissions
+      contents: read
     strategy:
       matrix:
-        image: ["ubuntu:20.04", "ubuntu:22.04", "ubuntu:24.04"]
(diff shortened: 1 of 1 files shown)


jhasse:
https://docs.zizmor.sh/audits/#excessive-permissions
Doesn't this mean we "over-scoped" somewhere else? Sorry I don't really understand most of GitHub Actions 😅

userdocs:
So as far as I understand it, the default GitHub token perms are these when nothing is set (permissive)

https://docs.github.com/en/actions/reference/github_token-reference#permissions-for-the-github_token

As soon as you set one of them, all the rest default to none. (restricted)

https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token#defining-access-for-the-github_token-permissions

So it's always been permissive by default.

userdocs:
So i'll go with whatever you decide. My preference in workflows is restrictive by default.

jhasse:
I see. I agree restrictive would be better, I also see why GitHub went with the default.

So I should merge this PR as is?

userdocs:
I think so. It has the required permissions only and checkout thing is the same idea, restrictive by default instead having your token in the .git folder.

PR #2632, opened 2025-07-01
workflow update - concurrency to cancel outdated runs

userdocs:
cancel old runs when new pushes are made to a pr/branch.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -6,6 +6,10 @@ on:
   release:
     types: published
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   fedora:
     runs-on: [ubuntu-latest]
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -6,6 +6,10 @@ on:
   release:
     types: published
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
     runs-on: macos-13
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -6,6 +6,10 @@ on:
   release:
     types: published
 
+concurrency:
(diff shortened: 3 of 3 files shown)


userdocs:
I tested with two forks, to make sure to does not cancel runs from other pull requests.

image

jhasse:
Thanks :)

jhasse:
Okay we have a problem: The "cancelled" jobs appear as "failed" in GitHub's web UI. I'm also getting spammed with emails for every job that gets cancelled.

Any way to fix that or is it a GitHub Actions limitation?

userdocs:
As far as I know they are both platform limitations.

https://docs.github.com/en/actions/how-tos/managing-workflow-runs-and-deployments/managing-workflow-runs/canceling-a-workflow#steps-github-takes-to-cancel-a-workflow-run

> For steps that need to be canceled, the runner machine sends SIGINT/Ctrl-C to the step's entry process (node for javascript action, docker for container action, and bash/cmd/pwd when using run in a step). If the process doesn't exit within 7500 ms, the runner will send SIGTERM/Ctrl-Break to the process, then wait for 2500 ms for the process to exit. If the process is still running, the runner kills the process tree.

People asking to be able to supress cancelled notifications

https://github.com/orgs/community/discussions/13015

So i'm not sure there is a way to handle within the concurrency feature. 

Maybe a way would be a hack to make all steps skipped if a conditions is met but that would require spamming `if:` in all the workflows.

userdocs:
@jhasse so looking at the actions that were cancelled and the docs, all your merges into `master` were cancelled as the concurrency group was the same.

So the spam came from the multiple merges of different pr's we can maybe ignore master like this?

https://docs.github.com/en/enterprise-cloud@latest/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#example-only-cancel-in-progress-jobs-on-specific-branches

Since the aim was to cancel runs on pr branches?

jhasse:
That's all way too complicated for my me. I think this is something GitHub Actions should figure out in their defaults, therefore I reverted the PR for now.

Still thanks for looking into it :)

PR #2633, opened 2025-07-01
windows.yml - add matrix for arm64 runner

userdocs:
Add windows-11-arm partner image runner to build arm64 binaries. This allows for the binary testing steps to be run instead of skipped for arm64 due to the host being x64.

added hyphen to suffix so it becomes ninja-binary-archives-arm64 instead of ninja-binary-archivesarm64

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -8,16 +8,15 @@ on:
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


userdocs:
not sure if making use of server-2025/windows 11 arm64 has any unintended side effects for binaries.

but the workflow seems to work fine.

PR #2635, opened 2025-07-01
Update CMakeLists.txt to use C++14 too.

digit-google:
In PR #2560, configure.py was changed to force -std=c++14, but this was not done in the corresponding CMakeLists.txt, which still compiled Ninja for C++11.

This PR simply fixes the situation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -191,8 +191,8 @@ else()
 	endif()
 endif()
 
-target_compile_features(libninja PUBLIC cxx_std_11)
-target_compile_features(libninja-re2c PUBLIC cxx_std_11)
+target_compile_features(libninja PUBLIC cxx_std_14)
+target_compile_features(libninja-re2c PUBLIC cxx_std_14)
 
 #Fixes GetActiveProcessorCount on MinGW
 if(MINGW)


jhasse:
Why did you use `set(CMAKE_CXX_STANDARD ` instead? I don't know enough CMake to know the differences, but the other way is the "modern" one IIRC.

digit-google:
I don't know much either, I stumbled upon https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html doing a simple web search.

Apparently, there is also [CXX_STANDARD_REQUIRED](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD_REQUIRED.html) which can be set to `ON` to force a minimum level (see also [this page](https://matgomes.com/setting-cpp-standard-cmake/) which recommends it).

Another alternative is using something like [`target_compile_features(libninja PUBLIC cxx_std_14)
`](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD_REQUIRED.html) which is supposed to also apply to any dependents.

Let me know if you prefer any scheme, I don't have any preference.

jhasse:
Let's use the current and just change 11 to 14.

digit-google:
Thanks, done.

jhasse:
Thanks! :)

PR #2636, opened 2025-07-02
Convert \n to \r\n on Windows in rsp files, fix #2616

### LINKED ISSUE #2616: Ninja 1.13.0 with MSVC `link.exe` fails, apparently due to RSP line limit of 16,383 characters

jhasse:
Other files written will still use "wb" as introduced in 46f272d.

### DIFF
--- a/src/disk_interface.h
+++ b/src/disk_interface.h
@@ -51,9 +51,11 @@ struct DiskInterface: public FileReader {
   virtual bool MakeDir(const std::string& path) = 0;
 
   /// Create a file, with the specified name and contents
+  /// If \a crlf_on_windows is true, \n will be converted to \r\n (only on
+  /// Windows builds of Ninja).
   /// Returns true on success, false on failure
-  virtual bool WriteFile(const std::string& path,
-                         const std::string& contents) = 0;
+  virtual bool WriteFile(const std::string& path, const std::string& contents,
+                         bool crlf_on_windows) = 0;
 
   /// Remove the file named @a path. It behaves like 'rm -f path' so no errors
   /// are reported if it does not exists.
@@ -71,12 +73,13 @@ struct DiskInterface: public FileReader {
 struct RealDiskInterface : public DiskInterface {
   RealDiskInterface();
   virtual ~RealDiskInterface() {}
-  virtual TimeStamp Stat(const std::string& path, std::string* err) const;
-  virtual bool MakeDir(const std::string& path);
-  virtual bool WriteFile(const std::string& path, const std::string& contents);
-  virtual Status ReadFile(const std::string& path, std::string* contents,
-                          std::string* err);
-  virtual int RemoveFile(const std::string& path);
+  TimeStamp Stat(const std::string& path, std::string* err) const override;
+  bool MakeDir(const std::string& path) override;
+  bool WriteFile(const std::string& path, const std::string& contents,
+                 bool crlf_on_windows) override;
+  Status ReadFile(const std::string& path, std::string* contents,
+                  std::string* err) override;
+  int RemoveFile(const std::string& path) override;
 
   /// Whether stat information can be cached.  Only has an effect on Windows.
   void AllowStatCache(bool allow);
--- a/src/build.cc
+++ b/src/build.cc
@@ -840,7 +840,7 @@ bool Builder::StartEdge(Edge* edge, string* err) {
     if (!disk_interface_->MakeDirs((*o)->path()))
       return false;
     if (build_start == -1) {
-      disk_interface_->WriteFile(lock_file_path_, "");
+      disk_interface_->WriteFile(lock_file_path_, "", false);
       build_start = disk_interface_->Stat(lock_file_path_, err);
       if (build_start == -1)
         build_start = 0;
@@ -860,7 +860,7 @@ bool Builder::StartEdge(Edge* edge, string* err) {
   string rspfile = edge->GetUnescapedRspfile();
   if (!rspfile.empty()) {
     string content = edge->GetBinding("rspfile_content");
-    if (!disk_interface_->WriteFile(rspfile, content))
+    if (!disk_interface_->WriteFile(rspfile, content, true))
       return false;
   }
 
--- a/src/build_log_test.cc
+++ b/src/build_log_test.cc
@@ -227,23 +227,24 @@ TEST_F(BuildLogTest, DuplicateVersionHeader) {
 }
 
 struct TestDiskInterface : public DiskInterface {
-  virtual TimeStamp Stat(const std::string& path, std::string* err) const {
+  TimeStamp Stat(const std::string& path, std::string* err) const override {
     return 4;
   }
-  virtual bool WriteFile(const std::string& path, const std::string& contents) {
+  bool WriteFile(const std::string& path, const std::string& contents,
+                 bool crlf_on_windows) override {
     assert(false);
     return true;
   }
-  virtual bool MakeDir(const std::string& path) {
+  bool MakeDir(const std::string& path) override {
     assert(false);
     return false;
   }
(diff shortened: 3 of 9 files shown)


jhasse:
Untested.

@digit-android I'm not sure what failures the change from "w" to "wb" fixed though. Can you review?

src/disk_interface.h
```diff
@@ -51,9 +51,11 @@ struct DiskInterface: public FileReader {
   virtual bool MakeDir(const std::string& path) = 0;
 
   /// Create a file, with the specified name and contents
+  /// If \a crlf_on_windows is true, \n will be converted to \r\n (only on
+  /// Windows builds of Ninja).
   /// Returns true on success, false on failure
-  virtual bool WriteFile(const std::string& path,
-                         const std::string& contents) = 0;
+  virtual bool WriteFile(const std::string& path, const std::string& contents,
+                         bool crlf_on_windows) = 0;
```
digit-google, line 58: I suggest a different way to achieve this that will be far less instrusive (e.g. will keep most unit-tests unmodified), as well as more readable (boolean arguments can be hard to read at call sites):

1) Do not touch `WriteFile()` here, but instead add a `virtual WriteTextFile(...);` method with a default implementation that just calls `WriteFile()`.

2) Override it in `ReadDiskInterface::WriteTextFile()` on _WIN32 only, to perform the write the new option.

3) Modify the call site that writes the response file to use `WriteTextFile()`.

This also needs a new unit-test to verity that WriteTextFile() works as expected (with different expected results on Posix and Windows). Wdyt?

jhasse: > boolean arguments can be hard to read at call sites

Well that depends ;)
image

The extra method has it advantages as you pointed out: No need to touch the unit-tests. But it also adds some indirection and we would have to think about how not duplicate the code besides `crlf_on_windows ? "w" : "wb");`. Also I wouldn't name it `WriteTextFile` because it obscures what it is about: What does "text" file mean exactly? Are there any other differences except CRLF? Do we care about these? No. So I think a better name would be `WriteFileWithCrlfOnWindows`. And I think it would be clearer to simply use a bool for that.

No strong preference though, in the end I think it isn't important, just necessary that we fix this fast and release 1.13.1 before more people run into the regression on Windows.

> This also needs a new unit-test to verity that WriteTextFile() works as expected (with different expected results on Posix and Windows).

With that I agree regardless of bool / extra method :) Or a high-level test that uses link.exe.

digit-google: I frankly have no idea what your screenshot refers to. Some sort of Intellisense or a new C++ version?

Anyway, bool arguments are bad because `true` / `false` is usually vague at the call site. There are two better alternatives in practice: One is to use a `class enum` that forces the use of a named value without the risk of implicit conversions from bool, as in:

```
struct DiskInterface {
  /// Describe how to filter
  enum class TextFilter {
    None = 0,
    NewlinesAsCrLf,   // Replace \n by \r\n on output.
  };
  
  bool WriteFile(const std::string& path, const std::string& content, TextFilter text_filter = TextFilter::None);
};
```

This would make one call-site slightly more verbose, e.g. `WriteFile(path, content, TextFilter::NewlinesAsCrLf)`, and will leave the other untouched.

The other is to use function overloading with an "tag struct" type, as in:

```
struct DiskInterface {
  struct ReplaceNewlinesWithCrLf {};
  
  // Unmodified WriteFile() interface.
  virtual bool WriteFile(const std::string& path, const std::string& content) = 0;
  
  // Function overload, default implementation just calls regular WriteFile().
  virtual bool WriteFile(const std::string& path, const std::string& content, ReplaceNewlinesWithCrLf) {
    WriteFile(path, content);
  }
};
```

Which can be used at call sites with the following syntax `WriteFile(path, content, ReplaceNewlinesWithCrLf{})`

This is not much different from the `WriteFileText()` method implemented in the latest patch of PR #2623.

Which one would you prefer?

jhasse: > I frankly have no idea what your screenshot refers to. Some sort of Intellisense or a new C++ version?

Inlay Hints in Visual Studio Code.

> Anyway, bool arguments are bad because true / false is usually vague at the call site.

Sure, but not the end of the world. I think the methods you mentioned are nice, especially if there are multiple bool parameters or room for a third state in the future, but overkill in this case.

Issue #2454, opened 2024-05-23
missingdeps tool has some odd false positives


eli-schwartz:
https://gitlab.freedesktop.org/xorg/xserver

build rules:

```
[...]

PR #2638, opened 2025-07-05
Do not treat dummy targets as missing dependencies.

digit-google:
Dummy targets are outputs of phony edges that have no inputs. They are treated specially by Ninja at build time: if they do not exist, they are simply considered out-of-date, instead of a missing-input error.

This fixes issue #2454, where the `missingdeps` tool, did not treat dummy targets properly,
always reporting them as missing.

- Add Edge::has_dummy_outputs() method and use it in DependencyScan::RecomputeOutputsDirty() for clarity.
- Modify MissingDependencyScanner::ProcessNodeDeps() to ignore edges with dummy outputs.
- Add a corresponding unit-test.

### DIFF
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -282,7 +282,7 @@ bool DependencyScan::RecomputeOutputDirty(const Edge* edge,
   if (edge->is_phony()) {
     // Phony edges don't write any output.  Outputs are only dirty if
     // there are no inputs and we're missing the output.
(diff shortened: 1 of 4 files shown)


xrsojib:
[Uploading ninja.spec.txt…]()

xrsojib:
FREE fire

moritzx22:
I think this PR improves `ninja -t missingdeps` and is ready to merge.

PR #2643, opened 2025-07-25
[Windows] Prevent suspend during builds

lb90:
Prevent the system from going to sleep / suspend during builds

For now the implementation is Windows-only. The code uses APIs introduced in Windows 7

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -1863,6 +1863,8 @@ NORETURN void real_main(int argc, char** argv) {
 
     ninja.ParsePreviousElapsedTimes();
 
+    ContinuedExecution continued_execution(nullptr);
+
     ExitStatus result = ninja.RunBuild(argc, argv, status);
     if (g_metrics)
       ninja.DumpMetrics();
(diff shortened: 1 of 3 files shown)


jhasse:
I like it and at the same time I'm not sure if this is really needed. I certainly wouldn't expect a command line tool to block sleep. Also there might be long running commands someone runs via ninja.

digit-google:
As interesting as it is, I don't think that's Ninja responsibility to modify the global state of the machine (imagine if multiple programs would want to modify at it at the same time). And this can easily be handled by a specific command (on Windows) or wrapper script (on Unix) before / after the build.

scivision:
This should only be done by a script invoking the build command IMO, not within Ninja itself. This is quite simple as shown by [caffeinate.py](https://gist.github.com/scivision/d0e6dbebb88687791129ad722c90e68a) single standalone Python script.

lb90:
Hi everyone, sorry for the delay! Yeah a wrapper script works just fine, I think we can close this. Thank you!

PR #2646, opened 2025-08-05
[macOS] Prevent suspend during builds

lb90:
Same as #2643 but for macOS


scivision:
For my opinion, same as #2643 this is a nice idea, but, should be done outside Ninja with a little wrapper script like [caffeinate.py](https://gist.github.com/scivision/d0e6dbebb88687791129ad722c90e68a), which could be pared down even simpler than that example.

lb90:
Right, closing! :)

Issue #2599, opened 2025-05-20
depfile containing file with ampersand handled incorrectly


trombonehero:
Given the following C file:

```c
#include "a&b.h"
```

Compiling with `cc -c -MD foo.c` yields the following depfile:

```make
foo.o: foo.c a&b.h
```

If I use this depfile with `deps` and `depfile`, Ninja v1.11.1 (`1.11.1.git.kitware.jobserver-1` from MacPorts) will always rebuild `foo.o` due to parsing `a&b.h` as two file, `a` and `b.h`:

```
ninja explain: output a of phony edge with no inputs doesn't exist
ninja explain: a is dirty
ninja explain: output b.h of phony edge with no inputs doesn't exist
ninja explain: b.h is dirty
[1/1] cc -c -MD foo.c
```

Now, ampersands might be silly things to include in a C header file name, but they do come up in other things that I build with Ninja (e.g., image files included within a [Typst](https://typst.app) file). The C example was used to demonstrate that such things are not escaped in the GCC syntax.

moritzx22:
Merge request fixes this particular issue for character "&". [MR](https://github.com/ninja-build/ninja/pull/2601)

There is still a more general issue in the parsing of the deps file.
The following example assumes that "&" is still invalid:
`abc&cde`
is parsed as
`abc
cde`
The parser does not check if the separator is valid. The separator could be supposed to be something like a whitespace, but it can be something else as well. But that might be a separate issue.

trombonehero:
Awesome. Thanks!

PR #2655, opened 2025-08-19
feat: `zero` deps format

magistau:
This patch adds support for NUL-delimited dependency files, which is quite a simple format avoiding all the issues possible with `gcc`/`msvc` formats. It requires almost no parsing: just read NUL-terminated strings until the file is empty. Because of this, it doesn't depend on file paths being "nice" in any way and doesn't require any escaping for special characters like `$` or `\`. It is quite common in shell tooling, like `find -print0`, `xargs -0` and others.
GCC format, which should more accurately be described as "GCC version of Makefile subset", is quite problematic: #2599, #2590, #2568, #2357, #1856, #1825, #1774, #1542, #1227 and probably more. Compared to this, the only possible issue with the proposed format is lack of a NUL after the last entry, which I handled by adding a NUL if the file doesn't end in one.
The parser itself is implemented in 8 logical lines of code, compared to hundreds of lines MSVC and GCC need. Because of this simplicity, any project can easily implement a reader or a writer without any shortcomings. GCC format is impossible to get right, as even GCC itself was unable to do so (quoted from `libcpp/mkdeps.cc` from GCC 14.3.0):
```cpp
/* Apply Make quoting to STR, TRAIL.  Note that it's not possible to
   quote all such characters - e.g. \n, %, *, ?, [, \ (in some
   contexts), and ~ are not properly handled.  It isn't possible to
   get this right in any current version of Make.  (??? Still true?
   Old comment referred to 3.76.1.)  */
```
Typst contains a similar remark: https://github.com/typst/typst/blob/880e69b0c840cd0d2f0c51e72966ab72c008e713/crates/typst-cli/src/compile.rs#L524-L525
That led me to creating ~~typst/typst#6648~~ typst/typst#7022, adding NUL-delimited format, and that in turn led me here.

Note that the proposed format has no "official" standard, RFC or otherwise. Due to its simplicity, though, there are only two flavors: either you terminate your fail with a NUL (e.g. `foo␀bar␀baz␀`) or you don't (e.g. `foo␀bar␀baz`). The former is usually easier to implement due to being more consistent, and this situation is similar to how some old tools require that a text file ends with an LF and some don't. To be honest, I haven't seen any tools that omit the trailing NUL, but they might still exist. Anyway, the implementation here supports both "flavors" without any compatibility cost.

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -786,7 +786,7 @@ Ninja 1.3 can instead process dependencies just after they're generated
 and save a compacted form of the same information in a Ninja-internal
 database.
 
-Ninja supports this processing in two forms.
+Ninja supports this processing in three forms.
 
 1. `deps = gcc` specifies that the tool outputs `gcc`-style dependencies
    in the form of Makefiles.  Adding this to the above example will
@@ -810,6 +810,11 @@ rule cc
   deps = msvc
   command = cl /showIncludes -c $in /Fo$out
 ----
+
+3. `deps = zero` specifies that the tool outputs the dependencies as a list of
+   NUL-terminated paths, like what `find . -print0` would produce.
+   Although this format is not used by any known C/C++ compilers, it is
+   the easiest to produce from a script, so it works in ad hoc scenarios.
 
 If the include directory directives are using absolute paths, your depfile
 may result in a mixture of relative and absolute paths. Paths used by other
--- a/src/build.cc
+++ b/src/build.cc
@@ -1024,10 +1024,10 @@ bool Builder::ExtractDeps(BuildResult::CommandCompleted& result,
       // complexity in IncludesNormalize::Relativize.
       deps_nodes->push_back(state_->GetNode(*i, ~0u));
     }
-  } else if (deps_type == "gcc") {
+  } else if (deps_type == "gcc" || deps_type == "zero") {
     string depfile = result.edge->GetUnescapedDepfile();
     if (depfile.empty()) {
-      *err = string("edge with deps=gcc but no depfile makes no sense");
+      *err = string("edge with deps=") + deps_type + string(" but no depfile makes no sense");
       return false;
     }
 
@@ -1045,17 +1045,30 @@ bool Builder::ExtractDeps(BuildResult::CommandCompleted& result,
     if (content.empty())
       return true;
 
-    DepfileParser deps(config_.depfile_parser_options);
-    if (!deps.Parse(&content, err))
-      return false;
+    if (deps_type == "gcc") {
+      DepfileParser deps(config_.depfile_parser_options);
+      if (!deps.Parse(&content, err))
+        return false;
 
-    // XXX check depfile matches expected output.
-    deps_nodes->reserve(deps.ins_.size());
-    for (vector<StringPiece>::iterator i = deps.ins_.begin();
-         i != deps.ins_.end(); ++i) {
-      uint64_t slash_bits;
-      CanonicalizePath(const_cast<char*>(i->str_), &i->len_, &slash_bits);
-      deps_nodes->push_back(state_->GetNode(*i, slash_bits));
+      // XXX check depfile matches expected output.
+      deps_nodes->reserve(deps.ins_.size());
+      for (vector<StringPiece>::iterator i = deps.ins_.begin();
+           i != deps.ins_.end(); ++i) {
+        uint64_t slash_bits;
+        CanonicalizePath(const_cast<char*>(i->str_), &i->len_, &slash_bits);
+        deps_nodes->push_back(state_->GetNode(*i, slash_bits));
+      }
+    } else {
+      const char* ptr = content.c_str();
+      const char* const end = &content.back() + 1;
+      while (ptr < end) {
+        StringPiece piece = ptr;
+        // ptr is advanced first, because `piece.len_` is modified by `CanonicalizePath`
+        ptr += piece.size() + 1;
+        uint64_t slash_bits;
+        CanonicalizePath(const_cast<char*>(piece.str_), &piece.len_, &slash_bits);
+        deps_nodes->push_back(state_->GetNode(piece, slash_bits));
+      }
     }
 
     if (!g_keep_depfile) {
--- a/src/build.cc
+++ b/src/build.cc
@@ -1024,10 +1024,10 @@ bool Builder::ExtractDeps(BuildResult::CommandCompleted& result,
       // complexity in IncludesNormalize::Relativize.
       deps_nodes->push_back(state_->GetNode(*i, ~0u));
     }
-  } else if (deps_type == "gcc") {
+  } else if (deps_type == "gcc" || deps_type == "zero") {
     string depfile = result.edge->GetUnescapedDepfile();
     if (depfile.empty()) {
-      *err = string("edge with deps=gcc but no depfile makes no sense");
+      *err = string("edge with deps=") + deps_type + string(" but no depfile makes no sense");
       return false;
     }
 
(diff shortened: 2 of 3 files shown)


src/build.cc
```diff
@@ -1021,17 +1021,33 @@ bool Builder::ExtractDeps(CommandRunner::Result* result,
...
+      const char* ptr = content.c_str();
+      const char* const end = &content.back();
+      while (ptr < end) {
+        StringPiece piece = ptr;
+        ptr += piece.size() + 1;
+        uint64_t slash_bits;
+        CanonicalizePath(const_cast<char*>(piece.str_), &piece.len_,
+                         &slash_bits);
+        deps_nodes->push_back(state_->GetNode(piece, slash_bits));
+      }
```
elliotgoodrich: If you wanted to you could include the null character at the end of the `std::string` to avoid modification, but then you need to skip empty paths in your loop (which may be a good idea anyway?)

```suggestion
      const char* ptr = content.c_str();
      // Include the terminating '\0' in the range
      const char* const end = content.c_str() + content.size() + 1;
      while (ptr < end) {
        StringPiece piece = ptr;
        ptr += piece.size() + 1;
        if (piece.empty())
          continue;
        uint64_t slash_bits;
        CanonicalizePath(const_cast<char*>(piece.str_), &piece.len_,
                         &slash_bits);
        deps_nodes->push_back(state_->GetNode(piece, slash_bits));
      }
```

magistau: I'm not sure what one has to do with the other. It is true that `c_str()` returns a pointer to a NUL-terminated string, although I wasn't sure about that initially; but I'm not sure if there's a need to handle empty file names in a special way. It's an easy check, sure, but how is it different from any other invalid path?

elliotgoodrich: Yes they are independent things, but if the user's data already contains a null character at the end, then by parsing the extra one added by `std::string` we will end up with an empty `StringPiece` and I don't think we should be adding that as a dependency. You can handle that by skipping all empty `StringPiece`s, if you want to be more flexible and ignore empty paths from the user, or you could conditionally allow the null character and then not skip empty `StringPiece`s:

```cpp
      const char* ptr = content.c_str();
      const bool needNullTerminator = content.empty() || content.back() != '\0';
      const char* const end = ptr + content.size() + needNullTerminator;
```

magistau: It won't. The `ptr += ` line advances the pointer past the current piece's NUL. For non-terminal pieces, this is the first character of the next piece, and for the terminal piece it's the string's NUL, which is also pointed to by `end`. Now, the `ptr < end` condition fails, so there's no attempt to parse any sort of an empty piece. Same logic applies to cases where user's file doesn't end with NUL — then `ptr` ends up on `end+1` after reading the past piece, and the condition likewise fails.

elliotgoodrich: Sorry for any confusion, but when I wrote,

> If you wanted to you could include the null character at the end of the std::string to avoid modification, but then you need to skip empty paths in your loop (which may be a good idea anyway?)

it was in reference to the code snippet that I had suggested.  I don't think your original code adds in any empty `StringPiece` code that wasn't supplied by the user.

But instead of conditionally adding `\0`,

```cpp
      if (content.back() != '\0') {
        content.push_back('\0');
      }
```

we should instead conditionally use the null-terminator that is already there in `std::string`.

```cpp
      const char* ptr = content.c_str();
      const bool needNullTerminator = content.back() != '\0';
      const char* const end = ptr + content.size() + needNullTerminator;
```

(I realised that `content` is never empty at this point in the code so we I didn't need to put the check in).

magistau: Please see the latest version of the patch — I applied your suggestion without any checks (neither for "ends with NUL" nor for "is empty").

magistau:
@elliotgoodrich sorry for pinging but could I get any feedback on this?

magistau:
CC @jhasse as well, I don't know how should I proceed, I got no response on this PR for half a year, I got no response on IRC for 20 hours, and the mailing list requires a Google account for some reason which I would rather not share. So sorry if I'm being obnoxious, but this is a PR that adds eight significant lines of code, and it's been open for almost a year with only one spontaneous review which was addressed in due time and was never heard of since.

elliotgoodrich:
> @elliotgoodrich sorry for pinging but could I get any feedback on this?

hey @magistau - I don't have merge privileges on this repo, so I don't know if any approval from me carries much weight :) I gave some initial feedback because I like this change (and could have used it in another project) and I wanted to help so it would be in a better shape for when the owners take a look.

magistau:
cc @evmar @nico

src/build.cc
```diff
@@ -1045,17 +1045,30 @@ bool Builder::ExtractDeps(BuildResult::CommandCompleted& result,
...
-      deps_nodes->push_back(state_->GetNode(*i, slash_bits));
+      // XXX check depfile matches expected output.
+      deps_nodes->reserve(deps.ins_.size());
+      for (vector<StringPiece>::iterator i = deps.ins_.begin();
+           i != deps.ins_.end(); ++i) {
+        uint64_t slash_bits;
+        CanonicalizePath(const_cast<char*>(i->str_), &i->len_, &slash_bits);
+        deps_nodes->push_back(state_->GetNode(*i, slash_bits));
+      }
+    } else {
```
digit-google, line 1061: nit: Add `/* deps_type == "zero" */` for clarity here

src/build.cc
```diff
@@ -1045,17 +1045,30 @@ bool Builder::ExtractDeps(BuildResult::CommandCompleted& result,
...
+      const char* ptr = content.c_str();
+      const char* const end = &content.back() + 1;
+      while (ptr < end) {
+        StringPiece piece = ptr;
```
digit-google, line 1066: This assumes that there is a trailing zero at the end of `contents`, which is not checked everywhere. Use `memchr()` instead to ensure a read-overflow will never be performed at runtime. And this will support files which omit the final zero in their output.

digit-google (commented):
This looks good, but needs an update to the manual to describe the new feature and a proper regression test in `misc/output_test.py`

src/build.cc
magistau: Well, [cppreference](https://en.cppreference.com/cpp/string/basic_string/c_str) says that `c_str()` is NUL-terminated. `ptr` always points past a NUL, so after the loop reads the last filename it would be either on the NUL terminating the whole string (if the file has a trailing NUL) or out of bounds by one (if it doesn't), either way `ptr < end` will fail, so out of bounds read is not possible.

magistau: See also https://github.com/ninja-build/ninja/pull/2655#discussion_r2678606425

magistau: Also, `src/build_test.cc` includes tests for both trailing NUL and no trailing NUL.

magistau:
@digit-google I've updated the manual. Not sure what's there to test in `misc/output_test.py`, as this change doesn't affect the output and the functionality is already tested in `src/build_test.cc`.

src/build.cc
digit-google: Oops, you are perfectly right. For some reason I thought this loop was iterating over a StringPiece. Sorry about the noise.

doc/manual.asciidoc
```diff
@@ -811,6 +811,12 @@ rule cc
   command = cl /showIncludes -c $in /Fo$out
 ----
 
+3. `deps = zero` specifies that the tool outputs the dependencies as a list of
+   NUL-terminated paths, like what `find . -print0` would produce.
+   Although this format is not used by any known C/C++ compilers, it is
+   the easiest to produce from a script, so it works in ad hoc scenarios.
+   Also, it is produced by https://typst.app/[Typst typesetting system].
```
digit-google: nit: remove this link, as it's not useful here.

src/build.cc
digit-google: And the unit-tests are good enough for me too.

digit-google (approved):
LGTM, @jhasse please take look

doc/manual.asciidoc
magistau: Done. I thought it's a good idea to list some compiler producing this format like the other two formats have.

digit-google (approved):
ping to @jhasse

PR #2660, opened 2025-08-19
Document some additional `-t` tool flags

### LINKED ISSUE #2658: `ninja -t` helpers are inconsistent about whether filenames are quoted
consider the following two commands:
```
$ ninja -t targets rule frontmatter | rg '!.md.json'
.build/2018-02-14-Speed-Up!.md.json
$ ninja -t inputs | rg '!.md.json'
'.build/2018-02-14-Speed-Up!.md.json'
```
in `-t targets`, the file path is printed as a literal string. in `-t inputs`, shell quoting rules are applied.

i think `-t inputs` is in the wrong here: the path is not embedded inside a command string, so there's no reason to quote it. but fixing it is probably hard due to backwards compatibility. maybe ninja could add a `-t inputs-unquoted`? or `-t inputs` could take an `--unquoted` flag?

jyn514:
Fixes https://github.com/ninja-build/ninja/issues/2658

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -270,6 +270,8 @@ and a web browser will be opened. This can be changed as follows:
 ninja -t browse --port=8000 --no-browser mytarget
 ----
 +
+Use `--help` to see a full list of options.
++
 `graph`:: output a file in the syntax used by `graphviz`, an automatic
 graph layout tool.  Use it like:
 +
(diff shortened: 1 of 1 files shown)


digit-google:
I'd prefer not to document all tool options in the manual, just add a line that says that each tool may have custom options that change its behavior, and to call it with `--help` to see details.

Technically, this PR is missing a lot of other tool options anyway. Plus since some options are added over time, this makes phrases like "Available since Ninja 1.12" incomplete / misleading anyway in general, unless one wants to document it for *every* tool option. This would be a needless maintenance burden.

jyn514:
> I'd prefer not to document all tool options in the manual, just add a line that says that each tool may have custom options that change its behavior, and to call it with `--help` to see details.

unfortunately `--help` doesn't work for all tools:
```
$ ninja -t cleandead --help
Cleaning... 1 files.
```
so we'd have to document which tools support `--help`, which doesn't seem like any less of a burden.

> since some options are added over time, this makes phrases like "Available since Ninja 1.12" incomplete / misleading anyway in general, unless one wants to document it for _every_ tool option.

maybe we can put a note at the top of the section that "Available since" is only about the tool itself, not its options? i agree that documenting when options were introduced would be a maintenance burden.

digit-google:
Good point. I'd be in favor of adding `--help` support to all tools instead, except maybe the `urtle` one ;-)

jyn514:
OMG that's a *good* tool

PR #2673, opened 2025-09-17
Print an entry in the compilation database for each input file

### LINKED ISSUE #1590: the compilation database only contains the first file/output
In the the compilation database JSON dumped by `ninja -t compdb link`, I noticed the "file" field only has the first item instead of all items that serve as the inputs for a command. This is also true for the "output" field.

This seems to be intentional - in [ninja.cc](https://github.com/ninja-build/ninja/blob/master/src/ninja.cc)'s method `NinjaMain::ToolCompilationDatabase`
```c++
        printf("\",\n    \"file\": \"");
        EncodeJSONString((*e)->inputs_[0]->path().c_str()); // only the first item
        printf("\",\n    \"output\": \"");
        EncodeJSONString((*e)->outputs_[0]->path().c_str()); // only the first item
```

But only recording the first item for these two fields in the compilation database defies the purpose of building a compilation database. This weird behavior makes subsequent tooling unreliable.

Thanks!

dblsaiko:
Targets which use multiple input files only have their first input file show up in the compilation database. This can cause issues with tools that build a list of files to inspect from the compilation database. For example, when using the Swift programming language together with the SourceKit-LSP language server, only one source file from each Swift module gets processed correctly by the language server using the current compilation database output from Ninja, because Swift modules are compiled with one compiler invocation and therefore one Ninja target for the whole module.

This changes the compilation database formatter to emit one entry for each input file of a target instead of just the first input file, which makes tools able to pick up all the input files used in the build.

Fixes #1590.
See also https://github.com/mesonbuild/meson/pull/14264.

### DIFF
--- a/misc/output_test.py
+++ b/misc/output_test.py
@@ -513,8 +513,20 @@ options:
   }},
   {{
     "directory": "{b.path}",
+    "command": "cat in2 out1 out2",
+    "file": "out1",
+    "output": "out2"
+  }},
+  {{
+    "directory": "{b.path}",
     "command": "cat out2 out1 out3",
     "file": "out2",
+    "output": "out3"
+  }},
+  {{
+    "directory": "{b.path}",
+    "command": "cat out2 out1 out3",
+    "file": "out1",
     "output": "out3"
   }}
 ]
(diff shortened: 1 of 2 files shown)


dblsaiko:
Fixed the test.

jhasse:
thx :)

Issue #2662, opened 2025-08-27
segfault in `Plan::NodeFinished`


dseomn:
ninja 1.12.1-1 just started segfaulting with my build, but not every time. The source files are ~12GB and have some things I'd rather not post to GitHub anyway[^1], but I could upload build.ninja and all the subninja, dyndep, and dep files if that would be useful.

[^1]: Nothing too sensitive though, just stuff like public blog post comments by users other than myself. If it's useful, I could share the files with a few people. Also, almost all of the size is from images, so I could probably replace all of those with placeholders to get the size down.

```
$ DEBUGINFOD_URLS=https://debuginfod.debian.net gdb -- ninja
GNU gdb (Debian 16.3-1) 16.3
Copyright (C) 2024 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ninja...

This GDB supports auto-downloading debuginfo from the following URLs:
  <https://debuginfod.debian.net>
Enable debuginfod for this session? (y or [n]) y
Debuginfod has been enabled.
To make this setting permanent, add 'set debuginfod enabled on' to .gdbinit.
Reading symbols from /home/dseomn/.cache/debuginfod_client/98b4d7c6d48fc2ddac2e9b8a920b39fd04245608/debuginfo...
(gdb) r
Starting program: /usr/bin/ninja 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[0/1] INIT[Detaching after vfork from child process 429980]
[1/1] INIT
[0/10] SCAN htaccess.jinja[Detaching after vfork from child process 429983]
[0/10] SCAN posts/2014-06-29-boxborough-and-stow/index.html.jinja[Detaching after vfork from child process 429984]
[0/10] SCAN posts/feeds.atom.jinja[Detaching after vfork from child process 429986]
[...]

Issue #2653, opened 2025-08-16
dynamic output -> dynamic input does not trigger rebuild, depending on order of build.ninja statements


dseomn:
With this build file that has a file `out-1` that's both a dynamic output of `stamp-1` and a dynamic input of `stamp-2`:

```ninja
rule dd
  command = printf 'ninja_dyndep_version = 1\nbuild $stamp | $dyn_out: dyndep | $dyn_in\n' > $stamp.dd

rule dyn_copy
  command = cp $dyn_in $dyn_out && touch $stamp
  dyndep = $stamp.dd

build stamp-1.dd: dd
  stamp = stamp-1
  dyn_in = source
  dyn_out = out-1

build stamp-2.dd: dd
  stamp = stamp-2
  dyn_in = out-1
  dyn_out = out-2

build stamp-1: dyn_copy || stamp-1.dd
  stamp = stamp-1
  dyn_in = source
  dyn_out = out-1

build stamp-2: dyn_copy || stamp-1.dd stamp-2.dd
  stamp = stamp-2
  dyn_in = out-1
  dyn_out = out-2
```

Both the initial and a second build work correctly:

```
$ touch source && ninja -d explain -v
ninja explain: output stamp-1.dd doesn't exist
ninja explain: output stamp-1 doesn't exist
ninja explain: output stamp-2.dd doesn't exist
ninja explain: output stamp-2 doesn't exist
[1/4] printf 'ninja_dyndep_version = 1\nbuild stamp-1 | out-1: dyndep | source\n' > stamp-1.dd
ninja explain: loading dyndep file 'stamp-1.dd'
ninja explain: output stamp-1 doesn't exist
ninja explain: output stamp-2 doesn't exist
[2/4] printf 'ninja_dyndep_version = 1\nbuild stamp-2 | out-2: dyndep | out-1\n' > stamp-2.dd
ninja explain: loading dyndep file 'stamp-2.dd'
ninja explain: out-1 is dirty
[3/4] cp source out-1 && touch stamp-1
[4/4] cp out-1 out-2 && touch stamp-2
$ touch source && ninja -d explain -v
ninja explain: loading dyndep file 'stamp-1.dd'
ninja explain: output stamp-1 older than most recent input source (1755384471679080143 vs 1755384474423035509)
ninja explain: loading dyndep file 'stamp-2.dd'
ninja explain: out-1 is dirty
[1/2] cp source out-1 && touch stamp-1
[...]

Issue #2666, opened 2025-09-02
[dyndep] edge "flip" from dynamic dependencies ends up in a cycle


mathstuf:
Dynamic dependencies add edges to a build graph during the build. These edges are stored in `.ninja_deps` as well. If a build has an edge `a -> b` from dyndeps that, on an incremental build, flips to `b -> a`, the `.ninja_deps`-sourced historical edges are not pruned before adding the new edges. This ends up with `ninja` reporting a cycle in the graph.

Reproducer (via CMake; `ninja`-only not yet made):

`CMakeLists.txt`:

```cmake
cmake_minimum_required(VERSION 3.30)
project(cm-switch-deps CXX)

set(CMAKE_CXX_STANDARD 20)

add_library(hasmodules)
target_sources(hasmodules
  PRIVATE
    FILE_SET CXX_MODULES
      FILES atob.cpp btoa.cpp)
```

`atob.cpp`:

```cpp
export module a;
// import b;
```

`btoa.cpp`:

```cpp
export module b;
import a;
```

If you configure this with `cmake -S. -Bbuild -GNinja` and `ninja -C build`, it builds fine. If you then comment out the `import a;` and uncomment `import b;` statements and rebuild, ninja reports a cycle between `a.gcm -> b.gcm -> a.gcm`. I believe that `ninja` should, upon loading dyndeps for `a.gcm`, drop edges from `a.gcm` sourced from `.ninja_deps`. Not sure of the feasibility without research.

mathstuf:
I suspect there are "rhyming" issues where `export module b;` moves between files with different `import` statements. So a test case where `b.gcm` is made by a different compilation also needs to be considered and tested.

moritzx22:
# Root Cause Analysis
There is one problem related to issue #2666.
- The deps log will lead to the cycle detection
## ./ninja_deps
```
$ ninja
ninja: no work to do.
// change the cpp files
$ ninja && ninja -t restat
...
$ ninja
ninja: error: dependency cycle: CMakeFiles/hasmodules.dir/a.pcm -> CMakeFiles/hasmodules.dir/b.pcm -> CMakeFiles/hasmodules.dir/a.pcm
```
The dependency graph looks like: red nodes are dirty
![Image](https://github.com/user-attachments/assets/76ebee75-93e3-418a-97a6-ebb662591f4b)
[master branch]

The dependency graph shall look like:
[...]

PR #2675, opened 2025-09-21
This commit will fix issue#2653.

moritzx22:
This Pull request will reflect that dynamic dependencies may change the graph during recomputeDirty of nodes. A newly added dyndep output may be dirty and can cause other targets to be dirty as well. This shall take effect if those nodes has been visited before. A new algorithm to iterate over the graph to make nodes dirty has been added.
For a detailed problem description, see [issue#2653](https://github.com/ninja-build/ninja/issues/2653)

### DIFF
--- a/src/build_test.cc
+++ b/src/build_test.cc
@@ -4405,3 +4405,109 @@ TEST_F(BuildTest, ValidationWithCircularDependency) {
   EXPECT_FALSE(builder_.AddTarget("out", &err));
   EXPECT_EQ("dependency cycle: validate -> validate_in -> validate", err);
 }
+
+// Test case: Late dyndep merges two initially separate dependency graphs into a
+// single unified graph. This verifies 'dirty propagation' once a dyndep file
+// introduces a new dependency edge at build time.
+// https://github.com/ninja-build/ninja/issues/2653
+TEST_F(BuildTest, LateDynDepPropagateDirty) {
+  ASSERT_NO_FATAL_FAILURE(AssertParse(&state_, R"ninja(
+rule mkdyn
+  command = printf 'ninja_dyndep_version = 1\nbuild  a3 | input_b: dyndep |\n' > dd
+
+rule cp
+  command = cp $in $out
+
+rule cc
+  command = cp $in $out && touch input_b
+
+build dd: mkdyn
+
+build b1: cp input_b
+build b2: cp b1
+build b3: cp b2
+build b4: cp b3
+build b5: cp b4
+
+build a1: cp input_a
+build a2: cp a1
+build a3: cc a2 || dd
+  dyndep = dd
+build a4: cp a3
+build a5: cp a4
+)ninja"));
+
+  fs_.Create("input_b", "");
+  fs_.Create("b1", "");
+  fs_.Create("b2", "");
+  fs_.Create("b3", "");
+  fs_.Create("b4", "");
+  fs_.Create("b5", "");
+
+  fs_.Create("input_a", "");
+  fs_.Create("a1", "");
+  fs_.Create("a2", "");
+  fs_.Create("a3", "");
+  fs_.Create("a4", "");
+  fs_.Create("a5", "");
+  fs_.Create("dd",
+             "ninja_dyndep_version = 1\n"
+             "build  a3 | input_b: dyndep |\n");
+
+  string err;
+
+  fs_.Tick();
+  fs_.Create("input_a", "");
+
+  command_runner_.commands_ran_.clear();
+  state_.Reset();
+
+  EXPECT_TRUE(builder_.AddTarget("b5", &err));
+  EXPECT_EQ("", err);
+  err.clear();
+  EXPECT_TRUE(builder_.AddTarget("a5", &err));
+  EXPECT_EQ("", err);
+  err.clear();
+
+  EXPECT_TRUE(GetNode("a5")->dirty());
+  // After loading dyndep, dirty propagation should mark the entire B chain dirty.
+  EXPECT_TRUE(GetNode("b5")->dirty());
+  EXPECT_TRUE(GetNode("b4")->dirty());
+  EXPECT_TRUE(GetNode("b3")->dirty());
+  EXPECT_TRUE(GetNode("b2")->dirty());
+  EXPECT_TRUE(GetNode("b1")->dirty());
+
+  EXPECT_EQ(builder_.Build(&err), ExitSuccess);
+  EXPECT_EQ("", err);
+
+  EXPECT_EQ(10u, command_runner_.commands_ran_.size());
+
+  // Second scenario: test top‑level node detection in PropagateDirty.
+  // Now we only request b4 (not b5).
+  // This checks that PropagateDirty correctly identifies top‑level nodes
+  // even when they still have dependent edges.
+  command_runner_.commands_ran_.clear();
+  state_.Reset();
+
+  fs_.Tick();
+  fs_.Create("input_a", "");
+
+  EXPECT_TRUE(builder_.AddTarget("b4", &err));
+  EXPECT_EQ("", err);
+  err.clear();
+  EXPECT_TRUE(builder_.AddTarget("a5", &err));
+  EXPECT_EQ("", err);
+  err.clear();
+
+  EXPECT_TRUE(GetNode("a5")->dirty());
+  EXPECT_FALSE(GetNode("b5")->dirty());
+  EXPECT_TRUE(GetNode("b4")->dirty());
+  EXPECT_TRUE(GetNode("b3")->dirty());
+  EXPECT_TRUE(GetNode("b2")->dirty());
+  EXPECT_TRUE(GetNode("b1")->dirty());
+
+  EXPECT_EQ(builder_.Build(&err), ExitSuccess);
+  EXPECT_EQ("", err);
+
+  EXPECT_EQ(9u, command_runner_.commands_ran_.size());
+}
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -79,6 +79,67 @@ bool DependencyScan::RecomputeDirty(Node* initial_node,
   return true;
 }
 
+namespace {
+class PropagateDirty {
+ public:
+  /// Propagate "dirty" status through all given edges.
+  /// Any node that becomes dirty and has no further outgoing edges
+  /// is added to 'topLevelNodes'.
+  static void call(const std::vector<Edge*>& edges,
+                   std::vector<Node*>* topLevelNodes);
+
+ private:
+  PropagateDirty(std::vector<Node*>* topLevelNodes)
+      : top_level_nodes_(topLevelNodes) {}
+
+  /// returns true if at least one of its outputs was originally marked 'clean'.
+  bool process(Edge* edge);
+
+  std::vector<Node*>* top_level_nodes_;
+};
+
+void PropagateDirty::call(const std::vector<Edge*>& edges,
+                          std::vector<Node*>* topLevelNodes) {
+  PropagateDirty propagateDirtyDyndep(topLevelNodes);
+  for (Edge* outedges : edges)
+    propagateDirtyDyndep.process(outedges);
+}
+
+bool PropagateDirty::process(Edge* const edge) {
+  bool clean_out = false;
+  for (Node* const out : edge->outputs_) {
+    // Only propagate if the node is currently clean.
+    if (out->clean()) {
+      clean_out = true;
+      out->MarkDirty();
+      edge->outputs_ready_ = false;
+
+      // If the node has no outgoing edges, it is a top‑level target.
+      if (out->out_edges().empty()) {
+        top_level_nodes_->push_back(out);
+        continue;
+      }
+
+      // Track whether any dependent edges had clean outputs.
+      bool dependent_clean = false;
+
+      for (Edge* outEdge : out->out_edges()) {
+        if (outEdge && process(outEdge)) {
+          dependent_clean = true;
+        }
+      }
+
+      // If this node have outgoing edges, but none of those edges produced
+      // clean outputs, then this node is effectively a top‑level node in the
+      // dirty propagation chain.
+      if (!out->out_edges().empty() && !dependent_clean)
+        top_level_nodes_->push_back(out);
+    }
+  }
+  return clean_out;
+}
+}  // namespace
+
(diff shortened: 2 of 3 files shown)


moritzx22:
# Test Bench
Example build file for testing:
```
rule cp
  command = cp $in $out

rule cp_dynB
  command = cp $in $out && touch $stamp && touch $output
  dyndep = $stamp.dd

rule mkdyn
  command = printf 'ninja_dyndep_version = 1\nbuild  $st_out | $dyn_out: dyndep | $dyn_in\n' > $stamp.dd

build b1: cp  b
build b2: cp  b1
build b3: cp  b2
build b4: cp  b3
build b5: cp  b4

build a1: cp  a
build a2: cp  a1
build a3: cp_dynB a2 || stamp-a.dd
  stamp = stamp-a
  output = b
build a4: cp  a3
build a5: cp  a4

build stamp-a.dd: mkdyn
  stamp = stamp-a
  dyn_out = b
  st_out = a3
```
now do the following commands:
```
$ touch a b && ninja
$ touch a && ninja b5 a5
$ touch a && ninja b5 a5
```
In the master branch are issues, not the required processes are always started. The solution of this Pull Request works well.

# Analysis
the next figure reports the dependency graph without dyndep[noDyndep]:
noDyndep
[noDyndep]

The next figure reports the dependency graph with dyndep[complete], using this Pull Request, and target `a` is touched:
complete
[complete]

The red means `dirty`, blue means `clean`.

The next figure reports the dependency graph using master branch, and target `a` is touched:
master
[master]

It can be easily seen that targets with `b` are not considered to be dirty by the master branch.

## Scanning Algorithm Master branch

1. First target b5 will be added. There are still 2 independent graphs a and b, see figure [noDyndep]. Nothing in graph `b` is dirty, and nothing is detected.
2. Second target a5 is added. Now the dyndep is added, the graph dependency changes. Only nodes with an `a` are visited and are marked as dirty. The targets with an `b` are not marked as dirty.
3. no further action, as b5 has already been added, any `b` is missed to be marked as dirty.

The algorithm does not reflect that the graph may change. This issue occurs if a dyndep adds an output target (here `b`) and if this target is dirty and has other outputs.

## Proposed Scanning Algorithm enhancement

if the dyndep is applied, there is an additional iterating algorithm across the graph following the outputs. The outputs are only followed if they have been visited before. Therefore a node can now be dirty, clean and spare. This does prevent to mark nodes as dirty which are not requested to be built.
The command `$ touch a && ninja a5` would not follow the `b` targets.

The additional iterating over the graph starts here:  [iterating](https://github.com/ninja-build/ninja/pull/2675/files#r2366300501)

src/graph.cc
```diff
@@ -202,8 +236,16 @@ bool DependencyScan::RecomputeNodeDirty(Node* node, std::vector<Node*>* stack,
   // Finally, visit each output and update their dirty state if necessary.
   for (vector<Node*>::iterator o = edge->outputs_.begin();
        o != edge->outputs_.end(); ++o) {
-    if (dirty)
+    if (dirty) {
       (*o)->MarkDirty();
+      if (dyndepLoaded) {
```
moritzx22: Here the additional iterating of the graph may start...

moritzx22:
The recently pushed commit will address issue[#2641](https://github.com/ninja-build/ninja/issues/2641), it is a simular dyndep issue as #2653.

# Test Bench
Similar as reported in the issue, just one target was added `abc`
```
rule dyndep
  command = printf 'ninja_dyndep_version = 1\nbuild stamp | other: dyndep\n' > $out

rule build_with_dyndep
  command = touch $out other
  dyndep = $out.dd

rule depends_on_dyndep
  command = printf '%s: other\n' $out > $out.d && cp other $out
  depfile = $out.d
  
rule cp
  command = cp $in $out

build stamp.dd: dyndep
build stamp: build_with_dyndep || stamp.dd
build abc: cp stamp
build copy: depends_on_dyndep || abc
```

Following graph with proposed solution
a
Following graph without proposed solution (copy is not dirty and therefore not build)
a

blue is not build, red is build.
The following commands has been used to generate the graphics.
```
$ ninja -v -d explain
adding spaces to the build_with_dyndep command
$ ninja -v -d explain
```

moritzx22:
updates:
- fix for issue 2641
- fix and update in unittest

dseomn:
A previous version of this fixed #2662 (see https://github.com/ninja-build/ninja/issues/2653#issuecomment-3259696152), but I just tested https://github.com/ninja-build/ninja/compare/c74beb43695ff8cae955f05a6cbe1753aa9517f8...moritzx22:8e0cf13f9d2bafb03f5ff02e66862e45c53dac42 on ninja 1.13.2, and it doesn't fix #2662 anymore. (It doesn't claim to fix that either, so that's fine. Just FYI in case you thought it did still fix that.)

moritzx22:
> A previous version of this fixed #2662 (see [#2653 (comment)](https://github.com/ninja-build/ninja/issues/2653#issuecomment-3259696152)), but I just tested [c74beb4...moritzx22:8e0cf13f9d2bafb03f5ff02e66862e45c53dac42](https://github.com/ninja-build/ninja/compare/c74beb43695ff8cae955f05a6cbe1753aa9517f8...moritzx22:8e0cf13f9d2bafb03f5ff02e66862e45c53dac42) on ninja 1.13.2, and it doesn't fix #2662 anymore. (It doesn't claim to fix that either, so that's fine. Just FYI in case you thought it did still fix that.)

This PR was never intended to fix 2662. I previously created a separate branch [quickfixIssue2653](https://github.com/moritzx22/ninja/tree/quickfixIssue2653) as a quick workaround for issue 2663, without opening a PR for it. As described in this [comment](https://github.com/ninja-build/ninja/issues/2653#issuecomment-3240429907), that branch also happened to fix 2662 — likely by coincidence.
- Issue 2663 is well understood and is directly addressed by this PR.
- Issue 2662 is not well understood and is not targeted here.
  - not reproducible, difficult to address


Both the quickfix branch and this PR attempt to solve similar underlying problems (Certain dynamic dependencies are not loaded in time for correctly computing the dirty state of a node), but in very different ways:
 - the quickfix branch uses a broad “sledgehammer” approach (load all dynamic dependencies in a separate algorithm, prior to computing the dirty state of the nodes),
 - this PR applies a much more targeted fix for a specific issue.

jhasse:
@bradking What do you think?

mathstuf:
Adding a comment here to remind myself to review this for #2666 effects as well.

bradking:
The analysis and proposed solution in https://github.com/ninja-build/ninja/pull/2675#issuecomment-3316101189 ~~LGTM~~.  EDIT: After looking again, perhaps not.  See https://github.com/ninja-build/ninja/issues/2653#issuecomment-4085264303.  

I've not reviewed the implementation though.

mathstuf:
This does not solve #2666 (but wanted to check).

mathstuf:
I have verified that CMake's test suite for C++ modules and Fortran still work with this PR as well.

I've not reviewed the code.

moritzx22:
> This does not solve #2666 (but wanted to check).
 
- 2666
  - Root issue: the build graph is updated based on outdated depfiles, which can lead to false cycle detection.
    - This is not directly tied to dynamic dependency handling.
  - The problem also appears in the CMake/module area.
  - The underlying cause is already well understood.
  - In terms of impact, 2666 is more critical than PR 2675.
    - build will fail, and resolving the issue may require a full rebuild.
  - I agree with @mathstuf that PR 2675 is unrelated to 2666.

AJIOB:
Hi all. Do we have any plans to actualize & merge this PR?

moritzx22:
## changes in recent push
- Rebased the branch onto the latest master to resolve merge conflicts.
- Minor cleanups
  - Renamed identifiers for clarity.
  - Small code improvements.
  - Updated and clarified comments.
- Fixes for issue 2653:
  - The detection of top‑level nodes in `PropagateDirty` was incorrect in cases where a node still had outgoing edges but none of those edges produced clean outputs.
  - The detection logic has been revised so these nodes are now correctly identified as top‑level in the dirty‑propagation process.
  - Unit tests have been expanded and updated to cover these scenarios.

moritzx22:
Friendly ping to @bradking, do you see a chance to review the code?

bradking:
@moritzx22 let's simplify review by taking one issue at a time.  Please drop the part that fixes #2641.  That can be fixed in a follow-up PR.  Let's focus on #2653 here.  Please squash the test case into the main commit, and revise the message as follows:

<details><summary>Suggested commit message</summary>

```
dyndep: Build a portion of the graph made reachable by dynamic dependencies

Dynamic dependencies can modify the build graph during RecomputeDirty,
including adding new outputs or introducing new dependency edges.  A newly
added dyndep output may already be dirty and must be able to propagate
dirtiness to its dependents, even if those nodes were previously visited during
the initial dirty-marking pass.

Update the dirty-propagation logic to correctly handle graph changes introduced
by dyndep.  A new traversal algorithm ensures that all affected nodes are
revisited and marked dirty when necessary.

Fixes: #2653
```

</details>

moritzx22:
## Recent Push
- Issue 2641 has been removed at this stage
  - Note that certain graph‑related problems can only be resolved when issues 2641 and 2653 are fixed together
- Unit tests have been slightly modified
  - The Ninja build file for the unit tests can be copied and used directly on Linux‑like operating systems, where it is fully functional
- The commit message has been improved, and only one single commit remains

src/build_test.cc
```diff
@@ -4405,3 +4405,109 @@ TEST_F(BuildTest, ValidationWithCircularDependency) {
...
+  EXPECT_TRUE(GetNode("b1")->dirty());
+
+  EXPECT_EQ(builder_.Build(&err), ExitSuccess);
+  EXPECT_EQ("", err);
+
+  EXPECT_EQ(10u, command_runner_.commands_ran_.size());
+
+  // Second scenario: test top‑level node detection in PropagateDirty.
+  // Now we only request b4 (not b5).
+  // This checks that PropagateDirty correctly identifies top‑level nodes
```
bradking: Please use ascii `-` instead of a non-breaking hyphen:

```suggestion
  // Second scenario: test top-level node detection in PropagateDirty.
  // Now we only request b4 (not b5).
  // This checks that PropagateDirty correctly identifies top-level nodes
```

bradking:
I tested this locally, and it can cause build statements to run more than once.  Building the example in #2641 runs all three build statements twice.

AJIOB:
Looks like it's expected: you suggest to remove the #2641 fix from that PR.

So, is it OK for you, @bradking ?

bradking:
The problem also happens with both changes as of commit `4ac01c1618` too.  I just pointed at #2641 as one example.  It's a general problem.  Depending on evaluation order, I think the change here may cause already-built nodes to be considered dirty again.

In https://github.com/ninja-build/ninja/issues/2653#issuecomment-4085264303 I posted an analysis of that issue's use case showing that the change here may not be needed.

moritzx22:
Focusing on issue #2653, the newly added unit test already captures the failing behavior:

* When executed against the current `master` branch, the test fails.
* When executed against the proposed fix, the test passes and GitHub CI succeeds.

The test also validates several important correctness properties:

* It counts the number of executed  [jobs](https://github.com/moritzx22/ninja/blob/32d470da38299852898284944be4660047cd326b/src/build_test.cc#L4483) and confirms that all expected jobs are run exactly once.
* It verifies that the dirtiness of all nodes is computed correctly.
* After a full build, no nodes remain dirty.

To conveniently reproduce the `master` behavior using the proposed fix, you can temporarily comment out the following lines [L276-L277](https://github.com/moritzx22/ninja/blob/32d470da38299852898284944be4660047cd326b/src/graph.cc#L276-L277):
```cpp
if (dyndepLoaded)
  PropagateDirty::call(out->out_edges(), validation_nodes);
```

With these lines disabled, the algorithm behaves like the current `master` branch, and the unit test fails again. This demonstrates that the additional propagation step is required to resolve the issue.

As an alternative reproduction, the following minimal example works as a standalone [build.ninja](https://github.com/moritzx22/ninja/blob/32d470da38299852898284944be4660047cd326b/src/build_test.cc#L4415-L4437) on Linux:
<details>
<summary>Show full `build.ninja` file</summary>

```ninja
rule mkdyn
  command = printf 'ninja_dyndep_version = 1\nbuild  a3 | input_b: dyndep |\n' > dd

rule cp
  command = cp $in $out
rule cc
  command = cp $in $out && touch input_b

build dd: mkdyn

build b1: cp input_b
build b2: cp b1
build b3: cp b2
build b4: cp b3
build b5: cp b4

build a1: cp input_a
build a2: cp a1
build a3: cc a2 || dd
  dyndep = dd
build a4: cp a3
build a5: cp a4
```
</details>
On the current `master` branch, the build result depends on the order of targets, leading to inconsistent behavior:

```
$ touch input_a && ninja b5 a5
[5/5] cp a4 a5

$ touch input_a && ninja a5 b5
[10/10] cp b4 b5
```

A correct build should always execute 10 jobs. The proposed fix ensures consistent behavior regardless of target order: all 10 jobs are executed, and no nodes remain dirty after a single run.

moritzx22:
> Depending on evaluation order, I think the change here may cause already-built nodes to be considered dirty again.

Thanks for pointing this out. Could you provide a minimal example build along with the exact steps to reproduce the issue? Once I have that, I can analyze whether the change is indeed causing nodes to be marked dirty unexpectedly and determine the correct fix.

moritzx22:
> In https://github.com/ninja-build/ninja/issues/2653#issuecomment-4085264303 I posted an analysis of that issue's use case showing that the change here may not be needed.

Your analysis is specific to the example build in issue #2653. I provided a separate analysis here: https://github.com/ninja-build/ninja/pull/2675#issuecomment-3316101189,  based on a different and intentionally minimal build file.  This example contains essentially only a single dyndep output that produces a manifest output, and it is designed to isolate the exact behavior under discussion.

Your explanation does not show why the build example used in this PR would not demonstrate a real issue, nor how your reasoning from the other example can be transferred to this one. From my perspective, the example here is valid and accurately illustrates the problem, and I do believe there is a problem.

Could you please provide an analysis based directly on the build example in this PR, addressing:
 - why you believe Ninja’s behavior in this specific example is correct
 - in what way my analysis of this example is incorrect
 - how the example would need to be changed to not having issues

bradking:
Okay, I'll need some time to investigate further and post a more detailed analysis.

bradking:
See https://github.com/ninja-build/ninja/issues/2653#issuecomment-4135823988.

PR #2680, opened 2025-10-12
Only load depsfile if not dirty [Fix #2666]

### LINKED ISSUE #2666: [dyndep] edge "flip" from dynamic dependencies ends up in a cycle

moritzx22:
This Merge request is related to [#2666](https://github.com/ninja-build/ninja/issues/2666).

This pull request proposes to load the depsfile only if it is not dirty.
For a more detailed description, see [comment #2666](https://github.com/ninja-build/ninja/issues/2666#issuecomment-3394810561)

### DIFF
--- a/src/build_test.cc
+++ b/src/build_test.cc
@@ -685,6 +685,17 @@ bool FakeCommandRunner::StartCommand(Edge* edge) {
         fs_->WriteFile(edge->outputs_[i]->path(), content, false);
       }
     }
+  } else if (edge->rule().name() == "cp-deps") {
+    // like copy, additionally write to file d.d
+    assert(!edge->inputs_.empty());
+    assert(edge->outputs_.size() >= 1);
+    string content;
+    string err;
+    if (fs_->ReadFile(edge->inputs_[0]->path(), &content, &err) ==
+        DiskInterface::Okay) {
+      fs_->WriteFile(edge->outputs_[0]->path(), content, false);
+      fs_->Create("d.d", content);
+    }
   } else if (edge->rule().name() == "touch-implicit-dep-out") {
     string dep = edge->GetBinding("test_dependency");
     fs_->Tick();
@@ -1040,6 +1051,7 @@ TEST_F(BuildTest, DepFileMissing) {
   ASSERT_NO_FATAL_FAILURE(AssertParse(&state_,
 "rule cc\n  command = cc $in\n  depfile = $out.d\n"
 "build fo$ o.o: cc foo.c\n"));
+  fs_.Create("fo o.o", "");
   fs_.Create("foo.c", "");
 
   EXPECT_TRUE(builder_.AddTarget("fo o.o", &err));
@@ -1057,6 +1069,7 @@ TEST_F(BuildTest, DepFileOK) {
   Edge* edge = state_.edges_.back();
 
   fs_.Create("foo.c", "");
+  fs_.Create("foo.o", "");
   GetNode("bar.h")->MarkDirty();  // Mark bar.h as missing.
   fs_.Create("foo.o.d", "foo.o: blah.h bar.h\n");
   EXPECT_TRUE(builder_.AddTarget("foo.o", &err));
@@ -1090,6 +1103,7 @@ TEST_F(BuildTest, DepFileParseError) {
 "rule cc\n  command = cc $in\n  depfile = $out.d\n"
 "build foo.o: cc foo.c\n"));
   fs_.Create("foo.c", "");
+  fs_.Create("foo.o", "");
   fs_.Create("foo.o.d", "randomtext\n");
   EXPECT_FALSE(builder_.AddTarget("foo.o", &err));
   EXPECT_EQ("foo.o.d: expected ':' in depfile", err);
@@ -1126,6 +1140,7 @@ TEST_F(BuildTest, OrderOnlyDeps) {
   Edge* edge = state_.edges_.back();
 
   fs_.Create("foo.c", "");
+  fs_.Create("foo.o", "");
   fs_.Create("otherfile", "");
   fs_.Create("foo.o.d", "foo.o: blah.h bar.h\n");
   EXPECT_TRUE(builder_.AddTarget("foo.o", &err));
@@ -1244,6 +1259,7 @@ TEST_F(BuildTest, DepFileCanonicalize) {
 "build gen/stuff\\things/foo.o: cc x\\y/z\\foo.c\n"));
 
   fs_.Create("x/y/z/foo.c", "");
+  fs_.Create("gen/stuff/things/foo.o", "");
   GetNode("bar.h")->MarkDirty();  // Mark bar.h as missing.
   // Note, different slashes from manifest.
   fs_.Create("gen/stuff\\things/foo.o.d",
@@ -4318,6 +4334,8 @@ TEST_F(BuildTest, ValidationDependsOnOutput) {
   EXPECT_EQ("cat in2 > validate", command_runner_.commands_ran_[0]);
 }
 
+// Note: The graph in this unittest contains an issue
+// in terms of "ninja -t missingdeps"
 TEST_F(BuildWithDepsLogTest, ValidationThroughDepfile) {
   const char* manifest =
       "build out: cat in |@ validate\n"
@@ -4367,6 +4385,42 @@ TEST_F(BuildWithDepsLogTest, ValidationThroughDepfile) {
 
   {
     fs_.Create("in2", "");
+    fs_.Create("in", "");
+
+    State state;
+    ASSERT_NO_FATAL_FAILURE(AddCatRule(&state));
+    ASSERT_NO_FATAL_FAILURE(AssertParse(&state, manifest));
+
+    DepsLog deps_log;
+    ASSERT_TRUE(deps_log.Load(deps_log_file_.path(), &state, &err));
+    ASSERT_TRUE(deps_log.OpenForWrite(deps_log_file_.path(), &err));
+    ASSERT_EQ("", err);
+
+    Builder builder(&state, config_, NULL, &deps_log, &fs_, &status_, 0);
+    SafeRelease protect(&builder);
+    builder.command_runner_.reset(&command_runner_);
+
+    EXPECT_TRUE(builder.AddTarget("out2", &err));
+    ASSERT_EQ("", err);
+
+    EXPECT_EQ(builder.Build(&err), ExitSuccess);
+    EXPECT_EQ("", err);
+
+    // The depslog is loaded as out2 is not dirty due to the manifest,
+    // it is dirty only based on the depslog
+    // The out and validate actions should have been run as well as out2.
+    ASSERT_EQ(command_runner_.commands_ran_.size(), size_t(3));
+    // out has to run first, as both out2 and validate depend on it.
+    EXPECT_EQ("cat in > out", command_runner_.commands_ran_[0]);
+
+    deps_log.Close();
+  }
+
+  fs_.Tick();
+  command_runner_.commands_ran_.clear();
+
+  {
+    fs_.Create("in2", "");
     fs_.Create("in3", "");
 
     State state;
@@ -4388,10 +4442,10 @@ TEST_F(BuildWithDepsLogTest, ValidationThroughDepfile) {
     EXPECT_EQ(builder.Build(&err), ExitSuccess);
     EXPECT_EQ("", err);
 
-    // The out and validate actions should have been run as well as out2.
-    ASSERT_EQ(command_runner_.commands_ran_.size(), size_t(3));
-    // out has to run first, as both out2 and validate depend on it.
-    EXPECT_EQ("cat in > out", command_runner_.commands_ran_[0]);
+    // The depslog is not loaded as out2 is dirty due to manifest input in3
+    // The validate will not run
+    ASSERT_EQ(command_runner_.commands_ran_.size(), size_t(1));
+    EXPECT_EQ("cat in3 > out2", command_runner_.commands_ran_[0]);
 
     deps_log.Close();
   }
@@ -4473,4 +4527,40 @@ TEST_F(StateTestWithBuiltinRules, ComplexTargetPreserved) {
   ASSERT_NE(node, nullptr);
 
   EXPECT_EQ(node->path(), "foo %2F bar?baz&x=1");
+}
+
+// Verifies that a stale depfile does not introduce a dependency cycle.
+// https://github.com/ninja-build/ninja/issues/2666
+TEST_F(BuildTest, CycleWithOldDepfile) {
+  ASSERT_NO_FATAL_FAILURE(AssertParse(&state_, R"ninja(
+rule cp
+  command = cp $in $out
+rule cp-deps
+  command = cp $in $out && printf 'b: X' > d.d
+build b: cp-deps a
+  depfile = d.d
+build c: cp b
+build d: cp c
+)ninja"));
+
+  fs_.Create("a", "b: X");    // depfile content to be generated in this run
+  fs_.Create("d.d", "b: d");  // depfile from last run, cycle "d -> c -> b -> d"
+  fs_.Create("X", "");
+
+  string err;
+  ASSERT_TRUE(builder_.AddTarget("d", &err)) << " Error: " << err;
+  EXPECT_EQ("", err);
+
+  EXPECT_EQ(builder_.Build(&err), ExitSuccess);
+  EXPECT_EQ("", err);
+  err.clear();
+
+  ASSERT_EQ(3u, command_runner_.commands_ran_.size());
+  EXPECT_EQ("cp a b && printf 'b: X' > d.d", command_runner_.commands_ran_[0]);
+  EXPECT_EQ("cp b c", command_runner_.commands_ran_[1]);
+  EXPECT_EQ("cp c d", command_runner_.commands_ran_[2]);
+  std::string content;
+  fs_.ReadFile("d.d", &content, &err);
+  EXPECT_EQ("b: X", content);
+  EXPECT_EQ("", err);
 }
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -32,6 +32,211 @@
 
 using namespace std;
 
+namespace {
+
+/// execute hash only once in lifetime of object and only on request
+struct LazyEdgeCommandHash {
+  LazyEdgeCommandHash(const Edge* edge) : edge_(edge) {}
+  std::uint64_t operator()() {
+    if (!valid_) {
+      valid_ = true;
+      return command_ = BuildLog::LogEntry::HashCommand(
+                 edge_->EvaluateCommand(/*incl_rsp_file=*/true));
+    }
+    return command_;
+  }
+
+ private:
+  const Edge* edge_;
+  std::uint64_t command_ = 0;
+  bool valid_ = false;
+};
+
+/// Performance‑optimized helper class for recomputing the dirty state of
+/// outputs.
+class RecomputeOutputsDirtyCache {
+
+  /// Cached wrapper around BuildLog::LookupByOutput.
+  ///
+  /// This class behaves like an pointer to BuildLog::LogEntry; the pointer may
+  /// be nullptr if no entry exists. The lookup result is cached after the first
+  /// call to avoid repeated BuildLog lookups.
+  class CachedLogEntry {
+   public:
+    CachedLogEntry(){};
+
+    /// check for nullptr
+    bool is_valid() const { return entry_; }
+    /// When calling this function repeatedly for the same object,
+    /// the `buildLog` and `output` pointers must remain unchanged.
+    /// Stable pointer identity is required for correct caching behavior.
+    bool LookupByOutput(const BuildLog* buildLog, const Node* output);
+    const BuildLog::LogEntry* operator->() const { return entry_; }
+
+   private:
+    bool evaluated_ = false;
+    BuildLog::LogEntry* entry_ = nullptr;
+
+#ifndef NDEBUG
+    const Node* checkOutput_ = nullptr;
+#endif
+  };
+
+ public:
+  RecomputeOutputsDirtyCache(BuildLog* build_log,
+                             OptionalExplanations& explanations, Edge* edge)
+      : buildLog_(build_log), explanations_(explanations), edge_(edge),
+        logEntry_(edge->outputs_.size()) {}
+
+  /// Determines whether at least one output of edge 'edge_' is considered dirty.
+  ///
+  /// Returns:
+  /// - `true`  if at least one output is dirty
+  /// - `false` if all outputs are clean
+  bool all(const Node* most_recent_input);
+
+ private:
+  /// Checks whether `output` of `edge_` is dirty.
+  /// Returns true if dirty.
+  bool RecomputeOutputDirty(const Node* output, const Node* most_recent_input,
+                            CachedLogEntry& entry);
+
+  /// Dirty check for phony edges.
+  /// Returns true if dirty.
+  bool Phony(Node* output, const Node* most_recent_input) const;
+
+  const BuildLog* const buildLog_;
+  OptionalExplanations& explanations_;
+  const Edge* const edge_;
+
+  const bool isRestat_ = edge_->GetBindingBool("restat");
+  bool generator_ = false;
+  bool generatorValid_ = false;
+  std::vector<CachedLogEntry> logEntry_;
+  LazyEdgeCommandHash commandHash_ = LazyEdgeCommandHash(edge_);
+};
+
+bool RecomputeOutputsDirtyCache::CachedLogEntry::LookupByOutput(
+    const BuildLog* buildLog, const Node* output) {
+  if (evaluated_) {
+    assert(output == checkOutput_);
+    return entry_;
+  }
+  evaluated_ = true;
+  assert((checkOutput_ = output, true));
+  return (entry_ = buildLog->LookupByOutput(output->path()));
+}
+
+bool RecomputeOutputsDirtyCache::all(const Node* most_recent_input) {
+  for (std::size_t i = 0; i != edge_->outputs_.size(); ++i) {
+    const bool outputDirty =
+        edge_->is_phony()
+            ? Phony(edge_->outputs_[i], most_recent_input)
+            : RecomputeOutputDirty(edge_->outputs_[i], most_recent_input,
+                                   logEntry_[i]);
+    if (outputDirty)
+      return true;
+  }
+  return false;
+}
+
+bool RecomputeOutputsDirtyCache::Phony(Node* output,
+                                       const Node* most_recent_input) const {
+  // Phony edges don't write any output.  Outputs are only dirty if
+  // there are no inputs or validations and we're missing the output.
+  // If a phony target has inputs or validations, or the output exists,
+  // they are used for dirty calculation instead of this fallback.
+  if (edge_->inputs_.empty() && edge_->validations_.empty() &&
+      !output->exists()) {
+    explanations_.Record(output,
+                         "output %s of phony edge with no inputs doesn't exist",
+                         output->path().c_str());
+    return true;
+  }
+
+  // Update the mtime with the newest input. Dependents can thus call mtime()
+  // on the fake node and get the latest mtime of the dependencies
+  if (most_recent_input) {
+    output->UpdatePhonyMtime(most_recent_input->mtime());
+  }
+
+  // Phony edges are clean, nothing to do
+  return false;
+}
+
+bool RecomputeOutputsDirtyCache::RecomputeOutputDirty(
+    const Node* output, const Node* most_recent_input, CachedLogEntry& entry) {
+  // Dirty if we're missing the output.
+  if (!output->exists()) {
+    explanations_.Record(output, "output %s doesn't exist",
+                         output->path().c_str());
+    return true;
+  }
+
+  // If this is a restat rule, we may have cleaned the output in a
+  // previous run and stored the command start time in the build log.
+  // We don't want to consider a restat rule's outputs as dirty unless
+  // an input changed since the last run, so we'll skip checking the
+  // output file's actual mtime and simply check the recorded mtime from
+  // the log against the most recent input's mtime (see below)
+  bool used_restat = false;
+  if (isRestat_ && buildLog_ && entry.LookupByOutput(buildLog_, output)) {
+    used_restat = true;
+  }
+
+  // Dirty if the output is older than the input.
+  if (!used_restat && most_recent_input &&
+      output->mtime() < most_recent_input->mtime()) {
+    explanations_.Record(output,
+                         "output %s older than most recent input %s "
+                         "(%" PRId64 " vs %" PRId64 ")",
+                         output->path().c_str(),
+                         most_recent_input->path().c_str(), output->mtime(),
+                         most_recent_input->mtime());
+    return true;
+  }
+
+  if (buildLog_) {
+    if (!generatorValid_) {
+      generator_ = edge_->GetBindingBool("generator");
+      generatorValid_ = true;
+    }
+    if (entry.LookupByOutput(buildLog_, output)) {
+      if (!generator_ && commandHash_() != entry->command_hash) {
+        // May also be dirty due to the command changing since the last build.
+        // But if this is a generator rule, the command changing does not make
+        // us dirty.
+        explanations_.Record(output, "command line changed for %s",
+                             output->path().c_str());
+        return true;
+      }
+      if (most_recent_input && entry->mtime < most_recent_input->mtime()) {
+        // May also be dirty due to the mtime in the log being older than the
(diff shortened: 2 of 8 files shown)


moritzx22:
running the example from #2666 with the proposed solution, reports no cycle and builds as expected
```
...
// change the cpp files
$ ninja
[6/6] Linking CXX static library libhasmodules.a
$ ninja
ninja: no work to do.
```
the solution does work for the build from #2666 but it is still a draft.

src/build_test.cc
```diff
@@ -534,6 +534,17 @@ struct FakeCommandRunner : public CommandRunner {
   VirtualFileSystem* fs_;
 };
 
+class SafeRelase {
```
mathstuf:
```suggestion
class SafeRelease {
```

mathstuf (changes requested):
I think the test suite leak fix can be its own PR. Can a test case for the compelling scenario be added?

src/build_test.cc
moritzx22: done

moritzx22:
> think the test suite leak fix can be its own PR

new PR created: #2684. I will keep this PR unchanged, until PR2684 is merged and a rebase can be done.

moritzx22:
> Can a test case for the compelling scenario be added?

One cycle test for the depfile has been added. A similar test for dyndep is pending.
Numerous tests still fail because this PR does change some basic rules, ninja is designed to and this is reflected in the test suite.

moritzx22:
The dyndep issue is none. The dyndep file is already only loaded if it is not dirty. The respective commit has been removed.

moritzx22:
### Changes in the recent push
- more mature implementation
  - restat functionality is corrected
  - dependencies are only checked once
  - runtime performance optimization
- unit tests have been changed to comply with the new depsfile loading
  - unit tests runs without failing in Linux
  - one Windows only test does still fail
- tests with builds like llvm reported the expected behavior

moritzx22:
### ninja -t missingdeps
console output
```
$ninja -t missingdeps
... There might be build flakiness if any of the targets listed above
are built alone, or not late enough, in a clean output directory.
```
This essentially means that issues can occur if a depsfile is not loaded because it has not yet been generated. With this merge request, the condition is extended: if the corresponding target is considered dirty due to the manifest, the depsfile will also not be loaded.

As a result, this PR may increase the likelihood of build flakiness when missingdeps are present.
_Note: well‑designed Ninja builds should not produce missing dependencies._

At the time of writing, I do not see any additional negative impact introduced by the conceptual change in this PR.

src/graph.h
```diff
@@ -294,7 +295,9 @@ struct ImplicitDepLoader {
   /// Load implicit dependencies for \a edge.
   /// @return false on error (without filling \a err if info is just missing
   //                          or out of date).
-  bool LoadDeps(Edge* edge, std::string* err);
+  bool LoadDeps(Edge* edge, std::string* err,
+                std::array<std::size_t, 2>* ptr = nullptr);
```
mathstuf: Is there a better name than `ptr` available?

src/graph.cc
```diff
@@ -743,13 +938,42 @@ bool ImplicitDepLoader::LoadDepsFromLog(Edge* edge, string* err) {
...
+
+bool ImplicitDepLoader::LoadDepsFromLogTry(const Edge* edge,
+                                           string* err) const {
+  // NOTE: deps are only supported for single-target edges.
+  Node* output = edge->outputs_[0];
+  DepsLog::Deps* deps = deps_log_ ? deps_log_->GetDeps(output) : NULL;
+  if (!deps) {
+    err->clear();
+    // explanations_.Record(output, "deps for '%s' are missing",
+    //                      output->path().c_str());
```
mathstuf: These look like useful messages.

mathstuf (commented):
Looks OK to me, but someone else should also review (I did it mainly from the test cases).

src/graph.cc
```diff
@@ -743,13 +938,42 @@ bool ImplicitDepLoader::LoadDepsFromLog(Edge* edge, string* err) {
...
+    return false;
+  }
+
+   // Deps are invalid if the output is newer than the deps.
+  if (output->mtime() > deps->mtime) {
+    // explanations_.Record(output,
+    //                      "stored deps info out of date for '%s' (%" PRId64
+    //                      " vs %" PRId64 ")",
+    //                      output->path().c_str(), deps->mtime, output->mtime());
```
moritzx22: > These look like useful messages.

similar to this [comment](https://github.com/ninja-build/ninja/pull/2680/files/2cef9583b7c7ab88a9799f77115bc6bb51843bd2#r2599824597)

src/graph.h
moritzx22: renamed

src/graph.cc
moritzx22: uncommented

src/graph.cc
moritzx22: uncommented

src/missing_deps.cc
```diff
@@ -39,16 +39,17 @@ struct NodeStoringImplicitDepLoader : public ImplicitDepLoader {
         dep_nodes_output_(dep_nodes_output) {}
 
  protected:
-  virtual bool ProcessDepfileDeps(Edge* edge,
-                                  std::vector<StringPiece>* depfile_ins,
-                                  std::string* err);
+  bool ProcessDepfileDeps(Edge* edge, std::vector<StringPiece>* depfile_ins,
+                          std::string* err,
+                          std::array<std::size_t, 2>* input_range) override;
```
moritzx22: Marking this as `override` is a candidate for separate PR.
Without this commit, the unit-tests do pass. Test-coverage could be improved to make the unittests fail if the function is not overriden.

moritzx22:
rebased to master

src/missing_deps.cc
mathstuf: Make the base class pure virtual (it can keep the implementation)?

moritzx22: > Make the base class pure virtual (it can keep the implementation)?

Class `ImplicitDepLoader` would become an abstract base class, but it is still required to instantiate objects of this class. That would not work anymore.

moritzx22:
rebased to master

moritzx22:
Rebased to master in previous push.

digit-google (changes requested):
This PR is impossible to review properly as a stack of 8 commits with what looks like random changes with unclear commit messages, please squash / rebase this into something that is simpler to review. For example:

- one commit to add the constness changes + the cp-deps test rule implementation.

- second commit to change the LoadDepXXX() signatures with proper documentation of all new parameters, preferably without changing the implementation yet.

- a third commit that changes the implementation to change the behavior / fix the bug and modify the tests accordingly.

Each commit should have a clear commit message explaining its purpose and why things are changed in a certain way. I'll add some inline comments too.

src/graph.cc
```diff
@@ -31,6 +31,217 @@
 
 using namespace std;
 
+namespace {
+
+/// execute hash only once in lifetime of object and only on request
+struct hashCommand {
```
digit-google: Please follow existing coding conventions, i.e. struct/class names should use PascalCase (hashCommand -> HashCommand), and member variables should use trailing underscore (valid -> valid_). Moreover, call this LazyEdgeCommandHash for clarity.

src/graph.cc
```diff
@@ -31,6 +31,217 @@
...
+
+ private:
+  const Edge* edge_;
+  std::uint64_t command_ = 0;
+  bool valid = false;
+};
+
+/// class is similar to a pointer of BuildLog::LogEntry
+/// additionally the LookupByOutput is cached for performance reasons
+class LogEntryCache {
```
digit-google: nit: Name this CachedLogEntry for clarity since this is not a cache.

src/graph.cc
```diff
@@ -31,6 +31,217 @@
...
+ public:
+  LogEntryCache(){};
+
+  operator bool() const { return entry_; }
```
digit-google: Explain what this corresponds to and when it is safe to call, since it never looks at evaluated_, the meaning of the result value is ambiguous. Consider replacing this with `is_valid()` for clarity, bool operators can lead to surprising bugs.

src/graph.cc
```diff
@@ -31,6 +31,217 @@
...
+  bool LookupByOutput(const BuildLog* buildLog, const Node* output);
+  const BuildLog::LogEntry* operator->() const { return entry_; }
+
+ private:
+  bool evaluated_ = false;
+  BuildLog::LogEntry* entry_ = nullptr;
+};
+
+bool LogEntryCache::LookupByOutput(const BuildLog* buildLog, const Node* output) {
```
digit-google: This interface is ambiguous, because the function could in theory be called with different |output| values and will only return a result corresponding to the first call. You could implement something similar without a dedicated LogEntryCache class with a simple std::map<const Node*, const BuildLog::LogEntry*> instead inside RecomputeOutputsDirty_, which would be simpler / clearer.

src/graph.cc
```diff
@@ -31,6 +31,217 @@
...
+  if (evaluated_)
+    return entry_;
+  evaluated_ = true;
+  return (entry_ = buildLog->LookupByOutput(output->path()));
+}
+
+/// performance optimized to recompute the outputs
+class RecomputeOutputsDirty_ {
```
digit-google: Please do not use trailing underscores in class names (or even inside them). Call this RecomputeOutputsDirtyCache instead, or something similar. Also consider moving changes related to performance optimizations to their own commit so they can be reviewed more easily.

src/graph.cc
```diff
@@ -31,6 +31,217 @@
...
+ public:
+  RecomputeOutputsDirty_(BuildLog* buildLog, OptionalExplanations& explanations,
```
digit-google: coding style: please use snake_case for variable / member identifiers (buildLog -> build_log)

src/graph.cc
```diff
@@ -31,6 +31,217 @@
...
+                         Edge* edge)
+      : buildLog_(buildLog), explanations_(explanations), edge_(edge),
+        LogEntry_(edge->outputs_.size()) {}
+  bool all(const Node* most_recent_input);
```
digit-google, line 108: nit: document what these methods do.

src/graph.cc
```diff
@@ -31,6 +31,217 @@
...
+  // on the fake node and get the latest mtime of the dependencies
+  if (most_recent_input) {
+    output->UpdatePhonyMtime(most_recent_input->mtime());
+  }
+
+  // Phony edges are clean, nothing to do
+  return false;
+}
+
+// disable warning for windows
```
digit-google: Explain why these are needed exactly.

src/graph.cc
```diff
@@ -31,6 +31,217 @@
...
+
+// IF_FIRSTRUN (condition)
+// expands to:
+//    assert(!(!FIRSTRUN && condition))
+//    if (FIRSTRUN && condition)
+#define IF_FIRSTRUN(cond)                   \
+  assert(FIRSTRUN || !(cond)); /* NOLINT */ \
+  if (FIRSTRUN && (cond))      /* NOLINT */
+
+template <bool FIRSTRUN>
```
digit-google, line 221: I strongly recommend to getting rid of the template parameter, and adding a simple `first_run` function parameter instead.

src/graph.h
```diff
@@ -295,7 +296,9 @@ struct ImplicitDepLoader {
   /// Load implicit dependencies for \a edge.
   /// @return false on error (without filling \a err if info is just missing
   //                          or out of date).
-  bool LoadDeps(Edge* edge, std::string* err);
+  bool LoadDeps(Edge* edge, std::string* err,
+                std::array<std::size_t, 2>* input_range = nullptr);
```
digit-google: Please update the documentation to explain the purpose of this new `input_range` parameter. Consider using a simple `struct InputRange { size_t start; size_t end; }` definition to make this easier to read and understand. Clarify that this is an optional output parameter.

src/graph.cc
```diff
@@ -79,6 +290,46 @@ bool DependencyScan::RecomputeDirty(Node* initial_node,
...
+
+  // Visit all inputs before checking if any of them is ready.
+  // Newly encountered edges may load dyndep files and gain
+  // outputs that correspond to some of our inputs.
+  for (Node* i : edge->inputs_) {
+    if (!RecomputeNodeDirty(i, stack, validation_nodes, err))
+      return false;
+  }
+
+  const auto input_end = edge->inputs_.cend() - input_range[1];
```
digit-google: Is this computation correct here? Can you clarify the meaning of input_range[1]? From the name "range" it can be assumed that this would be the position of the first item after the range, but in this case, you would use `input_end = edge->inputs_.cbegin() + input_range[1]` instead.

If the value is a count instead, "input_span" might be a better name, but the computation would be `input_end = edge->inputs_.cbegin() + input_range[0] + input_range[1]` so I am puzzled as to what this code does.

src/graph.cc
```diff
@@ -146,70 +395,70 @@ bool DependencyScan::RecomputeNodeDirty(Node* node, std::vector<Node*>* stack,
...
-  // outputs that correspond to some of our inputs.
-  for (Node* i : edge->inputs_) {
-    if (!RecomputeNodeDirty(i, stack, validation_nodes, err))
-      return false;
   }
 
   // Load output mtimes so we can compare them to the most recent input below.
-  for (Node* o : edge->outputs_) {
+  for (vector<Node*>::iterator o = edge->outputs_.begin();
+       o != edge->outputs_.end(); ++o) {
```
digit-google: nit: Why regress here when the original code was perfectly fine?

src/graph.cc
```diff
@@ -146,70 +395,70 @@ bool DependencyScan::RecomputeNodeDirty(Node* node, std::vector<Node*>* stack,
...
+  RecomputeOutputsDirty_ recomputeOutputsDirty(build_log(), explanations_,
+                                               edge);
+  if (!dirty)
+    dirty = recomputeOutputsDirty.all(most_recent_input);
+
+  // only try to load the deps log if no rebuild is necessary
+  // if an rebuild is necessary the deps log is outdated for this target
+  if (!edge->deps_loaded_ && !dirty) {
+    // This is our first encounter with this edge.  Load discovered deps.
+    std::array<std::size_t, 2> newLinks{ 0, 0 };
```
digit-google: nit: `newLinks` doesn't mean anything here. Use something more specific here. `new_deps` maybe?

digit-google: Or more precisely `new_deps_range`

src/graph.cc
```diff
@@ -734,10 +912,14 @@ bool ImplicitDepLoader::ProcessDepfileDeps(
     node->AddOutEdge(edge);
   }
 
+  if (input_range)
+    (*input_range)[1] = std::distance(implicit_dep, edge->inputs_.end());
```
digit-google: oh, so you are storing the distance from the last implicit input to the end of the array. This data structure definitely is not a range. Consider storing the distance from the start for clarity.

moritzx22:
> This PR is impossible to review properly as a stack of 8 commits with what looks like random changes with unclear commit messages, please squash / rebase this into something that is simpler to review. ...

True. I will reorder and clean up the commits so that each one has a clear purpose and is easier to understand. Only the final commit introduces a functional change. All earlier commits are refactoring or cleanup and keep current master behavior. After restructuring the history, this separation will be clearer and the review much simpler. Most of the other comments will be incorporated as well.

src/graph.cc
moritzx22: done

src/graph.cc
moritzx22: done

src/graph.cc
moritzx22: renamed to `is_valid()`

src/graph.cc
moritzx22: done

src/graph.cc
moritzx22: done

src/graph.cc
moritzx22: done

src/graph.cc
moritzx22: MSVC warns (and errors) on constructs like:
```
if (false) {
// do some stuff
}
```
This is suppressed. Anyhow this is obsolete with the change to c++17 and the use of `if constexpr`

src/graph.cc
moritzx22: corrected

src/graph.cc
moritzx22: renamed to `new_deps`

src/graph.cc
moritzx22: The template parameter is `constexpr`, which gives the compiler the best opportunity to optimize the code. Conceptually, the template function represents two distinct functions and helps avoid code duplication. The runtime `if` is replaced with `if constexpr` in the next push, so the unused branch is removed entirely at compile time.

There are essentially three design options:
 - A template function  
   - Uses a `constexpr template parameter` to generate two optimized code paths without duplication.
   - Cleanly separates the regular function parameters from the compile‑time selection parameter.
 - A regular function with a runtime parameter  
   - Simpler interface
 - Two separate functions  
    - Maximum clarity, but duplicates code.

Please restate your preference.
If you still recommend avoiding the template parameter, I’d appreciate some more detail on why the template approach is undesirable in this context.

Note: The `if` is changed to `if constexpr` in the macro.

src/graph.h
moritzx22: replaced `std::array` with:
```
struct InputView {
  std::size_t offset_begin = 0;
  std::size_t offset_end = 0;
};
```

moritzx22: The original `conditional` was removed to improve performance.
The Call that previously used that `conditional` now uses a dummy object.
This path(missingdeps) isn't performance‑critical, so using a dummy is acceptable.

src/graph.cc
moritzx22: This data structure represents a subset of a container.
The usual approach starting from index 0 has a drawback:
the default‑constructed view should represent the entire container, which requires knowing its size.

In this context it would look  like:
```cpp
if (!RecomputeEdgesInputsDirty(node, InputView(), most_recent_input, dirty,
                                 stack, validation_nodes, err))
// would need to become
if (!RecomputeEdgesInputsDirty(node, InputView{0, node->in_edge()->inputs_.size()}, most_recent_input, dirty,
                                 stack, validation_nodes, err))
```
[graph.cc#L482](https://github.com/moritzx22/ninja/blob/0070603c01cb1febf33a24108e80d128882e39e2/src/graph.cc#L482)

I can implement the change, and it can certainly be expressed more cleanly than in the example above. However, in this particular context the change makes the code a bit more complicated. Before proceeding, I’d appreciate your advice.

src/graph.cc
moritzx22: Here are the reasons behind introducing this new class:

Requirements
 - Cache the result of LookupByOutput.
 - Store only a single pointer and a single boolean per cache entry.
 - Keep the cache in contiguous memory (std::vector) for compiler‑friendly access patterns.
 - Allocate the memory only once (vector sized in the constructor).
 - Allow lookup of the cached value for a given output at effectively zero cost.

`RecomputeOutputsDirty` is performance‑critical.
Its worst‑case scenario is a clean build, where no early exits occur and every output must be visited. `RecomputeOutputsDirty` will be invoked twice in this situation because the depfile is loaded here. This is exactly where the cache provides the most benefit.

To achieve this performance, the class relies on strict usage assumptions:
`LookupByOutput` must always be called with the same parameters for a given instance. Debug assertions enforce this, and the assumptions are documented in the code. These constraints allow the implementation to remain efficient.

`RecomputeOutputsDirtyCache` ensures these assumptions hold. It selects the correct output and its associated cache entry for processing. The `CachedLogEntry` type is defined in the private section to prevent accidental misuse outside the intended context.

A `std::map` could also be used to implement the cache, and it would likely be simpler to write, but I expect its performance to be worse, especially for clean builds.

Please advise.

src/graph.cc
moritzx22: example usage for clarity
```cpp
CachedLogEntry cached;
if(cached.is_valid()) cached->foo(); // nullptr
cached.LookupByOutput(build_log, output);  // assign value
if(cached.is_valid()) cached->foo(); // call foo()
cached.LookupByOutput(build_log, output);  // already cached
if(cached.is_valid()) cached->foo(); // call foo()

// a raw pointer instead
BuildLog::LogEntry* entry = nullptr;
if(entry) entry->foo(); // nullptr
entry = build_log.LookupByOutput(output);
if(entry.is_valid()) entry->foo(); // call foo()
entry = build_log.LookupByOutput(output); // second call
if(entry.is_valid()) entry->foo(); // call foo()

```

src/graph.cc
digit-google: Second, there is 

> This data structure represents a subset of a container. The usual approach starting from index 0 has a drawback: the default‑constructed view should represent the entire container, which requires knowing its size.
> 
Technically, this is neither a "view" nor a "subset" as these terms usually refer to objects that can be used directly to access individual items. This is not the case here: you just have a pair of numbers, whose interpretation requires additional information (in this case the exact and unmodified `inputs_` array they refer to). Hence using something like "range" in the name makes more sense. Another option is to store the  `Edge` pointer in the data structure (or at least a pointer to its edge->inputs_ array).

If you prefer a non-conventional layout / interpretation for the values, I strongly recommend making a custom class with human-friendly accessors to properly document its purpose and simplify its usage. E.g.

```
struct EdgeInputsRange {
  /// Create new instance covering all |edge| inputs.
  EdgeInputsRange(const Edge* edge);
  
  /// Create instance covering the [start_pos..end_pos) interval of |edge| inputs.
  EdgeInputsRange(const Edge* edge, size_t start_pos, size_t end_pos);

  size_t start_pos() const;
  size_t end_pos() const;

private:
  ...
};
```

> In this context it would look like:
> 
> ```c++
> if (!RecomputeEdgesInputsDirty(node, InputView(), most_recent_input, dirty,
>                                  stack, validation_nodes, err))
> // would need to become
> if (!RecomputeEdgesInputsDirty(node, InputView{0, node->in_edge()->inputs_.size()}, most_recent_input, dirty,
>                                  stack, validation_nodes, err))
> ```
> 
> [graph.cc#L482](https://github.com/moritzx22/ninja/blob/0070603c01cb1febf33a24108e80d128882e39e2/src/graph.cc#L482)
> 
> I can implement the change, and it can certainly be expressed more cleanly than in the example above. However, in this particular context the change makes the code a bit more complicated. Before proceeding, I’d appreciate your advice.

src/graph.cc
digit-google: This should go as a comment inside the source code, so that future maintainers now how / when to keep this.

moritzx22: sorry for not being clear. In the latest push it looks like

```cpp
if constexpr (false) {
// do some stuff
}
```
and no warning or error is reported by MSVC anymore. The disable warning stuff has been removed.

src/graph.cc
moritzx22: changed to 
```cpp
struct EdgeInputsRange {
  /// Create new instance covering all |edge| inputs.
  EdgeInputsRange(Edge* edge)
      : beg_(edge->inputs_.begin()), end_(edge->inputs_.end()),
        edge_(edge) {}

  using iterator = std::vector<Node*>::const_iterator;

  iterator begin() const { return beg_; }
  iterator end() const { return end_; }

  iterator beg_;
  iterator end_;

  /// The edge whose input range is being viewed.
  Edge* const edge_;
};
```
in recent push. See https://github.com/ninja-build/ninja/pull/2680#issuecomment-4209133391

moritzx22:
next push, see https://github.com/ninja-build/ninja/pull/2680#discussion_r3053771337

moritzx22:
### changes next push
- minor improvments
- ImplicitDepLoader::LoadDepFileTry()
  - use Stat instead of Readfile
- ImplicitDepLoader::LoadDepsFromLogTry()
  - call StatIfNecessary()

src/graph.cc
```diff
@@ -269,7 +295,8 @@ bool DependencyScan::VerifyDAG(Node* node, vector<Node*>* stack, string* err) {
 }
 
 bool DependencyScan::RecomputeOutputsDirty(Edge* edge, Node* most_recent_input,
-                                           bool* outputs_dirty, string* err) {
+                                           bool* const outputs_dirty,
```
digit-google, line 586: nit: drop the `const` here as it does nothing.

src/graph.cc
```diff
@@ -715,14 +744,16 @@ bool ImplicitDepLoader::LoadDepFile(Edge* edge, const string& path,
...
 
 bool ImplicitDepLoader::ProcessDepfileDeps(
-    Edge* edge, std::vector<StringPiece>* depfile_ins, std::string* err) {
+    EdgeInputsRange* input_range, std::vector<StringPiece>* depfile_ins,
+    std::string* err) {
   // Preallocate space in edge->inputs_ to be filled in below.
   vector<Node*>::iterator implicit_dep =
-      PreallocateSpace(edge, static_cast<int>(depfile_ins->size()));
+      PreallocateSpace(input_range->edge_, static_cast<int>(depfile_ins->size()));
+  input_range->beg_ = implicit_dep;
```
digit-google: Modifying members of this instance directly breaks encapstulation and is bad style. Use a constructor with explicit arguments for begin/end instead to make this clearer.

The input_range argument should appear after input arguments in the function signature, and the documentation should clarify that it is written to on success.

src/graph.h
```diff
@@ -295,7 +296,7 @@ struct ImplicitDepLoader {
   /// Load implicit dependencies for \a edge.
```
digit-google, line 297: Update the document to clarify the role of |input_range| here. It would be clearer if you keep the Edge* parameter as input, and make|input_range| and output-only parameter after it instead (this avoids dealing with partially-constructed objects which make the code harder to understand).

src/graph.h
```diff
@@ -304,17 +305,18 @@ struct ImplicitDepLoader {
  protected:
   /// Process loaded implicit dependencies for \a edge and update the graph
```
digit-google, line 310: Again clarify the rold of |input_range|. In this case it seems to be an output-only so should better appear after |depfile_ins| for clarity.

src/graph.h
```diff
@@ -304,17 +305,18 @@ struct ImplicitDepLoader {
...
   /// @return false on error (without filling \a err if info is just missing)
-  virtual bool ProcessDepfileDeps(Edge* edge,
+  virtual bool ProcessDepfileDeps(EdgeInputsRange* input_range,
                                   std::vector<StringPiece>* depfile_ins,
                                   std::string* err);
 
   /// Load implicit dependencies for \a edge from a depfile attribute.
```
digit-google, line 316: again, update documentation accordingly. Keep Edge* input and EdgeInputsRange* outputs separate for clarity if necessary.

src/graph.h
```diff
@@ -304,17 +305,18 @@ struct ImplicitDepLoader {
...
   /// @return false on error (without filling \a err if info is just missing).
-  bool LoadDepFile(Edge* edge, const std::string& path, std::string* err);
+  bool LoadDepFile(EdgeInputsRange* input_range, const std::string& path,
+                   std::string* err);
 
-  /// Load implicit dependencies for \a edge from the DepsLog.
+  /// Load implicit dependencies for \a input_range from the DepsLog.
   /// @return false on error (without filling \a err if info is just missing).
-  bool LoadDepsFromLog(Edge* edge, std::string* err);
+  bool LoadDepsFromLog(EdgeInputsRange* input_range, std::string* err);
```
digit-google: should this be a `const EdgeInputsRange*` or does the value mutates due to this function. Please clarify.

src/graph.h
```diff
@@ -324,9 +326,30 @@ struct ImplicitDepLoader {
...
 
+/// A lightweight range of an Edge's input nodes.
+///
+/// A default-constructed `EdgeInputsRange` spans the entire
+/// `edge->inputs_` vector.
+struct EdgeInputsRange {
+  /// Create new instance covering all |edge| inputs.
+  EdgeInputsRange(Edge* edge)
+      : beg_(edge->inputs_.begin()), end_(edge->inputs_.end()),
+        edge_(edge) {}
```
digit-google: Add an EdgeInputsRange(iterator, iterator) constructor and make beg_ and end_ private members.

moritzx22:
### Previous push
rebased to master

### Next push
incorporated comments, see [diff](https://github.com/ninja-build/ninja/compare/9bfeee77518052b4f94f03cd690228ddfee94651..2d73512a1b7c7b319cfd2fa706b9387b69a7544a)

src/graph.cc
moritzx22: done

src/graph.h
moritzx22:
```cpp
  /// Load implicit dependencies for \a edge.
  /// @return emtpy optional on error (without filling \a err if info is just
  ///                                  missing or out of date).
  /// @return newly added `EdgeInputsRange` on success, describing the
  ///         dependencies that were added to the graph.
  std::optional<EdgeInputsRange> LoadDeps(Edge* edge, std::string* err);
  //                        bool LoadDeps(Edge* edge, std::string* err);
```
new signature is more clear. Outputs are now separated from inputs.

src/graph.cc
moritzx22: Respective change has been reverted, comment should be obsolete. see [comment](https://github.com/ninja-build/ninja/pull/2680#discussion_r3089414978).

src/graph.h
moritzx22: see [comment](https://github.com/ninja-build/ninja/pull/2680#discussion_r3089414978)

src/graph.h
moritzx22: see [comment](https://github.com/ninja-build/ninja/pull/2680#discussion_r3089414978)

src/graph.h
moritzx22: Parameter shifted to return value, see [comment](https://github.com/ninja-build/ninja/pull/2680#discussion_r3089414978).

src/graph.h
moritzx22: Done.

src/graph.cc
digit-google: I am still seeing it?

digit-google (approved):
This is good, the code makes so much more sense now. Thanks for your patience!
@jhasse please take a look.

src/graph.cc
moritzx22: You're right, I overlooked that. I'll correct it. See https://github.com/ninja-build/ninja/pull/2680#issuecomment-4273506744

moritzx22:
Great to hear the code is clearer. Thanks for the reasonable comments!

moritzx22:
next push is minor, see https://github.com/ninja-build/ninja/pull/2680#discussion_r3087518316

src/graph.cc
```diff
@@ -32,6 +32,211 @@
...
+  bool Phony(Node* output, const Node* most_recent_input) const;
+
+  const BuildLog* const buildLog_;
+  OptionalExplanations& explanations_;
+  const Edge* const edge_;
+
+  const bool isRestat_ = edge_->GetBindingBool("restat");
+  bool generator_ = false;
+  bool generatorValid_ = false;
+  std::vector<CachedLogEntry> logEntry_;
```
moritzx22, line 115: One additional memory alignment optimization is possible: the size of each `CachedLogEntry` element in `logEntry_` can be reduced from 16 to 8 bytes by replacing the bool `evaluated_` field with a sentinel pointer. This halves the vector's memory footprint and improves cache efficiency for edges with many outputs.

Whether this is worth merging is debatable, most edges have only 1–3 outputs, so the benefit might be limited in practice.

See the [branch](https://github.com/moritzx22/ninja/tree/fix2666_alignment) and the [diff](https://github.com/moritzx22/ninja/compare/fix2666..fix2666_alignment).

moritzx22:
### Next Push
resolve conflicts, rebase to master.

PR #2684, opened 2025-10-21
Fix unit test crash, if test fails

moritzx22:
Fix unit test crash, if test fails.

Note: This is coming from PR [2680](https://github.com/ninja-build/ninja/pull/2680), but the issue is not strictly related to PR2680

### DIFF
--- a/src/build_test.cc
+++ b/src/build_test.cc
@@ -534,6 +534,17 @@ struct FakeCommandRunner : public CommandRunner {
   VirtualFileSystem* fs_;
 };
 
+class SafeRelease {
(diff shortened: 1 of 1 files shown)


mathstuf (approved):

jhasse:
Can you change the commit message to "Fix unit test crash, if test fails"? It isn't really about a memory leak if I understand it correctly.

moritzx22:
> Can you change the commit message to "Fix unit test crash, if test fails"?

Done

- Minor bug fix for windows, the CI should now succeed.

Issue #2682, opened 2025-10-17
Use estimated build time of subtree as scheduler edge weight heuristic


PowellNGL:
I've been testing this small change to the scheduler edge weight heuristic, and I've measured significant improvements in wall clock time in several real-world projects

The change is here: https://github.com/PowellNGL/ninja/commit/67e093c8afb9465772d6d1e764f28a930b536975

Currently, the scheduler only uses the depth of a subtree as the heuristic, this change instead uses the **total estimated build time of the subtree** as measured in previous runs. I've found several past discussions about using the estimated build time of the edge as a heuristic (e.g. https://github.com/ninja-build/ninja/pull/2177, https://github.com/ninja-build/ninja/pull/2019, https://github.com/ninja-build/ninja/issues/376) but I haven't seen anything about summing the entire subtree

### Test Results (various open source C++ projects)

Each measurement is the average of 3 builds on an AMD Ryzen 7 7800X3D with -j 12. 'old' means the latest version of ninja from master, 'new' means ninja built from my fork

|   Repo   |    Build Type    | Rebuild Type | Avg time (old) | Avg time (new) |  Change |
|----------|------------------|--------------|----------------|----------------|---------|
|  boost   |  Debug (ccache)  |    clean     |     0.56s      |     0.44s      | -21.01% |
|  boost   | Release (ccache) |    clean     |     0.50s      |     0.39s      | -20.79% |
| protobuf |     Release      | incremental  |     80.51s     |     71.73s     | -10.91% |
|  ninja   |  Debug (ccache)  |    clean     |     0.45s      |     0.42s      |  -5.22% |
|  ninja   |     Release      |    clean     |     9.71s      |     9.21s      |  -5.12% |
|  boost   |     Release      | incremental  |     9.28s      |     8.86s      |  -4.46% |
| protobuf |     Release      |    clean     |    140.13s     |    134.26s     |  -4.18% |
| protobuf |      Debug       | incremental  |     64.24s     |     61.96s     |  -3.56% |
|  boost   |      Debug       | incremental  |     8.10s      |     7.84s      |  -3.17% |
| protobuf |      Debug       |    clean     |    118.72s     |    115.64s     |  -2.59% |
|  boost   |     Release      |    clean     |     31.13s     |     30.34s     |  -2.56% |
|  boost   |      Debug       |    clean     |     26.38s     |     25.90s     |  -1.81% |
[...]

PR #2686, opened 2025-10-25
Include estimated build time of subtree in the edge weight heuristic

PowellNGL:
As described in https://github.com/ninja-build/ninja/issues/2682, this change uses the estimated build time of a subtree (measured during the previous run) as the scheduler heuristic instead of just its depth. This can make rebuilds faster by starting the critical path as early as possible


src/build.cc
```diff
@@ -468,7 +468,9 @@ namespace {
 // Heuristic for edge priority weighting.
 // Phony edges are free (0 cost), all other edges are weighted equally.
```
moritzx22: > // ... all other edges are weighted equally.

This comment is outdated and could be updated.

PR #2695, opened 2025-11-27
Speed up methods in `State`

elliotgoodrich:
Remove the double-lookup inside `paths_` in `State::GetNode` in the case where we are first calling this for a particular `path`.

Add overloads of `GetNode` that can take ownership of a `std::string`, which is the common case when calling from `ManifestParser` and we have a `std::string` returned by the `Evaluate` method going to waste.

To avoid writing an overload of `AddIn`, `AddOut`, and `AddValidation` taking a `std::string&&` by changing the signature to instead take a `Node *` and require caller to get this through `State::GetNode`.

### DIFF
--- a/src/graph.h
+++ b/src/graph.h
@@ -19,6 +19,7 @@
 #include <queue>
 #include <set>
 #include <string>
+#include <utility>
 #include <vector>
 
 #include "dyndep.h"
@@ -42,6 +43,8 @@ struct State;
 struct Node {
   Node(const std::string& path, uint64_t slash_bits)
       : path_(path), slash_bits_(slash_bits) {}
+  Node(std::string&& path, uint64_t slash_bits)
+      : path_(std::move(path)), slash_bits_(slash_bits) {}
 
   /// Return false on error.
   bool Stat(DiskInterface* disk_interface, std::string* err);
--- a/src/manifest_parser.cc
+++ b/src/manifest_parser.cc
@@ -339,7 +339,8 @@ bool ManifestParser::ParseEdge(string* err) {
       return lexer_.Error("empty path", err);
     uint64_t slash_bits;
     CanonicalizePath(&path, &slash_bits);
-    if (!state_->AddOut(edge, path, slash_bits, err)) {
+    Node *node = state_->GetNode(std::move(path), slash_bits);
+    if (!state_->AddOut(edge, node, err)) {
       lexer_.Error(std::string(*err), err);
       return false;
     }
@@ -361,7 +362,8 @@ bool ManifestParser::ParseEdge(string* err) {
       return lexer_.Error("empty path", err);
     uint64_t slash_bits;
     CanonicalizePath(&path, &slash_bits);
-    state_->AddIn(edge, path, slash_bits);
+    Node *node = state_->GetNode(std::move(path), slash_bits);
+    state_->AddIn(edge, node);
   }
   edge->implicit_deps_ = implicit;
   edge->order_only_deps_ = order_only;
@@ -374,7 +376,8 @@ bool ManifestParser::ParseEdge(string* err) {
       return lexer_.Error("empty path", err);
     uint64_t slash_bits;
     CanonicalizePath(&path, &slash_bits);
-    state_->AddValidation(edge, path, slash_bits);
+    Node *node = state_->GetNode(std::move(path), slash_bits);
+    state_->AddValidation(edge, node);
   }
 
   if (options_.phony_cycle_action_ == kPhonyCycleActionWarn &&
--- a/src/state.cc
+++ b/src/state.cc
@@ -92,12 +92,39 @@ Edge* State::AddEdge(const Rule* rule) {
   return edge;
 }
 
+Node* State::GetNode(const char* path, uint64_t slash_bits) {
+  return GetNode(StringPiece(path), slash_bits);
+}
+
 Node* State::GetNode(StringPiece path, uint64_t slash_bits) {
-  Node* node = LookupNode(path);
-  if (node)
-    return node;
-  node = new Node(path.AsString(), slash_bits);
-  paths_[node->path()] = node;
+  const std::pair<Paths::iterator, bool> result = paths_.try_emplace(path, nullptr);
+  Node*& node = result.first->second;
+  const bool isNew = result.second;
+  if (isNew) {
+    assert(node == nullptr);
+    // If `try_emplace` added a new node then we need to actually allocate a `Node`
+    // element and replace the existing value in the map.
+    node = new Node(path.AsString(), slash_bits);
+
+    // Additionally, as the key is a StringPiece, it needs to refer to a string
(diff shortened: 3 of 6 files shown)


src/manifest_parser.cc
```diff
@@ -198,7 +198,7 @@ bool ManifestParser::ParseDefault(string* err) {
     uint64_t slash_bits;  // Unused because this only does lookup.
     CanonicalizePath(&path, &slash_bits);
     std::string default_err;
-    if (!state_->AddDefault(path, &default_err))
+    if (!state_->AddDefault(std::move(path), &default_err))
```
digit-google: nit: AddDefault() takes a StringPiece as its first argument, so this std::move() doesn't do anything. No need to change its signature given that this is so rarely called in practice.

src/state.cc
```diff
@@ -92,12 +92,33 @@ Edge* State::AddEdge(const Rule* rule) {
...
+  return node;
+}
+
+Node* State::GetNode(std::string&& path, uint64_t slash_bits) {
+  const std::pair<Paths::iterator, bool> result = paths_.try_emplace(path, nullptr);
+  Node*& node = result.first->second;
+  const bool isNew = result.second;
+  if (isNew) {
+    assert(node == nullptr);
+    node = new Node(std::move(path), slash_bits);
```
digit-google, line 125: The Node() constructor takes a `const std::string&` as its first argument, so this will still perform a copy :-/ But it should be easy to fix by changing the constructor definition in `graph.h` with something that looks like:

```
Mode(std::string path, ...) : path_(std::move(path)), .... {}
```

This would also benefit the `Node(path.AsString(), ...)` call on line 105 above.

src/state.cc
```diff
@@ -92,12 +92,33 @@ Edge* State::AddEdge(const Rule* rule) {
...
-    return node;
-  node = new Node(path.AsString(), slash_bits);
-  paths_[node->path()] = node;
+  const std::pair<Paths::iterator, bool> result = paths_.try_emplace(path, nullptr);
+  Node*& node = result.first->second;
+  const bool isNew = result.second;
+  if (isNew) {
+    assert(node == nullptr);
+    node = new Node(path.AsString(), slash_bits);
+    // Update the key to refer to the long-lived string in the Node
```
digit-google: Oh this is smart but also a bit subtle. Can I ask you to add a comment explaining here that this cannot break the hash map's internals because the key type's equality and hash operations only care about the content of the StringPiece-pointed data, not the address it lives at. I think that would help for long-term maintenance.

src/manifest_parser.cc
elliotgoodrich: Yes true. I will revert this move.

There was no benefit to changing `AddDefault` since it only calls `LookupNode` and not `GetNode`, so there is only ever 1 lookup in the hash table.

src/state.cc
elliotgoodrich: I swear I did this - I just have no idea where it went :D

digit-google: Thanks, an additionnal Nost(std::string&&) constructor works too.

digit-google:
@jhasse please take a look

moritzx22:
> To avoid writing an overload of `AddIn`, `AddOut`, and `AddValidation` taking a `std::string&&` by changing the signature to instead take a `Node *` and require caller to get this through `State::GetNode`.

Changing the signatures of `AddIn`, `AddOut`, and `AddValidation` does not preserve the original semantics of these functions. Moving this responsibility to the caller would alter behavior and require widespread changes throughout the codebase, including updates to the unit tests. The scope of this PR should remain a pure speed optimization, without introducing semantic changes.

A simpler and semantically correct alternative is to keep the existing behavior and update the signatures to accept rvalue strings:
```cpp
void AddIn(Edge* edge, std::string&& path, uint64_t slash_bits);
bool AddOut(Edge* edge, std::string&& path, uint64_t slash_bits, std::string* err);
void AddValidation(Edge* edge, std::string&& path, uint64_t slash_bits);
```
This preserves the current beaviour of the functions, avoids unnecessary copies, and does not require overloads for lvalue semantics (which are not needed here). The impact on existing call sites is minimal, and the unit tests can remain unchanged. I also do not see any performance penalty compared to the currently proposed solution.

### Conflicts with PR #2713
As discussed in https://github.com/ninja-build/ninja/pull/2713#discussion_r2725630059, PR 2713 changes the semantics of `GetNode` from a two‑step creation process to a single‑step process. After that change, the following comment is no longer guaranteed to be true:
https://github.com/ninja-build/ninja/blob/cc60300ab94dae9bb28fece3c9b7c397235b17de/src/state.h#L110-L113
Because `GetNode` would no longer guarantee that a node originates from the manifest, shifting responsibility to the caller would break the assumptions that `AddIn`, `AddOut`, and `AddValidation` currently rely on.

src/graph.h
```diff
@@ -19,6 +19,7 @@
 #include <queue>
 #include <set>
 #include <string>
+#include <utility>
```
moritzx22, line 22: It’s not clear what requires this include.

elliotgoodrich:
> Changing the signatures of `AddIn`, `AddOut`, and `AddValidation` does not preserve the original semantics of these functions. Moving this responsibility to the caller would alter behavior and require widespread changes throughout the codebase, including updates to the unit tests. The scope of this PR should remain a pure speed optimization, without introducing semantic changes.

All necessary changes are in this PR and there are 10 call sites that need to be updated - including all those in unit tests. I don't think that is particularly widespread.

> ### Conflicts with PR #2713

My PR has been open and approved for 2 months now.  If another change is landed before this one then I will update it no problem, but before that point it doesn't make sense for me to guess what PRs are going to land and to make pre-emptive changes to avoid conflicts.

elliotgoodrich:
@jhasse @digit-google is this okay to land?

jhasse:
I don't have time to look at this.

elliotgoodrich:
Okay thank you - is this change then blocked until further notice?

src/state.cc
```diff
@@ -92,12 +92,39 @@ Edge* State::AddEdge(const Rule* rule) {
...
+
 Node* State::GetNode(StringPiece path, uint64_t slash_bits) {
-  Node* node = LookupNode(path);
-  if (node)
-    return node;
-  node = new Node(path.AsString(), slash_bits);
-  paths_[node->path()] = node;
+  const std::pair<Paths::iterator, bool> result = paths_.try_emplace(path, nullptr);
+  Node*& node = result.first->second;
+  const bool isNew = result.second;
```
digit-google, line 102: nit: prefer snake_case for variable names in the Ninja source code, e.g. `is_new`

src/graph.h
digit-google: Technically, any use of `std::move()` requires it. In practice, most other C++ standard library headers will include it anyway due to their template-based implementation, one could consider this an implementation detail. This looks fine to me.

PR #2696, opened 2025-11-27
Allow `EvalString` to retain capacity

elliotgoodrich:
Change the implementation of `EvalString` from
`std::vector<std::string>` (or `std::string` for 1-token strings) to just a `std::string`, where tokens are stored contiguously, prefixed by their size.  Introduce an `EvalStringBuilder` to create `EvalString` elements, which allows us to reduce the size of `EvalString` further because appending tokens requires an additional `std::size_t` for bookkeeping.  This has the added benefit that `EvalString` is now immutable, which is usually easier to reason about.

This implementation has a few performance improvements:
  * Smaller `sizeof` (roughly 50% of the previous `EvalString`)
  * Better cache locality of tokens
  * Retains all capacity on `Clear()`

Since all capacity can be retained in `Clear()`, it is possible to reuse `EvalString` (or `EvalStringBuilder`) objects and remove the need for many allocations.  In this commit we introduce `value_` to `ManifestParser`, which is reused in parsing variables, pools, rules, and default statements.  A future commit will create a class similar to `std::vector<EvalStringBuilder>` that will allow reusing `EvalStringBuilder` elements and this will avoid most allocations when constructing `EvalString`s.

The biggest change is that `EvalString` no longer has a `std::string` for the token, but instead returns a `StringPiece`.  This has a problem with `Env::LookupVariable` and several other functions that took a `std::string` to lookup values.  To avoid creating temporary `std::string` values, we change the comparator of `std::map` to allow looking up using both `StringPiece` and `std::string` and update the function signatures.

Now that `EvalString` has a public way to iterate through tokens, the `EvalString::Evaluate` method no longer needs to be on `EvalString`, and has been moved to a static method on `Env`. This has been templated on `Env` because there's no need for polymorphism and it gives the compiler more opportunity to devirtualize calls to `LookupVariable`.  The implementation has a shortcut where we can avoid switching on the type when we have a text token since it is either the last one or followed by a special token.

The implementation of `EvalString` has been adapted from the one here https://github.com/elliotgoodrich/trimja/blob/e6fb3d188ea018806e28be5b2d4a5d2a82d427a1/src/evalstring.h


src/manifest_parser.cc
```diff
@@ -215,11 +215,11 @@ bool ManifestParser::ParseEdge(string* err) {
   validations_.clear();
 
   {
-    EvalString out;
+    EvalStringBuilder out;
     if (!lexer_.ReadPath(&out, err))
       return false;
     while (!out.empty()) {
-      outs_.push_back(std::move(out));
+      outs_.emplace_back(std::move(out).str());
```
moritzx22: Using emplace_back doesn’t provide more benefit than push_back.
`std::move(out).str()` is already an `EvalString&&`, and outs_ stores `EvalString`.
So we’re inserting a fully constructed rvalue, not building it in place.

src/eval_env.cc
```diff
@@ -52,11 +52,11 @@ const Rule* BindingEnv::LookupRule(const string& rule_name) {
   return NULL;
 }
 
-void Rule::AddBinding(const string& key, const EvalString& val) {
+void Rule::AddBinding(const std::string& key, const EvalString& val) {
   bindings_[key] = val;
```
moritzx22, line 56: **nit**

This requires `EvalString::EvalString()`.
The best option would be `std::map::insert_or_assign`, but that’s only available in C++17.
In C++14 you can implement the same behavior manually using `find` and `emplace`.
`EvalString::EvalString()` could be made private so that only `EvalStringBuild` is allowed to construct an `EvalString`.

src/evalstring.h
moritzx22, line 1: Do you think some additional unittests for this file (and may be the evaluate function) would be appropriate?

elliotgoodrich: Can never have enough unit tests - I've added `evalstring_test.cc` to test `EvalString` and `EvalStringBuilder`.  I haven't tested `evaluate` though, there is already a decent amount of code testing evaluate based on the unit test failures I had when making this change.

src/manifest_parser.cc
elliotgoodrich: You're right - in this case it's going to be identical.  I made this change as I initially made `EvalStringBuilder` convertible to `EvalString` and avoided the `str()` method.  However, this resulted in some ambiguous function calls and I had to go back to an explicit method.

However, is there any reason to change it back if it's identical?  I generally think preferring `emplace_back` over `push_back` is a good way to work since it avoids potentially expensive temporaries.

src/eval_env.cc
elliotgoodrich: I agree that default ctor gives us the fastest way to do this pre-C++17.  Since `EvalString` is meant to be a value-semantic type, you can make the argument it should have a default constructor (from Elements of Programming), but that view is [debated](https://quuxplusone.github.io/blog/2018/05/10/regular-should-not-imply-default-constructible/).

src/manifest_parser.cc
elliotgoodrich: Also, I'm not a big fan of `std::move(builder).str()`, even though it may be the most obvious way of what's happening - IDK, perhaps a `builder.release()` or something would also work.

I would be happy to take suggestions on a prettier API!

moritzx22: > However, is there any reason to change it back if it's identical?

- changes to the code base should have a good reason
- `push_back` is more clear than `emplace_back` here

> I generally think preferring `emplace_back` over `push_back` is a good way to work since it avoids potentially expensive temporaries.

Antithesis: if it does yes, if not `push_back` is more clear(Opinionated).

moritzx22:
https://github.com/ninja-build/ninja/blob/c74beb43695ff8cae955f05a6cbe1753aa9517f8/src/ninja.cc#L1832-L1834

use a temporary to limit lifetime, and free memory early:
```cpp
if (!ManifestParser(&ninja.state_, &ninja.disk_interface_, parser_opts)
             .Load(options.input_file, &err)) {
```
if you think it is not related to your PR, just ignore...

src/manifest_parser.cc
moritzx22: > I would be happy to take suggestions on a prettier API!
### c++17
```cpp
// minimum c++17
class EvalStringBuilder {
...
operator EvalString() &&;
};

outs_.emplace_back(std::move(out));
// outs_.push_back(std::move(out)); // do not work
```
### c++14
I do not know better, but I would prefer the `builder.release()`. Current solution is fine as well.

src/evalstring_test.cc
```diff
@@ -0,0 +1,110 @@
...
+// limitations under the License.
+
+#include "evalstring.h"
+
+#include "test.h"
+
+using Token = EvalString::TokenType;
+
+namespace {
+
```
moritzx22, line 22: ### More pretty with literals
add here:
```cpp
std::pair<StringPiece, Token> operator"" _R(const char* str, std::size_t len) {
  return std::pair<StringPiece, Token>{ StringPiece(str, len), Token::RAW };
}

std::pair<StringPiece, Token> operator"" _S(const char* str, std::size_t len) {
  return std::pair<StringPiece, Token>{ StringPiece(str, len), Token::SPECIAL };
}
```
 allows to use
```cpp
std::vector<std::pair<StringPiece, Token>> data = { "a"_R, "b"_R, "c"_S };
 // std::vector<std::pair<StringPiece, Token>> data2 = { { "a", Token::RAW }, { "b", Token::RAW }, { "a", Token::SPECIAL }
```

src/manifest_parser.h
```diff
@@ -72,6 +71,8 @@ struct ManifestParser : public Parser {
   // subparser_ is reused solely to get better reuse out ins_/outs_/validation_.
   std::unique_ptr<ManifestParser> subparser_;
   std::vector<EvalString> ins_, outs_, validations_;
+  // value_ is reused while parsing variables.
+  EvalStringBuilder value_;
```
moritzx22, line 75: This variable is an replacement for a local scope variable. To prevent a misuse, and make it safe, I propose the following.
```cpp
// just calls clear() if an local object is "generated"
class EvalStringFactory{
public:
  // returns data_ and calls clear()
  EvalStringBuilder& make(){
    data_.clear(); 
    return data_;
  }
private:
EvalStringBuilder data_;
};
```

usage
```cpp
class X{ 
EvalStringFactory eval_;
};

void X::function()
{
  auto& myEvalStringBuilder = eval_.make();
  // myEvalStringBuilder is ready to use;
  // do sth.
}
```
Note: as a bonus, a check for debug builds could be inserted. This checks that only one generated variable is alive at the same time. That would require an action in the destructor of the generated object, and therefore an additional wrapper type.

src/manifest_parser.cc
elliotgoodrich: I think `emplace_back` should be idiomatic at this point. But I agree that it's probably confusing at minimum to make this change as part of this PR since it has no effect. I've reverted the `push_back` -> `emplace_back` changes :)

elliotgoodrich:
> https://github.com/ninja-build/ninja/blob/c74beb43695ff8cae955f05a6cbe1753aa9517f8/src/ninja.cc#L1832-L1834
> 
> use a temporary to limit lifetime, and free memory early:
> 
> ```c++
> if (!ManifestParser(&ninja.state_, &ninja.disk_interface_, parser_opts)
>              .Load(options.input_file, &err)) {
> ```
> 
> if you think it is not related to your PR, just ignore...

In any other function I would agree, but `real_main` doesn't return and instead always calls `exit` because ninja wants to exit ASAP and this allows us to skip calling destructors.

src/manifest_parser.h
elliotgoodrich: I think this is a good change and could also be applied to `ins_`, `outs_`, and `validations_` that are similarly used.  I would probably prefer to not make this change in this PR and write a separate generic components that could be used here and in other source files and make the change afterwards in one go.

src/string_piece_util.cc
```diff
@@ -76,3 +77,11 @@ bool EqualsCaseInsensitiveASCII(StringPiece a, StringPiece b) {
 
   return true;
 }
+
+bool StringPieceLess::operator()(const StringPiece& lhs,
+                                 const StringPiece& rhs) const {
+  return std::lexicographical_compare(lhs.begin(),
+                                      lhs.end(),
+                                      rhs.begin(),
+                                      rhs.end());
```
digit-google, line 86: What is the purpose of this change since the result is the same as using StringPiece::operator<, except that the former will always use a fast memcmp() call, while this implementation may be slower, depending on how the C++ library is implemented?

Even if this was a valid change, move this to an independent commit and add some unit test for this.

src/evalstring.natvis
```diff
@@ -0,0 +1,52 @@
...
+    <Intrinsic Name="rawLength" Expression="*(const EvalString::Offset*)pos_"/>
+    <Intrinsic Name="mask" Expression="(EvalString::Offset)1 &lt;&lt; (sizeof(EvalString::Offset) * 8 - 1)"/>
+    <Intrinsic Name="isVariable" Expression="(rawLength() &amp; mask()) != 0"/>
+    <Intrinsic Name="length" Expression="rawLength() &amp; ~mask()"/>
+    <DisplayString Condition="isEnd()">{{ end }}</DisplayString>
+    <DisplayString Condition="!isEnd() &amp;&amp; isVariable()">{{ var = {pos_ + sizeof(EvalString::Offset),[length()]s8b} }}</DisplayString>
+    <DisplayString Condition="!isEnd() &amp;&amp; !isVariable()">{{ text = {pos_ + sizeof(EvalString::Offset),[length()]s8} }}</DisplayString>
+  </Type>
+</AutoVisualizer>
+
```
digit-google, line 52: What is this file?

src/evalstring_test.cc
digit-google: I don't think there is a good reason to introduce custom string literals in this code base. While this can make some code prettier, it introduces a mini-language with non-obvious semantics, making everything more difficult to review for multiple maintainers.

While this could be justified if there was a heavy use case to simplify a large code base, but this is clearly not the case here.

src/evalstring.h
```diff
@@ -0,0 +1,111 @@
...
+  /// Return whether the held EvalString has any tokens.
+  bool empty() const;
+
+  /// Clear the held EvalString.
+  void Clear();
+
+  /// Append (or extend if the last token is already raw text) a raw text
+  /// token to the end of the held EvalString.
+  /// @pre `!text.empty()`
+  void AddText(StringPiece text);
```
digit-google, line 92: nit: It is common in Builder patterns to return an `EvalStringBuilder&`  reference here and other methods, which will allow chaining method calls, which is very handy, especially in unit-tests which can do things like:

```
EvalStringBuilder builder;
builder.AddText("foo").AddSpecial("bar");
```

src/evalstring.h
```diff
@@ -0,0 +1,111 @@
...
+
+  /// Append a special variable token to the end of the held EvalString.
+  /// @pre `!text.empty()`.
+  void AddSpecial(StringPiece text);
+
+  /// Return a const reference to the held EvalString.
+  const EvalString& str() const &;
+  operator const EvalString&() const;
```
digit-google, line 100: nit: The purpose of a Builder is to create a frozen / immutable object, so returning references to the builder's state is generally a bad idea and makes the code more fragile. A common name for the method that returns the newly-constructed instance is also "Build()" or "Take()" (the latter clearing the builder at the same time) and makes for clearer code.

src/evalstring.h
```diff
@@ -0,0 +1,111 @@
...
+
+  /// Return a copy of the EvalString by move-constructing the held string.
+  /// @post After calling this method `this->empty() == true`.
+  EvalString str() &&;
```
digit-google, line 104: nit: The name `str()` is misleading, just use a Build() method that returns a new EvalString instance. With the chaining mentioned above, you can write tests simply with statements like:

```
EvalString s = EvalStringBuilder().AddText("foo").AddSpecial("bar").Build();
```

src/dyndep_parser.cc
```diff
@@ -14,14 +14,15 @@
 
 #include "dyndep_parser.h"
 
-#include <vector>
-
 #include "dyndep.h"
+#include "evalstring.h"
```
digit-google, line 18: To follow conventions of the actual code, please call this file `eval_string.h` instead.

src/evalstring.cc
```diff
@@ -0,0 +1,200 @@
...
+// meaning that an EvalStringBuilder that is constantly reused will be very
+// unlikely to allocate.
+
+namespace {
+
+const EvalString::Offset leadingBit =
+    static_cast<EvalString::Offset>(1)
+    << (std::numeric_limits<EvalString::Offset>::digits - 1);
+
+EvalString::Offset clearLeadingBit(EvalString::Offset in) {
```
digit-google, line 44: nit: respect coding conventions by capitalizing the first letter of functions, e.g. `ClearLeadingBit()`.

In this case, it might be simpler to make EvalString::Offset a struct using sized bitfields, e.g.:

```
struct Offset {
  bool is_special : 1;
  unsigned size : 31;

  Offset() : is_special(0), size(0) {}
  Offset(bool a_special, size_t a_size) : is_special(a_special), size(static_cast<unsigned>(a_size)) {
    assert(static_cast<size_t>(size) == a_size);
  }
};
```

The rest of the code will be considerably clearer, and the compiler will handle all the bit fiddling for you.

src/evalstring.h
```diff
@@ -0,0 +1,111 @@
...
+#define NINJA_EVALSTRING_H_
+
+#include <string>
+#include <utility>
+
+class EvalStringBuilder;
+struct StringPiece;
+
+/// A tokenized string that contains variable references.
+/// Can be evaluated relative to an Env.
```
digit-google, line 23: Please update this documentation to clarify the purpose of this class, in particular:

- EvalString instances are immutable, and non-empty ones must be created by an EvalStringBuilder.

- Their main use is to be evaluated into a final string by an Env::Lookup

- An EvalString models a sequence of (token_string token_kind) pairs, and only supports range-based iteration is supported through begin() and end() methods, which return forward-only iterators.

I also recommend defining an `EvalString::Token` struct type to hold the type and string for each token, which will make uses clearer compared to std::pair::first and std::pair::second.

src/manifest_parser.h
digit-google, line 75: nit: This name is misleading, call this `value_builder_` for clarity.

digit-google: There is nothing wrong with this variable (apart the name, see below :-)). Please don't make things more complicated than necessary. In general, if a type has delicate usage semantics that would justify what is proposed here, this is generally the sign of a bad design. If that class was part of a third-party library, that would be a good idea, but this is not the case here.

src/eval_env.h
```diff
@@ -20,46 +20,42 @@
...
-  enum TokenType { RAW, SPECIAL };
-  typedef std::vector<std::pair<std::string, TokenType> > TokenList;
-  TokenList parsed_;
-
-  // If we hold only a single RAW token, then we keep it here instead of
-  // pushing it on TokenList. This saves a bunch of allocations for
-  // what is a common case. If parsed_ is nonempty, then this value
-  // must be ignored.
-  std::string single_token_;
+  template <typename E>
```
digit-google, line 36: nit: This template function should be a template method of EvalString instead. Also keep it simple, performance will be dominated by the lookups in env->LookupVariable(), there is little evidence that the fallthrough improves anything except micro benchmarks. E.g.

```
std::string result;
bool last_was_string = false;
for (const auto& token : *this) {
  if (token.is_special) {
   result.append(env->LookupVariable(token.str));
  } else {
    result.append(token.str.str_, token.str.len_);
  }
}
return result;
```

Also please document that the `E` type should provide a `LookupVariable(StringView)` method for clarity.

src/graph.h
```diff
@@ -190,8 +190,8 @@ struct Edge {
   std::string EvaluateCommand(bool incl_rsp_file = false) const;
 
   /// Returns the shell-escaped value of |key|.
-  std::string GetBinding(const std::string& key) const;
-  bool GetBindingBool(const std::string& key) const;
+  std::string GetBinding(const StringPiece& key) const;
```
digit-google, line 193: nit: Pass StringPiece by value, it's simpler and has the same performance in nearly all cases.

src/dyndep_parser.cc
digit-google: I am stopping my comments here, this PR is far too large and should be split into several commits for readbility.

elliotgoodrich: Thank you for your review, I will break this up into:
  1. Changing methods in `graph.h` to take `StringPiece` instead of `std::string`
  2. Add `EvalString`
  3. Add `EvalString` to manifest parser

src/eval_env.h
elliotgoodrich: I'll resubmit this as a method on `EvalString`.

For `EvalString`s that alternative between text and variables (presumably most of them as the only other alternative is multiple consecutive variables) the fallthrough trick simplifies work for the branch predictor as the switch will almost always be for raw text.

src/evalstring.cc
elliotgoodrich: Will do, thanks.

src/evalstring.h
elliotgoodrich: Will make those changes.

src/evalstring.natvis
elliotgoodrich: It's debugging visualizer for MSVC in the same way you can get gdb pretty printers (https://learn.microsoft.com/en-us/visualstudio/debugger/create-custom-views-of-native-objects?view=visualstudio)

src/string_piece_util.cc
elliotgoodrich: I've just double checked and there is no `operator<` for `StringPiece` that I can find, which is probably why I added this here.

[Clang](https://github.com/llvm/llvm-project/blob/db2f0f87b9263262dcc16a8517f71414f018f264/libcxx/include/__algorithm/lexicographical_compare.h#L76) and [MSVC](https://github.com/microsoft/STL/blob/c865e85d77b9c802359212e8371c65a2f9d44722/stl/inc/xutility#L6095) both have fast branches for using `memcmp` when possible and have the appropriate logic for deciding order when `memcmp == 0` and the strings have different lengths.

I'll bring this out into a smaller PR.

elliotgoodrich:
This is being closed in favor of smaller PRs. The first of which is introducing `StringPieceLess` https://github.com/ninja-build/ninja/pull/2716.

PR #2697, opened 2025-12-05
Make sure to carry the old uid and gid on file replace

jerome-pouiller:
This is a refresh of https://github.com/ninja-build/ninja/pull/1588 opened in 2019 (but the bug is still there).

The normal workflow is to run "ninja all && sudo ninja install". Unfortunately, ninja may rebuild .ninja_log or .ninja_deps. In this case, they are created with root as owner. Once done, user can't build without being root.

The patch below carry the original owner of the file (as "cp -a" does).

Fixes: #1302

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -1022,6 +1022,41 @@ bool Truncate(const string& path, size_t size, string* err) {
   return true;
 }
 
+bool ReplaceContent(const string& file_dst, const string& new_content,
+                    string* err) {
+#ifndef _WIN32
+  struct stat old_file;
+  bool found_uid_gid = true;
+
+  if (stat(file_dst.c_str(), &old_file) < 0) {
+    found_uid_gid = false;
+  }
+#endif
+
+  if (platformAwareUnlink(file_dst.c_str()) < 0) {
+    *err = strerror(errno);
+    return false;
+  }
+
+  if (rename(new_content.c_str(), file_dst.c_str()) < 0) {
+    *err = strerror(errno);
+    return false;
+  }
+
+#ifndef _WIN32
+  // apply uid and gid again so we always stay as the uid and gid of the first
+  // ninja invocation that created the file at file_dst
+  if (found_uid_gid) {
+    if (chown(file_dst.c_str(), old_file.st_uid, old_file.st_gid)) {
+      Warning("Reapplying previous uid and gid failed with: %s uid: %d gid: %d",
+              strerror(errno), old_file.st_uid, old_file.st_gid);
+    }
+  }
+#endif
+
+  return true;
+}
+
 int platformAwareUnlink(const char* filename) {
 	#ifdef _WIN32
 		return _unlink(filename);
--- a/src/util.h
+++ b/src/util.h
@@ -117,6 +117,16 @@ bool Truncate(const std::string& path, size_t size, std::string* err);
 #define PATH_MAX _MAX_PATH
 #endif
 
+
+/// ReplaceContent the content of \a file_dst with the content of \a
+/// new_content. On platforms other than Windows, the file pointed to by \a
+/// file_dst will keep the same uid and gid, if possible. Reason for this is
+/// that permissions on Windows are more complex than on Linux, fetching
+/// permissions on Windows would involve an non-trivial amount of code to fetch
+/// users and have fallback paths
+bool ReplaceContent(const std::string& file_dst, const std::string& new_content,
+                    std::string* err);
+
 #ifdef _WIN32
 /// Convert the value returned by GetLastError() into a string.
 std::string GetLastErrorString();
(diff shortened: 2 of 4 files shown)


jerome-pouiller:
Note the issues pointed out by @jonesmz in the original PR already exist in the current code. In order to keep things atomic (and also because I am lazy and I am not sure of the real benefit of these improvements) this PR don't to implement these suggestions.

jerome-pouiller:
v2:
  - Fix warnings raised by Fedora

src/build_test.cc
```diff
@@ -534,6 +534,17 @@ struct FakeCommandRunner : public CommandRunner {
   VirtualFileSystem* fs_;
 };
 
+class SafeRelease {
+  public:
+  ~SafeRelease() {
+    builder_->command_runner_.release();
```
mathstuf, line 540: Can `builder_` ever be `nullptr`?

mathstuf: Looks like "no" (local to this file and is always used via `&builder`). Can we use a `Builder&` instead in that case?

src/util.h
```diff
@@ -117,6 +117,15 @@ bool Truncate(const std::string& path, size_t size, std::string* err);
 #define PATH_MAX _MAX_PATH
 #endif
 
+/// ReplaceContent the content of \a file_dst with the content of \a
+/// new_content. On platforms other than Windows, the file pointed to by \a
+/// file_dst will keep the same uid and gid. Reason for this is that permissions
+/// on Windows are more complex than on Linux, fetching permissions on Windows
+/// would involve an none trivial amount of code to fetch users and have
+/// fallback paths
```
mathstuf:
```suggestion
/// ReplaceContent the content of \a file_dst with the content of \a
/// new_content. On platforms other than Windows, the file pointed to by \a
/// file_dst will keep the same uid and gid, if possible. Reason for this is that permissions
/// on Windows are more complex than on Linux, fetching permissions on Windows
/// would involve an non-trivial amount of code to fetch users and have
/// fallback paths
```

src/util.cc
```diff
@@ -1022,6 +1022,41 @@ bool Truncate(const string& path, size_t size, string* err) {
...
+  if (rename(new_content.c_str(), file_dst.c_str()) < 0) {
+    *err = strerror(errno);
+    return false;
+  }
+
+#ifndef _WIN32
+  // apply uid and gid again so we always stay as the uid and gid of the first
+  // ninja invocation that created the file at file_dst
+  if (found_uid_gid) {
+    if (chown(file_dst.c_str(), old_file.st_uid, old_file.st_gid)) {
```
mathstuf, line 1050: Should we check for `getuid() == 0` first? Are there systems that perform installation as a non-build user that is not `root`?

src/build_test.cc
jerome-pouiller: For an unknown reason (maybe someone has force pushed the main branch), this commit has been included in my PR while the `main` branch already includes it. I am going to rebase my branch on `main` so it will disappear.

src/util.cc
jerome-pouiller: I have consider that `chown()` should issue a warning whatever the cause of the failure (uid != 0 or anything else we don't have in mind). So, the case you mention is covered by the current error handling. 

If you don't agree I can change the code.

jerome-pouiller:
v3:
  - rebased on master/
  - fixed comment

src/util.cc
mathstuf: I think the only issue is when you're running the build as a non-root user when another user has already built it. Not much to be done in that case.

There's also the "spoiled by root" case that this is fixing; not sure that needs considered.

jerome-pouiller: Currently, we issue a warning in these two scenarios. Do you think we should behave differently?

mathstuf: The warning just isn't that helpful. It'll be `EPERM` and two ids. Perhaps if we split it into:

- non-root user: warning; it isn't expected to work
- root user: stronger warning or error? it is expected to work

mathstuf:
Note that `run0` makes this moot for its users with the new (as of v259) `--empower` option to gain rights, but keep the original uid/gid just for this purpose. See [this LWN article](https://lwn.net/SubscriberLink/1051235/b7e924da61121fc6/).

The PR still makes sense; not everyone uses (or can) `run0` instead of `sudo`.

jerome-pouiller:
@jhasse, could you approve the github workflow (which is stale for 2 weeks)?

jerome-pouiller:
This PR is now ready since 1 month ago. Does anyone have a comment?

(cc: @jhasse @jonesmz @marcelhollerbach @Psirus)

jhasse:
Good job, thank you very much :)

PR #2704, opened 2026-01-01
Reduce `sizeof(Node)`

elliotgoodrich:
Reduce the size of `enum ExistenceStatus` from an `int` to `char` and then move an `int` to take up the 4 bytes of padding after the `bool` members.

This reduces `sizeof(Node)` from 144 to 136 on 64-bit (MSVC).


moritzx22:
This reduces `sizeof(Node) `from 120 to 112 on 64-bit (linux gcc).

PR #2705, opened 2026-01-01
Reduce `sizeof(Edge)`

elliotgoodrich:
Reduce the size of `enum VisitMark` from an `int` to `char` and then move this and the `Jobserver::Slot` (a `std::int16_t`) next to the `bool` members to fill up padding.

This reduces `sizeof(Edge)` from 196 to 184 on 64-bit (MSVC).


moritzx22:
This reduces `sizeof(Edge)` from 168 to 160 on 64-bit (linux gcc).

PR #2706, opened 2026-01-05
Speed up `/showIncludes` parsing

elliotgoodrich:
Achieve a 3-4x speed-up in parsing MSVC's `/showIncludes` by
removing almost all temporary `std::string` values.

`clparser_perftest.exe` before:

    Parse 16384 times in 2777ms avg 169.5us

and after:

    Parse 65536 times in 3122ms avg 47.6us

In general, most `std::string` parameters were changed to either
`StringPiece` or `std::string*` and the algorithm changed to an in-place
one.  Occassionally we change to a `const std::string&` when we require
a null-terminated string, but this is on private methods that we call
only with a `std::string`.

Most of the algorithms are similar, the biggest changes are in
`CLParser::Parse`, where we reuse the same string to emit the filtered
output, and `CLParser::Relativize`, where we introduce
`StringPieceRange` that is a range that lazily splits a string by a
separator. This can be fed into `std::mismatch` to give the right place
where we need to edit the string.  Most of the time we can avoid
splitting the entire string and instead just update the prefix by
removing directories and inserting "../".

There is still a temporary string required in
`CLParser::IsSystemInclude`, until we have C++17 and can use
`std::search` with a custom equality predicate.  Until then, we pass in
a temporary string to use and store this as a member of `CLParser` so
that we can avoid most allocations and deallocations here.

`operator==` for `StringPiece` was made into a friend free function so
we can write `"" == StringPiece()` and hit this overload - previously we
could only write `StringPiece() == ""`.

Add `StringPiece::substr`, which has the same behavior as
`std::string::substr` when exceptions are disabled (i.e. aborting when
`pos > size()`.  This will make it easy to swap to `std::string_view` in
the future.

### DIFF
--- a/src/clparser.cc
+++ b/src/clparser.cc
@@ -19,11 +19,11 @@
 #include <string.h>
 
 #include "metrics.h"
+#include "string_piece.h"
 #include "string_piece_util.h"
 
 #ifdef _WIN32
 #include "includes_normalize.h"
-#include "string_piece.h"
 #else
 #include "util.h"
 #endif
@@ -32,98 +32,151 @@ using namespace std;
 
 namespace {
 
-/// Return true if \a input ends with \a needle.
-bool EndsWith(const string& input, const string& needle) {
-  return (input.size() >= needle.size() &&
-          input.substr(input.size() - needle.size()) == needle);
-}
+/// Iterate through lines of a string (delimeted by "\r", "\n", or
+/// "\r\n"), removing all lines by default unless `KeepCurrent()`
+/// is called. Note that kept lines will be be changed to be
+/// separated by just "\n" regardless of the original line separator.
+struct InPlaceLineRemover {
+  InPlaceLineRemover(std::string* str)
+      : str_(str), out_(str->begin()), end_(str->size()), inStart_(0),
+        inEnd_(std::min(str_->find_first_of("\r\n", inStart_), end_)),
+        copyOnNext_(false) {}
 
-}  // anonymous namespace
+  ~InPlaceLineRemover() { str_->erase(out_, str_->end()); }
+
+  /// Return the current line.
+  StringPiece Current() const {
+    return StringPiece(str_->data() + inStart_, inEnd_ - inStart_);
+  }
+
+  /// Keep the current line instead of removing it from the string.
+  void KeepCurrent() { copyOnNext_ = true; }
+
+  /// Return whether we have iterated through all lines.
+  bool HasNext() const { return inStart_ != end_; }
+
+  /// Move to the next line and remove the current line from the
+  /// string unless `KeepCurrent()` has been called.
+  /// \pre `HasNext() == true`
+  void Next() {
+    assert(HasNext());
+
+    // No need to check bounds as `str[end_] == '\0'` and we won't
+    // advance further.
+    std::string::size_type nextStart = inEnd_;
+    nextStart = (*str_)[nextStart] == '\r' ? nextStart + 1 : nextStart;
+    nextStart = (*str_)[nextStart] == '\n' ? nextStart + 1 : nextStart;
+
+    if (copyOnNext_) {
+      const StringPiece line = Current();
+      out_ = std::copy(line.begin(), line.end(), out_);
+      *out_++ = '\n';
+      copyOnNext_ = false;
+    }
+
+    inStart_ = nextStart;
+    inEnd_ = std::min(str_->find_first_of("\r\n", inStart_), end_);
+  }
+
+private:
+  std::string* str_;
+  std::string::iterator out_;
+  std::string::size_type end_;
+  std::string::size_type inStart_;
+  std::string::size_type inEnd_;
+  bool copyOnNext_;
+};
+
+} // anonymous namespace
 
 // static
-string CLParser::FilterShowIncludes(const string& line,
-                                    const string& deps_prefix) {
-  const string kDepsPrefixEnglish = "Note: including file: ";
-  const char* in = line.c_str();
-  const char* end = in + line.size();
-  const string& prefix = deps_prefix.empty() ? kDepsPrefixEnglish : deps_prefix;
+StringPiece CLParser::FilterShowIncludes(const StringPiece& line,
+                                         const StringPiece& deps_prefix) {
+  const StringPiece kDepsPrefixEnglish = "Note: including file: ";
+  const char* in = line.str_;
+  const char* end = in + line.len_;
+  const StringPiece prefix = deps_prefix.empty() ? kDepsPrefixEnglish : deps_prefix;
   if (end - in > (int)prefix.size() &&
-      memcmp(in, prefix.c_str(), (int)prefix.size()) == 0) {
-    in += prefix.size();
+      memcmp(in, prefix.str_, (int)prefix.len_) == 0) {
+    in += prefix.len_;
     while (*in == ' ')
       ++in;
-    return line.substr(in - line.c_str());
+    return line.substr(in - line.str_);
   }
   return "";
 }
 
 // static
-bool CLParser::IsSystemInclude(string path) {
-  transform(path.begin(), path.end(), path.begin(), ToLowerASCII);
+bool CLParser::IsSystemInclude(StringPiece path, std::string* temp) {
+  // TODO: Use std::search in C++17 to avoid temporaries
+  temp->resize(path.size());
+  transform(path.begin(), path.end(), temp->begin(), ToLowerASCII);
   // TODO: this is a heuristic, perhaps there's a better way?
-  return (path.find("program files") != string::npos ||
-          path.find("microsoft visual studio") != string::npos);
+  return (temp->find("program files") != string::npos ||
+          temp->find("microsoft visual studio") != string::npos);
 }
 
 // static
-bool CLParser::FilterInputFilename(string line) {
-  transform(line.begin(), line.end(), line.begin(), ToLowerASCII);
+bool CLParser::FilterInputFilename(const StringPiece& line) {
+  const size_t maxFilenameSuffix = 3;
+  if (line.len_ < maxFilenameSuffix + 1) {
+    return false;
+  }
+  const auto dot = std::find(
+    std::make_reverse_iterator(line.end()),
+    std::make_reverse_iterator(line.end()) + maxFilenameSuffix + 1,
+    '.');
+  const size_t extLength = dot - std::make_reverse_iterator(line.end());
+  if (extLength > maxFilenameSuffix) {
+    return false;
+  }
+  char BUFFER[maxFilenameSuffix] = {};
+  transform(line.end() - extLength, line.end(), BUFFER, ToLowerASCII);
+  const StringPiece suffix(BUFFER, extLength);
   // TODO: other extensions, like .asm?
-  return EndsWith(line, ".c") ||
-      EndsWith(line, ".cc") ||
-      EndsWith(line, ".cxx") ||
-      EndsWith(line, ".cpp") ||
-      EndsWith(line, ".c++");
+  return suffix == "c" ||
+         suffix == "cc" ||
+         suffix == "cxx" ||
+         suffix == "cpp" ||
+         suffix == "c++";
 }
 
 // static
-bool CLParser::Parse(const string& output, const string& deps_prefix,
-                     string* filtered_output, string* err) {
+bool CLParser::Parse(string* output, const StringPiece& deps_prefix,
+                     string* err) {
   METRIC_RECORD("CLParser::Parse");
 
   // Loop over all lines in the output to process them.
-  assert(&output != filtered_output);
-  size_t start = 0;
   bool seen_show_includes = false;
 #ifdef _WIN32
   IncludesNormalize normalizer(".");
 #endif
 
-  while (start < output.size()) {
-    size_t end = output.find_first_of("\r\n", start);
-    if (end == string::npos)
-      end = output.size();
-    string line = output.substr(start, end - start);
-
-    string include = FilterShowIncludes(line, deps_prefix);
+  std::string normalized;
+  for (InPlaceLineRemover lines(output); lines.HasNext(); lines.Next()) {
+    const StringPiece line = lines.Current();
+    const StringPiece include = FilterShowIncludes(line, deps_prefix);
     if (!include.empty()) {
       seen_show_includes = true;
-      string normalized;
 #ifdef _WIN32
+      normalized = "";
       if (!normalizer.Normalize(include, &normalized, err))
         return false;
 #else
       // TODO: should this make the path relative to cwd?
-      normalized = include;
+      normalized.assign(include.str_, include.len_);
       uint64_t slash_bits;
       CanonicalizePath(&normalized, &slash_bits);
 #endif
-      if (!IsSystemInclude(normalized))
+      if (!IsSystemInclude(normalized, &temp_))
         includes_.insert(normalized);
     } else if (!seen_show_includes && FilterInputFilename(line)) {
       // Drop it.
       // TODO: if we support compiling multiple output files in a single
       // cl.exe invocation, we should stash the filename.
     } else {
-      filtered_output->append(line);
-      filtered_output->append("\n");
+      lines.KeepCurrent();
     }
-
-    if (end < output.size() && output[end] == '\r')
-      ++end;
-    if (end < output.size() && output[end] == '\n')
-      ++end;
-    start = end;
   }
 
   return true;
--- a/src/includes_normalize-win32.cc
+++ b/src/includes_normalize-win32.cc
@@ -19,6 +19,7 @@
 #include "util.h"
 
 #include <algorithm>
+#include <cassert>
 #include <iterator>
 #include <sstream>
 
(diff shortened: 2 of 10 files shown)


src/clparser.cc
```diff
@@ -32,98 +32,115 @@ using namespace std;
...
-          input.substr(input.size() - needle.size()) == needle);
-}
+struct StringChopperGuard {
+  StringChopperGuard(std::string* str) : str_(str), out_(str_->begin()) {}
+
+  ~StringChopperGuard() { str_->erase(out_, str_->end()); }
+
+  std::string* str_;
+  std::string::iterator out_;
+};
```
digit-google, line 88: nit: The purpose of this class and how to use it properly is really non trivial.
I recommend using a different interface that makes it easier to use and understand (especially for future reviewers). For example, a simple GetNextLine() / KeepCurrentLine() pair of methods to iterate over the content, or a function that takes an `std::function<bool(StringPiece)>` to decide which lines to keep.

At a minimum, please provide a comment explaining how to use the class, and add a method such as KeepLine() to avoid the caller to modify `out_` directly.

src/string_piece.h
```diff
@@ -67,6 +69,14 @@ struct StringPiece {
     return len_ == 0;
   }
 
+  /// Return a the substring [pos, pos + length), where length is
+  /// the smaller of count and size() - pos.
+  StringPiece substr(size_t pos = 0, size_t count = -1) const {
+    if (count > len_ - pos)
+      count = len_ - pos;
```
digit-google, line 82: this is missing an `if (pos > len_) { pos = len_; count = 0 }` check first to avoid overflows.

src/clparser.cc
elliotgoodrich: That's a much better design. I'll update the code to something like `GetNextLine()` and `KeepCurrentLine()`, it'll pull out a bunch of parsing code and make what the loop is doing much clearer.

src/string_piece.h
elliotgoodrich: [`std::string_view::substr`](https://en.cppreference.com/w/cpp/string/basic_string_view/substr.html) throws an exception if `pos > size()`, which I didn't think appropriate in the ninja codebase.  Considering `StringPiece` will probably be replaced with `std::string_view` at some point, I think the transition would be easier if we had `assert(pos <= len_)` instead of changing the input.  This way any correct use of `StringPiece::substr` would be a non-throwing case of `std::string_view::substr`.

But I'll defer to whatever approach you think reasonable for ninja since I am not currently calling `substr` where `pos > len_`.

digit-google: Note that Ninja is always compiled without exceptions, so even using `std::string_view` will not have the behavior you describe. The libc++ source for [std::string_view::substr()](https://github.com/llvm/llvm-project/blob/ed36bff24662737c9c7a3b9ab3c0ecaf46587e42/libc/src/__support/CPP/string_view.h#L175) shows that its implementation does two checks and actually never raises an exception.

Meanwhile `assert()` will be removed from release builds so is not a good idea, and the generated code is also smaller and faster with two simple checks and no conditional branch to a function like `Fatal()` anyway.

digit-google: Ooops, sorry, I was wrong, the link I gave is not for the libc++ version of std::string_view, but the replacement class of the same name used by the llvm `libc` (a different library). The correct link is [this one](https://github.com/llvm/llvm-project/blob/8246257cac48457b5dc90d8b20c1fe6b7537d790/libcxx/include/string_view#L466) which shows that libc++ will throw an exception, or even abort if `pos > size()` in this specific case.

I still think two small integer checks will lead to smaller code though, but do as you want, as long as the condition is properly checked. Do not use `assert()` though.

src/clparser.cc
elliotgoodrich: I changed to removing by default and then an extra method `KeepCurrent()`, which parallels what the previous code did so instead of 

```cpp
      filtered_output->append(line);
      filtered_output->append("\n");
```

we have

```cpp
      lines.KeepCurrent();
```

src/clparser.cc
```diff
@@ -32,98 +32,151 @@ using namespace std;
...
-      filtered_output->append(line);
-      filtered_output->append("\n");
+      lines.KeepCurrent();
     }
-
-    if (end < output.size() && output[end] == '\r')
-      ++end;
-    if (end < output.size() && output[end] == '\n')
-      ++end;
-    start = end;
```
elliotgoodrich, line 126: My initial change didn't actually handle lone `'\r'` and `'\n'`, but it does now and I have added a new test in `clparser_test.cpp` to cover line endings.

src/string_piece.h
elliotgoodrich: I've put it as a `Fatal`, which will match the `std::string_view::susbtr` behavior when compiling without exceptions.  This will make swapping over easier in the future.

src/clparser.cc
```diff
@@ -32,98 +32,151 @@ using namespace std;
...
   return "";
 }
 
 // static
-bool CLParser::IsSystemInclude(string path) {
-  transform(path.begin(), path.end(), path.begin(), ToLowerASCII);
+bool CLParser::IsSystemInclude(StringPiece path, std::string* temp) {
+  // TODO: Use std::search in C++17 to avoid temporaries
+  temp->resize(path.size());
+  transform(path.begin(), path.end(), temp->begin(), ToLowerASCII);
```
digit-google, line 113: nit: Can you use std::transform() here, we want to get rid of `using namespace std`, and this is ore consistent with the rest of your changes.

src/clparser.cc
```diff
@@ -32,98 +32,151 @@ using namespace std;
...
 
-/// Return true if \a input ends with \a needle.
-bool EndsWith(const string& input, const string& needle) {
-  return (input.size() >= needle.size() &&
-          input.substr(input.size() - needle.size()) == needle);
-}
+/// Iterate through lines of a string (delimeted by "\r", "\n", or
+/// "\r\n"), removing all lines by default unless `KeepCurrent()`
+/// is called. Note that kept lines will be be changed to be
+/// separated by just "\n" regardless of the original line separator.
```
digit-google, line 38: I don't think there is any need to support \r line terminators, since only the original MacOS (pre OSX) used that. Supporting \r\n and \n should be sufficient and will allow avoiding using the slow std::find_first_of_call() in the code below (instead, just find the next \n and verify if it follows an optional \r, which will be faster). What do you think?

src/clparser.cc
```diff
@@ -32,98 +32,151 @@ using namespace std;
...
+  /// string unless `KeepCurrent()` has been called.
+  /// \pre `HasNext() == true`
+  void Next() {
+    assert(HasNext());
+
+    // No need to check bounds as `str[end_] == '\0'` and we won't
+    // advance further.
+    std::string::size_type nextStart = inEnd_;
+    nextStart = (*str_)[nextStart] == '\r' ? nextStart + 1 : nextStart;
+    nextStart = (*str_)[nextStart] == '\n' ? nextStart + 1 : nextStart;
```
digit-google, line 68: nit: avoid repeated `(*str_)[index]` calls, each one must perform a small small-string-sso optimization check, you can just use `str_->data()` as a base pointer pointer instead here given the comment on line 64.

src/clparser.cc
```diff
@@ -32,98 +32,151 @@ using namespace std;
...
+  StringPiece Current() const {
+    return StringPiece(str_->data() + inStart_, inEnd_ - inStart_);
+  }
 
-}  // anonymous namespace
+  /// Keep the current line instead of removing it from the string.
+  void KeepCurrent() { copyOnNext_ = true; }
+
+  /// Return whether we have iterated through all lines.
+  bool HasNext() const { return inStart_ != end_; }
```
digit-google, line 56: This is better but I must say I find this API confusing because there is very little indication when / how to use these methods exactly. I recommend simplifying this and providing a usage example in the top-comment, e.g. something that looks like:

```
StringPiece  current_line;
InPlaceLineRemover remover(&output):
while (remove.GetCurrentLine(&current_line)) {
    if (... some condition)
        remover.KeepCurrentLine();
}
remover.Finish();
```

Should be easier to use and understand. And will allow you to manage less state in the implementation. Also please clarify in the documentation that |output| is modified in-place and should not be touched by the loop body until the Finish() call.

src/string_piece.h
digit-google: Works for me. Thanks.

src/includes_normalize.h
```diff
@@ -24,17 +24,17 @@ struct StringPiece;
...
   /// Normalize path relative to |relative_to|.
-  IncludesNormalize(const std::string& relative_to);
+  IncludesNormalize(const StringPiece& relative_to);
 
   // Internal utilities made available for testing, maybe useful otherwise.
-  static std::string AbsPath(StringPiece s, std::string* err);
-  static std::string Relativize(StringPiece path,
-                                const std::vector<StringPiece>& start_list,
-                                std::string* err);
+  static void AbsPath(std::string* s, std::string* err);
```
digit-google, line 30: Please change the name to `MakeAbsolutePath()` to clarify that this modifies |*s| in-place. Return false on error, true on success to simplify error checking.

src/includes_normalize.h
```diff
@@ -24,17 +24,17 @@ struct StringPiece;
...
+  static void Relativize(std::string* abs_path,
+                         const std::vector<StringPiece>& start_list,
+                         std::string* err);
```
digit-google, line 33: Similarly, return bool for success/failure here. Or in this specific case, it seems your new implementation never returns an error, so drop the |err| parameter and return void.

src/includes_normalize-win32.cc
```diff
@@ -124,86 +128,167 @@ bool IsFullPathName(StringPiece s) {
...
-      break;
-    }
+void IncludesNormalize::Relativize(
+    std::string *abs_path, const vector<StringPiece>& start_list, string* err) {
+  AssertIsAbsoluteInDebug(*abs_path);
+
+  const StringPieceRange path_list(*abs_path, '/');
+  const auto diff =
+      std::mismatch(path_list.begin(), path_list.end(), start_list.begin(),
+                    start_list.end(), EqualsCaseInsensitiveASCII);
```
digit-google, line 243: I must say that this code is a hard to understand and verify.

In particular because StringPieceRange, when passed an absolute file path as input, will return an empty StringPiece as the first item during iteration. For the loop (before your change) and the std::mistmatch call to work properly requires that `start_list` also starts with an empty StringPiece, which is absolutely not obvious here. 

It turns out that the only calls to this function pass the value of `split_relative_to_` which is itself the result of `SplitStringView(relative_to_, '/')` which *does* indeed create a vector with a first empty StringPiece as its first item. Hence that code is correct, but this is quite fragile as this precondition is not commented anywhere.

It seems `split_relative_to_` is useful for repeated calls to Normalize() though, so keeping it, instead of calling StringPieceRange(relative_to_, '/') within this function is probably better for performance.

What would you think about a helper class to hide these gory details and clarify usage, e.g.:

```
/// Helper class to convert absolute paths into paths relative to a fixed reference path.
/// Useful to perform repeated operations in a loop with minimal allocations.
///
/// Usage is:
///    RelativePathMaker maker(from_path_abs);
///    for (...) {
///       std::string path = GetAbsolutePath(...);
///       maker.AbsoluteToRelativeInPlace(&path);
///       ...  // path is now relative to from_path_abs
///    }
class RelativePathMaker {
  /// Create instance, input path must be absolute.
  RelativePathMaker(StringView from_path_abs)
    : from_path_split_(SpltStringPiece(from_path_abs)) {}

  /// Convert an absolute path into a path relative to the recorded reference path.
  /// Modifies |*abs_path| in-place.
  void AbsoluteToRelativeInPlace(std::string* abs_path) const;

  /// A variant that takes a StringPiece as input, and generates a new relative path string.
  std::string AbsoluteToRelative(StringPiece abs_path) const {
    std::string result = abs_path;
    AbsoluteToRelativeInPlace(&result);
    return result;
  }

private:
  std::vector<StringPiece> from_path_split_;
};
```

src/includes_normalize-win32.cc
```diff
@@ -124,86 +128,167 @@ bool IsFullPathName(StringPiece s) {
...
+  const size_t sections_to_replace = start_list.end() - diff.second;
+  const StringPiece dotdot = "../";
+  const size_t bytes_to_write = sections_to_replace * dotdot.size();
+  const size_t bytes_to_delete = diff.first->str_ - abs_path->data();
```
digit-google, line 247: Technically, this `diff.first` value being the result of an std::mistmatch() call could point to `path_list.end()` iterator value. In general, dereferencing an `end()` iterator is undefined behavior in C++, so this code looks suspicious.

In that case, and based on the current `StringPieceRange` implementation, `diff.first->str_` will point to the end of the `StringPiece` value passed to the `StringPieceRange` constructor.

In general, dereferencing a `StringPiece::end()` pointer  is also undefined behavior, as there is no guarantee that this points to any addressable memory. In this specific case, this works because it points to the buffer of and `std::string` which must store a terminating \0 byte, but this is a very subtle implementation detail.

So please to check against path_list.end() explicitly before dereferencing it, e.g.:

```
const size_t bytes_to_delete = (diff.first == path_list.end()) ? 0 : (diff.first->str_ - abs_path->data());
```

Also add a comment explaining that `diff.first->str_` points to the first character after the common path prefix, as this is not obvious from the lines above.

However, maybe not keep this variable after all (see comment below).

src/includes_normalize-win32.cc
```diff
@@ -124,86 +128,167 @@ bool IsFullPathName(StringPiece s) {
...
+  if (bytes_to_write > bytes_to_delete) {
+    // We can insert chars at any position from
+    // [begin(), begin() + bytes_to_delete) so we choose the last
+    // element to minimize copying
+    abs_path->insert(abs_path->begin() + bytes_to_delete,
+                     bytes_to_write - bytes_to_delete, '\0');
+  } else if (bytes_to_write < bytes_to_delete) {
+    abs_path->erase(abs_path->begin(),
+                    abs_path->begin() + bytes_to_delete - bytes_to_write);
+  }
```
digit-google, line 257: This is hard to follow. I suggest simplifying (e.g. don't use iterators if you don't have to) and using better variable names to make this easier to follow, e.g.:

```
// The length of the common path prefix, in abs_path characters.
const size_t common_prefix_len =
    (diff.first == path_list.end()) ? abs_path->size() : diff.first->str_ - abs_path->data();

// The number of ../ to be inserted at the start of the result, corresponding
// to the number of path segments after the common prefix from start_list.
const size_t dotdot_count = start_list.end() - diff.second;

const size_t dotdot_len = 3 * dotdot_count;

// The following must remove the common prefix characters, then prepend
// a sequence of |dotdot_count| "../" segments into the result. The end result
// is [<dotdot_len>][<non_common_path>].

// First relocate the non common path to the right position, removing
// or inserting bytes if needed.
if (common_prefix_len > dotdot_len) {
  abs_path->erase(0, common_prefix_len - dotdot_len);
} else if (common_prefix_len < dotdot_len) {
  abs_path->insert(0, dotdotlen - common_prefix_len, '\0')
}

// Now write the ../ sequence in place.
for (char* data = abs_path->data(), size_t count = dotdot_count; count > 0; count--) {
  memcpy(data, "../", 3);
  data += 3;
}
```

src/clparser.cc
```diff
@@ -32,98 +32,151 @@ using namespace std;
...
+bool CLParser::FilterInputFilename(const StringPiece& line) {
+  const size_t maxFilenameSuffix = 3;
+  if (line.len_ < maxFilenameSuffix + 1) {
+    return false;
+  }
+  const auto dot = std::find(
+    std::make_reverse_iterator(line.end()),
+    std::make_reverse_iterator(line.end()) + maxFilenameSuffix + 1,
+    '.');
+  const size_t extLength = dot - std::make_reverse_iterator(line.end());
```
digit-google, line 129: I am sorry but this is ridiculous use of iterators for a maxFilenameSuffix size of 3! Just do a loop in a lambda, e.g.:

```
auto find_path_ext(StringPiece path, size_t max_len) -> StringPiece {
  max_len += 1;  // Include dot in computations below.
  if (path.size()  >= max_len) {
    for (const char* pos = path.end(), limit = pos - max_len; pos > limit; --pos) {
      if (pos[-1] == '.')
        return StringPiece(pos, end - pos);  // do not include dot in result.
    }
  }
  return {};
};

const size_t max_ext_len = 3;  // Maximum extension length

StringPiece raw_ext = find_path_ext(line, max_ext_len);
if (raw_ext.empty())
  return false;

char lower_ext[max_ext_len] = {};
std::transform(raw_ext.begin(), raw_ext.end(), lower_ext, ToLowerASCII);
StringPiece ext(lower_ext, raw_ext.size());

return (ext == "c" || .ext == "cc" || .....);
```

src/string_piece.h
digit-google: However, this PR is becoming huge, please split it in separate commits that address different features (e.g. one for the string_piece.h + string_piece_utils.h modification), another one for the clparser one, and another one for the includes_normalize one... This will make everything easier to review.

elliotgoodrich:
Thank you @digit-google for all these reviews.  I have split this into 2 PRs for the moment.  There are more changes not yet up as a pull request, but they depend on these landing first.

PR #2711, opened 2026-01-15
Stat symlinks correctly on windows

jlonnberg:
Solves a problem where symlinks were incorrectly stat:ed on windows.
The current stat will use the mtime for the symlink and not the target file, caused builds to skip edges as the timestamp of the symlink is not updated when the target file is updated.


jhasse:
Thanks!

Issue #2573, opened 2025-02-19
Missing implicit dependencies in dyndep files are silently ignored instead of causing an error


stanleyj-edsn:
When using **dyndep**, if a file listed as an implicit dependency in the dyndep file is missing and there is no rule to create it, Ninja v1.12.1 does not produce an error but instead proceeds with the build as if everything is fine.  

However, running Ninja with `-d explain` correctly detects the missing file but still does not fail the build.  

Given `build.ninja`:

```ninja
build output.pdf: dita-to-pdf output.ditamap || output.pdf.dd 
# assume dita-to-pdf does simply `touch $out` for now
  dyndep = output.pdf.dd

```

Given `output.pdf.dd`:  
   ```ninja
   ninja_dyndep_version = 1
   build output.pdf : dyndep | thisfilexists.md mymissingfile.md
```

#### Expected Behavior
- Ninja should **fail with an error** if an implicit dependency listed in the dyndep file is missing and has no rule to generate it.  

#### Actual Behavior
- Ninja does not fail and proceeds with the build:  
  ```sh
  $ ninja output.pdf
  [1/1] touch output.pdf
  ```
- Running with `-d explain` shows that Ninja **knows** the file is missing:  
  ```sh
  $ ninja -v -d explain output.pdf
  ninja explain: loading dyndep file 'output.pdf.dd'
  ninja explain: mymissingfile.md has no in-edge and is missing
  ninja explain: mymissingfile.md is dirty
  [1/1] touch output.pdf
  ```

It could also be this is correct behavior, and I misunderstood how dyndep works. But how can I ensure ninja fails early, because it is especially important in a build pipeline where I want to be clear towards the developers that compilation will not work (and why, in this case because a file is missing and there is no way to build it) prior to getting an error from the compiler. Thanks in advance

eli-schwartz:
This looks like the behavior for an order-only `||` dependency.

> Every statement must specify exactly one explicit output and must use the rule name dyndep. The `| imp-outs...` and `| imp-ins...` portions are optional.


> The build statements in a dyndep file must have a one-to-one correspondence to build statements in the ninja build file that name the dyndep file in a dyndep binding. No dyndep build statement may be omitted and no extra build statements may be specified.

So... In a dyndeps rule, implicit inputs are ending up treated as order-only inputs instead?

moritzx22:
> **Expected Behavior**
[...]

PR #2713, opened 2026-01-18
Detection of missing implicit dependencies in dyndep files

moritzx22:
## Related issue
https://github.com/ninja-build/ninja/issues/2573

## Detailed Description
https://github.com/ninja-build/ninja/issues/2573#issuecomment-3764077423

## Example Build
```ninja
rule dyndep
  command = printf 'ninja_dyndep_version = 1\nbuild output.pdf : dyndep | thisfilexists.md  spare.md\n' > $out

rule touch
  command =  touch $out

build output.pdf: touch output.ditamap || output.pdf.dd 
  dyndep = output.pdf.dd

build output.pdf.dd: dyndep
```
### Graph
graph
File `spare.md` is missing.
### Proposed solution
```
$ ninja
ninja: build stopped: 'spare.md', needed by dyndep 'output.pdf.dd', missing and no known rule to make it.
```
The missing file is detected, ninja reports the missing file.

### Current Master Branch
```
$ ninja
[1/1] touch output.pdf
```
The missing file is silently ignored.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -825,6 +825,17 @@ bool Builder::StartEdge(Edge* edge, string* err) {
   if (edge->is_phony())
     return true;
 
+  // Late validation: verify that dyndep‑only inputs are present.
+  for (const auto input : edge->inputs_) {
+    if (input->missing() && input->generated_by_dyndep_loader()) {
+      string referenced;
+      referenced = ", needed by dyndep '" + edge->GetUnescapedDyndep() + "',";
+      *err = "'" + input->path() + "'" + referenced +
+             " missing and no known rule to make it";
+      return false;
+    }
+  }
+
   int64_t start_time_millis = GetTimeMillis() - start_time_millis_;
   running_edges_.insert(make_pair(edge, start_time_millis));
 
@@ -943,6 +954,13 @@ bool Builder::FinishCommand(CommandRunner::Result* result, string* err) {
         }
       }
     }
+
+    // Refresh output file status to support later dyndep existence checks.
+    // https://github.com/ninja-build/ninja/issues/2573
+    for (auto o : edge->outputs_) {
+      o->MarkExists();
+    }
+
     if (node_cleaned) {
       record_mtime = edge->command_start_time_;
     }
--- a/src/graph.h
+++ b/src/graph.h
@@ -77,6 +77,14 @@ struct Node {
     return exists_ == ExistenceStatusExists;
   }
 
+  void MarkExists() {
+    exists_ = ExistenceStatusExists;
+  }
+
+  bool missing() const {
+    return exists_ == ExistenceStatusMissing;
+  }
+
   bool status_known() const {
     return exists_ != ExistenceStatusUnknown;
   }
--- a/src/build_test.cc
+++ b/src/build_test.cc
@@ -3367,6 +3367,61 @@ TEST_F(BuildTest, DyndepMissingAndNoRule) {
   EXPECT_EQ("loading 'dd': No such file or directory", err);
 }
 
+TEST_F(BuildTest, DyndepMissingInput) {
+  // Check that the build system detects when a dyndep-provided input is missing
+  ASSERT_NO_FATAL_FAILURE(AssertParse(&state_,
+                                      R"ninja(
+rule touch
+  command = touch $out $out.imp
+build out: touch || dd
+  dyndep = dd
+)ninja"));
+
+  fs_.Create("dd", R"ninja(
+ninja_dyndep_version = 1
+build out | out.imp: dyndep | tmp.imp missing.imp
+)ninja");
+
+  fs_.Create("tmp.imp", "");
+
+  string err;
+  EXPECT_TRUE(builder_.AddTarget("out", &err));
+  ASSERT_EQ("", err);
+  EXPECT_EQ(builder_.Build(&err), ExitFailure);
+  EXPECT_EQ("'missing.imp', needed by dyndep 'dd', missing and no known rule to make it", err);
+  EXPECT_TRUE(command_runner_.commands_ran_.empty());
+}
+
+TEST_F(BuildTest, DyndepMissingInputUsedInDepFile) {
+  // The missing input cannot be detected because the dyndep‑introduced node is
+  // also marked as coming from a depfile.
+  ASSERT_NO_FATAL_FAILURE(AssertParse(&state_,
+                                      R"ninja(
+rule touch
+  command = touch $out $out.imp
+build out2: touch out1 || dd
+  dyndep = dd
+build out1: touch
+  depfile = out2.d
+)ninja"));
+
+  fs_.Create("dd", R"ninja(
+ninja_dyndep_version = 1
+build out2 | out2.imp: dyndep | tmp.imp missing.imp
+)ninja");
+  fs_.Create("out2.d", "out1: missing.imp\n");
+  fs_.Create("tmp.imp", "");
+
+  string err;
+  EXPECT_TRUE(builder_.AddTarget("out2", &err));
+  ASSERT_EQ("", err);
+  EXPECT_EQ(builder_.Build(&err), ExitSuccess);
+  EXPECT_EQ("", err);
+  ASSERT_EQ(2u, command_runner_.commands_ran_.size());
+  EXPECT_EQ("touch out1 out1.imp", command_runner_.commands_ran_[0]);
+  EXPECT_EQ("touch out2 out2.imp", command_runner_.commands_ran_[1]);
+}
+
 TEST_F(BuildTest, DyndepReadyImplicitConnection) {
   // Verify that a dyndep file can be loaded immediately to discover
   // that one edge has an implicit output that is also an implicit


moritzx22:
## Change in push
One previously unhandled scenario has been fixed:
when a missing node was marked as both depfile‑loaded and dyndep‑loaded, Ninja incorrectly reported it as missing due to the dyndep path, even though depfile loading should suppress that error. This bug is now resolved.

## Limitations
There remains one limitation: Ninja still cannot report a missing dyndep‑listed dependency of an input when that dependency also appears in a depfile. The dependency graph does not track loader information for the dependencies of inputs, so it is not possible to distinguish whether such a dependency originated from a depfile or a dyndep file. See [input](https://github.com/ninja-build/ninja/blob/cc60300ab94dae9bb28fece3c9b7c397235b17de/src/graph.h#L239-L246).

Support for tracking loader information at that level has not been added, as it may not be desirable.

src/graph.h
```diff
@@ -150,12 +164,18 @@ struct Node {
...
+  /// precedence. Depfile and deps log share the same status and are not tracked
+  /// separately.
+  ///
+  /// If node is loaded from depfile or deps log and do not have a producing
+  /// edge, the build should not abort if it is missing (as for regular source
+  /// inputs or inputs loaded from dyndep).
+  ///
+  /// There is no implicit default, callers must explicitly set an initial
+  /// value.
+  Loader generated_by_dep_loader_;
```
digit-google: That's a good change, but please use a default initializer here. I suggest making Loader::Manifest the first value in the enum, with value 0, and use it as the default.

src/graph.h
```diff
@@ -40,8 +40,10 @@ struct State;
 /// Information about a node in the dependency graph: the file, whether
 /// it's dirty, mtime, etc.
 struct Node {
-  Node(const std::string& path, uint64_t slash_bits)
-      : path_(path), slash_bits_(slash_bits) {}
+  enum Loader : char { Depfile, Dyndep, Manifest };
```
digit-google: nit: use an `enum class` here for clarity

src/graph.h
digit-google: Feel free to rename this to `loader_` too, as the `generated_by_dep_` prefix is misleading in the case of manifest-created nodes :)

src/state.cc
```diff
@@ -92,15 +92,39 @@ Edge* State::AddEdge(const Rule* rule) {
...
   if (node)
     return node;
-  node = new Node(path.AsString(), slash_bits);
+
+  node = new Node(path.AsString(), slash_bits, Loader);
   paths_[node->path()] = node;
   return node;
 }
 
+Node* State::GetNodeManifest(StringPiece path, uint64_t slash_bits) {
```
digit-google: This naming is really confusing, as this function does not return a manifest. Also the `Get` prefix hides the fact that this mutates the state of the node. Same goes for `GetNodeDeps()` and `GetNodeDyn()`.

Consider renaming the functions, for example:

- `GetNode()` -> `FindOrCreateNode()`
- `GetNodeManifest()` -> `MakeManifestNode()`
- `GetNodeDeps()` -> `MakeDepfileNode()`
- `GetModeDyn()` -> `MakeDyndepNode()`

Then you can keep `GetNode()` as is and won't have to change all the test code.

(you can also make `slash_bits` default-initialized to 0 to simplify the tests, but this is unrelated to the problem you're trying to fix here, though useful, so I would leave that in a separate PR).

src/state.h
```diff
@@ -103,7 +103,18 @@ struct State {
 
   Edge* AddEdge(const Rule* rule);
 
-  Node* GetNode(StringPiece path, uint64_t slash_bits);
+  /// For use in unit tests only; the node must already exist.
+  Node* GetNode(StringPiece path) const;
+  // Convenience helpers for creating or retrieving nodes with a specific
+  // loader.
```
digit-google, line 111: Related to the comment above about the names of these methods, please document that they mutate the state (and how) for each one, as this is subtle but important behavior.

src/build.cc
```diff
@@ -943,6 +954,12 @@ bool Builder::FinishCommand(CommandRunner::Result* result, string* err) {
         }
       }
     }
+
+    // Refresh output file status to support later dyndep existence checks.
```
digit-google, line 958: nice, can you add a comment pointing to issue #2573 here too, this will clarify why this is important.

src/graph.h
moritzx22: > That's a good change, but please use a default initializer here.

The constructor always sets loader_, so a default initializer wouldn’t change anything. Matching slash_bits_ might be strong enough reason to add one.
```cpp
struct Node{
Node(const std::string& path, uint64_t slash_bits, Loader loader)
      : path_(path), slash_bits_(slash_bits), loader_(loader) {}
...
std::string path_;
uint64_t slash_bits_ = 0;
// Loader loader_ = Loader::Manifest;
Loader loader_;
...
};
```
Introducing a default here might give the impression that Node can be constructed without specifying a loader, which isn’t the case. That’s why I avoided it, but I’m happy to adjust if it is desirable, please confirm.

> I suggest making Loader::Manifest the first value in the enum, with value 0, and use it as the default.

I will do. Historically the depfile acted as the implicit default, but changing the order doesn’t affect behavior.


> Feel free to rename this to loader_ too, as the generated_by_dep_ prefix is misleading in the case of manifest-created nodes :)

I will do.

src/graph.h
moritzx22: Appreciate the feedback - will do. I didn’t use a scoped `enum` here to stay consistent with the existing codebase. 

Would it be useful to open a separate administrative PR to migrate the existing `enums` to `enum class` for consistency?

src/state.cc
moritzx22: I renamed the functions to:
```cpp
/// For use in unit tests only; the node must already exist.
  Node* GetNode(StringPiece path) const;
  // Convenience helpers for creating or retrieving nodes with a specific loader.
  Node* FindOrCreateManifestNode(StringPiece path, uint64_t slash_bits);
  Node* FindOrCreateDepfileNode(StringPiece path, uint64_t slash_bits);
  Node* FindOrCreateDyndepNode(StringPiece path, uint64_t slash_bits);
private:
  Node* FindOrCreateNode(StringPiece path, uint64_t slash_bits, Node::Loader);
```
They’re a bit long, but they’re consistent and descriptive.

> Then you can keep GetNode() as is and won't have to change all the test code.

the old `GetNode(StringPiece path, uint64_t slash_bits)` was intentionally removed. Its responsibilities are now covered by the three `FindOrCreate…Node()` functions, and in practice it maps directly to `FindOrCreateDepfileNode()`.

The remaining `GetNode(StringPiece path) const` is not the successor of the historical `GetNode`. It’s a const lookup-only function that never creates nodes, so it doesn’t need slash_bits. It’s mainly used in unit tests, where the goal is often to check whether a node already exists.

The changes to the unit tests are inherent to the PR.  The new API doesn’t require callers to distinguish between looking up and creating a node, but it does make that distinction possible. In my view, this makes the tests cleaner and reinforces the clearer separation between lookup and creation.

If the intention is to keep the unittest changes small, `GetNode` can be replaced with `FindOrCreateDepfileNode`—this is already the pattern used in `deps_log_test.cc`.

> (you can also make slash_bits default-initialized to 0 to simplify the tests, but this is unrelated to the problem you're trying to fix here, though useful, so I would leave that in a separate PR).

I think that’s a good idea. I’d be in favor of including it in this PR as well, since the unit tests are already affected by the `GetNode` refactoring and would benefit from the simplification. I haven’t made the change.

src/state.h
moritzx22: I will add more comments.

src/build.cc
moritzx22: sure

src/state.cc
digit-google: Excellent names, thanks. And the associated comments are very helpful.

Regarding the enum class, they're not mandatory all, just a good way to clarify that these are enums. Most other enum values in the source code use prefixes to clarify the context. E.g. `LoaderManifest`, `LoaderDepfile`, etc... would be more clear where they are used and would be consistent with the rest of the project.

Personnaly, I am a fan of the `kTypeName` convention too because the `k` prefix (which stands for constant) makes it explicit that this is a value, and not a type name, but there is no point changing the current project's conventions any way.

src/graph.h
digit-google: See comment below, I don't think that's worth it, and to be honest the main project owner and sole submitter doesn't seem to have the time to do reviews of non-trivial PRs very often. I fear such a cosmetic change would stay in review limbo for a long time :-/ Thanks for the idea though.

digit-google:
This PR seems fine to me, I would prefer the enum values to use a `Loader` prefix, but I don't see that as a blocker. @jhasse what do you think?

jhasse:
> I would prefer the enum values to use a Loader prefix

It's an enum class, so that would mean `Loader::LoaderDyndep` which is a bit redundant.

> @jhasse what do you think?

I'm not that familiar with dyndeps, will have to take a closer look when I find the time. I trust your judgment and @bradking, @mathstuf.

digit-google:
> > I would prefer the enum values to use a Loader prefix
> 
> It's an enum class, so that would mean `Loader::LoaderDyndep` which is a bit redundant.
> 
I agree, but I meant either use a regular enum `LoaderDyndep`, to be consistent with the rest of the sources, or an enum class value like `Loader::Dyndep` :-)

PR #2716, opened 2026-01-21
Avoid string copies on lookup methods

elliotgoodrich:
Create a transparent comparator `StringPieceLess` that can allow us to lookup up values in a `std::map<std::string, V>` using a `StringPiece` without needing to create a temporary `std::string`.

Similarly, there are other methods on classes in `graph.h` that take a `std::string` that can work just fine with a `StringPiece` and avoid more temporaries.

### DIFF
--- a/src/string_piece_util.cc
+++ b/src/string_piece_util.cc
@@ -15,8 +15,10 @@
 #include "string_piece_util.h"
 
 #include <algorithm>
+#include <cstring>
 #include <string>
 #include <vector>
+
 using namespace std;
 
 vector<StringPiece> SplitStringPiece(StringPiece input, char sep) {
@@ -75,4 +77,16 @@ bool EqualsCaseInsensitiveASCII(StringPiece a, StringPiece b) {
   }
 
   return true;
+}
+
+bool StringPieceLess::operator()(StringPiece lhs,
+                                 StringPiece rhs) const {
+
+  const std::size_t min_length = std::min(lhs.size(), rhs.size());
+  const int cmp = std::memcmp(lhs.str_, rhs.str_, min_length);
+  if (cmp == 0) {
+    return lhs.size() < rhs.size();
+  } else {
+    return cmp < 0;
+  }
 }
--- a/src/string_piece_util.h
+++ b/src/string_piece_util.h
@@ -30,4 +30,16 @@ inline char ToLowerASCII(char c) {
 
 bool EqualsCaseInsensitiveASCII(StringPiece a, StringPiece b);
 
+/// A comparator for StringPiece that can be used in
+/// `std::map<std::string, V, StringPieceLess>` that will allow using a
+/// StringPiece as a key.  Note that this compares elements of the
+/// strings as `unsigned char`, which means that implementations that
+/// have a signed `char` may give an unexpected (yet valid) ordering
+/// when iterating through a `std::map` containing negative `char`s.
+struct StringPieceLess {
+  using is_transparent = void;
+
+  bool operator()(StringPiece lhs, StringPiece rhs) const;
+};
+
 #endif  // NINJA_STRINGPIECE_UTIL_H_
--- a/src/eval_env.cc
+++ b/src/eval_env.cc
@@ -18,8 +18,8 @@
 
 using namespace std;
 
-string BindingEnv::LookupVariable(const string& var) {
-  map<string, string>::iterator i = bindings_.find(var);
+std::string BindingEnv::LookupVariable(StringPiece var) {
+  const auto i = bindings_.find(var);
   if (i != bindings_.end())
     return i->second;
   if (parent_)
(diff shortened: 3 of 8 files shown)


src/string_piece_util.h
```diff
@@ -30,4 +30,13 @@ inline char ToLowerASCII(char c) {
 
 bool EqualsCaseInsensitiveASCII(StringPiece a, StringPiece b);
 
+/// A comparator for StringPiece that can be used in
+/// `std::map<std::string, V, StringPieceLess>` that will allow using a
+/// StringPiece as a key.
+struct StringPieceLess {
+  using is_transparent = void;
+
+  bool operator()(const StringPiece& lhs, const StringPiece& rhs) const;
```
digit-google: Should this take StringPiece value arguments instead?

https://quuxplusone.github.io/blog/2021/11/09/pass-string-view-by-value/

src/string_piece_util.cc
```diff
@@ -76,3 +76,11 @@ bool EqualsCaseInsensitiveASCII(StringPiece a, StringPiece b) {
 
   return true;
 }
+
+bool StringPieceLess::operator()(const StringPiece& lhs,
+                                 const StringPiece& rhs) const {
+  return std::lexicographical_compare(lhs.begin(),
+                                      lhs.end(),
+                                      rhs.begin(),
+                                      rhs.end());
```
digit-google: nit: There is no guarantee that the `std::lexicographical_compare()` implementation will be efficient compared to the use of `std::min()` and `::memcmp()`, which can use optimized SIMD instructions. See https://godbolt.org/z/4xcn46rGM

jhasse:
Not sure if we should put in this work and not directly go to `std::string_view`.

digit-google:
The benefit of this great PR is to improve performance, possibly quite significantly, by removing temporary string buffer heap allocations. I.e. it does change the generated code significantly.

Switching to `std::string_view` may require a massive change. It's a good idea but I recommend to do that in a different PR that does *not* change behavior, as this will be much easier to review for correctness.

Also have we switched to C++17 yet (a requirement for std::string_view)? I thought C++14 was still the required minimum at least for the base Ninja code (tests are already on C++17 due to GTest). Please correct me if I'm wrong.

src/string_piece_util.h
elliotgoodrich: Yes, missed this one.

src/string_piece_util.cc
elliotgoodrich: In my experience the standard libraries do a great job of falling back to the fast C methods and I would have argued this.  However, you're right about GCC! Even though GCC can fallback to memcmp - they fail to do so for `char` because it's not unsigned https://godbolt.org/z/4s8jhWE7T :(

This has got to be a bug as their `std::char_traits<char>::compare` does use it and the implementation of `std::lexicographical_compare` for MSVC and Clang use `memcmp`.

I'll file a bug and swap this to `memcmp` in the meantime.

elliotgoodrich: This isn't a bug. Compilers can only use memcmp in this case when `char` is signed on the platform (swap your godbolt to GCC ARM and it'll use memcmp) as they differ when the value of char is negative.

Learn something new every day!

digit-google: Oh right. Nice catch. Thanks for the reminder!

In the context of Ninja, our strings are UTF-8 so `std::memcmp()` probably makes more sense then (it always compares bytes as unsigned char values). I'd suggest a comment in the definition of the struct to clarify that it does utf8 code unit (i.e. unsigned char) comparisons.

Also means that when switching to std::string_view, we might want to keep this custom comparator instead of relying on std::less<>.

digit-google:
Thanks Elliot, latest version seems great (feel free to add the comment referenced above or not), @jhasse would you consider approving this?

src/string_piece_util.cc
elliotgoodrich: @digit-google Sure thing. I've added an attempt at documenting this and pushed it in the commit. Let me know if you want any revisions.

```cpp
/// A comparator for StringPiece that can be used in
/// `std::map<std::string, V, StringPieceLess>` that will allow using a
/// StringPiece as a key.  Note that this compares elements of the
/// strings as `unsigned char`, which means that implementations that
/// have a signed `char` may give an unexpected (yet valid) ordering
/// when iterating through a `std::map` containing negative `char`s.
```

elliotgoodrich:
@digit-google / @jhasse Is this commit okay to review and land? If not, is there someone else who I could ask to review? I would like to submit the followup changes from https://github.com/ninja-build/ninja/pull/2696 to speed up `EvalString` and reduce the number of allocations it need.

digit-google:
Hi @elliotgoodrich , sorry if I don't have much time for reviewing Ninja PRs these days, but this one still looks good to me :) @jhasse  please take a look.

jhasse:
Thanks to you both :)

jhasse:
damn, the CI fails after the merge. Can you have a look @elliotgoodrich and create a PR that fixes it so that I don't have to revert this?

elliotgoodrich:
@jhasse I have a fix https://github.com/ninja-build/ninja/pull/2809 that should work.

Thank you very much @jhasse and @digit-google for all the reviews. I will push the remaining optimization for `EvalString` :)

If your CI works with Github's merge queues, it could be worth turning them on to avoid PR clashes.

PR #2718, opened 2026-01-23
Add a small speed up to `PathDecanonicalized`

elliotgoodrich:
Avoid a mandatory allocation in `PathDecanonicalized` by taking an output `std::string*` to append to. Improve its performance by swapping `strchr` with `std::string::find`, which can take advantage of knowing the length of the string.

We can also return as soon as we run out of forwardslashes that need to be turned into backslashes, instead of iterating through all forwardslashes.

If backslashes are the common path separator in ninja build files on Windows, it may be worthwhile canonicalizing paths into backslashes-only, which would now allow us to skip the loop entirely in `PathDecanonicalized` in the case of paths with no forwardslashes.

Rename this function to `AppendPathDecanonicalized` to reflect its new functionality.

Running `manifest_parser_perftest.exe` on Windows gives initial timings of:

    min 549ms  max 613ms  avg 573.1ms

and timings after this change of:

    min 534ms  max 597ms  avg 559.9ms

which is a 2-3% improvement.

### DIFF
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -484,18 +484,20 @@ string EdgeEnv::LookupVariable(const string& var) {
 std::string EdgeEnv::MakePathList(const Node* const* const span,
                                   const size_t size, const char sep) const {
   string result;
+  std::string temp_path;
   for (const Node* const* i = span; i != span + size; ++i) {
     if (!result.empty())
       result.push_back(sep);
-    const string& path = (*i)->PathDecanonicalized();
     if (escape_in_out_ == kShellEscape) {
+      temp_path.clear();
+      (*i)->AppendPathDecanonicalized(&temp_path);
 #ifdef _WIN32
-      GetWin32EscapedString(path, &result);
+      GetWin32EscapedString(temp_path, &result);
 #else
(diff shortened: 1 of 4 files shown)


src/graph.cc
```diff
@@ -580,18 +582,29 @@ bool Edge::maybe_phonycycle_diagnostic() const {
...
-  for (char* c = &result[0]; (c = strchr(c, '/')) != NULL;) {
-    if (slash_bits & mask)
+  char* data = &(*output)[0];
+  char* end = data + output->size();
+  char* c = end - path.size();
+  while (slash_bits != 0) {
+    c = static_cast<char*>(memchr(c, '/', end - c));
+
+    // If `memchr` returns null then there are too many bits in `slash_bits`,
+    // which happens when we pass `~0` in `build.cc` instead of calculating it.
```
elliotgoodrich: Found https://github.com/ninja-build/ninja/blob/cc60300ab94dae9bb28fece3c9b7c397235b17de/src/build.cc#L997-L1001 the other day and added this in to avoid problems until `slash_bits` is calculated correctly here.

PR #2722, opened 2026-02-07
Add `StringPiece::substr` method

elliotgoodrich:
Add `StringPiece::substr` with almost identical behavior as
`std::string_view::substr`.   The only difference is what happens when
`pos > size()`, which is `StringPiece::substr` will call `Fatal` and
`std::string_view::substr` will throw or terminate (depending on whether
exceptions are enabled).

In ninja, where exceptions are disabled, these methods are basically identical and it should be no harder to migrate `StringPiece` to `std::string_view` after this.

`operator==` was moved to be a free function so that we can write both

```cpp
StringPiece("abc") == "abc";
"abc" == StringPiece("abc");
```

as currently we can't write the second line because we can't implicitly convert `"abc"` to a `StringPiece`.

This changes is needed for future changes to reduce the number of `std::string` copies for performance reasons.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -291,6 +291,7 @@ if(BUILD_TESTING)
     src/missing_deps_test.cc
     src/ninja_test.cc
     src/state_test.cc
+    src/string_piece_test.cc
     src/string_piece_util_test.cc
     src/subprocess_test.cc
     src/test.cc
--- a/configure.py
+++ b/configure.py
@@ -660,6 +660,7 @@ if gtest_src_dir:
         'manifest_parser_test',
         'ninja_test',
         'state_test',
+        'string_piece_util',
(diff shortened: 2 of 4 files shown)


elliotgoodrich:
This has been pulled out of https://github.com/ninja-build/ninja/pull/2706

digit-google:
Thanks, this looks good to me, but can you add some unit-tests for it for good measure? No need to test the Fatal() case in `substr()` here.

elliotgoodrich:
> Thanks, this looks good to me, but can you add some unit-tests for it for good measure? No need to test the Fatal() case in `substr()` here.

No problem @digit-google - done!

digit-google (approved):
Thank you, @jhasse can you take a look!

PR #2723, opened 2026-02-07
Fix includes normalize win32

lygstate:
Replace https://github.com/ninja-build/ninja/pull/2552

### DIFF
--- a/src/includes_normalize-win32.cc
+++ b/src/includes_normalize-win32.cc
@@ -14,28 +14,60 @@
 
 #include "includes_normalize.h"
 
-#include "string_piece.h"
-#include "string_piece_util.h"
-#include "util.h"
+#include <string.h>
+
+#include <windows.h>
 
 #include <algorithm>
 #include <iterator>
 #include <sstream>
 
-#include <windows.h>
+#include "string_piece.h"
+#include "string_piece_util.h"
+#include "util.h"
 
 namespace {
 
+constexpr const std::size_t MAX_PATH_LONG = 32767;
+
 bool InternalGetFullPathName(const StringPiece& file_name, char* buffer,
                              size_t buffer_length, std::string* err) {
-  DWORD result_size = GetFullPathNameA(file_name.AsString().c_str(),
-                                       buffer_length, buffer, NULL);
-  if (result_size == 0) {
-    *err = "GetFullPathNameA(" + file_name.AsString() + "): " +
-        GetLastErrorString();
+  // IMPORTANT: Using GetFullPathNameA() with a long paths will fail with
+  // "The filename or extension is too long" even if long path supported is
+  // enabled. GetFullPathNameW() must be used for this function to work!
+  buffer[0] = '\0';
+  // Convert to wide filename first.
+  std::wstring wide_filename;
+  if (!ConvertUTF8ToWin32Unicode(file_name, &wide_filename, err))
     return false;
-  } else if (result_size > buffer_length) {
-    *err = "path too long";
+
+  // Call GetFullPathNameW()
+  DWORD wide_full_size = GetFullPathNameW(wide_filename.c_str(), 0, NULL, NULL);
+  if (wide_full_size == 0) {
+    *err = "GetFullPathNameW(" +
+           std::string(wide_filename.begin(), wide_filename.end()) +
+           "): " + GetLastErrorString();
+    return false;
+  }
+
+  // NOTE: wide_full_size includes the null-terminating character.
+  std::wstring wide_path;
+  wide_path.resize(static_cast<size_t>(wide_full_size - 1));
+  DWORD wide_full_size2 =
+      GetFullPathNameW(wide_filename.c_str(), wide_full_size,
+                       const_cast<wchar_t*>(wide_path.data()), NULL);
+  if (wide_full_size2 == 0) {
+    *err = "GetFullPathNameW(" + file_name.AsString() + "): " + GetLastErrorString();
+    return false;
+  }
+
+  // Convert wide_path to Unicode.
+  int utf8_size =
+      WideCharToMultiByte(CP_UTF8, 0, wide_path.c_str(), wide_path.size() + 1,
+                          buffer, buffer_length, NULL, NULL);
+  if (utf8_size <= 0 || utf8_size >= buffer_length) {
+    *err = "WideCharToMultiByte(" + std::string(wide_path.begin(), wide_path.end()) +
+           "): " + GetLastErrorString();
     return false;
   }
   return true;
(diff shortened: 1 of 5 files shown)


src/includes_normalize-win32.cc
```diff
@@ -12,34 +12,65 @@
...
+                             size_t buffer_length, std::string* err) {
+  // IMPORTANT: Using GetFullPathNameA() with a long paths will fail with
+  // "The filename or extension is too long" even if long path supported is
+  // enabled. GetFullPathNameW() must be used for this function to work!
+  buffer[0] = '\0';
+  // Convert to wide filename first.
+  std::string filename_str = file_name.AsString();
+  std::wstring wide_filename;
+  if (!ConvertUTF8ToWin32Unicode(filename_str, &wide_filename, err))
     return false;
```
elliotgoodrich: You can avoid creating `filename_str` if `ConvertUTF8ToWin32Unicode` took a `StringPiece` instead of a `std::string`.

src/includes_normalize-win32.cc
```diff
@@ -12,34 +12,65 @@
...
-#include <windows.h>
+#include "includes_normalize.h"
+#include "string_piece.h"
+#include "string_piece_util.h"
+#include "util.h"
 
-using namespace std;
+#define _MAX_PATH_LONG 32767
 
 namespace {
```
elliotgoodrich: Identifiers starting with underscore and followed with a capital letter are reserved for the compiler.

```suggestion
namespace {

constexpr const std::size_t MAX_PATH_LONG = 32767;
```

src/includes_normalize-win32.cc
```diff
@@ -12,34 +12,65 @@
...
 
 #include <algorithm>
 #include <iterator>
 #include <sstream>
 
-#include <windows.h>
+#include "includes_normalize.h"
+#include "string_piece.h"
+#include "string_piece_util.h"
+#include "util.h"
```
elliotgoodrich, line 27: [Related Header](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) to come first for include directives.

```suggestion
#include "includes_normalize.h"

#include <string.h>
#include <windows.h>

#include <algorithm>
#include <iterator>
#include <sstream>

#include "string_piece.h"
#include "string_piece_util.h"
#include "util.h"
```

src/includes_normalize_test.cc
```diff
@@ -138,33 +137,34 @@ TEST(IncludesNormalize, LongInvalidPath) {
...
+  std::string forward_slashes(kExactlyMaxPath);
   replace(forward_slashes.begin(), forward_slashes.end(), '\\', '/');
   // Make sure a path that's exactly _MAX_PATH long is canonicalized.
   EXPECT_EQ(forward_slashes.substr(cwd_len + 1),
             NormalizeAndCheckNoError(kExactlyMaxPath));
 }
 
-TEST(IncludesNormalize, ShortRelativeButTooLongAbsolutePath) {
-  string result, err;
+TEST(IncludesNormalize, ShortRelativeButLongAbsolutePath) {
```
elliotgoodrich: Should at least one of these tests check for paths over `_MAX_PATH_LONG`?

src/includes_normalize-win32.cc
```diff
@@ -12,34 +12,65 @@
...
+                          buffer, buffer_length, NULL, NULL);
+  if (utf8_size <= 0 || utf8_size >= buffer_length) {
+    *err = "WideCharToMultiByte(" + std::string(wide_path.begin(), wide_path.end()) +
+           "): " + GetLastErrorString();
+    buffer[0] = '\0';
     return false;
   }
+  buffer[utf8_size] = '\0';
   return true;
 }
```
elliotgoodrich, line 74: Minor: You could let `WideCharToMultiByte` null terminate the output for you by passing the terminating null character in `wide_path`


```suggestion
  int utf8_size =
      WideCharToMultiByte(CP_UTF8, 0, wide_path.c_str(), wide_path.size() + 1,
                          buffer, buffer_length, NULL, NULL);
  if (utf8_size <= 0 || utf8_size >= buffer_length) {
    *err = "WideCharToMultiByte(" + std::string(wide_path.begin(), wide_path.end()) +
           "): " + GetLastErrorString();
return false;
}
return true;
}
```

lygstate:
cc @jhasse

src/includes_normalize-win32.cc
```diff
@@ -14,30 +14,60 @@
...
+  if (wide_full_size == 0) {
+    *err = "GetFullPathNameW(" +
+           std::string(wide_filename.begin(), wide_filename.end()) +
+           "): " + GetLastErrorString();
+    return false;
+  }
+
+  // NOTE: wide_full_size includes the null-terminating character.
+  std::wstring wide_path;
+  wide_path.resize(static_cast<size_t>(wide_full_size - 1));
```
zufuliu, line 55: I think `-1` can be omitted, it's not bad to include extra NUL at the end.
here is no need to use `static_cast` for widen unsigned conversion (`DWORD/unsigned long` => `size_t`).
<del>The document for `GetFullPathNameW` doesn't say `[out] LPWSTR  lpBuffer` parameter can be `NULL`, needs test to confirm.</del>
https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfullpathnamew
Edit: the function will check `[in]  DWORD   nBufferLength` parameter (zero in the call) first.

src/includes_normalize-win32.cc
```diff
@@ -14,30 +14,60 @@
...
+  DWORD wide_full_size2 =
+      GetFullPathNameW(wide_filename.c_str(), wide_full_size,
+                       const_cast<wchar_t*>(wide_path.data()), NULL);
```
zufuliu, line 58: I think `const_cast<wchar_t*>` is not need.

src/includes_normalize-win32.cc
lygstate: You mis-understading the size parameter of std::wstring::resize, the `std::wstring` always have NUL at the end. but the `size` parameter would not count it.

zufuliu: Yes, but it does not harm to have extra NUL at string end, `wide_path.resize(wide_full_size)` is simple.

I come up with an optimization idea, but not sure whether worth it.
1. init `std::wstring wide_path(MAX_PATH, 0);`,  `MAX_PATH` should works most of time, then do first `wide_full_size = GetFullPathNameW()` call.
2. if result `wide_full_size` large than `MAX_PATH`, resize `wide_path.resize(wide_full_size)`, then do second `wide_full_size = GetFullPathNameW()` call.
3. if `wide_full_size` is zero, store error message.

elliotgoodrich: There is a benchmark of this functionality at `clparser_perftest.exe` that you can use.

Also FYI, I have a [PR](https://github.com/ninja-build/ninja/pull/2725) to optimize the performance of `IncludesNormalize` that will probably conflict with this PR. If this lands first I don't mind rebasing mine and looking at the performance of it.

PR #2726, opened 2026-02-10
Make full use of parallelism in jobserver mode

### LINKED ISSUE #2683: Not making full use of parallelism in jobserver mode
When ninja runs as a jobserver client, it first starts as many commands as tokens it can acquire and then waits for the first pending command to finish. Only then will it again try to acquire tokens and start more commands.

If the jobserver has multiple clients (could be multiple ninja processes or a mix of build systems), it's possible that a token becomes available before the next pending command (of a given ninja process) finishes. In the time between the token becoming available and the next command finishing, the parallelism set by the jobserver is not fully utilized.

@digit-google Is this a known limitation? Are there any plans to improve this? I'm not familiar with the ninja code base, but it may be possible to add the jobserver FIFO to the `ppoll()` or `pselect()` call to allow reacting when a jobserver token becomes available.

jamesl-10:
Fixes: #2683. Also mentioned in the #1139 - specifically [here](https://github.com/ninja-build/ninja/issues/1139#issuecomment-2389078694).

This change addresses the Ninja client's inability to react to jobserver tokens becoming available - in particular, in `SubprocessSet::DoWork()`. The client will watch the jobserver for token if it has more work planned and react accordingly. This is accomplished by adding the POSIX FIFO's file descriptor to the set of file descriptors listened to in `ppoll() / pselect()`.


src/build.cc
```diff
@@ -769,6 +769,10 @@ ExitStatus Builder::Build(string* err) {
     // See if we can reap any finished commands.
     if (pending_commands) {
       CommandRunner::Result result;
+
+      // Tell command runner that if jobserver tokens become available while
+      // waiting, it should notify us - but only if we have more work to do.
+      command_runner_->WatchForJobserverTokens(plan_.work_ready());
       if (!command_runner_->WaitForCommand(&result) ||
```
juergbi: This separate `WatchForJobserverTokens()` method that is called exactly once before each `WaitForCommand()` call to affect the behavior of the latter seems a bit odd to me. Wouldn't it be possible and more straight forward to add a boolean (or maybe flags) parameter to `WaitForCommand()` instead?

src/subprocess-posix.cc
```diff
@@ -280,12 +280,15 @@ SubprocessSet::SubprocessSet() {
 
 // Reaps console processes that have exited and moves them from the running set
 // to the finished set.
-void SubprocessSet::CheckConsoleProcessTerminated() {
+void SubprocessSet::CheckConsoleProcessTerminated(WorkStatus* workStatus) {
```
digit-google: nit: use snake_case (i.e. `work_status`) for variable names to be consistent with the rest of the code.

src/subprocess.h
```diff
@@ -102,9 +102,18 @@ struct SubprocessSet {
...
+  struct WorkStatus {
+    bool subproc_finished = false;             // A subprocess has finished
+    bool jobserver_token_available = false;    // A jobserver token is available
+
+    bool finished() const { return subproc_finished || jobserver_token_available; }
+  };
+
   Subprocess* Add(const std::string& command, bool use_console = false);
-  bool DoWork();
+  bool DoWork(WorkStatus* workStatus = nullptr);
```
digit-google: Please fix the definitions in subprocess-win32.cc to unbreak the Windows build.

src/build.cc
digit-google: Good point, modifying the WaitForCommand() base method to take an additional default-initialized `watch_jobserver` parameter would make this code more readable. Alternatively, add another virtual method named WaitForCommandOrJobserverToken() in the base class whose default implementation simply calls WaitForCommand() and ignores the second parameter, and which can be overriden only in RealCommandRunner. This should minimize the changes to the rest of the code (e.g. the DryrunCommandRunner and some unit-tests).

digit-google (commented):
Thank you, this is really an excellent change. Apart from the review nits, can you rebase your PR into a series of small clean commits. E.g. one for the jobserver.h and subprocess.h changes, another one for the build.cc and real_command_runner.cc changes + the final regression test.

src/build.cc
jamesl-10: Yep this was one of the points I was a bit iffy on - changed it to have the `WaitForCommandOrJobserverToken()` virtual method in the base class.

src/subprocess.h
jamesl-10: Sounds good. One thing I want to ask - should we set the `work_status` even though it's not strictly used in the Windows build? Or should we just void it out for now?

juergbi: Given that `SubprocessSet` already keeps state in the instance, e.g., the queue of finished subprocesses, might it make more sense to move the `jobserver_token_available` flag into the `SubprocessSet` struct, eliminating the need for the `WorkStatus` struct, or am I missing something?

And then `!HasFinished() && !jobserver_token_available` would be the loop condition in `WaitForCommandOrJobserverToken()`, which could also replace the separate `if (subprocs_.HasFinished())` before the loop, as far as I can tell.

And we could then hopefully also limit the preprocessor conditional in `WaitForCommandOrJobserverToken()` to setting the jobserver fd.

src/subprocess-posix.cc
```diff
@@ -362,6 +372,12 @@ bool SubprocessSet::DoWork() {
   if (IsInterrupted())
     return true;
 
+  if (jobserver_fd_ >= 0) {
+    if (fds.back().revents && work_status) {
```
juergbi: `fds.back()` seems a bit fragile to me. I think it would be more robust to have a local variable that stores the index of the jobserver `pollfd` in the `fds` vector.

src/build.h
digit-google: For clarity, can you document the valid states for this struct, now that it became non-trivial, e.g:

- Command completed (edge != nullptr, status is one of ExitSuccess / ExitFailure, and output is the comand's output).
- User interruption (edge == null, status == ExitInterrupted)
- Jobserver Token available (edge == null, status == ExitSuccess)

Using an enum for the state might be a win for clarity, dont' know what @jhasse thinks about this.

src/subprocess-posix.cc
digit-google: Agreed, while it has to be at the end of the vector, otherwise the assertion on line 391 will fail, this is subtle (and should be documented in a comment I think). A variable to hold the index might be nice too.

src/subprocess.h
digit-google: Things would be considerably simpler if DoWork() simply returned an enum that indicates the reason for its exit (e.g. NoWorkToDo / CommandCompleted / UserInterruption / JobserverTokenAvailable). I also don't see a good need for the preprocessor conditionals in build.cc apart from setting the Jobserver file descriptor which is Posix specific.

src/build.h
jhasse: An algebraic type would be best. They aren't that nice in C++ though, i.e. you have to `struct Foo {};` the different states and the use a `std::variant`. I've used that for my work on using file hashes in addition to mtimes, PR coming up soon.

jhasse:
Can you update the title and the description to make it clearer what this PR is about?

src/subprocess-posix.cc
jamesl-10: Addressed this, please take a look at the new changes!

src/subprocess.h
jamesl-10: Addressed this, please take a look at the new changes!

One thing to note is that I made it a bitmap enum, since any one of CommandCompleted / UserInterruption / JobserverTokenAvailable can happen. I preferred this approach since that would allow us to have more granularity in the `RealCommandRunner` (or any caller of `SubprocessSet::DoWork()` really) in handling the different outcomes. For example in `WaitForCommand`, I use this to explicitly prioritize handling interrupts first, then reaping subprocesses, then finally jobserver token availability. Open to any alternative suggestions though!

src/build.h
jamesl-10: Just a quick update on this - I agree that using algebraic types is probably the most sound, and I have done a fair bit of investigative / experimental work towards this approach. However, the change is a bit more involved than I initially thought, since the other runners (i.e. `FakeCommandRunner` and `DryRunCommandRunner`) will need to conform to the new `std::variant` logic / states as well. It should be doable, but the enum approach would be easier to implement and will clearly convey the states for the runner.

Do we still want to move forward with ADT approach? I'm happy to proceed with it if we think it's worth the effort. And yes, I agree that the algebraic types - and in general, FP in C++ - is not as clean as other languages ;)

jamesl-10: Just to circle back on this - is `std::variant` supported? The CI is failing to build - maybe C++17 is not supported for Ninja?

Also, I was able to build / run the WIP commit with CMake locally, but not with the Ninja bootstrap, for reference.

digit-google: Ninja is still building with C++14, so no std::variant<> template is usable for now. For this simple case, I suggest just creating a custom type with an enum to distinguish the different result types. Something like:

```
struct Result {
  enum {
    kProcessCompleted = 0,
    kUserInterruption,
    kJobserverTokenAvailable,
  } Variant;
  
  Variant variant() const { return variant_ };
  
  bool has_process_completed() const { return variant_ == kProcessCompleted; }
  bool has_user_interruption() const { return variant_ == kUserInterruption; }
  bool has_job_token_available() const { return variant_ == kJobTokenAvailable; }

  // The result of a process completion.
  struct ProcessCompleted {
    ExitStatus status = ExitSuccess;
    std::string output;
    Edge* edge = nullptr;
  };

  // Return ProcessCompleted value, only valid if has_process_completed() returns true.  
  const ProcessCompleted& process_completed() const {
    assert(variant_ == kProcessCompleted);
    return completed_;
  }
  
  // Static helpers to create new instances.
  static Result MakeProcessCompleted(ExitStatus, std::string, Edge*);
  static Result MakeUserInterruption();
  static Result MakeJobserverTokenAvailable();
    
 private:
  Variant variant_ = kProcessCompleted;
  
  // Avoid a union here to save writing explicit destructor / copy / move operators.
  ProcessCompleted completed_;  // Only valid if state_ == kProcessCompleted.
};
```

jhasse: > Just to circle back on this - is std::variant supported? The CI is failing to build - maybe C++17 is not supported for Ninja?

Not yet, I will update them.

src/build_result.h
```diff
@@ -0,0 +1,97 @@
...
+  constexpr inline bool interrupted() const {
+    return std::holds_alternative<Interrupted>(state_);
+  }
+
+  constexpr inline bool jobserver_token_available() const {
+    return std::holds_alternative<JobserverTokenAvailable>(state_);
+  }
+
+  constexpr inline bool command_completed() const {
+    return std::holds_alternative<JobserverTokenAvailable>(state_);
```
juergbi:
```suggestion
    return std::holds_alternative<CommandCompleted,>(state_);
```

src/build.cc
```diff
@@ -761,40 +762,56 @@ ExitStatus Builder::Build(string* err) {
...
-        }
-        return result.status;
+      if (result.jobserver_token_available()) {
+        // Note: currently we only react to jobserver tokens availability
+        // on non-Windows platforms.
+
+        // Jobserver token is available; start main loop over to try to
+        // acquire jobserver token.
+        continue;
       }
```
juergbi: In case a jobserver token is available and a command has completed, don't we want to process the command completion before we continue the loop? I.e., don't we want to move these lines further down?

digit-google: Currently only one condition can be returned at a time, which keeps everything simple. There is imho very little value in returning multiple event flags in the result (and handling them properly in the caller).

src/build.h
digit-google: > 
> Not yet, I will update them.

That's great news, please let us know as soon as possible :)

jamesl-10: > Ninja is still building with C++14, so no std::variant<> template is usable for now. For this simple case, I suggest just creating a custom type with an enum to distinguish the different result types. Something like:
> ...

Yeah this was what I was initially thinking as well. But since C++17 support should be landing soon, should I wait on that for now?

src/build.cc
```diff
@@ -761,27 +762,47 @@ ExitStatus Builder::Build(string* err) {
...
-      CommandRunner::Result result;
-
       // Tell command runner that if jobserver tokens become available while
       // waiting, it should notify us - but only if we have more work to do.
       const bool watch_jobserver = plan_.more_to_do();
-      if (!command_runner_->WaitForCommandOrJobserverToken(&result, watch_jobserver) ||
-          result.status == ExitInterrupted) {
+      BuildResult result =
+          command_runner_->WaitForCommandOrJobserverToken(watch_jobserver);
+      if (result.interrupted() || result.finished()) {
```
digit-google: It seems `result.finished()` here would be impossible since `pending_commands` was true. Maybe call Fatal("internal error") instead if this condition is detected?

src/build.cc
```diff
@@ -100,19 +101,19 @@ bool Plan::AddSubTarget(const Node* node, const Node* dependent, string* err,
...
-     // since there is no producing edge to add to the plan.
-     if (node->dirty() && !node->generated_by_dep_loader()) {
-       string referenced;
-       if (dependent)
-         referenced = ", needed by '" + dependent->path() + "',";
-       *err = "'" + node->path() + "'" + referenced +
-              " missing and no known rule to make it";
-     }
-     return false;
+    // Leaf node, this can be either a regular input from the manifest
```
digit-google: nit: Please do not reformat unrelated parts of the sources, this makes reviews more difficult. Use `git clang-format` to only reformat lines near your changes instead.

src/build_result.h
```diff
@@ -0,0 +1,115 @@
...
+struct BuildResult {
+  /// Command was completed with varying exit statuses
+  struct CommandCompleted {
+    Edge* edge;
+    ExitStatus status;
+    std::string output;
+
+    explicit CommandCompleted(Edge* edge, ExitStatus& status,
+                              std::string& output)
+        : edge{ edge }, status{ status }, output{ output } {}
```
digit-google: nit: to be consistent with the rest of the code, use parentheses for initializers, e.g. `field(field)`. Also make the `output` parameter an `std::string` or `std::string&&` to avoid needless copies.

src/build_result.h
```diff
@@ -0,0 +1,115 @@
...
+
+#include "exit_status.h"
+
+class Edge;
+
+/// Stores the result of executing a build command
+struct BuildResult {
+  /// Command was completed with varying exit statuses
+  struct CommandCompleted {
+    Edge* edge;
```
digit-google: nit: use default initializers, even if there is no default constructor yet.

src/build_result.h
```diff
@@ -0,0 +1,115 @@
...
+
+  constexpr inline bool jobserver_token_available() const {
+    return std::holds_alternative<JobserverTokenAvailable>(state_);
+  }
+
+  constexpr inline bool command_completed() const {
+    return std::holds_alternative<CommandCompleted>(state_);
+  }
+
+  constexpr inline ExitStatus status() const {
```
digit-google: nit: Call this `exit_status()` for clarity, especially since there is a Status class and `status` objects in other parts of the code base. Also document what this method does since the return values are not obvious based on the state.

src/build.cc
digit-google: Another alternative is to introduce a commit that only reformats existing files of interests, then add your changes to commits on top of that. Both makes reviews easier :)

src/build.cc
```diff
@@ -790,26 +811,6 @@ ExitStatus Builder::Build(string* err) {
         continue;
       }
```
digit-google, line 816: I recommend making an `else` branch here calling `Fatal("internal bug: unexpected BuildResult state")` to catch regressions.

src/build.cc
```diff
@@ -994,11 +996,9 @@ bool Builder::FinishCommand(CommandRunner::Result* result, string* err) {
   return true;
 }
 
-bool Builder::ExtractDeps(CommandRunner::Result* result,
-                          const string& deps_type,
-                          const string& deps_prefix,
-                          vector<Node*>* deps_nodes,
-                          string* err) {
+bool Builder::ExtractDeps(BuildResult::CommandCompleted* result,
```
digit-google: I recommend taking a mutable reference here and explaining in the function's documentation that it may modify result.output, for example to extract msvc dependency messages out of it.

src/build_result.h
```diff
@@ -0,0 +1,115 @@
...
+      return ExitInterrupted;
+    if (finished())
+      return ExitSuccess;
+
+    return ExitFailure;
+  }
+
+  bool success() const { return status() == ExitSuccess; }
+
+  CommandCompleted* GetCommand() {
```
digit-google: nit: Rename to GetCommandCompleted() for clarity and return a mutable reference or a new instance (by moving the content from the BuildResult value). Also please document that it is a runtime/fatal error to use this if command_completed() is false.

src/build_result.h
```diff
@@ -0,0 +1,115 @@
...
+    if (jobserver_token_available())
+      return ExitSuccess;
+    if (interrupted())
+      return ExitInterrupted;
+    if (finished())
+      return ExitSuccess;
+
+    return ExitFailure;
+  }
+
```
digit-google, line 115: nit: Similarly, document this function for the same reason as exit_status().

digit-google (approved):
Thanks, apart from the nits, this looks great and makes the code base easier to understand :) cc @jhasse

configure.py
```diff
@@ -366,7 +366,7 @@ def search_system_path(file_name: str) -> Optional[str]:  # type: ignore # Missi
               '-Wno-unused-parameter',
               '-fno-rtti',
               '-fno-exceptions',
-              '-std=c++14',
+              '-std=c++17',
```
digit-google, line 370: You will also need to modify CMakeLists.txt to get this to work with CI

src/build.cc
jamesl-10: Oh my bad, I think I ran `clang-format` and manually reverted the changes - guess I didn't catch all of it 😅

jamesl-10:
Thank you for all the comments @digit-google! I'll address the nits soon (pretty late where I am right now, will work more on this tomorrow).

One thing I wanted to clarify is that we shouldn't care about implicit jobserver slots being released, correct? I.e. in regards to the `SubprocessSet::DoWork` behavior. I figured no, since these don't actually claim any tokens from the FIFO. The reason for asking is that the current regression test is not fully testing these changes, since rather than claiming a FIFO token, I think it was acquiring implicit slots and thus the `ppoll / pselect` was not getting covered.

digit-google:
> Thank you for all the comments @digit-google! I'll address the nits soon (pretty late where I am right now, will work more on this tomorrow).

No problem, thanks again :)

> One thing I wanted to clarify is that we shouldn't care about implicit jobserver slots being released, correct?

Yes, there is only one implicit slot, that can be acquired / released by a single Subprocess instance globally, and that has nothing to do with the FIFO.

> I.e. in regards to the `SubprocessSet::DoWork` behavior. I figured no, since these don't actually claim any tokens from the FIFO. The reason for asking is that the current regression test is not fully testing these changes, since rather than claiming a FIFO token, I think it was acquiring implicit slots and thus the `ppoll / pselect` was not getting covered.

If you launch more than one subprocess in parallel, the non-first ones will acquire a real token. Maybe we need a unit test for SubprocessSet to verify that the jobserver_token_available() condition is properly returned then? Wdyt?

src/build.cc
jamesl-10: Referencing the original code - it seems `DryRunCommandRunner::WaitForCommand` returns false if the `finished_` queue is empty. Then, in the build loop, it checks the condition `!command_runner_->WaitForCommand(&result)` or `result.status == ExitInterrupted`, which are basically the same event in the `RealCommandRunner` case (seemingly redundant). But I'm wondering if it's not **actually** redundant, but handling the case where we use `DryRunCommandRunner` and return false?

I changed `DryRunCommandRunner::WaitForCommand` to return `BuildResult::Finished` if `finished_.empty()`. So if the above is true, then the logic should stay the same.

jamesl-10: Although it seems that it's more likely just a redundancy, since we should always start a command before waiting on it

jamesl-10: Coming back to this - so apparently, this was actually not a redundant check. I think there was some edge case interaction in `Subprocess::Finish() / Subprocess::TryFinish()` where, if I am understanding correctly, we "save" the exit status of some child subprocess (see `exit_status_` member variable) so that we can reap them later. But technically, these forgotten child processes would count as a "finished command", so `WaitForCommand` would return true - BUT it would exit with `ExitInterrupted` as the status.

Incidentally, this is actually also something that the current changes don't catch either (I assumed that we would always have the `Interrupted` struct returned in case of user interruptions) which is causing the CI to fail. I verified that indeed there are cases where the result is a `CommandCompleted`, but has `ExitInterrupted`. Will amend that in the next update. Not sure if it's worth investigating always returning an `Interrupted` struct? The interactions seem a bit complex.

jamesl-10:
> If you launch more than one subprocess in parallel, the non-first ones will acquire a real token. Maybe we need a unit test for SubprocessSet to verify that the jobserver_token_available() condition is properly returned then? Wdyt?

For now, I made a small unit test where we simulate the jobserver fd using a pipe, and simply verify that writing to the pipe will return `SubprocessSet::JobserverTokenAvailable` in `DoWork()`. I'm not sure if we still want to have the python test? Since it will test more end-to-end behaviour.

digit-google:
> For now, I made a small unit test where we simulate the jobserver fd using a pipe, and simply verify that writing to the pipe will return `SubprocessSet::JobserverTokenAvailable` in `DoWork()`.

This is perfectly appropriate, after all we only need to test that a token on the FIFO triggers exactly this behavior :)

>  I'm not sure if we still want to have the python test? Since it will test more end-to-end behaviour.
It's good to have both, there is a ton of plumbing between `SubprocessSet::DoWork()` and the final way commands are scheduled, so having a regression test here is a great thing too. Thanks again.

digit-google (approved):
LGTM, cc @jhasse

jamesl-10:
@digit-google @jhasse rebased this PR to the master branch to capture the C++17 PR upgrade and dropped the commit in this PR. Also added a small fix - I noticed the builds sometimes hang when using the jobserver mode, and I realized it's because in `build.cc`, we have `const bool watch_jobserver = plan_.more_to_do()`. Changing this to  `const bool watch_jobserver = plan_.work_ready()` seems to have fixed it.

I suspect this happens when we do have more work to do, but none that is currently queued up in `ready_` until we unblock it by completing the command. In any case, it makes more sense anyways to change this flag to `work_ready()` anyways (in fact, I thought that was already in the code, but must've reverted it at some point and missed it)

src/real_command_runner.cc
```diff
@@ -35,6 +37,7 @@ struct RealCommandRunner : public CommandRunner {
     }
   }
 
+  bool watch_for_jobserver_tokens_ = false;
```
jamesl-10: Nit: unused

src/build_result.h
```diff
@@ -0,0 +1,126 @@
...
+      return cc->status;
+    if (jobserver_token_available())
+      return ExitSuccess;
+    if (interrupted())
+      return ExitInterrupted;
+    if (finished())
+      return ExitSuccess;
+
+    return ExitFailure;
+  }
```
jamesl-10, line 114: Should I change this to use `std::visit`? Might be useful to validate that we hit all cases of the variant

src/build_result.h
```diff
@@ -0,0 +1,126 @@
...
+    ExitStatus status = ExitSuccess;
+  };
+
+  // Store the actual state as an algebraic data type
+  std::variant<
+    std::monostate,
+    CommandCompleted,
+    JobserverTokenAvailable,
+    Interrupted,
+    Finished> state_;
```
jamesl-10, line 66: Should `state_` be allowed to have the `std::monostate`?

misc/jobserver_test.py
```diff
@@ -130,6 +150,53 @@ def generate_build_plan(command_count: int) -> str:
...
+        wait_2_ms: int = 4000,
+        offset: int = 0,
+    ) -> str:
+
+    """Generate a Ninja build plan where each client gets one short and one long task.
+
+    - `id` identifies the client for debugging purposes.
+    - `wait_1_ms` and `wait_2_ms` control task durations.
+    - `offset` allows multiple plans without conflicting filenames.
+    """
```
jamesl-10: Nit: should probably update this comment

src/build.h
jamesl-10: Resolving since C++17 has landed in #2734.

src/build_result.h
digit-google: I recommend keeping things simple for now.

While this is a nice idea, the non-auto uses of std::visit() requires an adapter `overloaded` template which would have to be introduced somewhere. The syntax is a bit peculiar and the compiler errors are horrible in case of typos. So this can be performed in a future PR.

src/build_result.h
digit-google: This needs to be default-constructible, so I would say yes. The alternative is to add a new Init state which would be equivalent.

src/build_result.h
jamesl-10: Sounds good! Yeah that's why I didn't include it initially, just wanted to make sure we were on the same page :)

jamesl-10:
>  Also added a small fix - I noticed the builds sometimes hang when using the jobserver mode, and I realized it's because in `build.cc`, we have `const bool watch_jobserver = plan_.more_to_do()`. Changing this to `const bool watch_jobserver = plan_.work_ready()` seems to have fixed it.

Just to add on to this - the issue occurred when I had a build where the first command is a console process that unblocks additional commands on completion, which I ran with high parallelism (e.g., `-j16`). Console processes don’t use file descriptors, instead signalling completion via `SIGCHLD`, which we handle after unblocking signals in `ppoll()/pselect()`. So `plan_.more_to_do()` is true, but no processes with FDs are running, and no other commands are in `ready_`.

Then since `const bool watch_jobserver = plan_.more_to_do()` and we have high parallelism, there is always a job server token available while this process runs. Due to some quirk of `ppoll()`, we would always acknowledge the jobserver token availability over capturing the `SIGCHLD` from the completed console process. So we unblock `ppoll()`, return `WorkResult::JobserverTokenAvailable`, and continue the build loop, over and over again! The `SIGCHLD` remains unacknowledged (though not lost) because of the constant retries.

I verified the signal is not actually lost by setting a manual upper-limit on the number of token retries in `DoWork()` in code - e.g. after we see a jobserver token available after 100 times, stop waiting on the jobserver fd in `ppoll()`.

A sample `.ninja` file to trigger this issue (I invoked this with a Makefile using `-j<N>`:

```
rule console_job
  command = python3 -c "import time; [time.sleep(0.1) for _ in range(10)]"
  pool = console

# This job is logically blocked until A finishes.
rule follow_up_job
  command = echo "B (Follow-up) starting..." && touch $out

build a.stamp: console_job
build b.stamp: follow_up_job a.stamp

default b.stamp
```

@digit-google @juergbi just wanted to get another review/stamp given this change, and also if you foresee any other way we might get a weird interaction with `ppoll()` or `pselect()`? Or other behavior in the `SubprocessSet` class that maybe I didn't account for. Sorry for the long writeup, but just wanted to make the change clear :)

src/real_command_runner.cc
```diff
@@ -95,25 +96,53 @@ bool RealCommandRunner::StartCommand(Edge* edge) {
...
+  while (status == SubprocessSet::WorkResult::NoWork) {
+    status = subprocs_.DoWork();
+  }
 
-  delete subproc;
-  return true;
+  // Address interrupts first, then subprocesses finishing, then finally
+  // jobserver token availability
+  if (status == SubprocessSet::WorkResult::Interrupted) {
+    result->status = ExitFailure;
```
juergbi: I think this should be `ExitInterrupted` in this commit ("Handle jobserver token availability in runner"). This line is replaced in the follow-up 'algebraic data type' commit, however, it may still make sense to fix up this commit to avoid confusion during review and future bisecting or debugging.

jamesl-10: Fixed!

jamesl-10:
Friendly ping @digit-google and @jhasse if you can take a quick look? :)

src/subprocess.h
```diff
@@ -102,9 +102,19 @@ struct SubprocessSet {
   SubprocessSet();
   ~SubprocessSet();
 
+  // The result of DoWork().
```
digit-google: Thanks. Can we document the meaning of each value, and maybe sort them in priority order? For example something like:

```
The result of DoWork(), in increasing priority order. When several events happen at the same time, the highest value will be reported:

NoWork: If there is no work to do, or if ppoll()/pselect() was interrupted by a spurious signal or returned an error.

JobserverTokenAvailable: Indicates that a new job slot token is available from the jobserver queue. This can only be returned if SetJobserverFD() has been called.

SubprocFinished: Indicates that at least one subprocess completed since the last call.

Interrupted: Indicates that a user interrupt (SIGINT, SIGHUP or SIGTERM) has been detected.
```

src/real_command_runner.cc
```diff
@@ -110,40 +111,46 @@ bool RealCommandRunner::WaitForCommandOrJobserverToken(Result* result, bool watc
...
+      delete subproc;
+
+      build_result = BuildResult::CommandCompleted(edge, status, output);
+      break;
+    }
+    case SubprocessSet::WorkResult::JobserverTokenAvailable:
+      build_result = BuildResult::JobserverTokenAvailable{};
+      break;
+    default:
+      break;
```
digit-google, line 151: nit: Add a Fatal() call here since this should never happen due to the loop no line 120, and will make the code more explicit :)

misc/jobserver_test.py
```diff
@@ -105,6 +105,26 @@ def ninja_spawn(
...
         )
 
+    def ninja_popen(
+        self,
+        ninja_args: T.List[str],
+        prefix_args: T.List[str] = [],
+        extra_env: T.Dict[str, str] = {},
+        capture_output: bool = True,
+    ) -> subprocess.Popen[str]:
+        """Run Ninja command and return the Popen object (i.e. without blocking)."""
```
digit-google: nit: "Start Ninja command" would be clearer :)

misc/jobserver_test.py
```diff
@@ -130,6 +150,53 @@ def generate_build_plan(command_count: int) -> str:
...
 
+def generate_double_build_plan(
+        wait_1_ms: int = 500,
+        wait_2_ms: int = 500,
+        offset: int = 0,
+    ) -> str:
+
+    """Generate a Ninja build plan where each client gets two tasks.
+
+    - `id` identifies the client for debugging purposes.
```
digit-google: nit: this doesn't seem to exist anymore

misc/jobserver_test.py
```diff
@@ -130,6 +150,53 @@ def generate_build_plan(command_count: int) -> str:
     )
     return result
 
+def generate_double_build_plan(
+        wait_1_ms: int = 500,
+        wait_2_ms: int = 500,
+        offset: int = 0,
+    ) -> str:
+
+    """Generate a Ninja build plan where each client gets two tasks.
```
digit-google: nit: Clarify this, e.g. "Generate a Ninja build plan that invokes the test helper script in two parallel commands, each one with its own duration argument.

misc/jobserver_test.py
```diff
@@ -130,6 +150,53 @@ def generate_build_plan(command_count: int) -> str:
...
+
+    - `id` identifies the client for debugging purposes.
+    - `wait_1_ms` and `wait_2_ms` control task durations.
+    - `offset` allows multiple plans without conflicting filenames.
```
digit-google: nit: Clarify: "start index of the output files generated by this plan", allows multiple build plans to run in parallel without conflicts.

src/subprocess_test.cc
```diff
@@ -264,3 +264,33 @@ TEST_F(SubprocessTest, ReadStdin) {
...
+  uint8_t token = 0;
+  ssize_t ret = ret = ::write(mock_jobserver_fd[1], &token, 1);
+  ASSERT_EQ(1, ret);
+  close(mock_jobserver_fd[1]);
+
+  // Assert we detected the jobserver token availability.
+  Subprocess* subproc = subprocs_.Add("sleep 1");
+  SubprocessSet::WorkResult result = subprocs_.DoWork();
+  ASSERT_EQ(SubprocessSet::WorkResult::JobserverTokenAvailable, result);
+
```
digit-google, line 319: To avoid busy looping below, either read the token from the pipe, or call `subprocs_.SetJobserverFD(-1)` . This will ensure DoWork() only returns when the process completes.

src/subprocess_test.cc
```diff
@@ -264,3 +264,33 @@ TEST_F(SubprocessTest, ReadStdin) {
...
+  // Setup a mock jobserver fd by using a pipe.
+  // Give the read end of the pipe to SubprocessSet.
+  int mock_jobserver_fd[2];
+  ASSERT_EQ(pipe(mock_jobserver_fd), 0);
+  subprocs_.SetJobserverFD(mock_jobserver_fd[0]);
+
+  // Trigger mock jobserver token availability by
+  // writing a byte to the write end of the pipe.
+  uint8_t token = 0;
+  ssize_t ret = ret = ::write(mock_jobserver_fd[1], &token, 1);
```
digit-google: nit: remove double assignment :-)

src/subprocess_test.cc
```diff
@@ -264,3 +264,33 @@ TEST_F(SubprocessTest, ReadStdin) {
   ASSERT_EQ(1u, subprocs_.finished_.size());
 }
 #endif  // _WIN32
+
+#ifndef _WIN32
+TEST_F(SubprocessTest, JobserverTokenAvailable) {
+  // Setup a mock jobserver fd by using a pipe.
+  // Give the read end of the pipe to SubprocessSet.
+  int mock_jobserver_fd[2];
+  ASSERT_EQ(pipe(mock_jobserver_fd), 0);
```
digit-google: nit: Use scoped classes to ensure the pipe handles are released when this function exits.

misc/jobserver_test.py
```diff
@@ -130,6 +150,53 @@ def generate_build_plan(command_count: int) -> str:
...
+    result += "build %s: long\n" % span_output_file(offset)
+    result += "build %s: short\n" % span_output_file(offset + 1)
+
+    result += "build all: phony %s\n" % " ".join(
+        [span_output_file(n + offset) for n in range(2)]
+    )
+
+    return result
+
+def get_span(build_dir: str, command_count: int) -> T.List[T.Tuple[int, int]]:
```
digit-google: nit: This function doesn't return a single span, so would be better named `get_spans()`, which would be possible if the one below was renamed to `get_spans_for_multiple_builds()` for clarity. Documentation would be nice.

misc/jobserver_test.py
```diff
@@ -130,6 +150,53 @@ def generate_build_plan(command_count: int) -> str:
...
+    return get_spans([build_dir], [command_count])
+
+def get_spans(build_dirs: T.List[str], command_counts: T.List[int]) -> T.List[T.Tuple[int, int]]:
```
digit-google: nit: Rename to `get_spans_for_multiple_builds()` or something clearer, and document the arguments and the return value, as well as the implied assumptions (regarding the offsets used to name the outputs). What would you think about sorting the result, which would simplify both the documentation and call sites?

misc/jobserver_test.py
```diff
@@ -290,6 +349,83 @@ def run_ninja_with_jobserver_pipe(args):
...
+
+                |---A---|                    (Client 1, 2000ms)
+                |---B---|                    (Client 1, 2000ms)
+                |-------C-------|            (Client 2, 4000ms)
+                           |-------D-------| (Client 2, 4000ms)
+
+                Make sure D starts after A and B finishes, but before C finishes.
+                """
+
+                spans: T.List[T.Tuple[int, int, int]] = get_spans([b1.path, b2.path], [client_tasks, client_tasks])
```
digit-google: nit: `T.List[T.Tuple[int, int]]`, or you could just drop the type hint here, since it will come from the function call.

misc/jobserver_test.py
```diff
@@ -290,6 +349,83 @@ def run_ninja_with_jobserver_pipe(args):
             max_overlaps = compute_max_overlapped_spans(b.path, task_count)
             self.assertEqual(max_overlaps, 1)
 
+    @unittest.skipIf(_PLATFORM_IS_WINDOWS, "These test methods do not work on Windows")
+    def test_jobserver_client_with_posix_fifo_token_efficiency(self):
+        # Due to implicit slots, the max number of parallel tasks is task_count + num_clients - 1
+        task_count = 2
+        client_tasks = 2
+
+        try:
```
digit-google: nit: I recommend using contextlib.ExitStack() to handle the cleanup handlers instead of a try..finally block, as this makes things easier to follow and maintain over time.

src/real_command_runner.cc
```diff
@@ -94,26 +96,61 @@ bool RealCommandRunner::StartCommand(Edge* edge) {
...
+      Subprocess* subproc = subprocs_.NextFinished();
+      ExitStatus status = subproc->Finish();
+      std::string output = subproc->GetOutput();
+      std::map<const Subprocess*, Edge*>::iterator e =
+          subproc_to_edge_.find(subproc);
+      Edge* edge = e->second;
+      subproc_to_edge_.erase(e);
+      delete subproc;
+
+      build_result = BuildResult::CommandCompleted(edge, status, output);
```
digit-google: nit: use std::move(output) to avoid one string copy :)

digit-google (commented):
Thanks again for the updates.

I have ported this stack of commits to my local copy of the Fuchsia Ninja version, and will be testing it next week extensively. We only use jobserver mode in specific build configurations (that need to launch 20+ sub-builds) which are quite long, e.g. around 20 minutes, so this will take time. But early testing shows no immediate regression or blocked builds so far, which is a great sign.

I'll send updates when this is done, but so far everything looks very good (barring the nits of course :-))

src/subprocess_test.cc
jamesl-10: Good catch, thanks!

misc/jobserver_test.py
jamesl-10: That sounds good to me, done!

jamesl-10:
Thanks again for the comments! Addressed all the nits, please keep us updated on how the testing goes :)

digit-google:
Hello, I could complete my benchmarking on the Fuchsia build, and results are very good :)

Summary is that build times are reduced on average from 2% to 12% with this PR applied to the Fuchsia Ninja sources. Mostly by reducing the variability of total build times (i.e. the minimum build time over 10 runs of each is about the same, but the maximum build time varies a lot less, and is closer to the minimum with the fix!).

@jhasse, please please take a look at the change here and approve this PR if you can.

Longer version:

Benchmarking consisted in running Fuchsia builds of the same sources, in three distinct build configurations that all launch 6, 12 ans 24 Ninja sub-builds in parallel with a controlling jobserver. Remote actions were disabled, so everything was run locally. For the record, these build configurations take around 12, 9 and 22 minutes each, on a lightly loaded workstation with 96 cores, plus plenty of RAM and disk.

Two Ninja binaries were compared, both compiled from the [Fuchsia Ninja source repository](https://fuchsia.googlesource.com/third_party/github.com/ninja-build/ninja): one version corresponding to the current tip-of-tree, and another version compiled with the fixes in this PR ported, rather than cherry-picked, since there are now significant differences between the upstream and Fuchsia sources.

Each build configuration was run 10 times from the exact same initial state (which included preliminary build steps that do not involve Ninja sub-builds). And the minimum, maximum and average times were collected.

```
Build configuration        fuchsia-ninja-tot   fuchsia-ninja-jamesl10


6 parallel sub-builds   min       0:11:57               0:11:56
                        max       0:13:06               0:12:09
                        average   0:12:40               0:12:02

                        diff     -0:00:38 (-5.00%)

12 parallel sub-builds  min       0:09:33               0:09:06
                        max       0:10:59               0:09:09
                        average   0:10:26               0:09:07

                        diff     -0:01:19 (-12.62%)

24 parallel sub-builds  min       0:21:32               0:21:30
                        max       0:24:26               0:22:38
                        average   0:22:32               0:22:02

                        diff     -0:00:30 (-2.22%)

```

@jamesl-10 , thank you for this excellent work

jhasse:
I trust you, so I will let this sit for a while and then merge if nothing comes up :)

Also thanks from my side to everyone involved!

jamesl-10:
Thank you so much for the detailed benchmarks! And for all the support on this PR too :)

juergbi:
We've been using this branch internally for a month for thousands of builds and haven't seen any issues.

Issue #663, opened 2013-10-11
Throttle rate of output messages


fischman:
Building a large project with ninja output going to an emacs _compilation_ buffer is painful b/c emacs doesn't do so well with 10k-line-buffers.

IWBN to have a rate-limit on progress messages, such as "don't emit a progress message more often than once per <N> seconds".  Could be done with an env var or a cmd-line option to ninja (I weakly prefer env var).

WDYT, worth writing a patch for?

fischman:
I've been out of the ninja game for years so unsubscribing from this.  Feel free to close if no longer useful, or @ me if you want me to see a followup.

bradneuman:
I recently started using cmake with ninja in emacs and encountered this same problem. Built-in support for this would be much nicer, but I took the approach of creating a quick-and-dirty "throttling" script in python here: https://gist.github.com/bradneuman/3f2bcb3b3a31f00129f69fead4d94c5f

YMMV and you may have to change the (very simple) regex I used there, but for now it did the trick for me. It just throttles to one-per-second for lines matching a certain regex and prints everything else

jonesmz:
Here's one way that this could be implemented: https://github.com/ninja-build/ninja/pull/1631

(Personally, I think librari-fying ninja is a better approach, so that the lib can be loaded by an application that offers customized output as desired)

PR #2727, opened 2026-02-11
Allow UTF-8 emoji characters in NINJA_STATUS

### LINKED ISSUE #2728: Support UTF-8 encoded emojis in NINJA_STATUS environment variable
Similar to #713, it would be great if NINJA_STATUS accounted for the width of UTF-8 encoded emojis. I've written a fix for this in https://github.com/ninja-build/ninja/pull/2727 which works for me, so I can use:

```
  export NINJA_STATUS=$(print -P "%B%F{magenta}🧵%%r 🚀%%o/s %F{green}⏱️%%w ⌛%%W %F{cyan}%%f/%%t %F{yellow}%%p%b %F{reset}")
```

and the full terminal width is used. Ideally, in the patch, I would have used `wcwidth()` but I believe this isn't available on Windows, so the code assumes UTF-8 emojis are 2 characters in width (apart from control characters), which is true on iTerm2 and probably most other terminal emulators.

I've added new tests for this, and the existing tests continue to pass.

Comments appreciated. Thanks :)

jthackray:
Add UTF-8 parsing code to allow UTF-8 encoded emojis in NINJA_STATUS. In #2428, support was added for parsing ANSI colors, so that the correct terminal width was observed. This patch is similar in nature, and allows UTF-8 emojis to be used whilst still using the full width of the terminal window.

I've had to add quite a bit of UTF-8 parsing code, so we don't misparse UTF-8 glyphs. `ElideMiddle` now decodes UTF-8 codepoints and assigns terminal column widths with a small emoji-width heuristic, while treating ZWJ and variation selectors as zero-width.

This code assumes common emoji and pictograph ranges render as double-width and everything else is single-width unless it is a control character. I would use `wcwidth()`, but AFAIK, this isn't available on Windows.

I've added some new tests for this functionality, and all the existing tests continue to pass.

Fixes #2728

### DIFF
--- a/src/elide_middle.cc
+++ b/src/elide_middle.cc
@@ -15,7 +15,161 @@
 #include "elide_middle.h"
 
 #include <assert.h>
+#include <stdint.h>
 #include <string.h>
+
+namespace {
+
+size_t Utf8CharSize(const char* ptr, const char* end) {
+  // Decode the leading byte using UTF-8 prefix masks:
+  // 0xxxxxxx (0x80)      -> 1 byte
+  // 110xxxxx (0xE0/0xC0) -> 2 bytes
+  // 1110xxxx (0xF0/0xE0) -> 3 bytes
+  // 11110xxx (0xF8/0xF0) -> 4 bytes
+  const unsigned char lead = static_cast<unsigned char>(*ptr);
+  if (lead < 0x80)
+    return 1;
+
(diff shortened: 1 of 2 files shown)


jhasse:
Thanks for the PR! This adds quite a bit of code though for a non-default use-case.

Also we might want to think about fixing #663 first.

PR #2729, opened 2026-02-13
Fix spurious rebuilds for phony targets with validations

codesuki:
Phony targets with no inputs are considered dirty if their output is missing. However, phony targets used as anchors for validation dependencies (e.g. from GN groups) often have no regular inputs but should not force a rebuild if the validations are clean.

This change updates RecomputeOutputDirty to check if a phony target has validation dependencies. If so, it is treated similarly to a target with inputs: it is only dirty if its dependencies are dirty, not automatically dirty because the output file is missing.

This fixes an issue where targets depending on a phony validation anchor would rebuild on every run.

### DIFF
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -285,8 +285,8 @@ bool DependencyScan::RecomputeOutputDirty(const Edge* edge,
   if (edge->is_phony()) {
     // Phony edges don't write any output.  Outputs are only dirty if
     // there are no inputs or validations and we're missing the output.
-    // If a phony target has inputs or validations, they are used for dirty
-    // calculation instead of this fallback.
+    // If a phony target has inputs or validations, or the output exists,
+    // they are used for dirty calculation instead of this fallback.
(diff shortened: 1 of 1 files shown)


src/graph.cc
```diff
@@ -285,7 +285,8 @@ bool DependencyScan::RecomputeOutputDirty(const Edge* edge,
   if (edge->is_phony()) {
     // Phony edges don't write any output.  Outputs are only dirty if
     // there are no inputs and we're missing the output.
```
digit-google: nit: Can you add a segment in the comment that explains why validations_ is checked too.

digit-google (approved):
Thank you, this looks fine, @jhasse please take a look.

src/graph.cc
codesuki: I added a small comment. Please let me know if that clarifies it.

digit-google: Yes, thank you!

jhasse:
Thanks!

Issue #1718, opened 2020-01-08
Ninja unnecessarily runs gn just after it was run manually


ilor:
Somewhat similar to #1376, I am getting an unnecessary re-run of a ninja generator (gn) right after it was run. It's so far been slightly tricky to confidently reproduce in a minimal repro, but it tends to emerge in incremental rebuilds of Chromium for me.

When this issue happens, what I'm seeing is somewhat similar to #1376. I run gn to regenerate ninja files, then ninja decides it needs to re-run gn again based on a build.ninja timestamp, which appears to be incorrect compared to the filesystem state.

The following is a repro I was able to get in a Chromium build, which while far from "minimal" is at least something. I used a recent master ninja version but the issue is not new.
```
$ gn gen out --args="target_os=\"android\""
Done. Made 22794 targets from 1916 files in 5538ms
$ stat -c "%Y %n" out/build.ninja out/args.gn
1578329248 out/build.ninja
1578329243 out/args.gn
$ ninja -C out -d explain
ninja: Entering directory `out'
ninja explain: recorded mtime of build.ninja older than most recent input args.gn (1578329238290052140 vs 1578329243458020060)
[1/1] Regenerating ninja files
```
The mtime of build.ninja on disk is not older than args.gn.

Note however this was not the _first_ build in the `out` directory I did, I had to play with it a bit to enter the broken state. I'll try to figure out if I can get this to happen reliably.

ilor:
Looks like the issue starts appearing in a clean build dir when the following happen:
* Some `--args` are passed to `gn gen` (didn't repro without `--args`, args used didn't seem to matter)
* A change to some BUILD.gn file triggered a (correct) gn re-run by ninja
Ninja will then start causing unnecessary gn re-runs after each time `gn gen` is issued manually in this build dir.

Essentially, in a clean Chromium checkout with a recent ninja, I'm getting:
```
$ rm -rf out2
$ gn gen out2 --args="target_os=\"android\""
Done. Made 22794 targets from 1916 files in 5671ms
$ ninja -C out2  # expecting ninja to NOT re-run gn (OK)
ninja: Entering directory `out2'
[builds]
ninja: build stopped: interrupted by user.
$ touch BUILD.gn 
$ ninja -C out2  # expecting ninja to re-run gn (OK)
ninja: Entering directory `out2'
[1/1] Regenerating ninja files
[builds]
ninja: build stopped: interrupted by user.
$ gn gen out2 --args="target_os=\"android\""
[...]

PR #2730, opened 2026-02-16
"ninja -t restat" doesn't work with "builddir"

jhasse:
When specifying "builddir = foo" in the manifest ninja -t restat doesn't work because the tool skips parsing the manifest for performance reasons. Maybe we should parse it but stop as soon as we have the builddir which is often at the beginning of the file.

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -397,6 +397,9 @@ _Available since Ninja 1.11._
 
 `restat`:: updates all recorded file modification timestamps in the `.ninja_log`
 file. _Available since Ninja 1.10._
++
+The build manifest won't be parsed when running this tool. If you're using a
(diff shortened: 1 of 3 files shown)


jhasse:
Changing `RUN_AFTER_FLAGS` to `RUN_AFTER_LOAD` would fix this, but it would reintroduce #1718. See d986e4db5630cf1c5547e69b5556f006f7d3444a

jhasse:
> [...] but stop as soon as we have the builddir which is often at the beginning of the file.

Can't do that because the manifest might not contain builddir and then we parse it anyway. We'll have to have a "quick" parse which just checks for bindings or something.

jhasse:
Different fix: Since the usecase for `-t restat` is the build generator and it knows the builddir, add an optional `--builddir=DIR` option that can be used.

PR #2734, opened 2026-02-25
Bump C++14 to C++17

jhasse:
see #2726

PR #2737, opened 2026-02-26
Update MSVC C++ standard to C++17, see PR #2734.

zufuliu:
Only use `/std:c++17` to keep sync with `-std=c++17` for GCC and Clang.


jhasse:
Thanks!

PR #2744, opened 2026-03-18
Document `pool` option under "Rule Variables"

### LINKED ISSUE #2732: Documentation: `pool` not listed under "Rule variables"
https://ninja-build.org/manual.html#ref_rule

The following is valid:
```
rule foo
  command = true
  pool = console
```

IMO it would be nice if `pool` was added to the linked section, for completeness sake,
especially since the linked section says:
>  Here is a full list of special keys

Maybe an indepth description is unnecessary, and just a link to https://ninja-build.org/manual.html#ref_pool is enough

braydenkrus:
In response to and closes #2732. This simply gives a little bit of documentation for pool, as specified in the issue.

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -1032,6 +1032,9 @@ keys.
 `out`:: the space-separated list of files provided as outputs to the build line
   referencing this `rule`, shell-quoted if it appears in commands.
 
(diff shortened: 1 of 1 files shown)


jhasse:
Thanks!

PR #2747, opened 2026-03-20
fix: use == instead of = in newline escape version check

cuiweixie:
The condition used assignment (=) instead of comparison (==), which would always evaluate to true and break the version check logic.


digit-google (approved):
Thank you! @jhasse  please take a look.

jhasse:
Thanks! I've also added this to clang-tidy in 297718e2b5d3c6cb053154261d7acbee62e6c521 so it can never happen again.

Issue #2652, opened 2025-08-15
`ninja -t browse` doesn't URL-encode targets


dseomn:
When running `ninja -t browse` with the file below, the page http://localhost:8000/?foo links to http://localhost:8000/?foo%2Fbar instead of the correct http://localhost:8000/?foo%252Fbar

```
rule copy
  command = cp $in $out
build foo%2Fbar: copy foo
```

lklivingstone:
Hi, I tried a fix for this issue and it seems to work. If no one is working on it, I can open a PR.

PR #2750, opened 2026-03-23
Fix/browse url encode

lklivingstone:
Fixes: #2652 

### Root Cause Analysis (RCA)

The -t browse tool generates links to targets using their raw names.
For targets containing % (e.g., foo%2Fbar), the link was generated without URL encoding:

```
/?foo%2Fbar
```

When this URL is handled by the server, it is decoded using unquote(), which interprets %2F as /:

```
foo%2Fbar -> foo/bar
```

As a result, Ninja attempts to query a different target (foo/bar instead of foo%2Fbar), leading to lookup failures and broken navigation in the browse UI.

### Fix

The fix ensures that target names are properly URL-encoded when generating links by using quote():

```
quote(output)
```

This encodes % as %25, so:

```
foo%2Fbar -> foo%252Fbar
```

Now when the server decodes the URL once:

```
foo%252Fbar -> foo%2Fbar
```
The original target string is preserved, and Ninja can correctly resolve the target.

#### Test Added

Added a unit test:

```
TEST_F(StateTestWithBuiltinRules, ComplexTargetPreserved)
```

This test verifies that targets containing:
- spaces
- percent-encoded sequences (%2F)
- URL-reserved characters (?, &, =)

are preserved exactly during parsing and lookup.

### DIFF
--- a/src/browse.py
+++ b/src/browse.py
@@ -37,9 +37,9 @@ if sys.version_info >= (3, 2):
 else:
     from cgi import escape
 try:
-    from urllib.request import unquote  # type: ignore # Module "urllib.request" has no attribute "unquote"
+    from urllib.parse import quote, unquote
 except ImportError:
-    from urllib2 import unquote
+    from urllib import quote, unquote
 from collections import namedtuple
 from typing import Tuple, Any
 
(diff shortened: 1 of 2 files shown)


lklivingstone:
@digit-google Let me know if you need any other changes.

lklivingstone:
I hope this is ok.

digit-google:
> I hope this is ok.

Perfect, thank your patience :)

PR #2752, opened 2026-03-24
Detect & rebuild externally modified output files

philwo:
Add an `output_mtime` field to the build log that records the actual mtime of each output file after a successful build. On subsequent builds, if an output's current mtime differs from the recorded value, the output is marked dirty and rebuilt.

Previously, Ninja only stored `command_start_time_` in the log's mtime field (to detect inputs modified during a build) and had no record of the output's actual mtime. This meant externally modified outputs (e.g. `echo "corrupted" > out.txt`) went undetected.

The new field is appended as a 6th tab-separated column in .ninja_log. Old Ninja versions parse it harmlessly (strtoull stops at the tab), so no log version bump is needed. Old log entries default to output_mtime=0, which skips the check. Generator rule outputs are also excluded since they are expected to be user-edited.

### DIFF
--- a/tests/restat/test_restat_builddir.py
+++ b/tests/restat/test_restat_builddir.py
@@ -66,9 +66,9 @@ build output.txt: touch
             output_entry, "", "output.txt not found in build/.ninja_log"
         )
 
-        # Parse the log entry: start_time\tend_time\tmtime\toutput\tcommand_hash
+        # Parse the log entry: start_time\tend_time\tmtime\toutput\tcommand_hash\toutput_mtime
         parts = output_entry.split("\t")
-        self.assertEqual(len(parts), 5, f"Unexpected log format: {output_entry}")
+        self.assertEqual(len(parts), 6, f"Unexpected log format: {output_entry}")
         original_mtime = int(parts[2])
 
         # Wait a bit to ensure different mtime
@@ -116,7 +116,7 @@ build output.txt: touch
         # Parse the updated log entry
         updated_parts = updated_entry.split("\t")
         self.assertEqual(
-            len(updated_parts), 5, f"Unexpected updated log format: {updated_entry}"
+            len(updated_parts), 6, f"Unexpected updated log format: {updated_entry}"
         )
         updated_mtime = int(updated_parts[2])
 


philwo:
Hey, for context, I noticed this while analyzing how closely our Siso build tool in Chromium sticks to the behavior of Ninja. During these tests, I noticed that modified outputs aren't detected and rebuilt by Ninja. I'm not sure if you want to change the behavior of Ninja in this regard, but if you also consider this a bug, then this should fix it.

jhasse:
It's actually intentional AFAIK. I'm not sure if we aren't opening Pandora's box with this change, @evmar might know.

One use-case for me is to quickly change an output (i.e. an asset file) as a quick and dirty hack to test something - while still being able to recompile other binaries. I wouldn't want ninja to override my changes to the asset output in that case.

philwo:
Ah, I totally understand how having this possibility can be nice. :) We actually got a feature request from a Chromium developer for Siso for exactly that use case, so we implemented a flag that causes it to temporarily not regenerate modified outputs: http://crrev.com/c/6191399/

Making this an explicit flag ensures that after the quick hack is no longer needed, the build automatically converges to the "correct" output state again once you stop passing the flag. That way there's no risk of forgetting that an output was manually modified, then the next day wondering why the binary doesn't do what the source says it should do. :grin:

If that approach were something that you think would work for Ninja, maybe a `-d tainted` debug mode would be the most "ninja"-way to implement it? (I'm not attached to the "tainted" name, maybe there's a better one.. suggestions welcome.)

Of course, if you or Evan think that the current behavior is working as intended for Ninja, that's also totally fine.

jhasse:
There are other cases where Ninja skips some checks (e.g. don't stat all files in some situations, requiring `ninja -t restat`). This means that a "corrupted" state can become unnoticed - even after this PR. So I'm not sure if we should even pretend? We could think about a mode where Ninja really checks everything it can and the default stays the way it is now?

evmar:
I can't really say what is "intended" behavior, I don't think I ever really thought this through.  I think thinking through the use cases and evaluating against them is probably the best you can do.

I believe the reason for the existing `command_start_time_` is for more than detecting modified inputs, it's also for the whole 'restat' behavior.  Suppose you have B as an output of build command A.  When A is out of date, you run the command, but if the build command doesn't write B, then the next build thinks A is still out of date.

In fact, it is "more correct" (for some meaning of correct that doesn't mean necessarily good or desirable, but rather I guess just more precise) to not record when the build command ran, but rather the exact mtime of B after the build ran, whether B ends up with a time in the past or up to date.  Then you can just use that to detect whenever B changes.  More recently I did a revisit of Ninja where I made every build depend [on the exact mtime of all inputs and outputs](https://github.com/evmar/n2/blob/main/src/hash.rs).  In practice this ends up being pretty sensitive to different projects' expectations of Ninja.  I recall one thing it breaks is that meson has a build step where [the step overwrites one of its inputs](https://github.com/search?q=repo%3Aevmar%2Fn2%20meson&type=code) when it runs, for example.

PR #2753, opened 2026-03-24
Restore "explain" output when a dyndep file is loaded

### LINKED ISSUE #2758: ninja no longer explains when dyndep files are loaded
With this project:

```console
$ cat build.ninja
rule touch
  command = touch $out $out.imp
build out: touch || dd
  dyndep = dd

$ cat dd
ninja_dyndep_version = 1
build out | out.imp: dyndep
```

ninja 1.12.1 explains when it loads the depfile:

```console
$ ninja -v -d explain
ninja explain: loading dyndep file 'dd'
ninja explain: output out doesn't exist
[1/1] touch out out.imp
```

ninja 1.13.2 does not:

```
$ ninja -v -d explain
ninja explain: output out doesn't exist
[1/1] touch out out.imp
```

This was regressed by #2067.

bradking:
Since #2067 the "loading dyndep file '...'" explanation has not been printed.  Instead the explanation has been recorded for the dyndep file's node *after* it has been brought up-to-date, and therefore never printed.

This explanation is generated just before its operation runs, so we can print it immediately instead of buffering it.

Fixes: #2758


bradking:
@jhasse this fixes a regression in `ninja -d explain` output with `dyndep` bindings that I noticed while working on #2749.

src/dyndep.cc
```diff
@@ -38,7 +39,8 @@ bool DyndepLoader::LoadDyndeps(Node* node, DyndepFile* ddf,
   node->set_dyndep_pending(false);
 
   // Load the dyndep information from the file.
-  explanations_.Record(node, "loading dyndep file '%s'", node->path().c_str());
+  if (status_)
+    status_->Explain("loading dyndep file '%s'", node->path().c_str());
```
digit-google: this is really confusing, imo, because now there are two ways for Ninja to print explanations. One in this line, and the other in status_printer.cc which prints the recorded information. Is there a way to make this more consistent and easier to understand?

digit-google (commented):
Can you provide a reproduction case that better explains the issue, as well as a regression test for this?

src/dyndep.cc
digit-google: I think it might make more sense to modify the `Explanations` class to support this directly, as well as move the output logic there, e.g.:

- Add an Explanations::ExplainEdge(Edge*) method that prints the explanation for a given edge, if there is one. This could be called from status_printer.cc.

- Add an Explanations::ExplainDyndepLoad(Node*) which could be called here.

This would avoid injecting the Status directly into the DyndepLoader and the DependencyScan classes too, keeping the change simpler in the rest of the code. Wdyt?

bradking: Thanks.  I've switched to that approach.  While at it I also factored out the `Explanations` implementation into its own source.

bradking:
> a reproduction case...a regression test for this?

I opened #2758 to document the problematic case, and added a corresponding test to `misc/output_test.py`.

While working on this I also noticed #2759, but that behavior has been present since `dyndep` was introduced and is not a regression.  I've left that as a FIXME comment in the test here, and will defer a fix to a later PR.

src/explanations.cc
```diff
@@ -0,0 +1,126 @@
...
+  }
+}
+
+void Explanations::LookupAndAppend(const void* item,
+                                   std::vector<std::string>* out) {
+  auto it = map_.find(item);
+  if (it == map_.end()) {
+    return;
+  }
+
```
AJIOB, line 81: nit: please, reserve the output before the pushing

src/explanations.cc
```diff
@@ -0,0 +1,126 @@
...
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "graph.h"
+#include "status_printer.h"
+
+namespace {
+void Explain(const char* msg, va_list ap) {
+  fprintf(stderr, "ninja explain: ");
```
AJIOB:
```suggestion
  fputs("ninja explain: ", stderr);
```

nit

src/explanations.cc
```diff
@@ -0,0 +1,126 @@
...
+  vfprintf(stderr, msg, ap);
+  fprintf(stderr, "\n");
```
AJIOB:
```suggestion
  fputs("\n", stderr);
```

nit

src/status.h
```diff
@@ -40,6 +40,7 @@ struct Status {
   /// (which is the default).
   virtual void SetExplanations(Explanations*) = 0;
 
+  virtual void NewLine() = 0;
```
digit-google, line 44: nit: Document what this does :)

digit-google (approved):
Thank you!

digit-google (approved):
Thanks

PR #2754, opened 2026-03-24
Document more precisely how a dyndep file may affect the build graph

bradking:
Only the part of the graph that depends on the edges with a dyndep binding may be affected by content loaded from the dyndep file. Otherwise the dyndep file could change the build graph in a way that causes already up-to-date build statements to become out-of-date.


bradking:
@jhasse this documents existing requirements for `dyndep` files that were not explicitly called out before.

PR #2756, opened 2026-03-26
Improve readability of dyndep, depfile, and validation node logic

bradking:
This changes no functionality, but makes RecomputeNodeDirty easier to follow:

* Queue an edge's validation nodes before *all* inputs

  Previously a dyndep binding could cause one input to be traversed
  before queuing the validation nodes of the current edge.  This does
  not affect correctness, but it makes the code easier to follow.

* Explicitly load a node's mtime before comparing it to the deps log

  Previously we were relying on RecomputeNodeDirty to do
  this before LoadDeps.  Avoid this non-local assumption.

* Simplify logic that updates an edge from dyndep and depfile bindings

  RecomputeNodeDirty checks both dyndep and depfile bindings when an edge
  is first encountered.  Consolidate them into a single conditional block.
  This simplification was missed in
  #1521.

### DIFF
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -105,6 +105,14 @@ bool DependencyScan::RecomputeNodeDirty(Node* node, std::vector<Node*>* stack,
   if (!VerifyDAG(node, stack, err))
     return false;
 
+  // Store any validation nodes from the edge for adding to the initial
+  // nodes.  Don't recurse into them, that would trigger the dependency
+  // cycle detector if the validation node depends on this node.
+  // RecomputeDirty will add the validation nodes to the initial nodes
+  // and recurse into them.
+  validation_nodes->insert(validation_nodes->end(),
+      edge->validations_.begin(), edge->validations_.end());
+
   // Mark the edge temporarily while in the call stack.
   edge->mark_ = Edge::VisitInStack;
   stack->push_back(node);
@@ -159,14 +167,6 @@ bool DependencyScan::RecomputeNodeDirty(Node* node, std::vector<Node*>* stack,
       dirty = edge->deps_missing_ = true;
     }
   }
-
-  // Store any validation nodes from the edge for adding to the initial
-  // nodes.  Don't recurse into them, that would trigger the dependency
-  // cycle detector if the validation node depends on this node.
-  // RecomputeDirty will add the validation nodes to the initial nodes
-  // and recurse into them.
-  validation_nodes->insert(validation_nodes->end(),
-      edge->validations_.begin(), edge->validations_.end());
 
   // Visit all inputs before checking if any of them is ready.
   // Newly encountered edges may load dyndep files and gain
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -748,6 +748,11 @@ bool ImplicitDepLoader::LoadDepsFromLog(Edge* edge, string* err) {
     return false;
   }
 
+  // Load the output's mtime if we haven't already.
+  if (!output->StatIfNecessary(disk_interface_, err)) {
+    return false;
+  }
+
   // Deps are invalid if the output is newer than the deps.
   if (output->mtime() > deps->mtime) {
     explanations_.Record(output,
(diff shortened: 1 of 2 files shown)


bradking:
@jhasse @digit-google I made these changes while working on #2749, but since they are only refactoring, decided to defer them to this follow-up PR.

src/graph.cc
```diff
@@ -82,13 +82,20 @@ bool DependencyScan::RecomputeDirty(Node* initial_node,
 bool DependencyScan::RecomputeNodeDirty(Node* node, std::vector<Node*>* stack,
                                         std::vector<Node*>* validation_nodes,
                                         string* err) {
+  auto const statIfNecessary = [this, err](Node* node) -> bool {
+    if (err) {
+      err->clear();
+    }
```
digit-google: I am not sure this is necessary at all. Let me explain why.

The implied contract for error reporting is that |*err| should only be touched in case of an error, i.e. when the function returns false. It is up to the high-level caller to clear the error variable if he want to reuse it after a failed call (e.g. in a unit-test that verified a failure before running another call).

This behavior is pretty standard, e.g. it is similar to how `errno` is updated by C library implementations, and comes down to avoiding repeatedly clearing the error string in normal call chains that do _not_ error.

And yes, it is not ideal, and something that is usually never properly documented in our comments (but should be).

I think https://github.com/ninja-build/ninja/commit/27e545a5a1695be26ffe7144a903492e7a013d45 was misguided, i.e. the unit-test that didn't clear the error properly should have been fixed instead.

Accepting this change seems harmless, but now |*err| might or might not be cleared even if the function returns true (because it could have exited on line 96 before the call in 98 which forces the clear), making its semantics hard to reason about.

So I think it would be simpler to keep the original semantics to implement them consistently in the whole source tree (and eventually properly document error passing everywhere). Wdyt?

digit-google: Also this lambda hides the fact that it mutate s |*err|, call sites would be clearer if it took the |err| pointer as an argument to make this explicit.

src/graph.cc
```diff
@@ -151,17 +153,12 @@ bool DependencyScan::RecomputeNodeDirty(Node* node, std::vector<Node*>* stack,
           return false;
       }
     }
-  }
 
-  // Load output mtimes so we can compare them to the most recent input below.
-  for (Node* o : edge->outputs_) {
-    if (!statIfNecessary(o))
+    // Load the first output's mtime.  It's needed in LoadDeps.
+    if (!statIfNecessary(edge->outputs_.front()))
```
digit-google: This seems undocumented (and fragile). Would it make more sense to move this to ImplicitDepLoader::LoadDepsFromLog() where it is used?

bradking: Good idea.  I added a commit to do that.

src/graph.cc
bradking: With the fix to `LoadDepsFromLog` my original motivation for the lambda is gone.  I've dropped that part of the change.

I agree with your analysis about the `err` clearing, but I consider that out of scope for this PR.  If you want to do it first I can rebase.  Otherwise, please do it afterward.

src/graph.cc
```diff
@@ -176,6 +165,15 @@ bool DependencyScan::RecomputeNodeDirty(Node* node, std::vector<Node*>* stack,
       return false;
   }
 
+  // Load output mtimes so we can compare them to the most recent input below.
+  for (Node* o : edge->outputs_) {
+    if (err) {
+      err->clear();
```
jhasse: I prefer `*err = "";` because it's clearer (I didn't even know std::string had a clear method).

src/graph.cc
```diff
@@ -176,6 +165,15 @@ bool DependencyScan::RecomputeNodeDirty(Node* node, std::vector<Node*>* stack,
...
+    }
+    if (!o->StatIfNecessary(disk_interface_, err))
```
jhasse: while we're at it, can you add {}?

jhasse (approved):

digit-google (approved):
Thank you, LGTM!

Issue #2379, opened 2024-01-31
ninja build files lacks of line feed


jdai:
similar problem: https://stackoverflow.com/questions/46970462/how-to-enable-multiline-logs-instead-of-single-line-progress-logs

![ninja](https://github.com/ninja-build/ninja/assets/778292/61cbc9ef-bede-4836-a492-0eadb3fdf923)

I read lineprinter.cc, woul dit be the 1st line below lacks of "\n"?
    printf("%s", to_print.c_str());
    printf("\x1B[K");  // Clear to end of line.
    fflush(stdout);
#endif

    have_blank_line_ = false;
  } else {
    printf("%s\n", to_print.c_str());
  }

============
PS E:\repos\king> $PSVertionTable; $Host
Name             : ConsoleHost
Version          : 5.1.22621.2506
InstanceId       : ad82b908-ad36-4994-8241-d9be350b2825
UI               : System.Management.Automation.Internal.Host.InternalHostUserInterface
CurrentCulture   : en-US
![ninja](https://github.com/ninja-build/ninja/assets/778292/68850ea3-2c95-4f72-8a32-67ca5cd8261a)
CurrentUICulture : en-US
PrivateData      : Microsoft.PowerShell.ConsoleHost+ConsoleColorProxy
DebuggerEnabled  : True
IsRunspacePushed : False
Runspace         : System.Management.Automation.Runspaces.LocalRunspace
================
OS Name Microsoft Windows 11 Enterprise
Version 10.0.22631 Build 22631
=============
PS E:\repos\king> git --version
git version 2.43.0.windows.1

digit-google:
This is working as intended.

The purpose of the code you are showing is to replace the content of the _current_ line, without jumping to the next one. This is how Ninja displays its status when it is run in an interactive terminal (but not when invoked from a script, or when its output is piped to another command).

You can use `ninja ... | cat` on Posix to avoid this (in this case each status line is printed in its own separate line), or `ninja .. > CON`  on Windows (same result).

Another way is to declare `TERM=dumb` in your environment.

davidgiven:
The problem with using something line `| cat` or `TERM=dumb` is that it causes any coloured compiler output not to be coloured.

Is there a way to persuade ninja to show the status messages one per line without doing this? I've tried putting an escaped newline in the `description` variable for my commands, but this seems to be getting stripped off...

PR #2763, opened 2026-04-04
Check `NO_COLOR` flag before `CLICOLOR_FORCE`

### LINKED ISSUE #2295: allow disabling colors in output
Please allow disabling colors in output.

Forcing colors in output is supported by setting `CLICOLOR_FORCE=1` (https://bixense.com/clicolors/) but forcing no colors in the output by setting `CLICOLOR=0` (as mentioned on the same page) is not supported. See also #1475.

braydenkrus:
In response to, and should close #2295. This is in progress, and I still need to properly test and format these changes.


src/line_printer.cc
```diff
@@ -50,6 +50,10 @@ LinePrinter::LinePrinter() : have_blank_line_(true), console_locked_(false) {
 #ifdef _WIN32
   // Try enabling ANSI escape sequence support on Windows 10 terminals.
   if (supports_color_) {
+    char* no_color = getenv("NO_COLOR");
+    if (no_color != NULL && std::string(no_color) != "0") {
```
digit-google: nit: According to https://bixense.com/clicolors/, the simple fact that `NO_COLOR` is set should be enough to turn off color output, i.e. no interpretation of the "0" value should be necessary.

src/line_printer.cc
```diff
@@ -59,8 +63,13 @@ LinePrinter::LinePrinter() : have_blank_line_(true), console_locked_(false) {
...
   if (!supports_color_) {
-    const char* clicolor_force = getenv("CLICOLOR_FORCE");
-    supports_color_ = clicolor_force && std::string(clicolor_force) != "0";
+    char* no_color = getenv("NO_COLOR");
+    if (no_color != NULL && std::string(no_color) != "0") {
+      supports_color_ = false;
+    } else {
+      const char* clicolor_force = getenv("CLICOLOR_FORCE");
+      supports_color_ = clicolor_force && std::string(clicolor_force) != "0";
+    }
```
digit-google, line 81: nit: It's unfortunate that, even in the original code, CLICOLOR_FORCE is only tested on Unix, and never on Windows. What would you think about using the same logic for both systems, e.g. by moving the detection code to helper functions (e.g. EnvHasNoColor() and EnvHasCliColorForce()) and calling them consistently?

digit-google (commented):
Please squash this into a single commit. It might be useful to add a link to https://bixense.com/clicolors/ in a comment explaining what rules the probes follow. @jhasse, do you think a check for `CLICOLOR` would be useful here? I see that there is also `FORCE_COLOR`  described at https://force-color.org/

braydenkrus:
Hello @digit-google! I'm on board with your advice. Also, there may be value in switching from `CLICOLOR_FORCE` to `FORCE_COLOR`, as the latter seems to be the most up-to-date (based on the message on top of https://bixense.com/clicolors/), and https://force-color.org/ has an example implementation showing its relation to `NO_COLOR` (which also appears to give `FORCE_COLOR` priority over `NO_COLOR`, even if `NO_COLOR` is set).

digit-google:
> Hello @digit-google! I'm on board with your advice. Also, there may be value in switching from `CLICOLOR_FORCE` to `FORCE_COLOR`, as the latter seems to be the most up-to-date (based on the message on top of https://bixense.com/clicolors/), and https://force-color.org/ has an example implementation showing its relation to `NO_COLOR` (which also appears to give `FORCE_COLOR` priority over `NO_COLOR`, even if `NO_COLOR` is set).

Good point, this seems important too. @khasse what do you think, since you wrote https://bixense.com/clicolors/ ?

braydenkrus:
@digit-google I just commited the changes you described. Also, if we switch to `FORCE_COLOR`, the logic for the flags will change a little bit, but is not too big of an issue.

jhasse:
I think we should support FORCE_COLOR in addition to CLICOLOR_FORCE in order not to break workflows.

As for NO_COLOR: I never understood the need for it. Why not `| cat` ninja? Or set the terminal to display every color as black? Why wouldn't you want colors at all as a human?

braydenkrus:
Hello @jhasse! I agree with supporting FORCE_COLOR and CLICOLOR_FORCE together. For NO_COLOR, `| cat` should be sufficient, making support for NO_COLOR unnecessary. However, I think it could still be worth having for the sake of completeness, and I'd like to hear your thoughts on that.

braydenkrus:
Just submitted a commit that adds FORCE_COLOR functionality, and comments out NO_COLOR functionality (I wanted to keep the code in case we end up using it).

braydenkrus:
I was looking at Issue #2379, and was wondering if that has any impact on anything here.

src/line_printer.cc
```diff
@@ -166,3 +184,18 @@ void LinePrinter::SetConsoleLocked(bool locked) {
     line_buffer_.clear();
   }
 }
+
+bool LinePrinter::EnvHasNoColor() {
+    char* no_color = getenv("NO_COLOR");
+    return no_color != NULL;
```
digit-google, line 190: nit: it would be better to recognize "0" as a special value that disables the feature, as seen in the example of https://force-color.org. The same is true for the other variables.

I suggest moving this to a different file like `ansi_color.h / ansi_color.cc` since this has nothing to do with the LinePrinter class implementation. These can come up with their trivial unit-tests in `ansi_color_test.cc` as well.

braydenkrus: Just sent over a PR with these changes!

src/ansi_color.cc
```diff
@@ -0,0 +1,55 @@
...
+
+bool EnvHasNoColor() {
+  #ifdef _WIN32
+  char buf[1024];
+  DWORD len = GetEnvironmentVariableA("NO_COLOR", buf, sizeof(buf));
+  return len > 0 && std::string(buf) != "0";
+  #else
+  char* no_color = std::getenv("NO_COLOR");
+  return no_color && std::string(no_color) != "0";
+  #endif
```
jhasse: I think std::getenv is standard and also supported on Window, no?

braydenkrus: Yes, it should be. I was matching what I was using for the test cases (where I use GetEnvironmentVariableA and SetEnvironmentVariableA together, as mismatching seems to cause issues), but it should be just fine here.

jhasse (changes requested):
Can you remove the tests? They are a bit unnecessary (not checking any complex logic, just that getenv/setenv works).
Then you can also make it simpler and remove ansi_color.h/cc and move the three functions to line_printer.cc.

jhasse:
As the methods are only used in line_printer.cc and don't use any state, they can just be normal functions (put them into an anonymous namespace).

Let's just uncomment the NO_COLOR code.

braydenkrus:
I just realized that NO_COLOR wouldn't get checked under usual circumstances on UNIX (at least in the way its checked on Windows), so I sent another PR just now.

EDIT: Though now I do realize that the check for NO_COLOR in the WIN32 block is unnecessary, but I don't think its a good idea for me to send a ton of commits in a row.

jhasse:
> but I don't think its a good idea for me to send a ton of commits in a row.

that's no problem :)

jhasse (approved):
Thank you :)

PR #2764, opened 2026-04-11
Fix deps log recovery failing to remove corrupt records

### LINKED ISSUE #2703: ninja is always recompiling everything
ninja version: 1.13.2
cmake version: 4.2.1

I'm not sure if it's related but I always get a "premature end of file; recovering" error from ninja when it runs. CMake v4 bug? In any case, if ninja runs the build to completion rerunning it recompiles everything it just built. I don't know how to debug this or what you need from me to reproduce it yourself.

Ninja 1.12.1 still gives the "premature end of file" warning, but doesn't exhibit the "always rebuild everything" behavior.

skeeto:
In DepsLog::Load(), the file offset used for truncation-based recovery was advanced past each record immediately after reading its bytes from disk, but before validating the record's semantic correctness (e.g. the path checksum or node-id range checks). When a record was structurally complete but semantically invalid, recovery truncated the file to an offset that included the bad record, so it was never removed.

This created an infinite cycle: every build would load the deps log, hit the corrupt record, "recover" by truncating everything after it (but not the record itself), then append new dependency records that would be destroyed by the next build's recovery. The result was spurious rebuilds on every invocation.

Move the offset update to the end of the loop body, after all validation has passed. Now when a record fails validation, the offset still points before it, and truncation correctly removes it.

Fixes #2703

* * *

I ran into endless "premature end of file; recovering" when building CMake. So I ran Ninja under a debugger and got to the bottom of it.

### DIFF
--- a/src/deps_log.cc
+++ b/src/deps_log.cc
@@ -203,7 +203,6 @@ LoadStatus DepsLog::Load(const string& path, State* state, string* err) {
       read_failed = true;
       break;
     }
-    offset += size + sizeof(size);
 
     if (is_deps) {
       if ((size % 4) != 0) {
@@ -268,6 +267,7 @@ LoadStatus DepsLog::Load(const string& path, State* state, string* err) {
       node->set_id(id);
       nodes_.push_back(node);
     }
+    offset += size + sizeof(size);
   }
 
   if (read_failed) {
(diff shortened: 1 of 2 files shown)


nico:
Thanks for the patch!

src/deps_log_test.cc has an "invalid manifest" test. Would it be possible to add a similar test that tests for this case?

skeeto:
Good idea. I've pushed a test: 93d37d0. It fails if I revert the fix.

PR #2765, opened 2026-04-12
Fall back to $builddir/target when target lookup fails

jhasse:
When $builddir is set in the manifest, passing "foo" on the command line now checks for a target named "foo" first, then falls back to "$builddir/foo" before reporting an error.

With Ninja's configure.py this means one can use `ninja libninja-re2c.a` to build `build/libninja-re2c.a` now.

### DIFF
--- a/src/ninja.cc
+++ b/src/ninja.cc
@@ -332,7 +332,9 @@ Node* NinjaMain::CollectTarget(const char* cpath, string* err) {
 
   Node* node = state_.LookupNode(path);
   if (!node && !build_dir_.empty()) {
-    string builddir_path = build_dir_ + "/" + path;
+    // if 'foo' is not found, try to fallback to $build_dir/foo when build_dir
+    // is set in the manifest
+    std::string builddir_path = build_dir_ + "/" + path;
     uint64_t builddir_slash_bits;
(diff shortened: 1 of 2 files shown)


src/ninja.cc
```diff
@@ -331,6 +331,16 @@ Node* NinjaMain::CollectTarget(const char* cpath, string* err) {
   }
 
   Node* node = state_.LookupNode(path);
+  if (!node && !build_dir_.empty()) {
+    string builddir_path = build_dir_ + "/" + path;
```
digit-google, line 335: nit: Use `std::string` instead of `string` (we should really get rid of `using namespace std`).

src/ninja.cc
```diff
@@ -331,6 +331,16 @@ Node* NinjaMain::CollectTarget(const char* cpath, string* err) {
   }
 
   Node* node = state_.LookupNode(path);
+  if (!node && !build_dir_.empty()) {
```
digit-google, line 334: nit: Can you add a comment here explaining this like in your description (e..g "if 'foo' is not found, try to fallback to ${build_dir}/foo when build_dir is set in the Ninja manifest").

digit-google (approved):

Issue #2242, opened 2023-01-23
Add a tool to measure the time of each step


cjdb:
It'd be nice to analyse the time taken (and possibly other stats from Unix's `/usr/bin/time`) for each job that ninja completes, and have it written to a log file.

```
$ # Example
$ ninja -t time
$ cat .ninja-time.log
[1/3] clang++ ...
real:
user:
sys:
[2/3] clang++ ...
real:
user:
sys:
[3/3] ld.lld ...
real:
user:
sys:
```

Hi-Angel:
In terms of an external utility, apparently there's a [ninjatracing](https://github.com/nico/ninjatracing) tool which converts `.ninja_log` to Chromium's `about:tracing` format, so the times taken may be viewed in a browser.

githubbin765:
This can be accomplished with linux native features, a cron job. Is this something your team would still like help with?

digit-google:
If you only care about the start/end times, they are already availablein the `.ninja_log` file, which is what `ninjatracing` and other tools use. The format is basic enough you can parse it with tools like `awk` or `sed`.

Issue #267, opened 2012-04-12
Feature request -- print elapsed time


ramosian-glider:
Ninja is so fast that I always run it under the time command to see how fast it actually is.
Can ninja please print the time it's been working for itself?

Hi-Angel:
> You can use "-t stats" to print some timing information.

There does not seem to be such a tool. It says a `unknown tool 'stats'`, and doing a `ninja -t list` does not show anything that would seem to be statistics-related.

kiwixz:
The flag you are looking for is `-d stats`.
```
$ ninja -d stats
ninja: no work to do.
metric          	count 	avg (us) 	total (ms)
.ninja parse    	1     	438.0   	0.4
.ninja_log load 	1     	193.0   	0.2
.ninja_deps load	1     	1585.0  	1.6
node stat       	1078  	2.4     	2.6
node stat       	1078  	2.3     	2.5

path->node hash load 0.99 (1102 entries / 1109 buckets)
```

PR #2766, opened 2026-04-12
%q to show time spent per edge

### LINKED ISSUE #2044: NINJA_STATUS format to show time spent per edge
NINJA_STATUS has `%o` to show overall finished rate of edges/per second, `%c` to show current edges/second, and `%e` elapsed time in seconds.

These are all helpful to diagnose slow builds, however, in some instances it would be really handy to see actual seconds taken per command. For example, custom non-compiler actions that run complex workflows, very slow linker execution on _certain platforms_ and so on. Would a `%q` or something be reasonable to show this ?

braydenkrus:
This should be able to close #2044. When %q is included in NINJA_STATUS, the time spent on an edge is printed. I still need to fix the formatting, which I will do sometime soon.


braydenkrus:
I have noticed issues #2242 and #267, should I change my implementation with these in mind?

PR #2768, opened 2026-05-05
misc: Add new fuzzers to be consumed by OSS-Fuzz

DavidKorczynski:
Adds additional fuzzing harnesses to be consumed by OSS-Fuzz. The goal is to increase code coverage, for which a recent coverage report is: https://storage.googleapis.com/oss-fuzz-coverage/ninja/reports/20260503/linux/report.html

### DIFF
--- a/misc/build_log_fuzzer.cc
+++ b/misc/build_log_fuzzer.cc
@@ -1,0 +1,35 @@
+// Copyright 2026 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <stdint.h>
+#include <string>
+#include <stdio.h>
+#include <unistd.h>
+#include "build_log.h"
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  char temp_path[] = "/tmp/ninja_build_log_fuzzer.XXXXXX";
+  int fd = mkstemp(temp_path);
+  if (fd < 0) return 0;
+  
+  write(fd, data, size);
+  close(fd);
+  
+  BuildLog build_log;
+  std::string err;
+  build_log.Load(temp_path, &err);
+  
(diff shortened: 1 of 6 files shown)


misc/build_log_fuzzer.cc
```diff
@@ -0,0 +1,35 @@
...
+
+#include <stdint.h>
+#include <string>
+#include <stdio.h>
+#include <unistd.h>
+#include "build_log.h"
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  char temp_path[] = "/tmp/ninja_build_log_fuzzer.XXXXXX";
+  int fd = mkstemp(temp_path);
```
elliotgoodrich, line 23: I think fuzzing is awesome and it's great to see more use of it in ninja.

Random note, in the future with some refactoring, we can probably get the BuildLog working without needing to write to disk. Which will increase the number of iterations we can get from the fuzzer.

And just linking some work I did in this area - I was able to roundtrip items from the build log [in my fuzzing](https://github.com/elliotgoodrich/trimja/blob/e6fb3d188ea018806e28be5b2d4a5d2a82d427a1/src/logreader.fuzz.cpp#L1) to additionally check for logic errors.

misc/build_log_fuzzer.cc
```diff
@@ -0,0 +1,35 @@
...
+  if (fd < 0) return 0;
```
elliotgoodrich, line 24: If we're ever unable to use the filesystem on OSSFuzz, this will silently hide it. Is it worth aborting to flag this issue?

misc/graph_fuzzer.cc
```diff
@@ -0,0 +1,103 @@
...
+    if (it != files_.end()) {
+      *contents = it->second;
+      return Okay;
+    }
+    return NotFound;
+  }
+  int RemoveFile(const std::string& path) override { return 0; }
+
+  std::map<std::string, TimeStamp> mtimes_;
+  std::map<std::string, std::string> files_;
```
elliotgoodrich, line 47: Minor: As we don't need the ordering

```suggestion
  std::unordered_map<std::string, TimeStamp> mtimes_;
  std::unordered_map<std::string, std::string> files_;
```

misc/graph_fuzzer.cc
```diff
@@ -0,0 +1,103 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <stdint.h>
+#include <string>
+#include <vector>
+#include <map>
```
elliotgoodrich, line 18:
```suggestion
#include <unordered_map>
```

PR #2770, opened 2026-05-08
Add PID to .ninja_lock, warn if another instance is running in builddir

### LINKED ISSUE #2762: Objects marked as dirty on NFS/Lustre FS from version 1.12 on.
From version 1.12 on, Ninja grants as "dirty" object files that have juste been created.
This seems to be a timestamp issue, that can be managed with ninja -t restat, from what I investigated.
Would it be possible to have an option so that after an effective build, Ninja runs -t restat ?
Or, would it be possible to step back to Ninja 1.11.1 behaviour (and prior releases) under some option ?
I took hours to get to this point, unable to understand why Ninja did not behave as expected, recompiling all the time.
I was on the edge to forget about Ninja, while being a newbie.

jhasse:
The warning might seem like a regression, but technically the use case of running two ninjas in the same build directory was never really supported - it works accidentally and we might break it at any time.

Although not the intention of this, this should fix #2762, because now the lock file isn't empty anymore, which was a problem on some network filesystems where the mtime wasn't updated in that case.

# TODO

* [ ] Accessing the file is kind of racy
* [ ] PIDs might be reused?
* [ ] Should .ninja_lock be deleted when Ninja quits?

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -612,6 +612,27 @@ Builder::Builder(State* state, const BuildConfig& config, BuildLog* build_log,
   string build_dir = state_->bindings_.LookupVariable("builddir");
   if (!build_dir.empty())
     lock_file_path_ = build_dir + "/" + lock_file_path_;
+
+  // If a lock file from a previous (or concurrent) ninja invocation exists,
+  // see whether that process is still alive and warn the user.
+  std::string stat_err;
+  if (disk_interface_->Stat(lock_file_path_, &stat_err) > 0) {
+    std::string lock_contents;
+    std::string read_err;
+    if (disk_interface_->ReadFile(lock_file_path_, &lock_contents, &read_err) ==
+        FileReader::Okay) {
+      char* end = nullptr;
+      long pid = strtol(lock_contents.c_str(), &end, 10);
+      if (end != lock_contents.c_str() && pid > 0 &&
+          static_cast<int>(pid) != GetPid() &&
+          IsProcessRunning(static_cast<int>(pid))) {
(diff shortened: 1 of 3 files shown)


src/build.cc
```diff
@@ -612,6 +612,27 @@ Builder::Builder(State* state, const BuildConfig& config, BuildLog* build_log,
...
+    if (disk_interface_->ReadFile(lock_file_path_, &lock_contents, &read_err) ==
+        FileReader::Okay) {
+      char* end = nullptr;
+      long pid = strtol(lock_contents.c_str(), &end, 10);
+      if (end != lock_contents.c_str() && pid > 0 &&
+          static_cast<int>(pid) != GetPid() &&
+          IsProcessRunning(static_cast<int>(pid))) {
+        Warning("another ninja process (pid %ld) seems to be running in this "
+                "build directory; if it is not, delete %s",
+                pid, lock_file_path_.c_str());
```
digit-google, line 683: I think this should be a Fatal() error instead. Two concurrent Ninjas are likely to break everything, from the `.ninja_deps` and `.ninja_log` files, and the content of output files when both launch the same command that want to write to the same output using a tool that doesn't support that well (e.g. without atomically renaming the output file).

digit-google (approved):

src/build.cc
jhasse: Unfortunately ninja calling ninja again on the same build dir is used in the wild. And it "works" right now, a fatal error would be a hard regression.

When thinking about it even the warning is too much. We need to correctly fix this (the second ninja process communicates to the first process which targets it wants to run).

PR #2772, opened 2026-05-09
Add --status flag with Ninja-style variable expansion

jhasse:
Introduces a `--status FMT` command-line flag that configures the progress status using Ninja's regular `$var`/`${var}` syntax with descriptive variable names ($finished, $total, $progress, $elapsed, etc.) instead of the `%`-escapes used by `NINJA_STATUS`. When passed, it takes precedence over `NINJA_STATUS`; the env-var path is left unchanged for backwards compatibility.

Continuation of #1180, but uses the opportunity to introduce some backwards incompatibility changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@
 *.pdb
 *.ilk
 /build*/
-/build.ninja
+/build*.ninja
 /ninja
 /ninja.bootstrap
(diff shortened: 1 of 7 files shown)


digit-google (approved):
Nice!

PR #2773, opened 2026-05-11
Support OSC 9;4 progress bars in NINJA_STATUS

### LINKED ISSUE #2555: OSC ConEmu escape codes generation for progress reporting in compatible terminals
Would it make sense to emit [OSC 9;4; ANSI sequences](https://conemu.github.io/en/AnsiEscapeCodes.html#ConEmu_specific_OSC)? If the terminal supports those it can display a progress bar in its UI.

bogner:
This adds `%b` and `%B` placeholders for `NINJA_STATUS` that will generate progress bars using `OSC 9;4` sequences for percentage of finished edges and time elapsed, respectively.

The `OSC 9;4` sequences are originally from ConEmu, but are supported by various terminals on linux and mac as well at this point.

Fixes #2555

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -242,6 +242,8 @@ specified by `-j` or its default)
 `%w`:: Elapsed time in [h:]mm:ss format. _(Available since Ninja 1.12.)_
 `%W`:: Remaining time (ETA) in [h:]mm:ss format. _(Available since Ninja 1.12.)_
 `%P`:: The percentage (in ppp% format) of time elapsed out of predicted total runtime. _(Available since Ninja 1.12.)_
+`%b`:: OSC 9;4 progress bar for percentage of finished edges. _(Available since Ninja 1.14.)_
+`%B`:: OSC 9;4 progress bar for percentage of time elapsed out of predicted total runtime. _(Available since Ninja 1.14.)_
 `%%`:: A plain `%` character.
 
 The default progress status is `"[%f/%t] "` (note the trailing space
--- a/src/status_printer.cc
+++ b/src/status_printer.cc
@@ -327,6 +327,18 @@ string StatusPrinter::FormatProgressStatus(const char* progress_status_format,
         break;
       }
 
+      // OSC 9;4 progress bar for percentage of edges
+      case 'b': {
+        int percent = 0;
+        if (finished_edges_ != 0 && total_edges_ != 0)
+          percent = (100 * finished_edges_) / total_edges_;
+        int state = finished_edges_ == total_edges_ ? 0 : 1;
+
+        snprintf(buf, sizeof(buf), "\x1b]9;4;%i;%i\x7", state, percent);
+        out += buf;
+        break;
+      }
+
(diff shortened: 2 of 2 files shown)


jhasse:
Thanks!

Please add `_(Available since Ninja 1.14.)_` to both.

Also: Should we maybe add `%B` to the default?

bogner:
> Please add `_(Available since Ninja 1.14.)_` to both.

Done

> Also: Should we maybe add `%B` to the default?

Maybe? The OSC 9;4 sequence seems to be pretty widely supported, but given that it isn't standard I worry that making it the default could be a bad experience on some terminals.

kbroulik:
> but given that it isn't standard I worry that making it the default could be a bad experience on some terminals.

Flatpak has it opt in because of that, see [FLATPAK_TTY_PROGRESS](https://docs.flatpak.org/en/latest/flatpak-command-reference.html#:~:text=FLATPAK_TTY_PROGRESS)

> FLATPAK_TTY_PROGRESS
> May be set to 0 to disable reporting machine-readable progress to the terminal. This feature can be disabled because it uses the OSC 9;4 sequence, which some terminal emulators interpret as a popup notification.

Edit: hmm, older revisions of the documentation say “set 1 to enable”. Maybe it was opt in at some point but is opt out now?

Edit 2: Indeed, they made it enabled by default again https://github.com/flatpak/flatpak/pull/6393

digit-google:
> > but given that it isn't standard I worry that making it the default could be a bad experience on some terminals.
> 
> Flatpak has it opt in because of that, see [FLATPAK_TTY_PROGRESS](https://docs.flatpak.org/en/latest/flatpak-command-reference.html#:~:text=FLATPAK_TTY_PROGRESS)
> 
> > FLATPAK_TTY_PROGRESS
> > May be set to 0 to disable reporting machine-readable progress to the terminal. This feature can be disabled because it uses the OSC 9;4 sequence, which some terminal emulators interpret as a popup notification.
> 
> Edit: hmm, older revisions of the documentation say “set 1 to enable”. Maybe it was opt in at some point but is opt out now?
> 
> Edit 2: Indeed, they made it enabled by default again [flatpak/flatpak#6393](https://github.com/flatpak/flatpak/pull/6393)

Thanks for looking into this. It's interesting that they essentially waited 10 months to turn the default.

In the case of this CL, since neither formatters are part of the default NINJA_STATUS, this can be considered opt-in and safe, since it can only be enabled voluntarily by Ninja users.

digit-google (approved):
Do we need a regression test for this in output_test.py? @jhasse, wdyt?

jhasse:
> Do we need a regression test for this in output_test.py? @jhasse, wdyt?

I think it isn't needed, because most of the logic is somewhere else (i.e. the terminal parsing it).

> but given that it isn't standard I worry that making it the default could be a bad experience on some terminals.

What could happen? AFAIK unknown codes are just ignored by terminals.

So I really think this should be activated by default, otherwise nearly no one will notice it.

digit-google:
Agreed, the %b and %B  directives are not part of the default status prefix, so supporting them by default seems reasonable.

PR #2776, opened 2026-05-15
Avoid needless CTest targets

scivision:
use enable_testing() to avoid spurious test targets, as we aren't using CDash or those targets


jhasse:
I don't understand this, sorry. Can you elaborate why this is an improvement?

PR #2780, opened 2026-05-24
Binary Manifest Cache

### LINKED ISSUE #2779: Binary Manifest cache
### Summary
Introduce a binary cache for the parsed build manifest to significantly speed up subsequent builds when the manifest has not changed.

### Motivation
Parsing the ASCII manifest is a significant bottleneck — the parse function accounts for a substantial share of total build time. Two main costs:
1. **Binary format parsing** is faster than ASCII parsing
2. **Path operations** like `canonicalize()` are expensive and currently re-executed on every parse — the binary cache stores the already-resolved paths

This especially helps with **frequent rebuilds** where the manifest is unchanged (e.g. incremental development cycles).

### Proposed Behavior
- If **no binary manifest exists**: parse the ASCII manifest as usual, then write a binary cache file
- If **a binary manifest exists**: load the binary cache instead of the ASCII manifest
- If the generator produces a **new manifest**: fall back to ASCII parsing and regenerate the cache (binary cache becomes stale)

### Activation
Opt-in via a new flag:
```
ninja --cache
```

### Non-Goals
- No changes to build file syntax or semantics
- No changes to generators
- No new build features — strictly a performance optimization
- Cache is not intended to be portable across machines or architectures

### Expected Impact
- Parse step faster on cache hit
- Significant speedup for frequent rebuilds when manifest is stable

moritzx22:
# Binary Manifest Cache (`--cache`)

### Description
This PR introduces a binary cache for the Ninja build manifest to improve rebuild performance. This is a pure performance optimization with no functional changes to the build system.

Fixes: #2779

### Changes

- Parse the ASCII manifest on first run and serialize it into a binary cache file
- On subsequent runs, load the binary cache instead of re-parsing the ASCII manifest
- If the manifest is regenerated (i.e. `build.ninja` is newer than the cache), invalidate and rebuild the cache
- Introduce a new opt-in flag `--cache`
  - If the flag is not provided, caching is disabled

Example:
```console
ninja --cache
```

### Manifest file
The cache file is `.build.ninja.bin`

### Rebuild Performance (LLVM benchmark, cache hit)

```console
$ time ninja -d stats
ninja: no work to do.
metric          	count 	avg (us) 	total (ms)
.ninja parse    	1     	196055.0	196.1
...
real	0m0,221s
```

```console
$ time ninja -d stats --cache
metric             	count 	avg (us) 	total (ms)
node stat          	2169  	0.7     	1.5
.ninja parse binary	1     	91528.0 	91.5
...
real	0m0,119s
```

### Rebuild Performance (LLVM benchmark, cache miss)

```console
$ rm ./build.ninja-bin && time ninja -d stats --cache
.ninja parse         	1     	194182.0	194.2
write binary manifest	1     	65286.0 	65.3
...
real	0m0,285s
```
## Performance

Writing the binary cache introduces an overhead of approximately 35% of the ASCII parsing time.

On cache hits, the binary manifest achieves approximately **50% faster parsing** compared to the ASCII implementation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -124,6 +124,7 @@ check_platform_supports_browse_mode(platform_supports_ninja_browse)
 
 # Core source files all build into ninja library.
 add_library(libninja OBJECT
+        src/binary.cc
 	src/build_log.cc
 	src/build.cc
 	src/clean.cc
@@ -269,6 +270,7 @@ if(BUILD_TESTING)
 
   # Tests all build into ninja_test executable.
   add_executable(ninja_test
+    src/binary_test.cc
     src/build_log_test.cc
     src/build_test.cc
     src/clean_test.cc
--- a/configure.py
+++ b/configure.py
@@ -537,7 +537,8 @@ n.newline()
 
 n.comment('Core source files all build into ninja library.')
 objs.extend(re2c_objs)
-for name in ['build',
+for name in ['binary',
+             'build',
              'build_log',
              'clean',
              'clparser',
@@ -644,6 +645,7 @@ if gtest_src_dir:
         test_variables += [('pdb', 'ninja_test.pdb')]
 
     test_names = [
+        'binary_test',
         'build_log_test',
         'build_test',
         'clean_test',
--- a/src/binary.cc
+++ b/src/binary.cc
@@ -1,0 +1,627 @@
+// Copyright 2026 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "binary.h"
+
+#include <cassert>
+#include <cstdint>
+#include <iostream>
+#include <string_view>
+#include <vector>
+
+#include "eval_env.h"
+#include "graph.h"
+#include "hash_map.h"
+#include "metrics.h"
+#include "state.h"
+
+namespace {
+
+/// Increment this whenever the binary format changes incompatibly.
+constexpr uint16_t kBinaryFormatVersion = 1;
+
+// Thin wrapper holding a const reference to T. operator<< writes the value's
+// raw bytes to the stream (little-endian on x86, matching the host byte order).
+template <class T>
+struct Bin {
+  explicit Bin(const T& value) : value_(value) {}
+  const T& value_;
+};
+
+// BinVal stores by value so typed helpers can safely cast from a temporary.
+template <class T>
+struct BinVal {
+  explicit BinVal(T value) : value_(value) {}
+  T value_;
+};
+
+template <class T>
+std::ostream& operator<<(std::ostream& os, const BinVal<T>& a) {
+  os.write(reinterpret_cast<const char*>(&a.value_), sizeof(a.value_));
+  return os;
+}
+
+template <class U> BinVal<int8_t>   binInt8  (U v) { return BinVal<int8_t>  (static_cast<int8_t>  (v)); }
+template <class U> BinVal<uint8_t>  binUInt8 (U v) { return BinVal<uint8_t> (static_cast<uint8_t> (v)); }
+template <class U> BinVal<uint16_t> binUInt16(U v) { return BinVal<uint16_t>(static_cast<uint16_t>(v)); }
+template <class U> BinVal<int32_t>  binInt32 (U v) { return BinVal<int32_t> (static_cast<int32_t> (v)); }
+template <class U> BinVal<uint32_t> binUInt32(U v) { return BinVal<uint32_t>(static_cast<uint32_t>(v)); }
+template <class U> BinVal<uint64_t> binUInt64(U v) { return BinVal<uint64_t>(static_cast<uint64_t>(v)); }
+
+// Writes a length-prefixed string: uint32_t byte count followed by raw bytes.
+struct StringBin {
+  explicit StringBin(std::string_view text) : text_(text) {}
+  std::string_view text_;
+};
+
+std::ostream& operator<<(std::ostream& os, const StringBin& s) {
+  os << binUInt32(s.text_.size());
+  os.write(s.text_.data(), s.text_.size());
+  return os;
+}
+
+// Writes an EvalString token list: uint32_t count, then for each token a
+// length-prefixed string followed by a type byte ('R' = raw, 'S' = special).
+using TokenListBin = Bin<EvalString::TokenList>;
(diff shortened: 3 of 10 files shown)


digit-google:
Thanks, I have not reviewed everything in this CL because it looks like it was created with AI, which makes reviews depressing, but a few general comments around this, since this is not a bad idea in itself (for example, for Fuchsia we implemented a persistent server mode to keep the graph in memory instead, but this solution addresses the same problem in a different way).

- First, do not use C++ stdio for writing lots of small binary values. The implementation of std::ostream is *terrible* by design and cannot be improved, and you will get easily a 5% to 10% performance boost by just using a dedicated Output/InputStream classes that provides direct method to read and write fixed-size integers, path strings, etc.

- Assume that manifest files will be corrupted. Ninja should detect them and fallback to loading the real manifest instead in this case. Crashing is not acceptable. Do not use assert() at all, they are left out of production code.

- Due to this, there is little benefit in using file offsets directly to reference other graph items in the wired representation, because you *will* need some sort of map and runtime lookup to verify that each value corresponds to a real item for safety. Just using incremental unique ids for each item type instead, this will make the code simpler to understand and less fragile in case of schema modification.

- I have not seen a mechanism to verify that the binary manifest is stale (but I may have missed it). Can you clarify if that just checks the top-level manifest, or is capable of verifying also all the other .ninja file it included. This would require writing the paths and timestamps of these files directly into the manifest and be performed as an early check. Again, I may have missed it because my initial review has been very fast.

- The encoding / decoding for each of Node, Edge, Pool etc  should be dedicated methods of each class, and not be implemented in a completely separate location that will be hard to keep in sync with changes in graph.cc. This makes the implementation closer to the rest of the detail, making it less likely to break stuff. Of course extra context (like offset/id mapping tables) should be passed in addition to the Input/Output stream pointer.

I am also curious if you have metrics that compare the size of the binary manifest to that of the source text build plan files for medium to large projects.

moritzx22:
### File sizes manifest vs. binary cache
| Project | ASCII | Binary |
|---------|-------------|--------------|
| LLVM    | 57MB        | 28MB         |
| GRPC    | 5.1MB       | 4.2MB        |
| ninja     | 85KB       | 66KB            |

moritzx22:
> First, do not use C++ stdio for writing lots of small binary values. The implementation of std::ostream is terrible by design and cannot be improved, ...

I will stop using C++ iostreams. As you pointed out, their implementation is not well suited for this use case.

I'll switch to C stdio (`std::fread`/`std::fwrite`) with a 64 KB buffer to reduce the number of system calls and memory allocations. This noticeably reduces CPU time. The changes will be included in the upcoming push.

moritzx22:
> I have not seen a mechanism to verify that the binary manifest is stale (but I may have missed it). Can you clarify if that just checks the top-level manifest, or is capable of verifying also all the other .ninja file it included.

You did not miss it, good catch. This is still a TODO.

**Next push**: Write the paths of all included .ninja files to the beginning of the binary cache. Their timestamps can then be checked on every build to determine if the binary manifest is stale.

digit-google:
Before going further, it would be interesting to have @jhasse 's opinion on the relevance of this feature.

My take is that this adds a non-trivial amount of code to maintain, and the benefit (50% faster parsing time, after 35% slower parsing time on the first invocation to create the cache) seems rather "mid" to me.

By comparison, for Fuchsia we decided to implement persistent mode to solve the same problem (among other things), where the graph stays in memory between invocations. There are benefits and drawbacks to this approach too:

Pros:
- no parsing so incremental invocations start nearly immediately (for the record, a medium Fuchsia build plan is about 1.4 GiB of .ninja files, so this saves about 2.7s per invocation on powerful workstations).

Cons: 
- much higher complexity costs, which we amortize around dozens of other Fuchsia-specific features, but which may not make sense for upstream Ninja. We are also fortunate to only need to really support Linux.

- only handles depfiles, not dyndep files, because the in-memory graph needs to be restored to its "pristine" state for consistency. It is very easy to add tracking info for this for depfile-induced changes. On the other hand, the changes that dyndep files do to the graph during the build are much more invasive, and since we don't need them at all for Fuchsia, we didn't implement any back-tracking for this.

I don't think such a feature would make sense for upstream either, just mentioning that everything has trade-offs.
So @jhasse, what is you take on the binary manifest approach?

jhasse:
I agree that this adds too much complexity for too little gains. For me speed has never really been a problem, so I would also have very little incentive to maintain this.

e4lam:
Gemini pointed me at this to solve my large ninja file loads. Even just running `ninja -n build.ninja` takes like 1.8s on my machine for a no-op build on a hefty cmake project.

Is there some other solution to this where we can add some minimal upstream hooks for Ninja so that someone can try fuschia has done? I apologize if this is a bad question as I don't exactly know if fuschia's solution is generally available. I also need Linux, macOS, and Windows support.

digit-google:
> Gemini pointed me at this to solve my large ninja file loads. Even just running `ninja -n build.ninja` takes like 1.8s on my machine for a no-op build on a hefty cmake project.

FWIW, you can try:

- The [Android version of Ninja](https://android.googlesource.com/platform/external/ninja/+/refs/heads/main) which addresses this problem with a parallel parser (using a thread pool).

- The [Fuchsia version of Ninja](https://fuchsia.googlesource.com/third_party/github.com/ninja-build/ninja/+/refs/heads/main) which implements persistent mode instead.

Note that each version is only supported for their respective's projects goals (e.g. minimal or non-existent Windows support). They do not necessarily recent Ninja upstream. They are fully open-source though (you can even use Gerrit to submit patches).

There are quite a number of other Ninja forks / re-implementations that might be best suited to your need though.

> Is there some other solution to this where we can add some minimal upstream hooks for Ninja so that someone can try fuschia has done?  I apologize if this is a bad question as I don't exactly know if fuschia's solution is generally available. I also need Linux, macOS, and Windows support.

There are no custom hooks available, and they are, imho, generally a bad idea (e.g. everybody wants a different hook, evolving or fixing hook APIs without breaking stuff becomes very difficult. and Hyrum's law kicks in pretty fast). I know that @jhasse wanted to turn Ninja into a library some time ago, but I do not agree this would be useful long-term :-)

jhasse:
I'm actually not that much of a fan of ninja as a library, but I'm in favor of landing a persistent mode eventually.

moritzx22:
The persistent mode approach has better performance in terms of startup time, and it would be great to see that upstream.

However, the implementation seems to require fairly invasive changes that affect several parts of Ninja's algorithms.

The binary manifest implementation might not be completely trivial, but I don't think it is particularly complicated either, and it has a much more limited impact on Ninja's existing algorithms. Simplifying it a bit, it is mostly about serializing and deserializing objects from class `State`.

moritzx22:
Many of the comments above are incorporated.

Recent push and performance table for LLVM:
type | ASCII parse | ASCII parse without filedisk read | Binary parse | Binary parse without filedisk read | Binary write | ASCII file-size | Binary file-size
-- | -- | -- | -- | -- | -- | -- | --
ms | 190 ms  | 160 ms |  48 ms | 35 ms  |  21 ms | 57MB  |  22MB
% | 100% | 100%   |  25% | 21%  |  na |  100% |  38%

Note that other parts of Ninja might also benefit from improved cache locality. As a side effect, all `nodes` are stored in a contiguous memory area.

Issue #2769, opened 2026-05-08
Consider removing /wd4244 and /wd4267 warning suppressions for MSVC builds


Omotola:
### Summary
The MSVC build configuration in CMakeLists.txt disables warnings C4244 and C4267 via /wd4244 /wd4267. These warnings flag implicit integer conversions that can lead to data truncation (e.g., size_t → int). Removing these suppressions and fixing the underlying conversions would improve code safety and help downstream consumers meet security compliance requirements.

### Background
Several security scanning tools and compliance frameworks flag the suppression of these specific warnings as a security concern:

C4244: Conversion from 'type1' to 'type2', possible loss of data
C4267: Conversion from 'size_t' to 'type', possible loss of data
These conversions can mask integer truncation bugs that potentially lead to buffer overflows or other memory safety issues.

These conversions can mask integer truncation bugs that potentially lead to buffer overflows or other memory safety issues.

As Ninja is widely distributed (bundled with CMake, Visual Studio, various Linux distributions, CI/CD platforms), downstream consumers are increasingly required to demonstrate that shipped binaries are built with security-hardened compiler settings. Upcoming regulations like the EU Cyber Resilience Act (CRA) are also placing greater scrutiny on software supply chain security.

### Current state
[CMakeLists.txt line 28:](https://github.com/ninja-build/ninja/blob/master/CMakeLists.txt#L28)
```add_compile_options(/W4 /wd4100 /wd4267 /wd4706 /wd4702 /wd4244 /GR- /Zc:__cplusplus)```

### Proposal

- Remove /wd4244 and /wd4267 from the MSVC compile options
- Fix the resulting conversion warnings throughout the codebase (explicit casts, type changes, or static_cast where appropriate)

I'm happy to contribute a PR for this if the maintainers are open to the change. I can start with a scoping pass to estimate the number of warnings and propose fixes.

jhasse:
PR welcome :)

scivision:
These flags were added in #1876

jhasse:
ah thanks for the reminder! Yeah I think it's okay to leave those warnings for now.

Omotola:
@jhasse Sorry for the delay in creating a [PR ](https://github.com/ninja-build/ninja/pull/2781) and Thanks for the context on #1876. I see the issue was closed, and I'd like to ask you to reconsider, because the picture today is materially different from what was true in 2020.
[...]

PR #2781, opened 2026-05-26
Enable MSVC warnings C4244 and C4267 (#2769)

Omotola:
## Summary

This PR addresses #2769 by removing the `/wd4244` and `/wd4267` MSVC warning suppressions from the build configuration and fixing the underlying integer-conversion warnings throughout the codebase.

These warnings flag implicit conversions that can silently truncate data (e.g., `size_t → int`, `double → int`), which can mask bugs and are increasingly flagged by security scanners and compliance frameworks (the issue mentions the EU Cyber Resilience Act). Since Ninja is widely redistributed (CMake, Visual Studio, Linux distros, CI/CD platforms), having a clean build under these warnings makes life easier for downstream consumers that need security-hardened builds.

## What changed

**Build configuration**
- `CMakeLists.txt` — removed `/wd4244 /wd4267` from MSVC compile options
- `configure.py` — removed the matching `/wd4244` and `/wd4267` entries (kept in sync per the comment in `CMakeLists.txt`)

**Source fixes** — 16 sites across 13 files. All are minimal `static_cast` / lambda fixes; no behavioral changes except one type widening noted below.

- `src/build.h` — `RunningEdgeMap` value type changed from `int` to `int64_t`. The map stores millisecond timestamps from `GetTimeMillis()`, which is already `int64_t` at every use site; storing it as `int` was an unintended truncation.
- `src/dyndep.cc`, `src/graph.cc`, `src/missing_deps.cc` — explicit `static_cast<int>(container.size())` when accumulating sizes into `int` counters (`implicit_outs_`, `implicit_deps_`, `missing_dep_path_count_`).
- `src/status_printer.h` — explicit `int64_t → double` cast when pushing time samples into a `queue<double>`.
- `src/json.cc` — explicit `double → size_t` cast for `string::reserve(length * 1.2)`.
- `src/disk_interface.cc` — replaced `::tolower` in `std::transform` calls with an `unsigned char → char` lambda (the canonical fix for the well-known C4244 from `int(int)` in `std::transform` over `std::string`).
- `src/util.cc` — casts for the `getcwd` and `_chsize` calls (which take `int`/`long`).
- `src/real_command_runner.cc` — explicit `double → int` cast for `load_capacity`.
- `src/includes_normalize-win32.cc` — `size_t → DWORD` cast for `GetFullPathNameA`.
- `src/getopt.c` — C-style casts where `int optind` is assigned from `size_t` (`optwhere`, `permute_from`).
- `src/build_test.cc`, `src/includes_normalize_test.cc`, `src/manifest_parser_perftest.cc` — `size_t → int` casts at the test sites flagged by the warnings.

## Validation

- Configured and built with **MSVC 19.44** (Visual Studio 2022) via CMake + Ninja, Release configuration.
- After the changes, the full build produces **zero C4244 and zero C4267 warnings**.
- Ran the full unit test suite (`ninja_test.exe`): **427 of 428 tests pass**.
- The single failure (`DiskInterfaceTest.StatSymlink`) is **pre-existing on `master`** and unrelated to this PR — it requires Windows Developer Mode or admin privileges so that `CreateSymbolicLinkA` succeeds. I verified by stashing my changes, rebuilding pristine `master`, and re-running the test: it fails identically.
- The Python `misc/output_test.py` suite ran cleanly (all 22 tests skipped on Windows, which is the expected behavior — they are POSIX-only).

No regressions introduced.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,7 @@ if(MSVC)
 	string(REPLACE "/GR" "" CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
 	# Note that these settings are separately specified in configure.py, and
 	# these lists should be kept in sync.
-	add_compile_options(/W4 /wd4100 /wd4267 /wd4706 /wd4702 /wd4244 /GR- /Zc:__cplusplus)
+	add_compile_options(/W4 /wd4100 /wd4706 /wd4702 /GR- /Zc:__cplusplus)
 	add_compile_definitions(_CRT_SECURE_NO_WARNINGS)
 else()
 	include(CheckCXXCompilerFlag)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,7 @@ if(MSVC)
   string(REPLACE "/GR" "" CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
   # Note that these settings are separately specified in configure.py, and
   # these lists should be kept in sync.
-  add_compile_options(/W4 /wd4100 /wd4267 /wd4706 /wd4702 /wd4244 /GR- /Zc:__cplusplus)
+  add_compile_options(/W4 /wd4100 /wd4706 /wd4702 /GR- /Zc:__cplusplus)
   add_compile_definitions(_CRT_SECURE_NO_WARNINGS)
 else()
   include(CheckCXXCompilerFlag)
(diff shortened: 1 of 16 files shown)


digit-google (commented):
Thank you for this PR, the changes look good to me, but please rebase your commit on top of recent upstream, merges should not be part of PRs.

Omotola:
@digit-google Thanks for the review! I've rebased onto the current master. The PR now shows a single commit with no merge commits. Let me know if you'd like any other changes.

digit-google (approved):
Thanks. @jhasse please take a look.

PR #2783, opened 2026-05-28
Fix dyndep loading for multi-output dyndep generators

### LINKED ISSUE #2782: Race condition when build step has multiple dyndep outputs
Given the following ninja file:

```ninja
rule make_dyndep
  command = printf "ninja_dyndep_version = 1\nbuild a.pcm: dyndep | b.pcm\n" > a.dd; printf "ninja_dyndep_version = 1\nbuild b.pcm: dyndep | c.pcm\n" > b.dd
build a.dd b.dd: make_dyndep

rule touch
  command = touch $out
build a.pcm: touch ../a | a.dd
  dyndep = a.dd

rule touch_slow
  command = sleep .1 && touch $out
build c.pcm: touch_slow ../c

rule check_exists
  command = [ -f c.pcm ]
build b.pcm: check_exists ../b | b.dd
  dyndep = b.dd
```

Then assuming the directory is set up with

```sh
touch a b c;
mkdir build
```

then the command

```sh
ninja -C build -f ../build.ninja
```

fails with

```console
ninja: Entering directory `build'
[2/4] [ -f c.pcm ]
FAILED: [code=1] b.pcm 
[ -f c.pcm ]
[3/4] sleep .1 && touch c.pcm
```

The build graph before running `ninja` looks correct

Image

And after running ninja is also looks correct

Image

Interestingly, either of the following commands succeed

```sh
ninja -C build -f ../build.ninja -j1
```

```sh
ninja -C build -f ../build.ninja a.pcm
```

```sh
ninja -C build -f ../build.ninja b.pcm
```

The first succeeds because the builds happen to be scheduled in the correct order, which is understandable. The second build builds everything (`a.pcm` depends on all other steps) but ninja actually has the dependencies correct in that case. The third build builds enough that you would expect it to show the bug but it also succeeds. Rerunning a failing command also leads to succeed because at that point the missing file has come to exist.

If we split up the rule `make_dyndep` into two rules, the build also succeeds, as in

```ninja
rule make_dyndep_a
  command = printf "ninja_dyndep_version = 1\nbuild a.pcm: dyndep | b.pcm\n" > a.dd
build a.dd: make_dyndep_a

rule make_dyndep_b
  command = printf "ninja_dyndep_version = 1\nbuild b.pcm: dyndep | c.pcm\n" > b.dd
build b.dd: make_dyndep_b

rule touch
  command = touch $out
build a.pcm: touch ../a | a.dd
  dyndep = a.dd

rule touch_slow
  command = sleep .1 && touch $out
build c.pcm: touch_slow ../c

rule check_exists
  command = [ -f c.pcm ]
build b.pcm: check_exists ../b | b.dd
  dyndep = b.dd
```

If our rule generates a single dyndep file that multiple later steps depend on, that also succeeds:

```ninja
rule make_dyndep
  command = printf "ninja_dyndep_version = 1\nbuild a.pcm: dyndep | b.pcm\nbuild b.pcm: dyndep | c.pcm\n" > combined.dd
build combined.dd: make_dyndep

rule touch
  command = touch $out
build a.pcm: touch ../a | combined.dd
  dyndep = combined.dd

rule touch_slow
  command = sleep .1 && touch $out
build c.pcm: touch_slow ../c

rule check_exists
  command = [ -f c.pcm ]
build b.pcm: check_exists ../b | combined.dd
  dyndep = combined.dd
```

If we reverse the order of the listed outputs in `make_dyndep` so that `b.dd` is listed before `a.dd`, the build also works (but that might just be coincidence and a more complicated reproducer could still find a bug?)

```ninja
rule make_dyndep
  command = printf "ninja_dyndep_version = 1\nbuild a.pcm: dyndep | b.pcm\n" > a.dd; printf "ninja_dyndep_version = 1\nbuild b.pcm: dyndep | c.pcm\n" > b.dd
build b.dd a.dd: make_dyndep

rule touch
  command = touch $out
build a.pcm: touch ../a | a.dd
  dyndep = a.dd

rule touch_slow
  command = sleep .1 && touch $out
build c.pcm: touch_slow ../c

rule check_exists
  command = [ -f c.pcm ]
build b.pcm: check_exists ../b | b.dd
  dyndep = b.dd
```

moritzx22:
## Issue
Fixes: #2782

## Problem Analysis
When an edge produces a dyndep file, Ninja performs two steps:
 - update the dependency graph with the newly discovered dependencies
 - update the build plan and schedule additional work

For edges that generate multiple dyndep files, Ninja previously processed them one at a time by immediately updating the plan after each individual dyndep file was loaded.

This meant the scheduler could observe a partially updated dependency graph while later dyndep files from the same edge had not yet been processed. As a result, newly discovered dependencies could be missed during scheduling, leading to incorrect build ordering in parallel builds.

## Proposed Change 
This PR changes the behavior to first load and apply all generated dyndep files to the graph. Only after the graph is fully updated, the build plan is refreshed.

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -223,9 +223,8 @@ bool Plan::EdgeFinished(Edge* edge, EdgeResult result, string* err) {
   edge->outputs_ready_ = true;
 
   // Check off any nodes we were waiting for with this edge.
-  for (vector<Node*>::iterator o = edge->outputs_.begin();
-       o != edge->outputs_.end(); ++o) {
-    if (!NodeFinished(*o, err))
+  for (auto o : edge->outputs_) {
+    if (!NodeFinished(o, err))
       return false;
   }
   return true;
@@ -241,9 +240,8 @@ bool Plan::NodeFinished(Node* node, string* err) {
   }
 
   // See if we we want any edges from this node.
-  for (vector<Edge*>::const_iterator oe = node->out_edges().begin();
-       oe != node->out_edges().end(); ++oe) {
-    map<Edge*, Want>::iterator want_e = want_.find(*oe);
+  for (auto oe : node->out_edges()) {
+    map<Edge*, Want>::iterator want_e = want_.find(oe);
     if (want_e == want_.end())
       continue;
 
--- a/src/build.cc
+++ b/src/build.cc
@@ -222,6 +222,10 @@ bool Plan::EdgeFinished(Edge* edge, EdgeResult result, string* err) {
   want_.erase(e);
   edge->outputs_ready_ = true;
 
+  // If this edge provides dyndep info, load it now.
+  if (builder_ && !builder_->LoadDyndeps(edge->outputs_, err))
+    return false;
+
   // Check off any nodes we were waiting for with this edge.
   for (auto o : edge->outputs_) {
     if (!NodeFinished(o, err))
@@ -231,14 +235,6 @@ bool Plan::EdgeFinished(Edge* edge, EdgeResult result, string* err) {
 }
 
 bool Plan::NodeFinished(Node* node, string* err) {
-  // If this node provides dyndep info, load it now.
-  if (node->dyndep_pending()) {
-    assert(builder_ && "dyndep requires Plan to have a Builder");
-    // Load the now-clean dyndep file.  This will also update the
-    // build plan and schedule any new work that is ready.
-    return builder_->LoadDyndeps(node, err);
-  }
-
   // See if we we want any edges from this node.
   for (auto oe : node->out_edges()) {
     map<Edge*, Want>::iterator want_e = want_.find(oe);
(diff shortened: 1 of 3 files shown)


src/build.cc
```diff
@@ -1064,16 +1058,32 @@ bool Builder::ExtractDeps(BuildResult::CommandCompleted& result,
...
-bool Builder::LoadDyndeps(Node* node, string* err) {
-  // Load the dyndep information provided by this node.
-  DyndepFile ddf;
-  if (!scan_.LoadDyndeps(node, &ddf, err))
-    return false;
-
-  // Update the build plan to account for dyndep modifications to the graph.
-  if (!plan_.DyndepsLoaded(&scan_, node, ddf, err))
-    return false;
+bool Builder::LoadDyndeps(const std::vector<Node*>& nodes, std::string* err) {
```
davidstone, line 1061: You can make this more efficient and simpler with

```c++
struct DyndepPair {
  DyndepFile file;
  Node* node;
};

bool Builder::LoadDyndeps(const std::vector<Node*>& nodes, std::string* err) {
  std::vector<DyndepPair> dyndep_vector;

  // first update scan_ for all loaded dyndep files
  for (auto node : nodes) {
    if (!node->dyndep_pending())
      continue;
    DyndepFile ddf;
    if (!scan_.LoadDyndeps(node, &ddf, err))
      return false;
    dyndep_vector.push_back({std::move(ddf), node});
  }

  // update the plan
  for (const auto & dyndep : dyndep_vector) {
    if (!plan_.DyndepsLoaded(&scan_, dyndep.node, dyndep.file, err))
        return false;
    }
  }
  return true;
}
```

davidstone:
It would be nice to see a test to show that the issue is actually fixed, especially if we can get an end-to-end test like what I included in my issue report.

src/build.cc
moritzx22: Nice.
See next push.

moritzx22:
The next push adds a unit test using the build file from the issue report.

src/build.cc
```diff
@@ -1064,16 +1058,27 @@ bool Builder::ExtractDeps(BuildResult::CommandCompleted& result,
...
+    DyndepFile ddf;
+    if (!scan_.LoadDyndeps(node, &ddf, err))
+      return false;
 
+    dyndeps.emplace_back(node, std::move(ddf));
+  }
+
+  // update the plan
+  for (const auto& dyndep : dyndeps) {
+    if (!plan_.DyndepsLoaded(&scan_, dyndep.first, dyndep.second, err))
```
bradking, line 1079: This loop recomputes the graph readiness repeatedly for every dyndep file.

#2789 proposes an alternative fix that loads all dyndep files and then recomputes graph readiness once.

PR #2785, opened 2026-06-05
Update lockfile mtime by using utimensat() if possible

### LINKED ISSUE #2784: on tmpfs, workaround-able bug in Darwin's fopen() leads to false "dirty" estimate
\[The following report was produced by the use of artisanal methods only.]

Hi all,

The following minimal test case was reduced from a ninja build file generated during execution of one of the tests in the cmake test suite that fails when the testing directory is on a tmpfs on macOS.

With a directory that contains only (1) the following `build.ninja`:

```ninja
rule compile_c
  command = /usr/bin/cc -c $in -o $out

rule link_c
  command = /usr/bin/cc $in -o $out

build t.o: compile_c t.c
build exe: link_c t.o
```

...and (2) the source file t.c (which is just `int main(){}`), and if we do an incremental build, as we do in the 2nd command in:

```sh
ninja -v
ninja -v
```

...then the expected & desired output for the 2nd command is:

```
ninja: no work to do.
```

...but if the directory with this test case is on a tmpfs on macOS (again, starting with no output files), then the output for the 2nd command is:

```
[1/1] /usr/bin/cc t.o -o exe
```

In that 2nd run, after ninja enters:

```cpp
DependencyScan::RecomputeOutputDirty(const Edge* edge,
                                     const Node* most_recent_input,
                                     const string& command,
                                     Node* output)
```

...with `most_recent_input = t.o`, and `output = exe`, we eventually reach the
`return true;` statement in:

```cpp
  if (build_log()) {
    // [...]
    if (entry) {
      // [...]
      if (most_recent_input && entry->mtime < most_recent_input->mtime()) {
        // [...]
        return true;
      }
    }
    // [...]
  }

```
...and so we have reached the incorrect conclusion that `exe` is dirty.

Here, `entry` is the entry from `.ninja_log` that represents `exe` (so its `mtime` was originally provided by a call to stat(2) during the previous ninja run), and `most_recent_input`'s mtime was provided by a call to stat(2) of `t.o` during the current run.

The comparison between the log entry and the direct filesystem query is desired (and introduced in https://github.com/ninja-build/ninja/commit/04d886b11041bb59d01df794cce7a1e8cad2250d); the problem is that an incorrect mtime value was written to the log. So how did it get there?

In the first run, the entry was created by the call to `BuildLog::RecordCommand()` from `Builder::FinishCommand()`, which uses an mtime value that originally came from the stat(2) call in `Builder::StartEdge()` that was applied to the lock file (which is `.ninja_lock` in the build directory). The lock file is created by the first evaluation of the call to `WriteFile()` in `StartEdge()`, just before the call to stat().

This call to WriteFile(), in turn, calls fopen(".ninja_lock","wb"), and this is where things go wrong: when the lock file already exists before this call to fopen(), on a tmpfs on macOS, this call to fopen() will _NOT_ affect the st_mtime of the file. So in this case, Darwin's fopen() has deviated from [the behavior of fopen() specified in POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fopen.html):

> If mode is w, wb, w+, wb+, or w+b, and the file did previously exist, upon successful completion, fopen() shall mark for update the last data modification and last file status change timestamps of the file. 

(Note, this deviation seems to happen _only_ when the file in question is on a tmpfs. When the same thing plays out on an SSD, the above call to fopen() updates the lock file's mtime as specified in POSIX.)

Fortunately, there are a couple of workarounds:

First, we could explicitly update the lock file's mtime by using the POSIX function [`utimensat()`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/utimensat.html) (where available, as determined by `check_cxx_symbol_exists()` in `CMakeLists.txt` as we already do for ppoll()), ideally from within a new member function, `DiskInterface::UpdateFileModificationTime()`, which will use the old way of calling WriteFile() when utimensat() is not available.

The second potential workaround is to change the call to WriteFile() so that it writes a byte to the lock file.

Although it would be very amusing to fix this bug by inserting a single space character in `src/build.cc`, I would prefer the first option (i.e., I prefer the option where we use `utimensat()`), because in that case the code would be more self-documenting.

But I will create pull requests for both options soon, so that reviewers will have a concrete choice rather than a hypothetical.

JamesWidman:
\[This PR was produced by the use of artisanal methods only.]

On a tmpfs on macOS 26.5 (25F71), the use of

  DiskInterface::WriteFile(".ninja_lock", "", false)

...had no effect in all evaluations after the first, and this is because of a bug in Darwin's implementation of fopen(). In particular, in the event where a file foo already exists, and is stored on a tmpfs filesystem, the call fopen("foo", "wb") fails to update the mtime of the file foo. (According to POSIX, this call to fopen() should update the mtime.)

In this change, if the system provides the POSIX function utimensat(), we will call that instead of calling fopen().

Fixes: #2784


JamesWidman:
This is the first of two alternative fix options mentioned in #2784; ﻿the second alternative is #2786.

jhasse:
used the other PR

PR #2786, opened 2026-06-05
Update lockfile mtime by writing a byte to it

### LINKED ISSUE #2784: on tmpfs, workaround-able bug in Darwin's fopen() leads to false "dirty" estimate

JamesWidman:
\[This PR was produced by the use of artisanal methods only.]

On a tmpfs on macOS 26.5 (25F71), the use of

  DiskInterface::WriteFile(".ninja_lock", "", false)

...has no effect in all evaluations after the first, and this is because of a bug in Darwin's implementation of fopen(). In particular, in the event where a file foo already exists, and is stored on a tmpfs filesystem, the call fopen("foo", "wb") fails to update the mtime of the file foo. (According to POSIX, this call to fopen() should update the mtime.)

In this change, we work around that bug by writing a single byte to the lock file.

Fixes: #2784

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -865,7 +865,12 @@ bool Builder::StartEdge(Edge* edge, string* err) {
     if (!disk_interface_->MakeDirs((*o)->path()))
       return false;
     if (build_start == -1) {
-      disk_interface_->WriteFile(lock_file_path_, "", false);
+      disk_interface_->WriteFile(lock_file_path_, ".", false);
+      // It's necessary to write at least one byte to the lock file because
+      // otherwise, certain filesystems (e.g. tmpfs on macOS; AWS FSx mounted
+      // via NFS) expose a bug in their fopen() implementations where they
+      // fail to update the lock file's mtime. (POSIX requires that fopen()
+      // update the file's mtime when the mode is "wb".)
       build_start = disk_interface_->Stat(lock_file_path_, err);
       if (build_start == -1)
         build_start = 0;


JamesWidman:
This is the second of two alternative fix options mentioned in #2784; ﻿the first alternative is #2785.

src/build.cc
```diff
@@ -858,7 +858,7 @@ bool Builder::StartEdge(Edge* edge, string* err) {
     if (!disk_interface_->MakeDirs((*o)->path()))
       return false;
     if (build_start == -1) {
-      disk_interface_->WriteFile(lock_file_path_, "", false);
+      disk_interface_->WriteFile(lock_file_path_, " ", false);
```
jhasse: Can use a non-whitespace byte? Also please add a comment why we do this.

JamesWidman: We can do that.

I was thinking of the utimensat() solution as the serious solution, and write-one-byte solution as the funny solution, so I selected ASCII code point 32 (decimal) because [Mel Brooks et al scientifically determined that 32 is the funniest number](https://www.youtube.com/watch?v=Mk8tuH0uwuo).

(To clarify: although it's funny that it works, I have seriously tested it and made sure that it actually works.)

But if write-one-byte is the serious solution then we can use a more serious number, so...

NUL (ASCII code point 0)?

JamesWidman: Or maybe a code point for a glyph?

The first such character is '!' (U+0021 EXCLAMATION MARK).

jhasse: I would use something simple and visible. `.` or `n`. `!` is also fine (the unicode stuff is confusing me though, you mean simply !, right?)

JamesWidman: Correct; I'll update it to `.` shortly (and add the explanatory comment).

JamesWidman:
Sorry for the lag; the process of reproducing [test failures like this](https://github.com/JamesWidman/ninja/actions/runs/31924879178/job/95110758763):

```
../../misc/jobserver_test.py .F...                                       [ 15%]
[...]
=================================== FAILURES ===================================
_____________ JobserverTest.test_jobserver_client_with_posix_fifo ______________

self = <jobserver_test.JobserverTest testMethod=test_jobserver_client_with_posix_fifo>

    @unittest.skipIf(_PLATFORM_IS_WINDOWS, "These test methods do not work on Windows")
    def test_jobserver_client_with_posix_fifo(self):
>       self._run_client_test([sys.executable, "-S", _JOBSERVER_POOL_SCRIPT])

../../misc/jobserver_test.py:329: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../misc/jobserver_test.py:273: in _run_client_test
    self.assertEqual(max_overlaps, task_count)
E   AssertionError: 2 != 4
```

...turns out to be a little slippery. But I finally managed to reproduce it both _with_ and also _without_ the proposed change of this PR.

Reproduction requires a pytest run to happen concurrently with multiple simultaneous runs of `ninja_test`. That seems to be why the failure occurs in only _some_ of github's automatic executions of the tests: sometimes, a sufficient number of instances of ninja_test complete before any of the containers start executing `misc/jobserver_test.py`, in which case the failure condition is not triggered and the tests appear to pass. 

For reproduction purposes, we can make a temporary edit in `.github/workflows/linux.yml` to the unit test run like so:

```yaml
      - name: Unit tests (GCC, Debug)
        run: for i in {1..10}; do ./build-gcc/Debug/ninja_test  --gtest_brief=1; done
```

...which will keep the unit tests running long enough for us to manually execute pytest a few times. Then we execute the workflow locally with the following [act](https://github.com/nektos/act) invocation:

```sh
act pull_request -r -j test -W '.github/workflows/linux.yml' --container-architecture linux/arm64
```

(The `-r` option is helpful as it'll keep the containers running, so that for subsequent runs, you can comment out the invocations of apt and cmake.)

And then, in one of the running containers, cd to `build-gcc/Debug` and:

```sh
pytest-3 --exitfirst -k test_jobserver_client_with_posix_fifo --color=yes ../..
```

You'll want to execute that pytest-3 invocation at a time when the status messages from `act` indicate that all containers are in the process of running the unit tests. On my machine (M2 Macbook pro, macOS 26.5.2 (25F84)), the pytest failure doesn't reproduce 100% of the time, but it's somewhere in the ballpark of 80% to 90%.

EDIT: the reproduction rate drops somewhat when the pytest run occurs in a container where no unit tests are being run simultaneously (but you still need _other_ containers that are simultaneously running the unit tests). E.g. when a container is running only this:

```sh
while true; do pytest-3 --exitfirst -k JobserverTest --color=yes ../..; done
```

...while the other containers are running the unit tests, about 20% of the pytest runs reproduce some failure. (I see about 20% failure regardless of whether the change from this PR is applied.)

I would have dug deeper than this, but since the failures can be reproduced _without_ the change proposed in this PR, I'm happy to hand this over to the jobserver experts.

digit-google:
Thank you for the thorough investigation. Unfortunately, I don't think  working-around the issue by changing / probing the container configuration is going to reliably get rid of the problem.

The issue is that these tests are *flaky by nature*, because they depend on the host kernel to implement process scheduling in a specific way (which corresponds to the usual case of lightly loaded development machine when the tests is normally run).

And flaky tests are really the root of all development frustration :-(

I can only see two practical ways to deal with this:

1) Do not run these tests in CI by default, for example by detecting `CI=1` or equivalent in the environment. They can run locally during development. This is not ideal but gets rid of the flakiness.

2) Build a version of Ninja that mocks the `SubprocessSet` class at runtime, to return deterministic results. For example these could come from a configuration file that a environment variable like `NINJA_MOCK_SUBPROCESS_CONFIG` could point to (existence of the value would trigger the behavior at runtime). Then run the tests with it only.

Ideally, support for this should *not* go into the Ninja release binary, but we could imagine a CI builder configuration to just enable this mode, and just run the corresponding tests. This would allow more regression tests to verify Ninja behavior into a mocked process environment, that could all run in a container with a single CPU core :)

That's a lot more work, obviously, but I think everything else is going to hide the problem again.

Wdyt?

jhasse:
can we move that discussion into a new issue?

PR #2788, opened 2026-06-08
Fix hyphenation and correct terminology in documentation

Broken-Diamond:


doc/manual.asciidoc
```diff
@@ -1088,9 +1088,9 @@ build myapp.exe: link a.obj b.obj [possibly many other .obj files]
 [[ref_rule_command]]
 Interpretation of the `command` variable
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Fundamentally, command lines behave differently on Unixes and Windows.
+Fundamentally, command lines behave differently on Unix and Windows.
```
braydenkrus, line 1140: Should Unix instead be written as UNIX? Though I have seen both used interchangably so I'm not sure which one would be the most correct choice.

doc/manual.asciidoc
braydenkrus: Apple uses macOS now instead of Mac OS X, so this can probably be changed.

braydenkrus: To stay consistent with the descriptions of the other rule variables, Used should be lowercase.

braydenkrus: Windows OS can be simplified to Windows.

braydenkrus (changes requested):
Your changes look good! I just went through and took note of anything else that could be corrected in the documentation.

doc/manual.asciidoc
Broken-Diamond: As far as I can tell, "UNIX" (the trademark) is used mainly for the original UNIX implementation. "Unix" is most commonly used now to specify any Unix-like system (BSDs, macOS, Linux, etc.). I'm good with whatever you'd prefer.

braydenkrus: In that case, Unix can be left as is.

digit-google:
The final change is good, but please squash this into a single commit rebased on top of current tip-of-tree.

Broken-Diamond:
There! I just went through and took note of anything else that could be corrected in the documentation.

Broken-Diamond:
Sorry, typo correction :-)

Broken-Diamond:
I squashed the changes into a single commit and rebased onto the current tip-of-tree. The new head is `8a1b0c9`, and CI is passing (unsurprisingly, because I didn't modify code). Ready for re-review.

digit-google (approved):
Thank you, @jhasse ptal

braydenkrus (approved):
Looks good to me!

PR #2789, opened 2026-06-09
dyndep: fix handling of multiple dyndep file outputs on one edge

### LINKED ISSUE #2782: Race condition when build step has multiple dyndep outputs

bradking:
Normally an edge with a dyndep binding cannot be scheduled until its dyndep file is loaded because we require the dyndep file to be an input. However, in the case that one edge produces multiple dyndep files as outputs, none of their dependent edges is blocked on its dyndep input because all the dyndep files are ready.  Load all of the dyndep files before recomputing graph readiness to avoid mistakenly scheduling an edge that appears to be ready only because its dyndep information has not been loaded yet.

Fixes: #2782

### DIFF
--- a/src/build.h
+++ b/src/build.h
@@ -19,6 +19,7 @@
 #include <map>
 #include <memory>
 #include <string>
+#include <unordered_map>
 #include <vector>
 
 #include "build_result.h"
@@ -85,9 +86,11 @@ struct Plan {
   void PrepareQueue();
 
   /// Update the build plan to account for modifications made to the graph
-  /// by information loaded from a dyndep file.
-  bool DyndepsLoaded(DependencyScan* scan, const Node* node,
-                     const DyndepFile& ddf, std::string* err);
+  /// by information loaded from a set of dyndep files.
+  bool DyndepsLoaded(DependencyScan* scan,
+                     const std::vector<Node*>& dyndep_nodes,
+                     const std::unordered_map<Edge*, Dyndeps>& dyndep_edges,
+                     std::string* err);
 
   /// Enumerate possible steps we want for an edge.
   enum Want
@@ -104,7 +107,9 @@ struct Plan {
 
  private:
   void ComputeCriticalPath();
-  bool RefreshDyndepDependents(DependencyScan* scan, const Node* node, std::string* err);
+  bool RefreshDyndepDependents(DependencyScan* scan,
+                               const std::vector<Node*>& dyndep_nodes,
+                               std::string* err);
   void UnmarkDependents(const Node* node, std::set<Node*>* dependents);
   bool AddSubTarget(const Node* node, const Node* dependent, std::string* err,
                     std::set<Edge*>* dyndep_walk);
@@ -239,8 +244,8 @@ struct Builder {
     scan_.set_build_log(log);
   }
 
-  /// Load the dyndep information provided by the given node.
-  bool LoadDyndeps(Node* node, std::string* err);
+  /// Load the dyndep information provided by the given edge's outputs.
+  bool LoadDyndeps(Edge* edge, std::string* err);
 
   State* state_;
   const BuildConfig& config_;
--- a/src/build.cc
+++ b/src/build.cc
@@ -222,6 +222,11 @@ bool Plan::EdgeFinished(Edge* edge, EdgeResult result, string* err) {
   want_.erase(e);
   edge->outputs_ready_ = true;
 
+  // Load dyndep info provided by this edge's outputs.
+  if (builder_ && !builder_->LoadDyndeps(edge, err)) {
+    return false;
+  }
+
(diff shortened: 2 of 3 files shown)


src/build.h
```diff
@@ -85,9 +85,11 @@ struct Plan {
   void PrepareQueue();
 
   /// Update the build plan to account for modifications made to the graph
-  /// by information loaded from a dyndep file.
-  bool DyndepsLoaded(DependencyScan* scan, const Node* node,
-                     const DyndepFile& ddf, std::string* err);
+  /// by information loaded from a set of dyndep files.
+  bool DyndepsLoaded(DependencyScan* scan,
+                     const std::vector<Node*>& dyndep_nodes,
+                     const std::map<Edge*, Dyndeps>& dyndep_edges,
```
davidstone: Is there a reason to use `std::map` instead of `std::unordered_map`?

bradking: Only that I was used to pre-c++11 development in Ninja.  Updated, thanks.

PR #2312, opened 2023-08-02
Update status on edge finished

ecatmur:
In the terminal, when an edge finishes update the status to display the first started (longest running) action. Usually this will be immediately overwritten with the next started action but if there is a bottleneck in the build or if it is reaching the end, this ensures that the console displays status of a build that is actually running and not one that has completed.
This ensures that the user attributes any delay to the (or an) action that is actually causing the slowdown and not unfairly to an action that has already finished just because it happened to be the last started.

To make this clear, also change the default NINJA_STATUS to include current actual parallelism level - this will reduce from number of cores to 1 as the bottleneck or end of build approaches.

Implementation stores currently running nodes in a deque, with NULL for interior completed nodes. This should be fairly efficient and allows future extension to e.g. display multiple parallel action names, as in #2249 and @orgads prototype for #111.

Partially resolves #111 (should be enough for most purposes).

PR #2791, opened 2026-06-13
Update status on edge finished

### LINKED ISSUE #111: Show progress when waiting
When we've been waiting for commands for more than a second with no output, print some "still waiting for xyz" text.

orgads:
In the terminal, when an edge finishes update the status to display the first started (longest running) action.

Usually this will be immediately overwritten with the next started action but if there is a bottleneck in the build or if it is reaching the end, this ensures that the console displays status of a build that is actually running and not one that has completed.

This ensures that the user attributes any delay to the (or an) action that is actually causing the slowdown and not unfairly to an action that has already finished just because it happened to be the last started.

To make this clear, also change the default NINJA_STATUS to include current actual parallelism level - this will reduce from number of cores to 1 as the bottleneck or end of build approaches.

Implementation stores currently running nodes in a deque, with null for interior completed nodes. This should be fairly efficient and allows future extension to e.g. display multiple parallel action names, as in

Partially resolves #111 (should be enough for most purposes).

### DIFF
--- a/misc/output_test.py
+++ b/misc/output_test.py
@@ -182,11 +182,11 @@ build b: echo
 build c: echo
   delay = 1
 ''', '-j3'),
-'''[1/3] echo c\x1b[K
+'''[1*3/3] echo c\x1b[K
 c
-[2/3] echo b\x1b[K
+[2*2/3] echo b\x1b[K
 b
-[3/3] echo a\x1b[K
+[3*1/3] echo a\x1b[K
 a
 ''')
 
@@ -199,7 +199,7 @@ build a: echo
 '''
         # Only strip color when ninja's output is piped.
         self.assertEqual(run(print_red),
-'''[1/1] echo a\x1b[K
+'''[1*1/1] echo a\x1b[K
 \x1b[31mred\x1b[0m
 ''')
         self.assertEqual(run(print_red, pipe=True),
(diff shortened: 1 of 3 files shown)


orgads:
Rebased version of:
- #2312 

See the discussion there.

I kept the original author name. Not sure if you have a policy for this kind of cases.

jhasse:
We don't do this for performance reasons. If a build has many edges it slows it down.

I think we need to implement some kind of throttling first before we can do this.

Also see #2516, might be the better way to fix this.

Regarding the NINJA_STATUS default change: That's a no-go without prior discussion.

13steinj:
>  If a build has many edges it slows it down.

How many is "many?" I didn't notice on what I thought was a pretty large build, but I could be overestimating scale here.

> Regarding the NINJA_STATUS default change: That's a no-go without prior discussion.

I told this to the original author too; I was afraid people rely on the default output format for scripts / grep. There is something to be said to knowing how many build actions are actively in flight though. Might want to explicity document some statusline examples?

jhasse:
> How many is "many?" I didn't notice on what I thought was a pretty large build, but I could be overestimating scale here.

I can't remember the issue and the number, but it was reported after we had a very similar change like this PR in the past.

PR #2793, opened 2026-06-13
Auto-detect system gtest on debian variants

orgads:


jhasse:
We shouldn't implement a whole build system in configure.py. If you want such detection, use the CMake build.

PR #2794, opened 2026-06-13
Initialize Options

orgads:


orgads:
I rebased #2237 and found this. Will you reconsider accepting #2237?

orgads:
Nevermind. I found that it is initialized by:
```c++
Options options = {};
```

PR #2795, opened 2026-06-13
Set C++17 requirement globally in CMake project

orgads:


jhasse:
why?

orgads:
To simplify and deduplicate.

jhasse:
AFAIK `cxx_std_17` is best practise. Let's keep it as is, as it isn't worth the risk to save 1 line.

PR #2796, opened 2026-06-14
Tests: Use context manager for Popen

orgads:
```
ResourceWarning: unclosed file <_io.BufferedReader name=3>
```


jhasse:
Why doesn't python do this automatically? Seems dirty to me (as a C++ programmer).

orgads:
🤷🏻‍♂️

orgads:
Here's a cleaner solution.

PR #2798, opened 2026-06-14
Add opt-in multi-line real-time build status display

### LINKED ISSUE #111: Show progress when waiting

orgads:
Add an opt-in `-m` status mode that shows currently running edges in a multi-line real-time status area on ANSI-capable terminals, while preserving the existing single-line output when the flag is not used.

Each status line shows the edge sequence number, description, and elapsed time. The multiline display is capped to the terminal height and 8 rows, and its width is clamped to 60 columns. Long job titles are trimmed from the left so the right side remains visible.

The last finished edge is kept as a temporary line above the active status block, overwritten by the next completion. Error output and command output are printed cleanly after clearing the status area.

Key changes:
- Add `-m` to enable multiline status explicitly.
- Keep default output unchanged when `-m` is not given.
- Add 100ms timeout to ppoll/pselect/IOCP so DoWork() returns periodically and refreshes status via Report().
- Detect terminal dimensions to cap status lines at min(rows - 1, 8).
- Clamp multiline width to 60 columns and trim long titles from the left.
- Track each edge's start order via Edge::sequence_ for per-line numbering.
- Overwrite the display block in place to reduce flicker.
- Show `... and N more` when running edges exceed the display limit.

Fixes #111.

### DIFF
--- a/misc/output_test.py
+++ b/misc/output_test.py
@@ -392,6 +392,57 @@ ninja: build stopped: subcommand failed.
                      flags="--status '<${finished}/${total}> '")
         self.assertEqual(output, '<1/1> echo a\x1b[K\ndo thing\n')
 
+    def test_multiline_flag(self) -> None:
+        long_title = 'component/subdir/very/long/job/title/that/should/keep/the/right_side/of/the_name'
+        output = run(
+            f'''rule echo
+  command = printf "done"
+  description = {long_title}
+
+build a: echo
+''',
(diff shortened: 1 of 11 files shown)


orgads:
It is very slow on Windows. I'll see if I can improve it. Changing to draft for now.

PR #2807, opened 2026-06-30
Musl Static Linux Binaries

Disservin:
Currently there only exists a glibc linked release for linux amd64/arm64, it'd be useful to also distribute statically linked musl builds.

Ran into a clang-tidy warning about `const uint8_t *p=(const uint8_t *)key;` in rapidhash, wasn't sure what the preferred solution is regarding third party packages, so I just excluded clang-tidy from checking thirdypart dir.

See test release at https://github.com/Disservin/ninja/releases (only lists the musl builds I canceled the other runs).
Corresponding Workflow Run https://github.com/Disservin/ninja/actions/runs/28441529239

Doesn't run address sanitizer because it's not available for alpine.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -19,4 +19,4 @@ WarningsAsErrors: '
   ,cppcoreguidelines-pro-type-cstyle-cast,
   ,bugprone-assignment-in-if-condition,
 '
-ExcludeHeaderFilterRegex: 'third_party/|generated/'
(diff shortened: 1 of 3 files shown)


jhasse:
not something we want to officially support, sorry.

Disservin:
> not something we want to officially support, sorry.

no problem 👍

PR #2809, opened 2026-07-01
Fix signature mismatch in `StatusFormatEnv`

elliotgoodrich:
`StatusFormatEnv` was added as an implementation of `Env`, but at the same time the signature of `LookupVariable` was changed.  This caused a compilation failure when both were merged.

### DIFF
--- a/src/status_printer.cc
+++ b/src/status_printer.cc
@@ -48,7 +48,7 @@ struct StatusFormatEnv : public Env {
   const string& description;
   StatusFormatEnv(const StatusPrinter* p, const string& d)
       : printer(p), description(d) {}
-  string LookupVariable(const string& var) override {
+  string LookupVariable(StringPiece var) override {
     if (var == "description")
       return description;
     return printer->FormatStatusVariable(var);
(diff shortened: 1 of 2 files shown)


src/status_printer.cc
```diff
@@ -506,7 +506,7 @@ string StatusPrinter::FormatStatusVariable(const string& name) const {
     return buf;
   }
 
-  Fatal("unknown variable '%s' in --status format", name.c_str());
+  Fatal("unknown variable '%s' in --status format", name.AsString().c_str());
```
elliotgoodrich, line 509: Potentially this is an extra string copy if we would have passed a `std::string` to `FormatStatusVariable`, but it is on the exceptional path so speed should not be an issue.

elliotgoodrich:
@jhasse for fixing `master`

src/status_printer.cc
digit-google: Agree that this doesn't matter in a case of a fatal crash :)

digit-google (approved):
Thanks, @jhasse your turn :)

PR #2811, opened 2026-07-05
Optimize `EvalString`

elliotgoodrich:
Change the implementation of `EvalString` from
`std::vector<std::string>` (or `std::string` for 1-token strings) to just a `std::string`, where tokens are stored contiguously, prefixed by their size.

This implementation has a few performance improvements:
  * Smaller memory footprint through reduction in `sizeof(EvalString)` and total count and size of heap allocations
  * Better cache locality of tokens
  * Retains all capacity on `Clear()`

Since all capacity can be retained in `Clear()`, it is possible to reuse `EvalString` objects and remove the need for
many allocations.  Future commits will reuse `EvalString` objects within `ManifestParser` to avoid allocations.

The implementation of `EvalString` has been adapted from the one here https://github.com/elliotgoodrich/trimja/blob/e6fb3d188ea018806e28be5b2d4a5d2a82d427a1/src/evalstring.h


elliotgoodrich:
This is part 2/3 of splitting up this large PR https://github.com/ninja-build/ninja/pull/2696. 

Part 1/3 has landed in https://github.com/ninja-build/ninja/pull/2716

src/eval_string.cc
```diff
@@ -0,0 +1,179 @@
...
+  return in & ~leadingBit;
+}
+
+EvalString::Offset setLeadingBit(EvalString::Offset in) {
+  return in | leadingBit;
+}
+
+bool hasLeadingBit(EvalString::Offset in) {
+  return in >> (std::numeric_limits<EvalString::Offset>::digits - 1u);
+}
```
digit-google, line 52: nit: The PR looks good but I would recommend letting the compiler handle the bit twiddling for clarity, with helper methods to encapsulate pointer casts. I.e. use a dedicated struct with bitfields such as:

```
struct TokenHeader {
  bool is_special : 1;
  size_t size : 31;

  static TokenHeader MakeText(size_t size) {
    TokenHeader result = { false, size };
    if (result.size != size)
      Fatal("token size too large: %zu", size);
    return result;
  }

  static TokenHeader MakeSpecial(size_t size) {
    TokenHeader result = { true, size };
    if (result.size != size)
      Fatal("token size too large: %zu", size);
    return result;
  }

  const char* AsConstCharPtr() const {
    return reinterpret_cast<const char*>(this);
  }

  char* AsCharPtr() const {
    return const_cast<char*>(AsConstCharPtr());
  }

  const TokenHeader* FromConstCharPtr(const char* ptr) {
    return reinterpret_cast<const TokenHeader*>(ptr);
  }

  TokenHeader* FromCharPtr(char* ptr) {
    return reinterpret_cast<TokenHeader*>(ptr);
  }
};
```

Then the rest will be much easier to read and understand.

src/eval_string.cc
```diff
@@ -0,0 +1,179 @@
...
+
+EvalString::const_iterator EvalString::end() const {
+  return const_iterator(tokens_.data() + tokens_.size());
+}
+
+void EvalString::AddText(StringPiece text) {
+  assert(!text.empty());
+  if (last_text_segment_len_ > 0) {
+    // If the last part was raw text we can extend it.
+    const EvalString::Offset newLength = last_text_segment_len_ + text.size();
```
digit-google, line 134: nit: this code does work because of the unstated detail that non-special segments do not have the leading bit set.  Either add a comment to make this clearer, or use a dedicated struct as suggested above.

src/eval_string.h
```diff
@@ -0,0 +1,105 @@
...
+
+#include <cstddef>
+#include <iterator>
+#include <string>
+#include <utility>
+
+#include "string_piece.h"
+
+/// A tokenized string that contains variable references.
+/// Can be evaluated relative to an Env.
```
digit-google: nit: Clarify that each EvalString can be seen as a sequence of (StringPiece, TokenType) pairs that can be iterated with begin() / end() methods.

digit-google (approved):

src/eval_string.cc
elliotgoodrich: I should have mentioned I tested this out after the last PR and it won't work on MSVC as it [doesn't guarantee 100% bit field packing](https://learn.microsoft.com/en-us/cpp/cpp/cpp-bit-fields?view=msvc-170) ([godbolt](https://godbolt.org/z/z6aTnEKve)).

src/eval_string.cc
elliotgoodrich: There's some documentation about this at the top of the source file, but I have improved this comment so users don't need to go looking for it.

src/eval_string.h
elliotgoodrich: I have updated the documentation to include this information - hopefully it's better now.

src/eval_string.cc
elliotgoodrich: The alternative is to store it as a `std::size_t` and a `bool` and use an extra byte per token. That way there's be no bit manipulation at the cost of a few extra bytes.

Otherwise I can't think of a way around the bit stuff.

digit-google: Oh my, thank you.  Let's forget about this for now then. What a sad joke.

src/eval_string.cc
digit-google: nit: I don't see the updated comment.

digit-google (approved):

src/eval_string.cc
elliotgoodrich: I just extended the comment here to this

```cpp
    // If the `last_text_segment_len` is positive, the last token was raw
    // text and we can just extend it.
```

and there is the one at the top of the source file

```cpp
// ... last_text_segment_len_ has the length of the last
// text section or 0 if the last section was not text. This allows us to jump
// back to the last token to extend it.
```

aljwfyabwshaj98-crypto (commented):
src/eval_string.cc

Issue #2709, opened 2026-01-13
Better support for globbing in generators


jhasse:
Globbing support in CMake:
```cmake
file(GLOB SOURCES src/*.cpp )
add_executable(foo ${SOURCES})
```
This results in two problems:
* When removing a .cpp file: Ninja fails because an input is missing.
* When adding a .cpp file: Linker error because of missing symbols. Or even worse, incremental build succeeds, but clean build would fail.

There's a workaround:
```cmake
file(GLOB SOURCES CONFIGURE_DEPENDS src/*.cpp )
add_executable(foo ${SOURCES})
```
But it's slow and causes other issues, e.g. #2507 and #2707.

I propose we implement a new attribute to the build manifest:
```ninja
glob = src/*.cpp
```

Ninja would then rerun that rule if glob changes. The result of the glob wouldn't be available.

Also I would suggest that we only support `*` and nothing else. Also we're always case-sensitive.
This simplifies the implementation.
Of course that makes the implementation not correct 100% of the time (at least with CMake), but currently it works 0% of the time. And perfect is the enemy is good.

If someone wanted to match CMake's exact behavior he could still use `CONFIGURE_DEPENDS`. Also I think that globbing is only used by pragmatists anyway ;)

mathstuf:
Is `*/.cpp` a typo? At first glance, it looks like "any file named `.cpp` in any (immediate?) directory". But perhaps there is another interpretation meant?

Note that CMake wouldn't be able to use this anyways without accepting globs directly in the source list in some way. At the time of the `file(GLOB)`, CMake has no idea what will be done with the results, and at the expansion site, there's no notion of the values coming from a `file(GLOB)`, nevermind what the glob pattern was. Given that there's no notion of a build rule to attach it to at generate time, even if this were to land, there is non-trivial design work on CMake's side to actually *use* it.

jhasse:
> Is */.cpp a typo?

Yes, sorry. Fixed it, thx :)

> At the time of the file(GLOB), CMake has no idea what will be done with the results, and at the expansion site, there's no notion of the values coming from a file(GLOB), nevermind what the glob pattern was.

Ah good point. Hm ... I guess then it could only be added for `CONFIGURE_DEPENDS`. Not sure if it's worth it then (because it might even be incorrect compared to the current implementation).

mathstuf:
[...]

PR #2815, opened 2026-07-22
Allow directory as inputs, i.e. input paths ending in /

jhasse:
See #2709.

### DIFF
--- a/doc/manual.asciidoc
+++ b/doc/manual.asciidoc
@@ -1176,6 +1176,28 @@ express the implicit dependency.)
 File paths are compared as is, which means that an absolute path and a
 relative path, pointing to the same file, are considered different by Ninja.
 
+[[dir_inputs]]
+Directory inputs
+~~~~~~~~~~~~~~~~
+
+_Available since Ninja 1.14._
+
+If an input path on a build line ends with a path separator (`/`), the input is
+treated as a directory rather than a regular file. When Ninja checks whether the
+build edge is up-to-date, it stats the directory itself and compares its
+modification time against the outputs.
+
+Most filesystems update a directory's modification time whenever entries are
+added to or removed from it, so a directory input is a convenient way to
+re-run a command when the *set* of files in a directory changes (for example,
+to regenerate an index when new source files are added). Changes to the
+contents of individual files inside the directory do not normally update the
+directory's modification time, so they will not by themselves trigger a rebuild
+- list the affected files explicitly (or via a depfile) if that is desired.
+
+A path with a trailing separator that exists but is not a directory is treated
+as missing.
+
 [[validations]]
 Validations
 ~~~~~~~~~~~
--- a/src/graph.h
+++ b/src/graph.h
@@ -110,6 +110,12 @@ struct Node {
     generated_by_dep_loader_ = value;
   }
 
+  /// True if this node refers to a directory (path was given with a trailing
+  /// slash in the manifest). When stat()ed, the directory's mtime is used to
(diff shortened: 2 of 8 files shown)


doc/manual.asciidoc
```diff
@@ -1176,6 +1176,28 @@ express the implicit dependency.)
...
+[[dir_inputs]]
+Directory inputs
+~~~~~~~~~~~~~~~~
+
+_Available since Ninja 1.14._
+
+If an input path on a build line ends with a path separator (`/`), the input is
+treated as a directory rather than a regular file. When Ninja checks whether the
+build edge is up-to-date, it stats the directory itself and compares its
+modification time against the outputs.
```
digit-google, line 1188: I think this phrase is confusing because Ninja *already* stats directory paths if there is no trailing slash. And I am sure that someone is using this in their build system, even if it is a fragile thing to do, so this behavior cannot be changed lightly.

A better phrasing might explain what the trailing slash does, e.g.:

"""
An input path with a trailing separator (e.g. `foo/bar/`) must always point to a valid directory, otherwise Ninja will complain with an error. By contrast, paths without it (e.g. `foo/bar`) can point to either a file or a directory.

In both cases, the timestamp used by Ninja corresponds to the file-system's directory entry itself (which does not change when the files within it are modified).
```

File: src/graph.h [hunk 2]
```diff
@@ -159,6 +165,11 @@ struct Node {
   /// can be loaded before the manifest.
   bool generated_by_dep_loader_ = true;
 
+  /// True if this node refers to a directory rather than a regular file.
+  /// Set when a manifest input/output path ends with a trailing slash. When
+  /// stat()ing, the path is treated as a directory and its mtime is used.
+  bool is_directory_ = false;
```
digit-google, line 171: nit: there is no need for a boolean flag. Just use `path_.back() == '/'` here.

doc/manual.asciidoc
```diff
@@ -1176,6 +1176,28 @@ express the implicit dependency.)
...
+Most filesystems update a directory's modification time whenever entries are
+added to or removed from it, so a directory input is a convenient way to
+re-run a command when the *set* of files in a directory changes (for example,
+to regenerate an index when new source files are added). Changes to the
+contents of individual files inside the directory do not normally update the
+directory's modification time, so they will not by themselves trigger a rebuild
+- list the affected files explicitly (or via a depfile) if that is desired.
+
+A path with a trailing separator that exists but is not a directory is treated
+as missing.
```
digit-google, line 1199: This doesn't seem to correspond to the implementation. If the path points to a file, it will be reported as an error ??

src/graph.h
digit-google: Actually scratch that, it depends if you want to treat `foo/` and `foo` in the same manifest as pointing to the same graph Node. From this code it looks like that if there is at least one use of `foo/` in the manifest for an input path, then the Node for `foo` is marked as a directory, even if other uses are not.

This is quite subtle, and also doesn't happen if `foo` is an output, or comes from a depfile, or a dyndep file.

It would probably make more sense to just check for the trailing separator when adding a new Node instead to cover all cases, otherwise this will lead to very hard-to-debug inconsistencies.

doc/manual.asciidoc
digit-google: I must say that I fail to see the benefit of this PR. I assumed it is to prepare for glob support but that would require another non-trivial change of logic in Ninja, so why introduce it here and not as a preliminary commit in a PR that does that?

PR #2821, opened 2026-08-09
Performance improvements RecomputeOutputsDirtyCache

moritzx22:
This PR adds several performance improvements to class `RecomputeOutputsDirtyCache`, see PR #2680. There are no functional changes.

- **Lazy allocation of `logEntry_`**
  - The cache vector is now allocated on first use in all() instead of unconditionally in the constructor, and is skipped entirely for phony edges, which never touch it.
- **Fast path for single-output edges**
  - Many edges have exactly one output
- **Lazy computation of `isRestat_`**
- **Shrink `CachedLogEntry` from 16 to 8 bytes**
  - Noticeable especially on edges with many outputs.

### DIFF
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -179,7 +179,7 @@ bool RecomputeOutputsDirtyCache::depfile(const Node* most_recent_input) {
   // Precondition: RecomputeOutputsDirtyCache::all() was previously called
   // with the same edge_->outputs_ as used here.
   assert(std::equal(checkOutputs_.begin(), checkOutputs_.end(),
-                    edge_->outputs_.begin()));
+                    edge_->outputs_.begin(), edge_->outputs_.end()));
 
   for (std::size_t i = 0; i != edge_->outputs_.size(); ++i) {
     assert(!edge_->is_phony());
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -97,8 +97,7 @@ class RecomputeOutputsDirtyCache {
  public:
   RecomputeOutputsDirtyCache(BuildLog* build_log,
                              OptionalExplanations& explanations, Edge* edge)
-      : buildLog_(build_log), explanations_(explanations), edge_(edge),
-        logEntry_(edge->outputs_.size()) {}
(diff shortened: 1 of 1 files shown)


src/graph.cc
```diff
@@ -105,16 +105,17 @@ class RecomputeOutputsDirtyCache {
...
     /// When calling this function repeatedly for the same object,
     /// the `buildLog` and `output` pointers must remain unchanged.
     /// Stable pointer identity is required for correct caching behavior.
     bool LookupByOutput(const BuildLog* buildLog, const Node* output);
     const BuildLog::LogEntry* operator->() const { return entry_; }
 
    private:
-    bool evaluated_ = false;
-    BuildLog::LogEntry* entry_ = nullptr;
+    static BuildLog::LogEntry* const kUnevaluated;
```
digit-google: nit: Just use a constant pointer value like `reinterpret_cast<const BuildLog::Entry*>(1)` instead of loading a pointer for .rodata on each call.

You cannot make such an expression `constexpr` due to the `reinterpret_cast<>` so, as a special case, just use a macro for simplicity instead. Call this `CACHED_LOG_ENTRY_INIT_SENTINEL` for clarity.

src/graph.cc
```diff
@@ -258,7 +285,7 @@ bool RecomputeOutputsDirtyCache::RecomputeOutputDirty(
   // output file's actual mtime and simply check the recorded mtime from
   // the log against the most recent input's mtime (see below)
   bool used_restat = false;
-  if (isRestat_ && buildLog_ && entry.LookupByOutput(buildLog_, output)) {
+  if (isRestat_() && buildLog_ && entry.LookupByOutput(buildLog_, output)) {
```
digit-google: So much complexity makes the code unnecessarily obscure and hard to follow, Just add a `is_restat_init_` boolean flag, and a `is_restat()` method that will check it before returning the result to perform.

Follow conventions by using `snake_case` and not `pascalCase`.

moritzx22: Done.

src/graph.cc
moritzx22: Done.

PR #2824, opened 2026-08-12
Document -k/-l flags and close browse socket on exit

### LINKED ISSUE #465: Ninja -t browse leaves socket open
Running ninja -t browse X then pressing Ctrl-C leaves the socket open. Rerunning ninja -t browse X causes a socket error.

NInja should either catch the ctrl-c and terminate python (the parent process), or support a ninja -t browsestop tool.

Note : I'm currently running 6a73d41072bafe2384aea20252c3faa6c1ceaa2a

### LINKED ISSUE #2757: Add testing and documentation for `-k` and `-l` flags
These 2 flags are not mentioned in the Manual page (though they are listed when `--help` is used, which may be sufficient as-is), and are not tested in the test suite. I have rough ideas for, and am willing to make a PR (or separate PRs) to fix these issues, which in turn will also fix issue #630.

victoriacampbellmake:
Fixes #2757
Fixes #465

## Summary

- Document `-k` and `-l` in the manual alongside other common command-line flags.
- Add a small test that verifies both flags appear in `ninja -h` output.
- Close the browse web server socket on exit (including Ctrl-C) so rerunning `ninja -t browse` does not fail with a socket error.

## Test plan

- [ ] Run `misc/flag_help_test.py` after building ninja.
- [ ] Start `ninja -t browse`, press Ctrl-C, and verify a second browse session can bind the port.


misc/flag_help_test.py
```diff
@@ -0,0 +1,25 @@
...
+
+class FlagHelp(unittest.TestCase):
+    def test_keep_going_and_load_average_in_help(self):
+        output = subprocess.check_output(
+            [NINJA_PATH, '-h'], stderr=subprocess.STDOUT, text=True
+        )
+        self.assertIn('-k N', output)
+        self.assertIn('keep going', output.lower())
+        self.assertIn('-l N', output)
+        self.assertIn('load average', output.lower())
```
digit-google: I think the value of this test is pretty low. Broad greping for "keep going" and "load average" could corresponds to anything. Besides, we don't really care about regression testing the `--help` message. I recommend dropping this from this PR.

victoriacampbellmake:
Thanks @digit-google — agreed, that `--help` grep test wasn't earning its keep. Dropped `misc/flag_help_test.py` from this PR.

digit-google:
Thanks, can you squash everything into a single commit, this will make the history much cleaner.

PR #2826, opened 2026-08-15
fix(windows): preserve UNC prefix after CanonicalizePath (#2800)

### LINKED ISSUE #2800: On windows, ```CanonicalizePath()``` doesn't seem to handle paths with a leading backslash, but without the drive letter.
A path is passed into  CanonicalizePath() as -
```\\foo\\bar\\baz\\patchlevel.h``` 

And is returned as 
```//foo/bar/baz/patchlevel.h```

Which ultimately ends up with errors such as ```ninja: error: FindFirstFileExA(//foo/bar/baz): The network path was not found.```

In this particular case, the path originates from a .res.d dependency file, which in turn gets canonicalized and then written into .ninja_deps. Thus the first run works ok, but on the second invocation of ninja, .ninja_deps is read back in and the FindFirstFileExA error occurs. Removing .ninja_deps works around the problem.

### PR DESCRIPTION
**Pull request by victoriacampbellmake**
﻿## Summary
Keeps Windows UNC prefixes as `\\` after path canonicalization so Win32 file APIs can open them (#2800).

`CanonicalizePath` rewrites every `\` to `/`. A UNC path such as `\\foo\bar\baz` therefore became `//foo/bar/baz`. On a later ninja run (e.g. after reading `.ninja_deps`), `FindFirstFileExA` treats that form as a network path and fails with "The network path was not found".

Rooted paths without a drive letter (`\foo` → `/foo`) are unchanged.

## Change
After slash normalization on Windows, if the result still starts with `//`, restore the first two characters to `\\`.

## Test
Updates `CanonicalizePath.PathSamplesWindows` for UNC inputs, including the `\\foo\bar\baz\patchlevel.h` case from the issue.

Closes #2800

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -305,8 +305,9 @@ void CanonicalizePath(char* path, size_t* len, uint64_t* slash_bits) {
 
   // UNC paths enter as "\\server\share\...". After the loop they look like
   // "//server/share/...". FindFirstFileExA treats that form as a network path
-  // and fails; keep the Win32 UNC prefix as backslashes (#2800).
-  if (*len >= 2 && start[0] == '/' && start[1] == '/') {
+  // and fails; restore the Win32 UNC prefix only when those separators were
+  // originally backslashes (#2800). Leave an already-forward "//..." alone.
+  if (*len >= 2 && start[0] == '/' && start[1] == '/' && (bits & 0x3) == 0x3) {
     start[0] = '\\';
     start[1] = '\\';
   }
--- a/src/util.cc
+++ b/src/util.cc
@@ -303,6 +303,14 @@ void CanonicalizePath(char* path, size_t* len, uint64_t* slash_bits) {
     }
   }
 
+  // UNC paths enter as "\\server\share\...". After the loop they look like
+  // "//server/share/...". FindFirstFileExA treats that form as a network path
+  // and fails; keep the Win32 UNC prefix as backslashes (#2800).
+  if (*len >= 2 && start[0] == '/' && start[1] == '/') {
+    start[0] = '\\';
+    start[1] = '\\';
+  }
+
   *slash_bits = bits;
 #else
   *slash_bits = 0;
--- a/src/util_test.cc
+++ b/src/util_test.cc
@@ -213,9 +213,14 @@ TEST(CanonicalizePath, PathSamplesWindows) {
   CanonicalizePath(&path);
   EXPECT_EQ("/foo", path);
 
+  // UNC prefix must stay as backslashes for Win32 file APIs (#2800).
   path = "\\\\foo";
   CanonicalizePath(&path);
-  EXPECT_EQ("//foo", path);
+  EXPECT_EQ("\\\\foo", path);
+
+  path = "\\\\foo\\bar\\baz\\patchlevel.h";
+  CanonicalizePath(&path);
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by victoriacampbellmake** [comment]
Pushed a follow-up: only rewrite leading `//` back to `\\` when those two separators were originally backslashes (`slash_bits & 0x3 == 0x3`), so an already-forward `//foo` path is left unchanged.

**Comment by kevle** [comment]
This will not fix paths using "`\\?\`" as an extended-length prefix from what I can tell, since those are not normalized by the Windows file I\O functions.

Once backslashes are allowed in paths at all, it might be reasonable to fully "canonicalize" paths to backslashes on Windows.

**Comment by jhasse** [comment]
what's the difference between UNC paths and network paths?

**Comment by kevle** [comment]
> what's the difference between UNC paths and network paths?

Network paths might be either referred to in the letter ( e.g. Z:\ ... )  format or as an UNC path from what I understand.

What is something else again however, is the extended-length prefix `\\?\`. This might also be used with UNC paths in the form `\\?\UNC\server\share\to\whatever`, but also for "regular" paths `\\?\C:\MyPath`, or `\\?\Volume{26a21bda-a627-11d7-9931-806e6f6e6963}\MyPath`.

Extended-length prefix paths have a few useful properties and peculiarities.

For Windows API's which support them, all string parsing is disabled ( this means path canonicalization and normalization, so conversion from `/ -> \` and `\foo\..\bar -> \bar` ), the path is sent straight to the file system. 
This allows software which runs on Windows versions with filesystem support for long paths to pass paths longer than `MAX_PATH` characters without any administrative intervention. 

Also, no Unicode conversion work is done from what I can tell, so paths should probably stored as WCHARS to use this.

So why would I want to use this troublesome format in the first place?

At the moment, long path ( so paths longer than ` MAX_PATH` 260 characters ) support requires the application manifest to contain `longPathAware` ( Ninja already has this, good! ) and a registry value has to be set: 
```regedit 
[HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\FileSystem]
"LongPathsEnabled"=dword:00000001
```

With extended-length prefix paths, Ninja only needs to be `longPathAware`, without requiring any changes in Windows configuration. 

In my use case I have generated a depfile in CMake containing long paths with extended-length prefixes, but due to Ninja canonicalization, these paths get mangled and Ninja `Stat` calls will fail due the forward slashes inserted in the path. This PR will not help, since the forward slashes are not actually part of the "Windows canonical" path. `\\?\C:/MyPath` will not resolve to `C:\MyPath`.

There is some evidence that this was already taken into consideration at some point in Ninja, see e.g. https://github.com/ninja-build/ninja/blob/b51a1e37c2fb89bbefa600bd155e1ce13983f09d/src/disk_interface.cc#L220


Windows resources related to this:

https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file

https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry

EDIT:

I forgot to mention why I think it might be worthwhile to discuss this in detail. 

If backslashes are allowed in the canonicalized form now, we might as well go all the way and also enable use of fully canonical Windows paths to fully support Windows features that are the remedy to some common problems, e.g. long paths on CI systems without the proper registry value.

jhasse:
> [...] e.g. long paths on CI systems without the proper registry value.

IMHO that's an anti-feature. CI systems should be properly configured.

PR #2828, opened 2026-08-23
Fix phonycycle=warn input handling

### LINKED ISSUE #1518: Parser bugs with phonycycle=warn compatibility mode
For compatibility with CMake 2.8.x and 3.0.x, Ninja provides a `phonycycle=warn` mode that recognizes a phony edge that erroneously contains its own output as an input and filters out the bad self-referencing input. The implementation has a couple of issues:
 - `Edge::maybe_phonycycle_diagnostic` doesn't check for order-only deps, but probably should.
 - In `ParseEdge`, when the duplicate node is removed from `edge->inputs_`, it's not removed from the corresponding `Node::out_edges_`.

For the first issue:

```
$ cat build.ninja
rule echo
  command = echo $out
build foo: echo
build bar: echo
build baz: phony foo bar || baz

$ ninja -t query baz
ninja: warning: phony target 'baz' names itself as an input; ignoring [-w phonycycle=warn]
baz:
  input: phony
    foo
    || bar
  outputs:
    baz
```

Ninja reports that `bar` is an order-only dep of `baz`, but it should be an ordinary dep like `foo`. The fix is trivial; just require that `order_only_deps_ == 0` in `Edge::maybe_phonycycle_diagnostic`.

Running Ninja on the same input file shows the second issue:

```
$ ninja
ninja: warning: phony target 'baz' names itself as an input; ignoring [-w phonycycle=warn]
ninja: error: could not determine root nodes of build graph
```

Ninja doesn't recognize `baz` as a root node because its `Node::out_edges_` fields still includes itself. If the manifest had other root nodes, it would build those, but not `baz`.

Removing the self-reference from `baz` by hand produces the expected output:

```
$ ninja
[1/2] echo bar
bar
[2/2] echo foo
foo
```

I would guess that CMake 2.8.x/3.0.x don't use order-only deps in its phony rules, so the first bug wouldn't affect it. Maybe the second one would.

pttydou:
Fixes #1518.

`phonycycle=warn` treated phony edges with order-only inputs as legacy CMake self-cycles. Removing a self input also left the edge in `Node::out_edges_`. This change limits compatibility filtering to edges without order-only inputs and removes the matching reverse edge when filtering.

Tests:
- `build-cmake\Debug\ninja_test.exe --gtest_filter=ParserTest.*`
- `build-cmake\Debug\ninja_test.exe --gtest_filter=ParserTest.PhonySelfReference*`
- `python misc/ninja_syntax_test.py`

### DIFF
--- a/src/graph.cc
+++ b/src/graph.cc
@@ -797,7 +797,7 @@ bool Edge::maybe_phonycycle_diagnostic() const {
   // of the form "build a: phony ... a ...".   Restrict our
   // "phonycycle" diagnostic option to the form it used.
   return is_phony() && outputs_.size() == 1 && implicit_outs_ == 0 &&
-      implicit_deps_ == 0;
+      implicit_deps_ == 0 && order_only_deps_ == 0;
 }
 
 // static
--- a/src/graph.h
+++ b/src/graph.h
@@ -116,6 +116,10 @@ struct Node {
   const std::vector<Edge*>& out_edges() const { return out_edges_; }
   const std::vector<Edge*>& validation_out_edges() const { return validation_out_edges_; }
   void AddOutEdge(Edge* edge) { out_edges_.push_back(edge); }
+  void RemoveOutEdge(Edge* edge) {
+    out_edges_.erase(std::remove(out_edges_.begin(), out_edges_.end(), edge),
+                     out_edges_.end());
+  }
   void AddValidationOutEdge(Edge* edge) { validation_out_edges_.push_back(edge); }
 
   void Dump(const char* prefix="") const;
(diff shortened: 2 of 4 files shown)


digit-google:
> Tests:
> 
> * `build-cmake\Debug\ninja_test.exe --gtest_filter=ParserTest.*`
> * `build-cmake\Debug\ninja_test.exe --gtest_filter=-DiskInterfaceTest.StatSymlink`
> * `build-cmake\Release\ninja_test.exe --gtest_filter=-DiskInterfaceTest.StatSymlink --gtest_brief=1`
> * `python misc/ninja_syntax_test.py`
> 
> The excluded Windows test requires symlink permission that is unavailable on this machine.

This change has nothing to do with symlinks so I am puzzled by the content of this section? Apart from that, the PR looks good, thank you.

digit-google (approved):
@jhasse ptal

pttydou:
Good point. That note was only documenting an unrelated limitation of my local Windows environment, so I removed it and kept the focused parser checks in the PR description.

PR #2832, opened 2026-09-09
Enable tweaking the scheduling order for first-time builds via explicit targets

kinke:
This is a stab at https://github.com/ninja-build/ninja/issues/232, allowing to tweak the scheduling order via e.g. `ninja <slow output> all`, for first-time builds without previous build times yet, such as fresh CI builds in ephemeral containers.

Implemented by overriding the weight of target-producing edges iff.
* multiple targets have been specified, and
* there is no previous build time for that edge yet.

This target-order-specific weight is stored in the upper 32 bits of the 64-bit weight, starting with 1<<32 for the last target's edge, 2<<32 for the 2nd-last target's, 3<<32 for the 3rd-last etc. It should thus always dominate normal weights (e.g., elapsed milliseconds for the previous build).

(Something similar to this was implemented in #2019, but not in accepted #2177. With #2686 having landed, I think this is the last remaining interesting piece from #2019.)

### DIFF
--- a/src/build.cc
+++ b/src/build.cc
@@ -541,11 +541,30 @@ void Plan::ComputeCriticalPath() {
 
   const auto& sorted_edges = topo_sort.result();
 
-  // First, reset all weights to 1.
+  // First, reset all weights to the previous elapsed time,
+  // falling back to 1 if unavailable, or 0 for phony edges.
   for (Edge* edge : sorted_edges)
     edge->set_critical_path_weight(EdgeWeightHeuristic(edge));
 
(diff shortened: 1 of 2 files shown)


kinke:
Oh, #232 is actually about adding a field in the build.ninja file IIUC. My number-one use case is for a CMake build, so no idea if CMake would ever support such an order/weight for its Ninja generator, if ninja supported that at one point. Adding explicit targets in the ninja command-line works today and doesn't require further tools support.

